PR #3, opened 2021-09-13
Add tube

TiagoCavalcante:
@rocky I'm having the error `no matching function 'draw(tube, pen)'` (or the error `no matching function 'surface(draw(tube, pen))'` when I change the code). But I believe that is because of a wrong Asymptote version or something like that.

I see `tube` in these 2 places:
- https://tex.stackexchange.com/questions/520338/the-transform-parameter-of-tube-in-asymptote
- https://asymptote.sourceforge.io/doc/three.html

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,8 @@ New builtins
 ++++++++++++
 
 * ``Guidermannian``
+* ``Cone``
+* ``Tube``
 
 Tensor functions:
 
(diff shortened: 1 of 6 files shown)


TiagoCavalcante:
image

mmatera (approved):

rocky (approved):
LGTM - we will deal with the asymptote side of things in the futre.

PR #4, opened 2021-09-14
Improving pattern matching

mmatera:
This PR contains several changes that  (after many tests) I think could be useful to accelerate the pattern matching mechanism.

* unroll any/all calls as loops in Cythonized modules: this avoids the overhead of a function call replacing it by a (C) `for` loop.
* `BaseExpression.get_head`  now avoids building a symbol and then look for its name. It saves two function calls.
* Now, `SameQ` first checks type, then `id`s, and then names in symbols.
* In `mathics.builtin.patterns.PatternTest`, if the condition is one of the most used tests (`NumberQ`, `NumericQ`, `StringQ`, etc) the `match` method is overwritten to specialized versions that avoid function calls.
* in the same aim, `mathics.core.patterns.AtomPattern` now specializes the comparison depending of the `Atom` type.


rocky:
@mmatera Would you attach suitable benchmarks please?

mmatera:
> @mmatera Would you attach suitable benchmarks please?

Benchmarks (I am going updating this as I get the numbers):

* docpipeline takes
    master: 0:04:54.701342
    faster_pattern_match: 0:04:22.623170

rocky (approved):
LGTM - merge when you feel this is ready.

Thanks for attaching benchmarks.

PR #7, opened 2021-09-16
Remove sort

TiagoCavalcante:


mathics/core/pattern.py
```diff
@@ -334,7 +334,6 @@ def yield_head(head_vars, _):
 
     def get_pre_choices(self, yield_func, expression, attributes, vars):
         if "System`Orderless" in attributes:
-            self.sort()
```
rocky, line 337: Why do you think this can be removed?

TiagoCavalcante: If that is orderless, why does the order matter? And the tests pass.

rocky: So maybe that should be if not "System`Orderless". 

A couple of important things. 

1. Don't assume that if test passing means that code is correct
2. If you see code that looks funny and more complicated make an effort to find out _why_ that code was added in the first place. Git blame might help here.

TiagoCavalcante:
@rocky I benchmarked it, but cannot see the difference. That is floating too much.

Almost nothing is `Orderless`, `Plus` and `Times` are, and I don't believe their pattern matching is going to change because of this.

@poeschko wrote that line 10 years ago.

rocky:
> @rocky I benchmarked it, but cannot see the difference. That is floating too much.
> 
> Almost nothing is `Orderless`, `Plus` and `Times` are, and I don't believe their pattern matching is going to change because of this.
> 
> @poeschko wrote that line 10 years ago.

That means he probably intended that for a reason. 

Ok. So if you don't mind let's leave this. It is not causing harm. Please let's not just change things for such small things and instead focus on the more important things. The chance of inadvertent breakage is too great and it takes us too much time and effort to track down subtle bugs that could be caused by this. 

At such time as the code is better organized and documented we may be in a better position. Right now we need to exercise more caution when changing code. 

If you want to add a _comment_ saying you think this is unnecessary that is fine.

TiagoCavalcante:
> Ok. So if you don't mind let's leave this. It is not causing harm. Please let's not just change things for such small things and instead focus on the more important things. The chance of inadvertent breakage is too great and it takes us too much time and effort to track down subtle bugs that could be caused by this.

I don't mind.

rocky:
Thanks @TiagoCavalcante ! I greatly appreciate it.

mathics/core/pattern.py
poeschko: Yes, canonicalizing before pattern matching was the idea here AFAIR.

PR #8, opened 2021-09-16
More string split tests and doc

rocky:
Add WL tests, revise doc
    
We should not be using #> except to test error messages. Instead, these should be moved to pytest where they run faster.
    
Revise documentation for StringSplit. Note that carriage returns are hard returns, not wrapped. They will appear in docs.
    
Describe more about why we are showing the examples shown.

Add summary text for more strings operations

### DIFF
--- a/mathics/builtin/string/operations.py
+++ b/mathics/builtin/string/operations.py
@@ -813,7 +813,7 @@ class StringSplit(Builtin):
       <dt>'StringSplit[$s$]'
       <dd>splits the string $s$ at whitespace, discarding the whitespace and returning a list of strings.
 
-      <dt>'StringSplit[$s$, "$pattern$"]'
+      <dt>'StringSplit[$s$, $pattern$]'
       <dd>splits $s$ into substrings separated by delimiters matching the string expression $pattern$.
 
       <dt>'StringSplit[$s$, {$p_1$, $p_2$, ...}]'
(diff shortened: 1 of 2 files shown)


mathics/builtin/string/operations.py
```diff
@@ -801,26 +810,30 @@ def apply(self, liststr, seps, evaluation):
...
-        <dd>splits $s$ at the delimiter $d$.
-    <dt>'StringSplit[$s$, {"$d1$", "$d2$", ...}]'
-        <dd>splits $s$ using multiple delimiters.
-    <dt>'StringSplit[{$s_1$, $s_2, ...}, {"$d1$", "$d2$", ...}]'
-        <dd>returns a list with the result of applying the function to
-            each element.
+      <dt>'StringSplit[$s$]'
+      <dd>splits the string $s$ at whitespace, discarding the whitespace and returning a list of strings.
+
+      <dt>'StringSplit[$s$, "$pattern$"]'
```
TiagoCavalcante, line 816: Maybe remove the quotes arounf `$pattern$`?

mathics/builtin/string/operations.py
```diff
@@ -868,6 +875,8 @@ class StringSplit(Builtin):
         "pysplit": "As of Python 3.5 re.split does not handle empty pattern matches.",
```
TiagoCavalcante, line 875: Now I see this. Is Python 3.5 supported?

TiagoCavalcante (approved):
LGTM

mathics/builtin/string/operations.py
rocky: Absolutely! Thanks for catching.

PR #9, opened 2021-09-16
Add clean-cython target

rocky:
@mmatera `make clean-cython && make`  is what should be used in between benchmark runs when you need to switch git SHA. 

This branch adds the `make clean-cython` part.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,8 @@ check: pytest gstest doctest
 
 #: Remove Cython-derived files
 clean-cython:
-	find mathics -name "*.so" -type f -delete
+	find mathics -name "*.so" -type f -delete; \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -53,9 +53,12 @@ install:
 check: pytest gstest doctest
 
 
+#: Remove Cython-derived files
+clean-cython:
+	find mathics -name "*.so" -type f -delete
```
mmatera, line 58: shouldn't also clean .c files?

rocky: I am not sure it not strictly necessary, but I guess it is better to be safe here. And this will work when there are no `.so` s but something else...

mmatera:
LGTM

TiagoCavalcante (approved):
LGTM

PR #10, opened 2021-09-17
Use Builtin class summary_text in `?`

rocky:
Recenty we have been filling out `summary_text` for Built-in Functions.

This text currently appears in Django when it is listed in a group. See for example "Operations on Strings".

This PR expands that use so that this short text also appears when `?xxx` or `Information[xxx, LongForm->False]` is requested.

### DIFF
--- a/mathics/builtin/assignment.py
+++ b/mathics/builtin/assignment.py
@@ -975,7 +975,7 @@ class Definition(Builtin):
         return self.format_definition(symbol, evaluation, grid=False)
 
 
-def _get_usage_string(symbol, evaluation, htmlout=False):
+def _get_usage_string(symbol, evaluation, is_long_form: bool, htmlout=False):
     """
     Returns a python string with the documentation associated to a given symbol.
     """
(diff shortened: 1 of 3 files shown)


rocky:
I would like to add a test for this, but currently we don't test this, and we need an output test which we don't have in pytest yet. 

So this will come later.

mathics/builtin/assignment.py
```diff
@@ -1000,6 +1000,8 @@ def _get_usage_string(symbol, evaluation, htmlout=False):
         bio = builtins.get(definition.name)
 
     if bio is not None:
+        if not is_long_form and hasattr(bio.builtin.__class__, "summary_text"):
```
TiagoCavalcante, line 1003: Looks great!

TiagoCavalcante (approved):

mmatera:
LGTM

rocky:
Am. Traveling. If someone would merge I would be grateful.

TiagoCavalcante:
@mmatera @rocky I want to add the test before merging. I'll do it until tommorrow.

rocky:
Thanks @TiagoCavalcante  ! Good work!

PR #11, opened 2021-09-18
Profiledocpipeline

mmatera:
In this PR another proposal for profiling changes is implemented. The idea is to profile `docpipeline.py` for different branches, and then compare which functions got an improvement in performance and which ones got slowed down.
What I am not sure is if this fit in mathics-core, or should be inside `mathics_benchmarks`

Update: In the current implementation there is a problem that I could not fix: `cProfile` does not keep track of fully qualified method names (let's say `mathics.buiiltin.arithfns.Plus.apply`), but just keep the name of the method, the file and line number where they are allocated ( `apply, mathics.builtins.arithfns/basic.py:366` ).
The problem is that when we want to compare two branches, the line numbers usually changes, and then the two versions can not be compared. @rocky or someone else: do you know if there is a way to translate `file:line funcname` into a fully qualified method name?


TiagoCavalcante:
> What I am not sure is if this fit in mathics-core, or should be inside mathics-benchmark

I believe this is mathics-benchmark.

PR #13, opened 2021-09-20
Faster n apply new

mmatera:
This PR inverts the roles of `N.apply` and `apply_N`, in a way that calling this last one does not imply passing through the pattern matching mechanism.
This PR also fixes a few bugs in the `Definition` object.

### DIFF
--- a/mathics/builtin/arithfns/basic.py
+++ b/mathics/builtin/arithfns/basic.py
@@ -555,6 +555,7 @@ class Power(BinaryOperator, _MPMathFunction):
                 y_err = y
             else:
                 y_err = apply_N(y, evaluation)
+                y_err = y_err if y_err.is_atom() else y_err.evaluate(evaluation)
             if isinstance(y_err, Number):
                 py_y = y_err.round_to_float(permit_complex=True).real
                 if py_y > 0:
--- a/mathics/builtin/numeric.py
+++ b/mathics/builtin/numeric.py
@@ -954,6 +954,8 @@ class N(Builtin):
 
     def apply_with_prec(self, expr, prec, evaluation):
         "N[expr_, prec_]"
+        if not prec.is_numeric(evaluation):
+            return
         val = apply_N(expr, evaluation, prec)
         if not isinstance(val, Number):
             val = val.evaluate(evaluation)
--- a/mathics/builtin/numeric.py
+++ b/mathics/builtin/numeric.py
@@ -44,6 +44,9 @@ from mathics.core.expression import (
     SymbolList,
     SymbolMachinePrecision,
     SymbolN,
+    SymbolPlus,
+    SymbolTimes,
+    SymbolPower,
     from_python,
 )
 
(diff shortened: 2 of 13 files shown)


rocky:
Benchmarks, please? (Any PR motivated by performance should have been benchmarks.)

mmatera:
@rocky, this is in progress. I just was able to make it works without apparent side effects...

rocky:
> @rocky, this is in progress. I just was able to make it works without apparent side effects...

Ok - then marked as a draft.  (Still, it is fine to give benchmarks while in progress as well as at the end.)

mathics/core/expression.py
```diff
@@ -594,6 +594,10 @@ def round_to_float(self, evaluation=None, permit_complex=False):
         if evaluation is None:
             value = self
         elif isinstance(evaluation, sympy.core.numbers.NaN):
+            print(
+                "why an evaluation parameter would be set to sympy.NaN and this should happend? "
+            )
+            assert False
```
rocky, line 600: This is essentially the same as raise `RuntimeError("your message here")1

mmatera: Indeed, but I also wanted to know why this check is done...

mmatera:
First benchmarks:


|              Expression                   |    Master     |    faster_n_apply_new  |  Improvement |
|------------------------------------------ | ------------  |  ----------------------| ------------ |
|`12382764626155151+475628472748234827.//N` |   	0.000874788 |   0.000815672          |    6.8%      |
|`12382764626155151-475628472748234827.//N`|	0.000668302 |	0.000561317          |    16.0 %    |
|`Table[BesselJ[1,x],{x,0,2,.1}]//N;`	|	0.0236843   |	0.0207219            |    12.2 %    |
|`Table[BesselJ[n,x],{n,0,5},{x,0,2,.1}]//N;`|	0.115693    |	0.0935047            |    18.7 %    |
|`Table[BesselJ[n,x]^2.1,{n,0,5},{x,0,2,.1}]//N;`|1.30566     |	1.17602              |    10 %      |
|`Table[Pi^(n+m),{n,0,5},{m,0,5}]//N;` 	      |  0.074508    |	0.0576915            |    22.6%     |   

Note: I just take the numbers using Time[expr;][[1]]. Later I can try to do the same with mathics-benchmarks

rocky:
> First benchmarks:
> 
> Expression	Master	faster_n_apply_new	Improvement
> `12382764626155151+475628472748234827.//N`	0.000874788	0.000815672	6.8%
> `12382764626155151-475628472748234827.//N`	0.000668302	0.000561317	16.0 %
> `Table[BesselJ[1,x],{x,0,2,.1}]//N;`	0.0236843	0.0207219	12.2 %
> `Table[BesselJ[n,x],{n,0,5},{x,0,2,.1}]//N;`	0.115693	0.0935047	18.7 %
> `Table[BesselJ[n,x]^2.1,{n,0,5},{x,0,2,.1}]//N;`	1.30566	1.17602	10 %
> `Table[Pi^(n+m),{n,0,5},{m,0,5}]//N;`	0.074508	0.0576915	22.6%
> Note: I just take the numbers using Time[expr;][[1]]. Later I can try to do the same with mathics-benchmarks

Thanks @mmatera !

mathics/builtin/arithfns/basic.py
```diff
@@ -555,6 +555,7 @@ def apply_check(self, x, y, evaluation):
                 y_err = y
             else:
                 y_err = apply_N(y, evaluation)
+                y_err = y_err if y_err.is_atom() else y_err.evaluate(evaluation)
```
TiagoCavalcante, line 558: I don't understand why we need to re-evaluate here and in other places.

mathics/builtin/numeric.py
```diff
@@ -62,8 +65,107 @@ def log_n_b(py_n, py_b) -> int:
...
+    """
+    try:
+        d = get_precision(prec, evaluation)
+    except PrecisionValueError:
+        return
+
+    if isinstance(expr, Number):
+        return expr.round(d)
+    if isinstance(expr, String):
+        return expr
```
TiagoCavalcante, line 88: Why accept strings here?

mmatera: Imagine you have the expression

`N[MyFunc[Pi^2, Method->"mymethod"]]`

Then, at some point, you have to apply `apply_N("mymethod")`. But then, you do not want to look for a pattern that converts "mymethod" to a number. Then, the best is to return the expression immediately.

mmatera:
Continues in #15

mathics/builtin/arithfns/basic.py
mmatera: I have explained that in the docstring of `apply_N`. In the previous implementation,
```
Expression(SymbolN, expr, prec).evaluate(evaluation)
```
first, ``expr`` is evaluated. Then, ``apply_with_prec()`` is called and evaluated, and then, the result is returned to  the `evaluate` method. If there were a change in the expression, then the returned value is reevaluated.
In the new implementation, this would be equivalent to
````
apply_N(expr.evaluate(evaluation), prec.evaluate(evaluation)).evaluate(evaluation)
```
There are then several  improvements using `apply_N` instead of the previous construction:

* Inside a Python routine, usually `expr` is written in its final form (if you want to pass  `expr=F[4]`, you do not construct it as `expr=F[1+3]`).
* The return value of `apply_N` are usually `Real` or `Complex` numbers. Then, if you call `evaluation(evaluate)` you would be wasting a function call. If the result of `apply_N` is not a number, a comparison shouldn't introduce a substantial overhead over the work of a full evaluation (implying pattern matching of rules, and several function calls). On the other hand, if you get a number,  a comparison is always cheaper than a function call.

mathics/builtin/numeric.py
TiagoCavalcante: Oh, I understand, `apply_N` is recursive. Thanks!

mmatera: Yep. This is why I can not put the conditional evaluation *inside* `apply_N`.

PR #14, opened 2021-09-21
Small tweaks

TiagoCavalcante:


mathics/builtin/inference.py
```diff
@@ -273,9 +272,6 @@ def get_assumption_rules_dispatch(evaluation):
     if val_consistent_assumptions == SymbolFalse:
         evaluation.message("Inconsistent assumptions")
 
-    if assumptions_list is None:
```
mmatera, line 276: Not sure about this one.

mathics/builtin/inference.py
```diff
@@ -298,10 +294,8 @@ def get_assumption_rules_dispatch(evaluation):
 
         if value:
             symbol_value = SymbolTrue
-            symbol_negate_value = SymbolFalse
```
mmatera, line 301: Sure that these are not necesary? I do not remember exactly why I put these lines...

mmatera:
Apart from the comments, LGTM

mathics/builtin/inference.py
TiagoCavalcante: `pred` isn't defined in this function, so if it be `None`, an undeclared variable exception would be raised.

mathics/builtin/inference.py
mmatera: In that case, go ahead...

mathics/builtin/inference.py
mmatera: OK, if the CI does not complains, then go ahead. At some point, I have to go through this code again anyway

rocky:
Benchmarks, please? Any PR that purports to improving performance should have benchmarks attached. Thanks.

mmatera:
This was just cleaning, I think it is OK. In any case, I will try to merge this on #13 to see if there are no conflicts.

PR #15, opened 2021-09-23
Faster n apply

mmatera:
(This is just a squashed version of #13)

This PR inverts the roles of N.apply and apply_N, in a way that calling this last one does not imply pass through the pattern matching mechanism.
This PR also fixes a few bugs in the Definition object.

Basic benchmarks:

First benchmarks:

|              Expression                   |    Master     |    faster_n_apply_new  |  Improvement |
|------------------------------------------ | ------------  |  ----------------------| ------------ |
|`12382764626155151+475628472748234827.//N` |   	0.000874788 |   0.000815672          |    6.8%      |
|`12382764626155151-475628472748234827.//N`|	0.000668302 |	0.000561317          |    16.0 %    |
|`Table[BesselJ[1,x],{x,0,2,.1}]//N;`	|	0.0236843   |	0.0207219            |    12.2 %    |
|`Table[BesselJ[n,x],{n,0,5},{x,0,2,.1}]//N;`|	0.115693    |	0.0935047            |    18.7 %    |
|`Table[BesselJ[n,x]^2.1,{n,0,5},{x,0,2,.1}]//N;`|1.30566     |	1.17602              |    10 %      |
|`Table[Pi^(n+m),{n,0,5},{m,0,5}]//N;` 	      |  0.074508    |	0.0576915            |    22.6%     |   

Note: I just take the numbers using Time[expr;][[1]]. Later I can try to do the same with mathics-benchmarks

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -44,6 +44,8 @@ Bugs
 * Improve Asymptote ``BezierCurve``implementation
 * ``Rationalize`` gives symmetric results for +/- like MMA does. If
   the result is an integer, it stays that way.
+* stream processing was redone. ``InputStream``, ``OutputStream`` and
+  ``StringToStream`` should all open, close, and assign stream numbers now
 
 4.0.0
 -----
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -42,6 +42,8 @@ Bugs
 
 * Fix and document better behavior of ``Quantile``
 * Improve Asymptote ``BezierCurve``implementation
+* ``Rationalize`` gives symmetric results for +/- like MMA does. If
+  the result is an integer, it stays that way.
 
 4.0.0
 -----
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,6 +31,8 @@ Enhancements
   future version may improve SVG handling of Density plots using
   elliptic density gratings in SVG. And/or we may define this in the
   JSON API.
+* SVG and Asymptote drawing now includes inline comments indicating
+  which Box Structures are being implemented in code
 
 Documentation
 .............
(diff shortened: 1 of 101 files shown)


CHANGES.rst
```diff
@@ -6,9 +6,8 @@ Internals
 
 * now `Expression.is_numeric()` accepts an `Evaluation` object as a parameter,
```
rocky, line 7: If the only think needed are the definitions, then why narrow the parameter to just the definition? 

@mmatera Your thoughts?

mmatera: @rocky, I just leave this in case we want to delegate something to the evaluation object, but maybe it would be clearer if the parameter is the `Definitions` object.

mathics/builtin/lists.py
```diff
@@ -2510,6 +2518,8 @@ def __init__(self, df, p, evaluation):
     def _compute_distance(self, i, j):
         p = self._p
         d = apply_N(self._df(p[i], p[j]), self._evaluation)
```
rocky, line 2520: I see this pattern of apply_N followed by an if statement where the more general evaluate needs to be done many times in this PR. 

Why not just do this inside apply_N ?  Or have two kinds of apply_N ?  

More broadly it is this kind of activity that I feel down the line makes the code more complex because it is an idiom that you sort of have to figure out by looking at, and it is not clear to me that duplicating the code everywhere is that much of a win over just doing it once in a function somewhere.

mathics/core/expression.py
```diff
@@ -3317,3 +3353,106 @@ def construct(leaves):
 
 def string_list(head, leaves, evaluation):
     return atom_list_constructor(evaluation, head, "String")(leaves)
+
+
+def apply_N(expr, evaluation, prec=SymbolMachinePrecision):
+    """
+    This function applies recursively `N-value` rules for convert symbols, expressions and exact numbers
+    into Real numbers with certain precision.
+    Notice that `apply_N(expr, evaluation, prec)` is not equivalent in general to call
```
rocky, line 3362: This is related to my comment before. Maybe this _should_ be made to be equivalent so that the
caller doesn't have to worry. 

Imagine if we wanted to make a mathics "compiler" to bytecode. One of the bytecode operations would be "N".

mmatera:
superseded by #139

PR #18, opened 2021-09-26
Split core expression additional tweaks

rocky:


rocky:
@mmatera both this and master fail quite a bit locally. Wondering what CI thinks.

mathics/__init__.py
```diff
@@ -8,23 +8,6 @@
 import numpy
 
 from mathics.version import __version__
-from mathics.core.symbols import Symbol
-from mathics.core.atoms import (
```
rocky, line 12: I believe these imports are allowing a lot of sloppiness in imports. 

Right now we don't have any non-mathics library uses. Although something like this down the line might make sense, I think we should come back to this after the code has been reorganized.

mathics/builtin/arithmetic.py
```diff
@@ -35,14 +35,11 @@
     from_mpmath,
     from_python,
 )
-from mathics.core.symbols import Symbol
+from mathics.core.symbols import Symbol, SymbolFalse, SymbolList, SymbolTrue
 from mathics.core.systemsymbols import (
```
rocky, line 39: I suspect those __init__ imports were allowing incorrect imports like this from occurring.

mathics/builtin/comparison.py
```diff
@@ -25,15 +25,13 @@
...
-from mathics.core.symbols import Symbol
+from mathics.core.symbols import Symbol, SymbolFalse, SymbolTrue
 from mathics.core.systemsymbols import (
-    SymbolFalse,
-    SymbolTrue,
     SymbolDirectedInfinity,
     SymbolInfinity,
     SymbolComplexInfinity,
 )
-from mathics.core.numbers import dps
```
rocky, line 36: `numbers` was rename to `number` because of a possible numpy conflict. See https://stackoverflow.com/questions/53668779/attributeerror-module-numbers-has-no-attribute-integral

After this change I was for example able to run `python mathics-core/core/atoms.py` without error.

mathics/builtin/arithmetic.py
mmatera: OK, I do not have clear what is the best way to do this. I thought that import Symbol* from systemsymbols was more consistent and easy to change/update in the future (let's say, using ed). To change all the modules to load symbols from their new place would have been easier if the different symbols had been in different import lines.

mathics/builtin/comparison.py
mmatera: excelent, thanks!

mmatera:
LGTM

mmatera (approved):

PR #19, opened 2021-09-27
Narrow description to just the mathics-core...

rocky:
and add links to mathics.org for a high-level view or mathics-omnibus
for a one-stop install process.


TiagoCavalcante (approved):
LGTM

PR #21, opened 2021-09-27
Improve minus performance

TiagoCavalcante:


mathics/builtin/arithfns/basic.py
```diff
@@ -289,22 +289,17 @@ def format_plus(self, items, evaluation):
         "Plus[items__]"
 
         def negate(item):
-            if item.has_form("Times", 1, None):
```
rocky, line 292: Why is it this we know that line can be removed?

rocky:
It looks like some code was moved around and a Form check removed. Why do we believe this is better? 

Are there examples and benchmarks that show a difference?

TiagoCavalcante:
> Are there examples and benchmarks that show a difference?

I'll attach the benchmarks, but before I guess there's more thing to do.

mathics/builtin/arithfns/basic.py
```diff
@@ -289,22 +289,17 @@ def format_plus(self, items, evaluation):
...
-                if isinstance(item.leaves[0], Number):
-                    neg = -item.leaves[0]
-                    if neg.sameQ(Integer1):
-                        if len(item.leaves) == 1:
-                            return neg
-                        else:
-                            return Expression("Times", *item.leaves[1:])
+            if isinstance(item.leaves[0], Number):
```
TiagoCavalcante, line 292: This already covers the `if item.has_form("Times", 1, None)`.

rocky: > This already covers the `if item.has_form("Times", 1, None)`.

To me this sounds like you are saying the same thing but in different words: the if test was removed because it is already covered. 

The if test checks if "Head "is "Times", and item has (at least?)  one leaf. If this is not checked, how do we know that this is the case. Or how do we know this check is not needed?

TiagoCavalcante:
image

*Note*: I'll give an detailed explain of the "if" today/tommorow.

mathics/builtin/arithfns/basic.py
```diff
@@ -289,22 +289,15 @@ def format_plus(self, items, evaluation):
...
-                    else:
-                        return Expression("Times", neg, *item.leaves[1:])
+            neg: Integer = -item.leaves[0]
+
+            if neg.sameQ(Integer1):
+                if len(item.leaves) == 1:
+                    return neg
                 else:
-                    return Expression("Times", -1, *item.leaves)
-            elif isinstance(item, Number):
```
TiagoCavalcante, line 304: It is never a `Number`.

mathics/builtin/arithfns/basic.py
TiagoCavalcante: If it already is `Times`, we are evolving its leaves in another `Times`, I'll attach the benchmarks.

TiagoCavalcante: And if it isn't, we are evolving in `Times` too, so same thing.

mathics/core/atoms.py
```diff
@@ -36,18 +36,20 @@
 @lru_cache(maxsize=1024)
 def from_mpmath(value, prec=None):
     "Converts mpf or mpc to Number."
+    if isinstance(value, mpmath.mpc):
```
TiagoCavalcante, line 39: This removes a function call.

mathics/builtin/arithfns/basic.py
TiagoCavalcante: That is never a `Number` because we only want to negate `x + -y` and `-y` is the same as `Times[-1, y]`.

mathics/builtin/arithfns/basic.py
TiagoCavalcante: If it is `Times` we are changing to another `Times`, and if it isn't, we are creating a `Times`, so same thing.

mathics/core/atoms.py
```diff
@@ -36,18 +36,20 @@
...
+        if value.imag == 0.0:
+            value = value.real
+        else:
+            real = from_mpmath(value.real, prec)
+            imag = from_mpmath(value.imag, prec)
+            return Complex(real, imag)
+
     if isinstance(value, mpmath.mpf):
```
mmatera, line 47: @TiagoCavalcante , also, you can improve a little bit more it, avoiding the `isinstance()` in case you already know that value is the imaginary part of an `mpmath.mpc`:

```
elif not isinstance(value, mpmath.mpf):
     raise TypeError(type(value))
if prec is None:
    return MachineReal(float(value))
else:
    return PrecisionReal(sympy.Float(str(value), prec))
```

mathics/builtin/arithfns/basic.py
rocky: All of these assumptions should be written in as comments. They might be wrong and they may change at some point.

TiagoCavalcante:
This didn't result in a perceptible difference, but I don't see anything more to do here.

image

mmatera:
@TiagoCavalcante , I think that the best we can do now is to try to understand how to improve the pattern matching mechanism. For that, probably we want to have a look at how it works in symja

TiagoCavalcante:
@mmatera maybe that is not the only problem, the recursivness is very slow, and the formatting seems to be slow too.

mmatera:
> @mmatera maybe that is not the only problem, the recursivness is very slow, and the formatting seems to be slow too.

Yep, in Python especially, but I do not see a way to avoid it here. The question is if we can reduce the overhead by improving the algorithm.

TiagoCavalcante:
@mmatera the unique algorithm that may be improved here is the one from `format_plus`, but I don't believe that is so bad. About the recursiveness, seems that there's a lot of inheritance in the pattern test classes (what leads to more function calls and also a lot of checking that may not be necessary).

rocky:
It feels to me like we are bouncing around from one place to another without following through on past or existing ideas.

At one point when it was decided to whack the guts out of numeric evaluation we were going to identify those primitives that are used the most or which are spending the most time.

In benchmarks I have a prototype for determining which operations are used the most commonly. That idea hasn't been fleshed out. 

Before that there is execute_fast() somewhere and that is another idea that hasn't been finished because we detoured into the numeric evaluation speedup and things related to that. 

It is not the "pattern matching mechanism" pers se, but the the evaluation _process_    This looseness of terminology I think can lead people astray. 

But if you want to focus on the evaluation process, I'd suggest _isolating_ that in some set of tests. We can set up a small and controlled number of definitions that exhibit the kinds of things that cause problems. In this test setup we can strip things down to _just the process of evaluation_ . All that matters here is that the expansion works and the right sequences are returned.

mmatera:
> It feels to me like we are bouncing around from one place to another without following through on past or existing ideas.
> 
I do not agree with this: there is a set of suspicious of dragging performance:
* frequently used primitives (N, NumberQ, NumericQ, SymbolQ)
* pattern matching routines (comparing strings instead of symbols)
* lookup definitions (innecesary recursivity, inefficiency in the choice of lookup keys)
* the evaluation loop: (using pattern matching where a direct call would be possible)

The problem is that many of them are entangled in the current implementation. One example is `applyN`: the current implementation is recursive, evaluates many times the same thing, on each recursion it uses the pattern matching mechanism, uses `NumericQ` and is called in many places (for example, in `Plus` and `Times` through `expr.numerify()`). What we are currently doing is to identify these primitives, splitting as much as possible to understand better their behavior, and decouple problems. The process is slow, because there are 10+ years of development and no one of us knows how all the parts works.

> At one point when it was decided to whack the guts out of numeric evaluation we were going to identify those primitives that are used the most or which are spending the most time.
> 

Yep, this is a part of the work. Now, we have PRs for identifying these primitives, and improving clarity and efficiency for some of them.

> In benchmarks I have a prototype for determining which operations are used the most commonly. That idea hasn't been fleshed out.
> 

Actually, it is a work in progress.

> Before that there is execute_fast() somewhere and that is another idea that hasn't been finished because we detoured into the numeric evaluation speedup and things related to that.
> 

@rocky,  I didn't dig in that line because I thought you had an idea about how to implement it. I think both directions are not incompatible.

> It is not the "pattern matching mechanism" pers se, but the the evaluation _process_ This looseness of terminology I think can lead people astray.
> 
mmm... when I talk about the "pattern matching mechanism", I am thinking of a particular piece of the evaluation process: take an `Expression` object, look into the definitions for a rule that matches with it, and apply the rule found. The evaluation process is wider: there is a recursive loop, caching strategies, built-in rules implemented in Python... but, by looking at the results of the recent benchmark experiments, it seems that the "pattern matching mechanism" takes a considerable part of the time required for evaluating simple expressions like 'a+b+2.4+a', which takes around 0.1s. It is a ridiculous large time for such a simple operation. 




> But if you want to focus on the evaluation process, I'd suggest _isolating_ that in some set of tests. We can set up a small and controlled number of definitions that exhibit the kinds of things that cause problems. In this test setup we can strip things down to _just the process of evaluation_ . All that matters here is that the expansion works and the right sequences are returned.

Yep, that is the idea. But to test individual parts of the process implies being able first to decomposing the process. That is what I am trying to do now.

PR #22, opened 2021-09-28
Fix tests

TiagoCavalcante:
This needs to be used with [mathics-scanner#35](https://github.com/Mathics3/mathics-scanner/pull/35) and [mathics-scanner#36](https://github.com/Mathics3/mathics-scanner/pull/36).


mmatera:
Maybe a change in setup.py is needed in order to pick the right version, isn't it?

TiagoCavalcante:
@mmatera those PRs need to be merged first.

mmatera:
In that case, go ahead when you feel ready

TiagoCavalcante:
@rocky using the installed version of Mathics Scanner with the latest PRs pass the tests (`make check -j3`). I think a new version of Mathics Scanner needs to be released (as @mmatera said).

TiagoCavalcante:
@rocky locally this passes, I believe a new version of Mathics Scanner needs to be released so this passes. Picking the Git egg version of Mathics Scanner isn't working.

test/test_parser/test_parser.py
```diff
@@ -93,9 +93,9 @@ def testSymbol(self):
     def testSpecialSymbol(self):
         self.check("\\[Pi]", "Pi")
         self.check("\\[Degree]", "Degree")
-        self.check("\\[ExponentialE]", "E")
-        self.check("\\[ImaginaryI]", "I")
-        self.check("\\[ImaginaryJ]", "I")
+        self.check("\\[ExponentialE]", "ⅇ")
+        self.check("\\[ImaginaryI]", "ⅈ")
+        self.check("\\[ImaginaryJ]", "ⅉ")
```
rocky, line 98: How do we know the system that this is run on supports unicode?

rocky:
> @rocky locally this passes, I believe a new version of Mathics Scanner needs to be released so this passes. Picking the Git egg version of Mathics Scanner isn't working.

Not convinced about that looking at the logs in the actions. It seems equally likely that the tests are saying that under certain situations people will have problems.

TiagoCavalcante:
@rocky yes, I guess they're related to the way Mathics Scanner is getting installed, I'll try installing through `git pull && make`.

rocky:
> @rocky yes, I guess they're related to the way Mathics Scanner is getting installed, I'll try installing through `git pull && make`.

If that works, then there is probably a Makefile target either in mathics-core or mathics-scanner that will build tables. 

(Alternatively, look for something like that)

PR #23, opened 2021-09-28
Assignment refactory - Stage 1

mmatera:
@rocky, as I promised, I started to reorganize `_SetOperator.assign_elementary`. Still, there is more work to do on this, but at least the basic cases were split. One question I have right now about how to continue is if we should put all the `process_assign_*` functions in a separate module, or if they should be member functions of `_SetOperator`. 

Also, `assign` should be reorganized in a further PR.

### DIFF
--- a/mathics/builtin/assignment.py
+++ b/mathics/builtin/assignment.py
@@ -26,7 +26,7 @@ from mathics.core.systemsymbols import (
 from mathics.core.atoms import String
 
 from mathics.core.definitions import PyMathicsLoadException
-from mathics.builtin.lists import walk_parts
+from mathics.algorithm.parts import walk_parts
 from mathics.core.evaluation import MAX_RECURSION_DEPTH, set_python_recursion_limit
 
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+indent_style = tab
+indent_size = 4
+
+[*.yml]
+indent_style = space
+indent_size = 2
+
+[*.py]
+indent_style = space
+indent_size = 4
--- a/.github/workflows/ubuntu-cython.yml
+++ b/.github/workflows/ubuntu-cython.yml
@@ -1,4 +1,4 @@
-name: Mathics (OSX)
+name: Mathics (ubuntu full with Cython)
 
 on:
   push:
@@ -8,13 +8,10 @@ on:
 
 jobs:
   build:
-    env:
-      NO_CYTHON: 1
-    runs-on: macos-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
-        os: [macOS]
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.9]
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
@@ -23,12 +20,11 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
-        brew install llvm@11
+        sudo apt-get update -qq && sudo apt-get install -qq liblapack-dev llvm-dev
         python -m pip install --upgrade pip
-        LLVM_CONFIG=/usr/local/Cellar/llvm@11/11.1.0/bin/llvm-config pip install llvmlite
     - name: Install Mathics with full dependencies
       run: |
-        make develop-full
+        make develop-full-cython
     - name: Test Mathics
       run: |
         make -j3 check
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -8,6 +8,8 @@ on:
 
 jobs:
   build:
+    env:
+      NO_CYTHON: 1
     runs-on: ubuntu-20.04
     strategy:
       matrix:
@@ -27,5 +29,4 @@ jobs:
         make develop-full
     - name: Test Mathics
       run: |
-        pip install -r requirements-dev.txt
         make -j3 check
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -8,6 +8,8 @@ on:
 
 jobs:
   build:
+    env:
+      NO_CYTHON: 1
     runs-on: windows-latest
     strategy:
       matrix:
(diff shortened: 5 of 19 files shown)


mathics/builtin/assignment.py
```diff
@@ -65,376 +74,577 @@ def get_symbol_list(list, error_callback):
...
+    focus = lhs.leaves[0]
+    tags = process_tags_and_upset_dont_allow_custom(
+        tags, upset, self, lhs, focus, evaluation
+    )
+    lhs, rhs = process_rhs_conditions(lhs, rhs, condition, evaluation)
+    rule = Rule(lhs, rhs)
+    for tag in tags:
+        if rejected_because_protected(self, lhs, tag, evaluation):
+            continue
+        count += 1
```
TiagoCavalcante, line 307: I believe this isn't necessary. We could instead return `len(tags) > 0`.

mathics/builtin/assignment.py
```diff
@@ -65,376 +74,577 @@ def get_symbol_list(list, error_callback):
...
+        defs.add_format(tag, rule, form)
+    return count > 0
```
TiagoCavalcante, line 345: Also, `len(tags) > 0`.

mathics/builtin/assignment.py
```diff
@@ -65,376 +74,577 @@ def get_symbol_list(list, error_callback):
...
+
+def process_assign_minprecision(self, lhs, rhs, evaluation, tags, upset):
+    lhs_name = lhs.get_name()
+    rhs_int_value = rhs.get_int_value()
+    # $MinPrecision = Infinity is not allowed
+    if rhs_int_value is not None and rhs_int_value >= 0:
+        max_prec = evaluation.definitions.get_config_value("$MaxPrecision")
+        if max_prec is not None and max_prec < rhs_int_value:
+            evaluation.message("$MinPrecision", "preccon", Symbol("$MinPrecision"))
+            raise AssignmentException(lhs, None)
```
TiagoCavalcante, line 454: I'm not sure, but I think that instead of `evaluation.message` and `AssignmentException` we could use `evaluation.error`.

mathics/builtin/assignment.py
```diff
@@ -65,376 +74,577 @@ def get_symbol_list(list, error_callback):
...
+
+    if len(lhs.leaves) not in (1, 2):
+        evaluation.message_args("N", len(lhs.leaves), 1, 2)
+        raise AssignmentException(lhs, None)
+    if len(lhs.leaves) == 1:
+        nprec = SymbolMachinePrecision
+    else:
+        nprec = lhs.leaves[1]
+    focus = lhs.leaves[0]
+    lhs = Expression(SymbolN, focus, nprec)
```
TiagoCavalcante, line 549: Maybe `apply_N` could be used here?

TiagoCavalcante (commented):
A great refactor!

mathics/builtin/assignment.py
mmatera: @TiagoCavalcante, that does not work: it is possible that all the tags are rejected because of symbol protection...

mathics/builtin/assignment.py
mmatera: same as before...

mathics/builtin/assignment.py
mmatera: instead of count, we can just set a flag to True, I do not know what is the best.

mathics/builtin/assignment.py
mmatera: Maybe. I just try to keep the logic of the code as similar as possible to the original in this stage. We can change that later.

mathics/builtin/assignment.py
mmatera: Nop: In this case, we need to keep the `N` unevaluated, because `apply_elementary` could be called both from `Set` as from `DelayedSet`.

mathics/builtin/assignment.py
rocky: When we are at the level of haggling over to use a boolean or an integer count, I suspect we are focusing on stuff that isn't going to matter that much. 

Yes, one is faster than the other, but not by that much to make a difference. Of course if you want to benchmark this and you come up with more than a one or two percent difference - great! Prove me wrong!

mmatera: @rocky, I do not say it for performance, but for clarity: I had to spend some time trying to figure out why we need a number, which in the end is not used.

TiagoCavalcante: @mmatera @rocky I believe here a boolean is better.

mathics/builtin/assignment.py
TiagoCavalcante: Ok.

mathics/builtin/assignment.py
rocky: > @rocky, I do not say it for performance, but for clarity: I had to spend some time trying to figure out why we need the number of modified tags, which in the end is not used. And for the @TiagoCavalcante`s comment, it happened the same to him.

Good - thanks for the clarification. I greatly appreciate the effort spent to make things more understandable. Thanks.

mmatera:
If there are no more comments over this PR, I will merge it.

rocky:
> If there are no more comments over this PR, I will merge it.

Give me one day to try to look over. If I don't look at it by tomorrow just merge.

mmatera:
> > If there are no more comments over this PR, I will merge it.
> 
> Give me one day to try to look over. If I don't look at it by tomorrow just merge.

If you want, I can leave this until the weekend.

rocky:
That'd be awesome - thanks!

rocky:
> LGTM - Thanks for giving me a chance to try to understand this better.
> 
> Here is something I'd like to see in a follow on.. This file is over 2K lines of code which always starts me thinking about things.
> 
> If you look at https://reference.wolfram.com/language/guide/Assignments.html you will see how this general topic is broken up, and nowaday we'd create a new directory called "assignments" and in that add submodules "set", "increment", "forms", and "clearing".
> 
> Also in that I would separate the internal module code (get_symbollist, find_and_tag_code) from the top-level builtins.
> 
> There is a lot of talk about what to Cythonize and what to target for speedup. Generally the top-level classes don't need this while those helper routines are more likely to do so.
> 
> So if we want to speed up _effective_ Cythonization, then a prerequisite would be to sort these out into different modules. However, my main motivation would be fore clarity and ease of understanding.

@mmatera If this is not on your roadmap, I'd like to split this up over this weekend.

mmatera:
@rocky, it is, but this week I was too busy to look at it. If you have the time and the will, go ahead.

PR #27, opened 2021-10-03
Lists overhaul

mmatera:
* move all the "algorithmic" routines in `mathics.core.lists` to an independent module in `mathics.algorithm.parts`
* cithonize `mathics.algorithm.parts`
* fix all the warnings produced by flake8
* Move `Exception` classes from `mathics.builtin.base` to `mathics.builtin.exceptions` to avoid circular references.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -81,6 +81,7 @@ else:
     EXTENSIONS_DICT = {
         "core": ("expression", "numbers", "rules", "pattern"),
         "builtin": ["arithmetic", "numeric", "patterns", "graphics"],
+        "algorithm": ["parts"],
     }
     EXTENSIONS = [
         Extension(
--- a/setup.py
+++ b/setup.py
@@ -81,7 +81,6 @@ else:
     EXTENSIONS_DICT = {
         "core": ("expression", "numbers", "rules", "pattern"),
         "builtin": ["arithmetic", "numeric", "patterns", "graphics"],
-        "algorithm": ["parts"],
     }
     EXTENSIONS = [
         Extension(
--- a/mathics/algorithm/parts.py
+++ b/mathics/algorithm/parts.py
@@ -1,0 +1,609 @@
+# -*- coding: utf-8 -*-
+
+"""
+Algorithms to access and manipulate elements in nested lists / expressions
+"""
+
+
+from mathics.core.expression import Expression
+from mathics.core.symbols import Symbol
+from mathics.core.atoms import Integer, from_python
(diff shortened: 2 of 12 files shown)


setup.py
```diff
@@ -81,6 +81,7 @@ def read(*rnames):
     EXTENSIONS_DICT = {
         "core": ("expression", "numbers", "rules", "pattern"),
         "builtin": ["arithmetic", "numeric", "patterns", "graphics"],
+        "algorithm": ["parts"],
```
rocky, line 84: I am a little concerned about the eagerness to use Cython instead of understanding exactly what's wrong and fixing it. 

As you have noted yourself, premature cython slows down development. 

Therefore independent of this I have in mind that we should add an environment variable to setup which is tested to determine whether we want Cython at all.

mmatera: I put in Cython this just to check that it would be possible. I can remove it now.

rocky: I realize what I write may come across as a bit complaining. To be clear, this isn't wrong. It is just somewhat annoying to me because it causes endless hours in pain in development for maybe only a few seconds speedup where what we really need is minutes of speedup.

It  goes in a direction that isn't as helpful as it might be intended to be if we were instead focused on the more the bigger performance gains.

mmatera: Done.

rocky: Thanks for your tolerance, understanding, and patience. Again, I will try to put in a PR so that we can develop in a way that doesn't require cython when it isn't wanted.

mmatera:
Once this PR be merged, the next step is to review the algorithm in `walk_parts`, in order to avoid the `set_position` / `position` juggling and make the code easier to understand and maintain.

rocky:
@mmatera so far this looks like a great improvement along making the code more modular and understandable so parts of it can be improved. 

Now that we have way to control running cython in development (and possibly in benchmarking) I am less concerned about the annoyance about caused by running Cython over everything. To my mind though there  still is the conceptual or philosophical dissatisfaction in the approach: when we don't understand something, rather than pick it apart and understand it, we take a shotgun approach which, as the analogy implies, has a bigger (and sometimes negative) effect on the surroundings than just the target you are shooting for. If you are going to use a shotgun then it would be nice to put up shields around that non-targarts first, no not afterwards.

mmatera:
@rocky, I have removed the cythonizing for the new module. I put it in the first place to check that cythonize it does not introduce problems afterward. In any case, the current implementation has a few problems, so we need to look at this again.

PR #28, opened 2021-10-03
Add a traced version of rules.py ...

rocky:
which tracks Builtins called and elapsed time in those routines.

### DIFF
--- a/mathics/core/rules.py
+++ b/mathics/core/rules.py
@@ -2,6 +2,7 @@
 # cython: language_level=3
 # -*- coding: utf-8 -*-
 
+import time
 
 from mathics.core.expression import Expression
 from mathics.core.symbols import strip_context, KeyComparable
@@ -10,6 +11,8 @@ from mathics.core.util import function_arguments
 
 from itertools import chain
 
+from collections import defaultdict
+function_stats = defaultdict(lambda: {"count": 0, "elapsed_microseconds": 0.0})
 
 class StopGenerator_BaseRule(StopGenerator):
     pass
@@ -126,10 +129,19 @@ class BuiltinRule(BaseRule):
         vars_noctx = dict(((strip_context(s), vars[s]) for s in vars))
         if self.pass_expression:
             vars_noctx["expression"] = expression
+        builtin_name = self.function.__qualname__.split(".")[0]
+        stat = function_stats[builtin_name]
+        ts = time.time()
+
+        stat["count"] += 1
         if options:
-            return self.function(evaluation=evaluation, options=options, **vars_noctx)
+            result = self.function(evaluation=evaluation, options=options, **vars_noctx)
         else:
-            return self.function(evaluation=evaluation, **vars_noctx)
+            result = self.function(evaluation=evaluation, **vars_noctx)
+        te = time.time()
+        elapsed = int((te - ts) * 1000)
+        stat["elapsed_microseconds"] += elapsed
+        return result
 
     def __repr__(self) -> str:
         return "<BuiltinRule: %s -> %s>" % (self.pattern, self.function)
@@ -147,3 +159,13 @@ class BuiltinRule(BaseRule):
         cls, name = dict["function_"]
 
         self.function = getattr(builtins[cls], name)
+
+def dump_tracing_stats():
+    for key_field in ("count", "elapsed_microseconds"):
+        print("count msecs  Builtin Name")
+        for name, statistic in sorted(function_stats.items(), key=lambda tup: tup[1][key_field], reverse=True):
+            print("%5d %6g %s" % (statistic["count"], statistic["elapsed_microseconds"], name))
+        print("")
+
+import atexit
+atexit.register(dump_tracing_stats)


rocky:
@mmatera @TiagoCavalcante 

Here is a new toy to play with. This code replaces the routine that calls apply so that we can count the number of times a built-in function is called and the time spent in that call.

Note that the time spent in a call like `Get` includes the time it spends in calling other builtins. 

To use just go into a `mathics` session or run a test (pytest, or doctest) and on exist statistics will appear. 

Attached are some timings for running:

* `mathics`  followed by an exit.
* `mathics` followed by `Needs["DiscreteMath`CombinatoricaV0.9`"]`
* `pytest -s test/test_compare.py`  (Note the `-s` option is important to see the trace)
[time-mathics.txt](https://github.com/Mathics3/mathics-core/files/7273941/time-mathics.txt)
[time-combinatorica-load.txt](https://github.com/Mathics3/mathics-core/files/7273942/time-combinatorica-load.txt)
[time-test-compare.txt](https://github.com/Mathics3/mathics-core/files/7273943/time-test-compare.txt)

What I like about all of this

* How easy it is to use
* How it is very targeted: you could stay in a mathics session for hours and the time you spend thinking isn't tracked until you do some evaluation
* How general it is: it works on in a mathics session, in pytest, doctest and probably even mathics Django

What remains to be done is to have this off to the side and figure out a way to monkey patch this in when it is desred.

I suspect that leaving it in all the time doesn't impact performance that much since the tracing code is pretty fast - even without running this through cython ;-) . 

However suspect it wouldn't be hard to monkey patch this in when desired. 

Ok - I'll stop here.  I am curious to learn what insights and information you all get out of this.

rocky:
Here is a doctest run. (I attach this with a little trepidation because I don't think doctest is the best or right kind of benchmark. But right now we haven't spent the time to create a good benchmark suite.)

[time-doctest.txt](https://github.com/Mathics3/mathics-core/files/7274003/time-doctest.txt)

rocky:
One other thing that might be done on option: we glom together all `apply()` methods for a given builtin. We could also have an option to split this out. 

More generally, in addition to an option for doing this individually or not, there might be an option to specify whether sorting is shown both for counts and elapsed microseconds, or just one or the other. 

Another direction to expand this is to do basically the same thing for evaluation time minus the time spent in the specific builtins.

rocky:
Two more todo ideas. Add two builtin functions "DumpStats[Options]" and "ClearStats[]" which dump statistics and clear out the statistics data accumulated.

mmatera:
> Two more todo ideas. Add two builtin functions "DumpStats[Options]" and "ClearStats[]" which dump statistics and clear out the statistics data accumulated.

I think that would be the best: defining a symbol that activates/deactivates tracing (let's say, `StartTracing[]/StopTracing[]`), and adding some builtins to dump the stats.

TiagoCavalcante (approved):

TiagoCavalcante:
@rocky thanks for doing this!

@raulcd

TiagoCavalcante:
@rocky thanks for doing this! Looks great!

@rocky @mmatera about keeping this in code, I think this shouldn't be there, as that don't adds anything for the end user. But yes, this would be easier to use this way.

mathics/core/rules.py
```diff
@@ -126,10 +131,19 @@ def do_replace(self, expression, vars, options, evaluation):
...
+
+        stat["count"] += 1
         if options:
-            return self.function(evaluation=evaluation, options=options, **vars_noctx)
+            result = self.function(evaluation=evaluation, options=options, **vars_noctx)
         else:
-            return self.function(evaluation=evaluation, **vars_noctx)
+            result = self.function(evaluation=evaluation, **vars_noctx)
+        te = time.time()
+        elapsed = int((te - ts) * 1000)
```
TiagoCavalcante, line 144: @rocky I'm just in doubt of the unit of seconds, look at [time](https://docs.python.org/3/library/time.html#time.time).

TiagoCavalcante:
@rocky one thing I don't understand: why 2 lists?

mmatera:
> @rocky thanks for doing this! Looks great!
> 
> @rocky @mmatera about keeping this in code, I think this shouldn't be there, as that doesn't add anything for the non-developer users. But yes, this would be easier to use this way.

Actually, WMA has something similar: (See: https://reference.wolfram.com/language/ref/Trace.html)

TiagoCavalcante:
Oh ok, let's keep it there.

mathics/core/rules.py
rocky: The subtraction gives the time sections. The granularity is in milliseconds which is given in a floating point number less than zero. So we multiply by 1000 to get it to be an integer number of milliseconds.

rocky:
> @rocky one thing I don't understand: why 2 lists?

One sorts by counts and the other sorts by elapsed time. When there are options given say in a dump routine you could specify either. 

Another option I just thought of and would like to have is to extend the code, is to somehow remove the accumulated time from builtins that were called underneath. That would reduce the time elapsed milliseconds of `Get`, `Needs` and `Quiet`.

rocky:
@TiagoCavalcante this is all the time I will have for this for about the next week. If you are interested feel free to extend it and finish it. 

Something that might be interesting to check is to see exactly how much overhead it adds when it is not used. 

Other activities are to start getting insights of the most used builtins to which need to be improved and then improve them.

TiagoCavalcante:
> Other activities are to start getting insights of the most used builtins to which need to be improved and then improve them.

@rocky I'm looking at them, but as you already said, isolated changes won't bring many benefits. Anyway, a 5% is still good.

rocky:
> Another option I just thought of and would like to have is to extend the code, is to somehow remove the accumulated time from builtins that were called underneath. That would reduce the time elapsed milliseconds of `Get`, `Needs` and `Quiet`.

Here is how you could implement this. Add a global counter that is incremented inside that `do_replace()` routine that has the other instrumentation. The do_replace method saves the old value of the counter before incrementing it. After the function returns it checks to see if the value has increased. If so, it  doesn't add to elapsed_microseconds

This of course makes its elapsed_microsecond not so useful for those "composite" builtins. But that's okay because you can get that without the option. And at first I think we want identify and focus on the primitive at the bottom.

Remember (I wish I didn't feel like it were necessary to mention this), we are not trying to solve all 32 or more classes of things that could be construed as classes of performance problems. Instead we should focus on one, fix that and then move on to the next one. Here, fixing performance of functions that are composites - that call other more primitive functions will come after fixing the performance of the primitive functions.

TiagoCavalcante:
@rocky about the unit of seconds:
```py
>>> x = time.time(); time.sleep(1); print(time.time() - x)
1.0014801025390625
```

Anyway, 1000 microseconds = 1 millisecond.

rocky:
> @rocky about the unit of seconds:
> 
> ```python
> >>> x = time.time(); time.sleep(1); print(time.time() - x)
> 1.0014801025390625
> ```

Right we want the .001 to be 1 unit. 

> 
> Anyway, 1000 microseconds = 1 millisecond.

Yes, I stand corrected. Please fix that in the next commit whenever/whoever does that.

TiagoCavalcante:
@rocky @mmatera what do you think of a `$Trace` which can be set to True and False, and instead of printing on exit, print between `In[]` and `Out[]` like a message?

rocky:
@mmatera is the expert on WL-compatible language issues.

TiagoCavalcante:
@mmatera @rocky I created the `Trace` builtin, though that's a bit different from WL's one.

TiagoCavalcante:
> The original code allowed sorting by elapsed time and count. Maybe an option can be added to add back this flexibility?

Doing it now.

> "custom" is a bit vague. How do you feel about something like "traced_do_replace?

Yes, makes sense.

rocky:
> > Note that atexit would be the preferred thing to use if when we add a --trace option to the CLI mathics and mathicsscript.
> 
> I think just one of the 2 (TraceOn/TraceOff or --trace) is fine. (I'd go with the 1st)

My preference is both since I can use the `--execute` option with `--trace` on a command line and get interesting stuff from batch. 

We don't have to do everything in one PR and probably shouldn't. On the other hand, the reason I mention all of this was just to get an ack.  

I can handle the `--trace` option from the CLI's in another PR.

PR #29, opened 2021-10-03
Allow a developer to overide to running Cython

rocky:
Set environment variable NO_CYTHON before running setup.py to avoid
running Cython even if it is installed.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+indent_style = tab
+indent_size = 4
+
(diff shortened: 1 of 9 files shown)


mmatera:
Excellent. Completely in favor! Can we also set the "No Cython" for git actions?

rocky:
> Excellent. Completely in favor! Can we also set the "No Cython" for git actions?

Sure.

TiagoCavalcante (approved):
LGTM

TiagoCavalcante:
@mmatera if Cython is not installed, it is not used, but I don't know if the Python workflow install that by default (but I guess that isn't installed).

PR #30, opened 2021-10-04
Small tweaks

TiagoCavalcante:
With this I want to use smaller classes as parent to other and remove the use of NumPy constants as they're 30-50% slower than their MPMath equivalent (which is also more precise).


mathics/builtin/numbers/constants.py
```diff
@@ -307,15 +265,15 @@ def apply_N(self, precision, evaluation, options={}):
...
+class E(_MPMathConstant, _SympyConstant):
     """
     <dl>
     <dt>'E'</dt>
         <dd>is the constant \u2107 with numerical value \u2243 2.71828.
     </dl>
 
     >> N[E]
-     = 2.71828
+     = 2.71828182845905
```
rocky, line 276: See above. This is doesn't follow the semantics of WL.

TiagoCavalcante: @rocky should I revert it?

rocky: I believe so.  When using mpmath we need to figure out how to make it match https://reference.wolfram.com/language/guide/PrecisionAndAccuracyControl.html 

Somewhere in old tickets this has come up before. @mmatera probably knows more about this than I do. 

Writing some sort of compatibility layer on top of mpmath in a lower-level language is one areas where the use of Cython, C, Rust, etc. may be  justified. 

However you should think of that akin to a project of its own like graphics3d (although probably not as extensive or as hard.)

TiagoCavalcante: > Writing some sort of compatibility layer on top of mpmath in a lower-level language is one areas where the use of Cython, C, Rust, etc. may be justified.

I guess Python is the best here, mpmath is written in Python.

TiagoCavalcante: > I believe so.

Ok. I'll remove it.

PR #31, opened 2021-10-04
Only sets build_ext to Cython when NO_CYTHON isn't set

TiagoCavalcante:


rocky:
This is just a formatting change, right? 

```
$ black setup.py 
All done! ✨ 🍰 ✨
1 file left unchanged.
```

TiagoCavalcante:
@rocky no, I changed the indentation so it is inside the "if NO_CYTHON isn't set".

rocky:
> @rocky no, I changed the indentation so it is inside the "if NO_CYTHON isn't set".

This then is a war of editors. My editor, GNU Emacs has an "comment region" and this is what it does. When that is uncommented it is indented exactly the way it should be for being inside the if.

rocky (approved):
LGTM - thanks for finding and fixing my buggy code.

PR #35, opened 2021-10-10
Allow white 3d points

TiagoCavalcante:
This replaces a temporary fix in the JSON function of the 3d Point
by a function-replacement in Point3DBox.
The temporary fix was converting the color from white to black. This was
done because the default color is white, and it should be black for the
points.


mmatera:
LGTM

mmatera (approved):

rocky (approved):
LGTM - sorry for the delay.

PR #37, opened 2021-10-11
adding test_parts unittest

mmatera:
This is related to #36. Here I put all the tests currently spread along doctests that are related to the part I would like to change.

### DIFF
--- a/test/test_parts.py
+++ b/test/test_parts.py
@@ -123,36 +123,87 @@ def test_read_access():
             "",
         ],
     ):
+        print("\n\n", str_expr, "\n", str_expected, "\n", msg)
         check_evaluation(
             str_expr, str_expected, msg, to_string_expr=False, to_string_expected=False
         )
 
 
+def test_weird():
+    for str_expr, str_expected, msg in (
+        # When an element is assigned, the element is what change, not the variable.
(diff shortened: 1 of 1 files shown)


rocky:
One other thing that has been in the back of my mind. When we talk about how pattern matching is dreadfully slow and all, I would not call this part of pattern matching. I would say that this code is doing expression traversal and substitution which is part of its evaluation process. Pattern matching in the evaluation process is about finding which apply methods to call. 

When I talk about vagueness and jumping around it is this kind of thing I mean. 

I think it right to focus on this and work on it. But the vagueness hasn't been helpful. 

The way I see things is we start to benchmark stuff, we find out that SetDelayed is taking 10 times longer in 1/10 the number of calls of MPMathFunction. So when we drill down into why SetDelayed is slow, we think that the expression traversal and replacement is either causing slowness or memory consumption. 

@mmatera Is this correct or is there something else going on?

mmatera:
> The way I see things is we start to benchmark stuff, we find out that SetDelayed is taking 10 times longer in 1/10 the number of calls of MPMathFunction. So when we drill down into why SetDelayed is slow, we think that the expression traversal and replacement is either causing slowness or memory consumption.

Sorry, I skipped this part in my previous answer, but yes, as I was saying before, I would blame the expression transversal and replacement, which is essentially repeating building strings to use them as keys for something that we could perfectly know where it was from the very beginig.

PR #38, opened 2021-10-11
Improve the looking of PrintTrace's docs

TiagoCavalcante:
The "\|" was shown in the Django documentation, so these tests were removed.


TiagoCavalcante:
@rocky now I see I didn't commit the removal of \|.

PR #40, opened 2021-10-12
defining SystemContextPath to make  MathicsSession(add_builtin=False)…

mmatera:
Without this, `MathicsSession(add_builtin=False)` is unable to do any evaluation.

### DIFF
--- a/mathics/core/definitions.py
+++ b/mathics/core/definitions.py
@@ -123,6 +123,29 @@ class Definitions(object):
             self.builtin.update(self.user)
             self.user = {}
             self.clear_cache()
+        else:
+            from mathics.core.rules import Rule
+            from mathics.core.atoms import String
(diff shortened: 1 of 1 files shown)


TiagoCavalcante:
@mmatera I think walk_parts changes should be in another PR, so maybe re-open this?

mmatera:
> @mmatera I think walk_parts changes should be in another PR, so maybe re-open this?

No, I  think I pick the wrong PR in my previous comment, sorry. 
Now, in master, we use the proxy variables to `Context` and `ContextsPath` which points to `Global` in case you do not load any builtin symbol. So, if you run
```
session=MathicsSession(add_builtin=False)
session.evaluate("F[a]")
```
you do not get an exception, but just `Expression(Symbol("F"), Symbol("a"))`

PR #43, opened 2021-10-17
Start to split out assignments into its own module

rocky:
This just breaks out what was one huge module into several along the organization that can be found in https://reference.wolfram.com/language/guide/VariablesAndFunctions.html or the [Mathematica 5 book](https://deptche.ccu.edu.tw/Chemistry/Chem_Math/Mathematica_V5_Book.pdf). 

The individual commits are organized in a reasonable way so that you can follow the progression of how things were split off piece by piece. I hope this makes it easier to follow, and I hope in the future we might think about splitting large changes in such a form. It also made it easier to work on. (That said, I admit this was easier to do because most of this is of a procedural and mechanical kind of activity. So it may be harder for other kinds of things. But it is worth thinking about how to organize sequence of commits to try to make them easier handle).

@mmatera I had a hard time understanding in the comments where major sections ended. in `mathics.builtin.assigment`.  There are comments at the beginning but not at the end. So without a lot of thought, I can't even tell if the sections are supposed to be nested either. 

We don't have to continue this kind of "tough love" coding - we can split up `builtins.assignments.internal` into more modules which make it clear where sections end and whether one section is nested in another. 

The other thing I should mention is that, it appears that we have an number of functions in the module `mathics.builtin.assignments.types` that don't appear anymore in Mathematica. Some were in Mathimatica 5 (under `Types of Values` which is where I got the section name and grouping of functions,  but `SubValues` isn't even in that. 

In general the functions listed in Mathematica Guides isn't unique: a function can be covered by more than one guide. So some of the organization is a bit arbitrary as to which section it appears in. And in the case of `LoadModule` which is a Mathics-specific function, it is a bit arbitrary under what section we have it. I suppose we could add a Mathics-specific section for all functions like this. Builtin-in Tracing Functions might appear here as well. What do you think though?

### DIFF
--- a/mathics/builtin/assignments/assignment.py
+++ b/mathics/builtin/assignments/assignment.py
@@ -23,6 +23,7 @@ class Set(BinaryOperator, _SetOperator):
     """
     <dl>
       <dt>'Set[$expr$, $value$]'
+
       <dt>$expr$ = $value$
       <dd>evaluates $value$ and assigns it to $expr$.
 
@@ -167,10 +168,10 @@ class SetDelayed(Set):
 class TagSet(Builtin, _SetOperator):
     """
     <dl>
-    <dt>'TagSet[$f$, $expr$, $value$]'
-    <dt>'$f$ /: $expr$ = $value$'
-        <dd>assigns $value$ to $expr$, associating the corresponding
-        rule with the symbol $f$.
+      <dt>'TagSet[$f$, $expr$, $value$]'
+
+      <dt>'$f$ /: $expr$ = $value$'
+      <dd>assigns $value$ to $expr$, associating the corresponding assignment with the symbol $f$.
     </dl>
 
     Create an upvalue without using 'UpSet':
@@ -196,6 +197,7 @@ class TagSet(Builtin, _SetOperator):
     messages = {
         "tagnfd": "Tag `1` not found or too deep for an assigned rule.",
     }
+    summary_text = "assigns a value to an expression, associating the corresponding assignment with the a symbol."
 
     def apply(self, f, lhs, rhs, evaluation):
         "f_ /: lhs_ = rhs_"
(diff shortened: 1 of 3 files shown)


mmatera:
LGTM

mmatera:
> @mmatera I had a hard time understanding in the comments where major sections ended. in `mathics.builtin.assigment`. There are comments at the beginning but not at the end. So without a lot of thought, I can't even tell if the sections are supposed to be nested either.

Sorry about that. I didn't realize that the end of the   `_SetOperator` was not clear enough. What I did es just to split the kilometric method in subfunctions, and comment what they do. But I think we can improve it as we iterate.

TiagoCavalcante:
> In general the functions listed in Mathematica Guides isn't unique: a function can be covered by more than one guide. So some of the organization is a bit arbitrary as to which section it appears in. And in the case of LoadModule which is a Mathics-specific function, it is a bit arbitrary under what section we have it. I suppose we could add a Mathics-specific section for all functions like this. Builtin-in Tracing Functions might appear here as well. What do you think though?

This is intuitive for people coming from WL, but I guess that isn't for people new to Mathics/WL.

> I hope this makes it easier to follow

Yes, it does.

Also, maybe add Python 3.10 to supported versions in README?

Last but not least: LGTM and thanks.

rocky:
> > In general the functions listed in Mathematica Guides isn't unique: a function can be covered by more than one guide. So some of the organization is a bit arbitrary as to which section it appears in. And in the case of LoadModule which is a Mathics-specific function, it is a bit arbitrary under what section we have it. I suppose we could add a Mathics-specific section for all functions like this. Builtin-in Tracing Functions might appear here as well. What do you think though?
> 
> This is intuitive for people coming from WL, but I guess that isn't for people new to Mathics/WL.
> 
> > I hope this makes it easier to follow
> 
> Yes, it does.

One of the things that I thought was cool about Mathematica probably even pre Mathematica 5 was how it was organized in a way that was made sense, grouped related mathematical functions together and described how they are related, where they are used and what they do.  This kind of thing unfortunately wasn't done in the older generations of symbolic mathematics like macsyma or IBM's version Scratchpad/ScratchPad2/Axiom. Or not done to the extent it was done in Mathematica.

Nowadays as I read the online docs, I personally feel that there is too much hype and advertising about how great and flexible and wonderful the stuff is without giving any details or examples of this awesomeness. (In other words, it is possibly bullshit.)

One of the things we can do, is go back and try put back some context of what is there, why and how to use. I am always looking for interesting examples of use of the functions. (BTW I see this is also a section in the Mathematica docs. I try to keep in mind that to use examples to be interesting and compelling examples when that is possible and isn't too cumbersome or require too much setup. 

Often if you compare Mathics' descriptions versus Mathematica's, I hope ours is clearer and informs better. (If, not let's change that to be so). 

So in this respect, I don't mind not following the Guide organization in those cases where we agree that some other organization is clearer.  Whether or not you understand Mathematica and WL, any clear organization is helpful. 

In summary, in my opinion start out with and are guided by Mathematica's more because it makes sense -- where we don't already have any. 
> 
> Also, maybe add Python 3.10 to supported versions in README?
> 

The badge come from repology. When the next mathics core releases happens and we emit something that works on 3.10, that badge will get updated automatically.

> Last but not least: LGTM and thanks.


Thanks for the kind works. But more important is thanks for all of the great you @TiagoCavalcante and @mmatera do to make things constantly do to make it better.

PR #44, opened 2021-10-19
Subexpression

mmatera:
@rocky, @TiagoCavalcante, this is an almost functional new version for `walk_parts`. According to my preliminary benchmarks, this version would be up to twice faster than the original (later I will add some specific tests and benchmarks).

Also, 
* The logic is (I hope) simpler and clearer than in the original code, and is based on objects instead of (nested) procedures.
* The behaviour of `walk_parts_new` is more consistent: in `walk_parts`, depending on the value passed to `assign`, it returns
the part (if `assign_list` is `None`) or the whole expression (if `assign_list` is not `None`). The new method always returns the referenced part.
* Does not require modifying the `Symbol` inside an expression, which helps to disentangle the code.

As it is now, it passes doctest but not the `combinatorica`  pytest. The problem seems to be that the new implementation does not work with the current implementation of caching expressions.  

Here is a minimal example of the problem:

This gives the expected behavior in one of the `test_combinatorica.py` tests that fail, but slightly rewritten to get the right behavior:
```
In[1]:= Needs["DiscreteMath`CombinatoricaV0.9`"]
Out[1]:= None

In[2]:= r=RotateLeft[Range[10],1]
Out[2]:= {2, 3, 4, 5, 6, 7, 8, 9, 10, 1}

In[3]:= ToCycles[r]
Out[3]:= {{2, 3, 4, 5, 6, 7, 8, 9, 10, 1}}

In[4]:= r
Out[4]:= {2, 3, 4, 5, 6, 7, 8, 9, 10, 1}
``` 

This is how is actually in the test, and the wrong behavior:
```
In[1]:= Needs["DiscreteMath`CombinatoricaV0.9`"]
Out[1]:= None

In[2]:= ToCycles[r=RotateLeft[Range[10],1]]
Out[2]:= {{2, 3, 4, 5, 6, 7, 8, 9, 10, 1}}

In[3]:= r
Out[3]:= {0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
``` 

For some reason, r is modified inside `ToCycles`. On the other hand, modifying `In[2]` in these ways does works:

* `ToCycles[r=RotateLeft[Range[10],1]; r]`
* `ToCycles[(r=RotateLeft[Range[10],1])[[All]]]`

So, the next step in this rabbit hole is to understand the caching mechanism...

In any case, I would like to know what do you think about the change in the framework.


mmatera:
Two more tests, now for "block" access:

# Old code
%timeit test_access_block_element()  -> 26 ms ± 669 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
%timeit test_update_block_element()  ->  95.7 ms ± 2.95 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

# New code


%timeit test_access_block_element() -> 36.6 ms ± 8.89 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
%timeit test_update_block_element() -> 35.2 ms ± 9.03 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)




```
def test_access_block_element():
    global global_idx, deep_array
    test1 = deep_array.copy()
    global_idx+=1 
    return walk_parts([test1],
               [Integer(1),
                Expression("System`Span", Integer(1),Integer(3),Integer(2)), 
                SymbolAll,
                Integer(2),
                Integer(1)],
               session.evaluation,
               None)
    

def test_update_block_element():
    global global_idx, deep_array
    test1 = deep_array.copy()
    global_idx+=1 
    return walk_parts([test1],
               [Integer(1),
                Expression("System`Span", Integer(1),Integer(3),Integer(2)), 
                SymbolAll,
                Integer(2),
                Integer(1)],
               session.evaluation,
               Integer(global_idx))
```

PR #51, opened 2021-10-25
rearrange from_python to first try if the argument is a BaseExpression

mmatera:
This is a much less drastic version of #49. This just makes that in `from_python`, the case of  Mathics expressions are tested as the first case.


TiagoCavalcante (approved):
LGTM

rocky:
LGTM

PR #52, opened 2021-10-25
Walk parts bench

mmatera:
@rocky, @TiagoCavalcante  here we go with a first (and very raw) unit test for walk_parts. When the framework gets clearer, maybe we can move this code to mathics-bench

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/52"></a>

### DIFF
--- a/mathics/core/systemsymbols.py
+++ b/mathics/core/systemsymbols.py
@@ -6,6 +6,7 @@ from mathics.core.symbols import PredefinedSymbol, Symbol
 Symbol_Context = Symbol("$Context")
 Symbol_ContextPath = Symbol("$ContextPath")
 SymbolAborted = PredefinedSymbol("$Aborted")
+SymbolAll = Symbol("All")
 SymbolAlternatives = Symbol("Alternatives")
 SymbolAssociation = Symbol("Association")
 SymbolAutomatic = Symbol("Automatic")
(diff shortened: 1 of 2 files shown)


TiagoCavalcante:
@mmatera should I move the tests to mathics-benchmark?

mmatera:
If you figure out where to put it, go ahead..

TiagoCavalcante:
@mmatera now mathics-benchmark supports Python code.

mmatera:
> @mmatera now mathics-benchmark supports Python code.

In that case, when I get some time I will see how to put this in mathics-benchmark. Let's keep this PR open until that, to avoid lose it.

mmatera:
@TiagoCavalcante, I rebase this to recover the relevant file. Maybe we can try just to copy to mathics-benchmark  to see how to integrate it.

TiagoCavalcante:
@mmatera the SymbolAll part can be merged, in mathics-benchmark using python is almost the same as using Mathics code, you just need to add this line in the benchmark file:
https://github.com/Mathics3/mathics-benchmark/blob/71bf69ff981978cb5bdb59bb9a486dd7d6cc9b94/benchmarks/example.yaml#L33

So yes, it is a kinda copy-paste.

mmatera:
Moved into mathics-benchmarks

PR #57, opened 2021-11-06
List fullform

mmatera:
This PR changes the behavior of MakeBoxes[List[___], FullForm] to make it closer to WMA. 
```
In[1] := FullForm[List[1,2,3]]
Out[1]:= {1,2,3}
```
This also helps to debug some code like the one in ``Series``, which can produce several nested lists.


TiagoCavalcante (approved):
LGTM - this makes sense.

PR #58, opened 2021-11-06
leaf.unevaluate -> unevaluated_leaves[leaf] in expression.evaluate_next

mmatera:
In `Expression.evaluate_next`, part of the algorithm requires to keep a register of what leaves were evaluated in certain stages. It was done by setting an "ad-hoc"  property `.unevaluate` in the Expression object. That was OK when `Symbol` was not a singleton. Now, two different leaves with the same symbol will have the same property, so this strategy becomes fragile. This PR is a first step for removing this fragility.


TiagoCavalcante (approved):
LGTM - security first!

mmatera:
As I explain in #122, this solution does not work :(

PR #60, opened 2021-11-09
Improve cylinders drawing

TiagoCavalcante:
Before the cylinders were always drawn around X+Y axis in Asymptote.
Now work as they do in mathics-threejs-backend, the cylinder is
drawn from the 1st point to the 2nd.
Special thanks to [Black Mild](https://tex.stackexchange.com/users/140722/black-mild)
by his [answer](https://tex.stackexchange.com/questions/615210/draw-cylinder-through-points/621249).

@rocky could you check if this is working properly?

### DIFF
--- a/mathics/format/asy.py
+++ b/mathics/format/asy.py
@@ -291,8 +291,6 @@ add_conversion_fn(Cuboid3DBox)
 def cylinder3dbox(self, **options) -> str:
     face_color = self.face_color.to_js()
 
-    # FIXME: currently always drawing around the axis X+Y
-    axes_point = (1, 1, 0)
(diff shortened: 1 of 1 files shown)


rocky:
> Before the cylinders were always drawn around X+Y axis in Asymptote. Now work as they do in mathics-threejs-backend, the cylinder is drawn from the 1st point to the 2nd. Special thanks to [Black Mild](https://tex.stackexchange.com/users/140722/black-mild) by his [answer](https://tex.stackexchange.com/questions/615210/draw-cylinder-through-points/621249).
> 
> @rocky could you check if this is working properly?

@TiagoCavalcante thanks - will try over the upcoming weekend.

PR #62, opened 2021-11-10
load mathics.builtin.box

mmatera:
This PR fixes some flakiness I found during the `Definitions` refactoring: `mathics.builtin.box.*` was not loaded during the startup.

### DIFF
--- a/mathics/builtin/__init__.py
+++ b/mathics/builtin/__init__.py
@@ -152,6 +152,7 @@ disable_file_module_names = (
 for subdir in (
     "arithfns",
     "assignments",
+    "box",
     "colors",
     "distance",
     "drawing",


mmatera:
@rocky, please confirm to me that this was a mistake and that we have no good reasons to do not to load these modules.

rocky:
> @rocky, please confirm to me that this was a mistake and that we have no good reasons to do not to load these modules.

I think all of those are imported as a result of importing other builtins via Python `import` statements. 

Because this is slow and non-standard, we use the dynamic loading only when there the modules wouldn't have otherwise have been imported by using Python's normal `import` mechanism.

mmatera:
> > @rocky, please confirm to me that this was a mistake and that we have no good reasons to do not to load these modules.
> 
> I think all of those are imported as a result of importing other builtins via Python `import` statements. Was there a particular Builtin that wasn't found?

Yes, all the built-in inside `mathics.builtin.box` are not loaded. I discovered this playing with the `Definitions` object. 

> 
> Because this is slow and non-standard, we use the dynamic loading only when there the modules wouldn't have otherwise have been imported by using Python's normal `import` mechanism.
> 
Indeed, at some point we need to thing about a better mechanism to load de built-in definitions. But it is a lot of work, and it just save a few seconds in the loading time.

> And that is yet another good reason why it is nice to separate top-level Builtins from the implementation and the boxing of those and the formatting - none of the things in the later category needs to be loaded in a non-standard way.

OK, but again, that is another step. My idea, in the long term, is to put all the complex formatting stuff in external modules (like `pymathics-asy` or `pymathics-matplotlib` and keep in the core the essential infrastructure for evaluation and rule application. In any case, we still have a long trip until there.

PR #63, opened 2021-11-11
ShowSteps

mmatera:
In this PR, a new ``System`ShowSteps`` symbol is introduced. If its value is set to `True`, all the steps in the evaluation are printed.
This could be helpful in order to understand how to optimize evaluation rules.


mathics/builtin/inout.py
```diff
@@ -65,6 +66,20 @@
...
 
 
+class ShowSteps(Predefined):
+    """
+    If this variable is set to `True`, the steps in the evaluation are shown.
+    """
+
+    name = "ShowSteps"
+    attributes = {
+        "Unprotected",
```
TiagoCavalcante, line 76: @mmatera, instead of letting it unprotected, is better to create a rule for setting this True or False (this is how $TraceBuiltins work).

mathics/builtin/specialfns/bessel.py
```diff
@@ -667,7 +667,9 @@ class SphericalBesselJ(_Bessel):
 
     attributes = ("Listable", "NumericFunction", "Protected")
 
-    rules = {"SphericalBesselJ[n_, z_]": "Sqrt[Pi / 2] / Sqrt[z] BesselJ[n + 0.5, z]"}
+    rules = {
+        "SphericalBesselJ[n_, z_]": "Times[Rational[1, 2], Power[Pi, Rational[1, 2]], Power[2, Rational[1, 2]], BesselJ[Rational[3, 2], z], Power[Power[z, -1], Rational[1, 2]]]"
```
TiagoCavalcante, line 671: @mmatera I think that adding the new feature shouldn't change this.

mathics/core/expression.py
```diff
@@ -734,6 +734,8 @@ def evaluate(self, evaluation) -> typing.Union["Expression", "Symbol"]:
 
         old_options = evaluation.options
         evaluation.inc_recursion_depth()
+        if evaluation.show_steps:
+            print("\t Evaluating: ", expr)
```
TiagoCavalcante, line 738: @mmatera this works only with Mathics CLI and mathics-script, not Mathics Django. I think it is better to use `evaluation.message`.

mathics/builtin/inout.py
```diff
@@ -65,6 +66,20 @@
 SymbolSubscriptBox = Symbol("SubscriptBox")
 
 
+class ShowSteps(Predefined):
+    """
+    If this variable is set to `True`, the steps in the evaluation are shown.
+    """
+
+    name = "ShowSteps"
```
TiagoCavalcante, line 74: I'm not sure what is the naming convention, but I feel like this should be `$ShowSteps`, and there should be also the function `ShowSteps` (like `TraceBuiltins`).

TiagoCavalcante (commented):
@mmatera if you want, I can continue this.

TiagoCavalcante (commented):
@mmatera if you want, I can continue this.

mathics/builtin/specialfns/bessel.py
mmatera: Yep, this should be in another PR

mathics/builtin/inout.py
mmatera: I thought about this as a configuration variable, but using an activating/deactivating function could also work.

mathics/core/expression.py
mmatera: My idea was to send it to a terminal, just for doing debugging tasks. But if you think it is useful in the graphics interface, go ahead...

mmatera:
> @mmatera if you want, I can continue this.

Yes, please, go ahead!

mathics/builtin/specialfns/bessel.py
rocky: I have removed it form this PR. 

If this kind of rule change is some sort of symbolic manipulation, we should consider implementing this in a more general way so that _whenever_ someone write one expression that rule is manipulated _automatically_ on entry.  One might consider adding a parameter to indicate whether automatic simplification (which I had started to spell symplification) should be applied.

rocky:
@mmatera @TiagoCavalcante I just had a chance to try this out locally. I like it!  I hope it is reworked in the way @TiagoCavalcante recommends.

rocky:
A feature for future improvement would be to use indentation to show nesting level: 

Instead of: 

```
In[5]:= Integrate[2 x + 3]
	 Evaluating:  System`Integrate[System`Plus[System`Times[2, Global`x], 3]]
	 Evaluating:  System`Integrate
	 Evaluating:  System`Plus[System`Times[2, Global`x], 3]
	 Evaluating:  System`Plus
	 Evaluating:  System`Times[2, Global`x]
	 Evaluating:  System`Times
	 Evaluating:  Global`x ...
```

```
In[5]:= Integrate[2 x + 3]
	 Evaluating:  System`Integrate[System`Plus[System`Times[2, Global`x], 3]]
	   Evaluating:  System`Integrate
	        Evaluating:  System`Plus[System`Times[2, Global`x], 3]
	          Evaluating:  System`Plus
	            Evaluating:  System`Times[2, Global`x]
	              Evaluating:  System`Times
	              Evaluating:  Global`x ...
```

TiagoCavalcante:
@mmatera this currently isn't working for me (neither your commit nor mine), gonna investigate it tomorrow, I'll push my changes by now.

rocky:
> @mmatera this currently isn't working for me (neither your commit nor mine), gonna investigate it tomorrow, I'll push my changes by now.

@TiagoCavalcante if you are referring to the CI failure that was a simple formatting check failure. 

There are a lot of things that you proposed which I see in this PR. The one that that isn't is using `evaluate.message()`  which is the wrong function and should be `evaluation.print_out()` instead I think.

I agree with you that since the output here is all things that a Mathics user has control over, it would be good to show it in output.

mmatera:
@TiagoCavalcante @rocky check this again when you have time.

mmatera:
imagen

rocky:
> Probably we want to use a specific print function instead `evaluation.print_out` to have the right indentation:
> 
> imagen

This gets back to a long standing problem we have in Forms.  There could be some sort of Code or Verbatim form.

I tried this in the CLI though, I and I love the indenting. And that the functional form. 

```In[1]:= ShowSteps[Binomial[6,1 + 2]]
    Evaluating: System`Binomial[6, System`Plus[1, 2]]
      Evaluating:  System`Binomial
      Evaluating: System`Plus[1, 2]
        Evaluating:  System`Plus
      -> 3
    -> 20
Out[1]= 20

In[2]:= ?$ShowSteps
enable or disable displaying the steps to get the result
Out[2]= Null
```

A suggestion for the short summary string might be:

"A Boolean variable which when set True traces Expression evaluation calls and returns"

And the summary string for `$TraceBuiltins` would change accordingly: 

"A Boolean Built-in variable when True collects function evaluation statistics"

mmatera:
> 
> This gets back to a long standing problem we have in Forms. There could be some sort of Code or Verbatim form.
> 
OK, that part can be handled elsewhere

> I tried this in the CLI though, I and I love the indenting. And that the functional form.
> 

Good!

> In[2]:= ?$ShowSteps
> enable or disable displaying the steps to get the result
> 
> A suggestion for the short summary string might be:
> 
> "A Boolean variable which, when True, traces Expression evaluation calls and returns"
> 
> And the summary string for `$TraceBuiltins` would change accordingly:
> 
> "A Boolean variable which, when True, collects function evaluation statistics"
> 

Done

Regarding `$ShowSteps`, I found that the right place to hold its value is in the `evaluation.definitions` object instead the `evaluation` object itself. `evaluation.definitions` exists during all the life of the Mathics kernel, while the evaluation object changes on each input.

TiagoCavalcante:
@mmatera LGTM. Thanks for finishing it!

PR #61, opened 2021-11-09
Attributes to symbol

mmatera:
In this PR, the way in which Attributes are stored in a definition is changed: instead of storing python strings, attributes are stored as symbols. This helps to do faster comparisons and accelerates both the transverse evaluation process as well as some aspects of the pattern matching algorithm.

Here are some benchmarks:

image

image

image

image

PR #64, opened 2021-11-13
Faster attributes

mmatera:
This PR contains the first part of the changes in the logic included in PR #61


TiagoCavalcante (approved):
LGTM

axkr:
Shouldn`t attributes be simply mapped from a "integer bit set" which is associated to each symbol?
- https://github.com/axkr/symja_android_library/blob/23785923ee5f5ef075692e1778c97759543ddf7b/symja_android_library/matheclipse-core/src/main/java/org/matheclipse/core/interfaces/ISymbol.java#L29

mmatera:
Yes, probably that would be faster. Something that we could do to keep the flexibility of the current implementation is to keep a bitmask in the definitions object that mirrors the stored attributes.

TiagoCavalcante:
@mmatera @rocky I think we could create a module with the attributes constants, then import them. Somethin like it:
current:
```
class F(Builtin):
    attributes = ("Protected", "ReadProtected")
```
new:
```
from mathics.attributes import Protected, ReadProtected

class F(Builtin):
    attributes = Protected | ReadProtected
```

If you agree, I can do this next week.

mmatera:
Looking at   https://reference.wolfram.com/language/ref/Attributes.html

I think, yes, it makes sense.

rocky:
Go for it!

TiagoCavalcante:
@mmatera I think all these changes except the last can the combined with attributes-bitset (I'm not sure about the last).

mmatera:
@TiagoCavalcante I think so. Let's try to merge it into another branch to see how they get on together.

mmatera:
This PR is deprecated.

PR #48, opened 2021-10-23
has_form

mmatera:
In this PR, following #47, I propose to modify the behavior of `expression.has_form` to take `Symbol`  references instead of strings as arguments. Again, this avoids having to check the name of the symbol(s) to be compared, and reduces the comparison to a "pointer-like" comparison. 

Benchmarks:

```
from mathics.core.expression import Expression
from mathics.core.symbols import Symbol

SymbolList = Symbol("List")
symboltst = Symbol("System`Internal`MySymboltest")
expr=Expression(symboltst,1,4,Symbol("None"), Symbol("True")) 
expr2=Expression(SymbolList,1,4,Symbol("None"), Symbol("True")) 
```

Old
-----

```
%timeit  expr.has_form("List", None)
```
678 ns ± 4.32 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

```
%timeit  expr.has_form("System`Internal`MySymboltest", None)
```

1.07 µs ± 6.79 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

```
%timeit  expr2.has_form("List", None)
```
746 ns ± 6.81 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

```
%timeit  expr2.has_form("System`Internal`MySymboltest", None)
```
996 ns ± 4.15 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

New
-------
```
%timeit  expr.has_form(SymbolList, None)
```
405 ns ± 2.2 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

```
%timeit  expr.has_form(symboltst, None)
```
449 ns ± 1.84 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

```
%timeit  expr2.has_form(SymbolList, None)
```
462 ns ± 2.18 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```
%timeit  expr2.has_form(symboltst, None)
```
409 ns ± 15.5 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

PR #65, opened 2021-11-13
Has form2

mmatera:
Superseeds #48


mathics/algorithm/parts.py
```diff
@@ -6,9 +6,9 @@
 
 
 from mathics.core.expression import Expression
-from mathics.core.symbols import Symbol
-from mathics.core.atoms import Integer
-from mathics.core.systemsymbols import SymbolInfinity
+from mathics.core.symbols import Symbol, SymbolList
+from mathics.core.atoms import Integer, from_python
```
TiagoCavalcante, line 10: I don't see where `from_python` is used.

mathics/builtin/box/compilation.py
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from mathics.builtin.base import BoxConstruct
+from mathics.version import __version__  # noqa used in loading to check consistency.
```
TiagoCavalcante, line 4: @mmatera @rocky do you think that checking consistency is still necessary. I think that Python wouldn't mess up with just some files. I think this would be either 0 or all.

mmatera: The only problem is to remove this line on each module...

TiagoCavalcante: I'll do that.

PR #66, opened 2021-11-14
Faster Plot

mmatera:
One of the slowest tests in `docpipeline` are the ones associated with Spherical Bessel functions. In particular, the ones that plot these functions. This was due to a wrong implementation of the numerification (is that the word?) of the functions to be plotted in
`mathics.builtin.drawing.Plot_`.
This PR
* Improves rules for SphericalBesselFunction(J/Y) to make the evaluation faster
* Fixes in the plot routine, that numerify expressions

### DIFF
--- a/mathics/builtin/drawing/plot.py
+++ b/mathics/builtin/drawing/plot.py
@@ -258,8 +258,7 @@ def compile_quiet_function(expr, arg_names, evaluation, expect_list):
                 return None
 
             return quiet_f
-
-    expr = Expression(SymbolN, expr)
+    expr = Expression(SymbolN, expr).evaluate(evaluation)
     quiet_expr = Expression(
         "Quiet",
         expr,


TiagoCavalcante:
@mmatera I don't understand why `SphericalBesselFunction` rule is different from `N @ SphericalBesselFunction`.

rocky:
@mmatera I just added a benchmark for this https://github.com/Mathics3/mathics-benchmark/tree/PlotSphericalBesselJ and I am not seeing any difference: 

report-Plot-SphericalBesselJ

What are you seeing?

mmatera:
These are my numbers with that test:
imagen

rocky:
oYes, when I run this `mathics -e 'Timing[Plot[SphericalBesselJ[1, x], {x, 0.1, 10}]]', I see what a big difference. But when I run from bench like this: 

```
python ./mathics_benchmark/compare.py -v Plot-SphericalBesselJ.yaml master faster-plot
find mathics -name "*.so" -type f -delete; \
find mathics -name "*.c" -type f -delete
find mathics -name "*.so" -type f -delete; \
find mathics -name "*.c" -type f -delete
```

I am seeing no difference. @TiagoCavalcante  ? 

One problem I had was that my branch in `mathics-core` didn't have f877f5 but I should have seen an error reporting that. 

And when I did pull that branch remove the results JSON I was still seeing the same thing. 

And then it took me something like 4 tries to come up with the command.

Sigh.

rocky (approved):
LGTM  But I wonder if this isn't a reflection of flaws in the code and if things were addressed properly in Mathics we wouldn't see such a discrepancy.

mathics/builtin/specialfns/bessel.py
```diff
@@ -691,7 +701,18 @@ class SphericalBesselY(_Bessel):
...
+        "SphericalBesselY[n_, z_]": """
+             Times[
+                Rational[1, 2], 
+                Power[Pi, Rational[1, 2]], 
+                Power[2, Rational[1, 2]], 
+                Power[z, Rational[-1, 2]], 
+                BesselY[Plus[Rational[1, 2], n], z]
+               ]
+             """,
+        "N[SphericalBesselY[n_,z_]]": "Sqrt[Pi/2]/Sqrt[z] * BesselY[N[1/2+n], N[z]]",
```
rocky, line 714: I'd suggest adding a comment here and in the other place where an additional rule is added to suggest why this is beneficial.

TiagoCavalcante:
@rocky what you need is `-p` and `-f`. By default `mathics-bench-compare` uses the existing JSON files (if they exist). `-f` ensure the benchmarks gonna be executed even if the JSON files exist. I'll add a comment to make this more clear.

mmatera:
> LGTM But I wonder if this isn't a reflection of flaws in the code and if things were addressed properly in Mathics we wouldn't see such a discrepancy.

OK, there are two things here. One is about the rules for evaluating Bessel functions. For them, probably yes: if the form we write the Builtin rule does not match to the fixed point that we expect the rule should achieve, but the evaluation is very fast, that should not matter too much. 
The other thing is about how we precompile an expression that should be numerically evaluated many times. This is what makes the largest difference here, and is not dependent on how the transversal evaluation or the pattern matching is implemented.  


Maybe this should be splitter in two PRs...

PR #68, opened 2021-11-14
Follow StackOverflow answer more carefully

rocky:
When I tried the branch that this is off of I saw
```
In[2]:= -: 2.1: syntax error
-: 1.37: syntax error
-: 1.24: syntax error
-: 1.24: syntax error
(2,0,0)
-: 1.17: syntax error
-: 2.1: syntax error
-: 1.51: syntax error
-: 1.24: syntax error.dev0, Style: inkpot, Mode: Emacs, Autobrace: True        
-: 1.39: syntax error
(1,1,0)
-: 1.17: syntax error
```

The changes here seem to get closer.  Also, it looked like we were drawing open cylinders.

In general if the solution has abstraction (variables for radius, height and endpoints), I don't think it a good idea to remove that in the reworking.

### DIFF
--- a/mathics/builtin/box/graphics3d.py
+++ b/mathics/builtin/box/graphics3d.py
@@ -776,12 +776,13 @@ class Cuboid3DBox(InstanceableBuiltin):
         pass
 
 
-class Cylinder3DBox(InstanceableBuiltin):
+class Cylinder3DBox(Graphics3DBox):
     """
     Internal Python class used when Boxing a 'Cylinder' object.
     """
 
     def init(self, graphics, style, item):
+        self.graphics = graphics
         self.edge_color, self.face_color = style.get_style(_Color, face_element=True)
 
         if len(item.leaves) != 2:
--- a/mathics/builtin/box/graphics3d.py
+++ b/mathics/builtin/box/graphics3d.py
@@ -776,13 +776,12 @@ class Cuboid3DBox(InstanceableBuiltin):
         pass
 
 
-class Cylinder3DBox(Graphics3DBox):
+class Cylinder3DBox(InstanceableBuiltin):
     """
     Internal Python class used when Boxing a 'Cylinder' object.
     """
 
     def init(self, graphics, style, item):
-        self.graphics = graphics
         self.edge_color, self.face_color = style.get_style(_Color, face_element=True)
 
         if len(item.leaves) != 2:
--- a/mathics/format/asy.py
+++ b/mathics/format/asy.py
@@ -43,6 +43,7 @@ from mathics.core.formatter import lookup_method, add_conversion_fn
 from mathics.format.asy_fns import (
     asy_add_bezier_fn,
     asy_add_graph_import,
+    asy_add_solids_import,
     asy_bezier,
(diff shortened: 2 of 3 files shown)


mathics/builtin/box/graphics3d.py
```diff
@@ -776,12 +776,13 @@ def _apply_boxscaling(self, boxscale):
         pass
 
 
-class Cylinder3DBox(InstanceableBuiltin):
+class Cylinder3DBox(Graphics3DBox):
```
TiagoCavalcante, line 779: `InstanceableBuiltin` is faster.

mathics/format/asy.py
```diff
@@ -294,29 +295,22 @@ def cylinder3dbox(self, **options) -> str:
...
-                    {self.radius},
-                    {distance},
-                    {tuple(point2)} - {tuple(point1)} // draw from point1 to point2
-                )), {rgb});"""
-            )
+            asy += f"real r={self.radius};\n"
+            asy += f"triple A={tuple(point1)}, B={tuple(point2)};\n"
+            asy += "real h=abs(A-B);\n"
+            asy += "revolution cyl=cylinder(A,r,h,B-A);\n"
+            asy += f"""draw(surface(cyl),{rgb});\n"""
```
TiagoCavalcante, line 309: Maybe single quote here?

TiagoCavalcante (commented):
@rocky I used the solution without abstraction because Asymptote is slow, and declaring variables instead of hardcoding a number is (or at least should be) slower. Also, I don't know if there are problems with variable re-declaration.
Anyway, that code wasn't working, and this one is :-)

mathics/builtin/box/graphics3d.py
rocky: > `InstanceableBuiltin` is faster.

But it is wrong. Cylinders are not unique, like the Number 5 is. And we need a graphics object to store whether we have imported `solds`.

mathics/format/asy.py
rocky: If this makes you happy go for it. I find this kind of thing worrying about one quote versus 3 positively stupid. Let black worry about stupid stuff like this.

rocky:
> @rocky I used the solution without abstraction because Asymptote is slow, and declaring variables instead of hardcoding a number is (or at least should be) slower. 

I don't know that Asymptote is slower for doing these kinds of computations. In fact it might be faster - for example it might have a faster `abs()` function between two points than the open-coded version in Python. 

Unless you have benchmarked these, please don't assume.  The first concern should be correctness and clarity.  


> Also, I don't know if there are problems with variable re-declaration. 

Go to http://asymptote.ualberta.ca/ and make some tests like: 

```
real H = 1.0;
real H = 2.0;
```

TiagoCavalcante:
@rocky `solids` is always imported. See https://github.com/Mathics3/mathics-core/blob/master/mathics/builtin/box/graphics3d.py#L580.

TiagoCavalcante:
@rocky also the `try except pass` seems not necessary.

PR #69, opened 2021-11-15
Stop importing __version__

TiagoCavalcante:
That was ued for checking consistency in loading, but that isn't
necessary in all files as Python wouldn't mess with just some
files, that'd be either no files or all files.


TiagoCavalcante:
@mmatera now this works.

mmatera:
LGTM

PR #70, opened 2021-11-17
Fix the distortion in Plot3D

TiagoCavalcante:
The bug was introduced in 0575f3d8eb574daf12eadf60b0786ea60839eb3c (thanks @rocky for finding it).
Plot3D was always drawing cube plots, even if the height was between -0.5 and 0.5 (supposing the range is -1 and 1), it was "distorted" to -1 and 1.

---

I know, it doesn't make any sense, I wish there were sense. Where did all the sense go?


rocky (dismissed):
<strike>LGTM - i tested fd85 locally using the gallery and it works.</strike> I had a merged version. No it is still broken.

PR #73, opened 2021-11-17
* Improved rules for SphericalBesselFunction(J/Y)

mmatera:
Benchmark:

``
Timing[Table[SphericalBesselJ[3,x],{x,0.00001,100,1.1}];]
``
with 6 runs,
master: 2.599 s  (sd .28s)
faster-bessel: 2.407 s (sd .219s)

### DIFF
--- a/mathics/builtin/specialfns/bessel.py
+++ b/mathics/builtin/specialfns/bessel.py
@@ -667,7 +667,17 @@ class SphericalBesselJ(_Bessel):
 
     attributes = ("Listable", "NumericFunction", "Protected")
 
-    rules = {"SphericalBesselJ[n_, z_]": "Sqrt[Pi / 2] / Sqrt[z] BesselJ[n + 0.5, z]"}
+    rules = {
+        "SphericalBesselJ[n_, z_]": """
+             Times[Rational[1, 2], 
+                   Power[Pi, Rational[1, 2]], 
+                   Power[2, Rational[1, 2]], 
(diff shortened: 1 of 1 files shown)


mathics/builtin/specialfns/bessel.py
```diff
@@ -667,7 +667,17 @@ class SphericalBesselJ(_Bessel):
...
+    rules = {
+        "SphericalBesselJ[n_, z_]": """
+             Times[Rational[1, 2], 
+                   Power[Pi, Rational[1, 2]], 
+                   Power[2, Rational[1, 2]], 
+                   Power[z, Rational[-1, 2]], 
+                   BesselJ[Plus[Rational[1, 2], n], z]
+             ]
+             """,
+        "N[SphericalBesselJ[n_,z_]]": "Sqrt[Pi/2]/Sqrt[z] * BesselJ[N[1/2+n], N[z]]",
```
rocky, line 679: I detect a pattern here. There is a part of me that feels like a whole bunch of rules like this should be collected together into an autoload file with _comments_ describing why they are useful. 

And then the the root cause is fixed, we would be in a better position to go relook at stuff like this in one place.

mmatera: The funny thing is that this seems to be useful in the current implementation. Let's keep this PR as a draft, since this does not have a very huge impact on performance.

mmatera:
I replace this by the Issue #91

PR #80, opened 2021-11-28
ShowSteps -> TraceEvaluation

rocky:
TraceEvaluation fits more regularly into the other Trace routines and is more likely to be found when running say `Names["Trace*"]` or looking this up from Django.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,17 +3,17 @@ CHANGES
 
 New Builtins
 ============
-* ``ShowStep=True`` makes that every step in the evaluation is printed out. If it is set to ``False`` the standard behaviour is recovered.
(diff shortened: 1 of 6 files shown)


mmatera:
What about TraceEvaluation?

rocky:
> What about TraceEvaluation?

That would be fine too. It is really _Expression_ evaluation that is getting traced. Are other kinds of evaluations other than those with expressions. Like evaluations in compiled code? Or evaluations kicked off from intermediate evaluations? 

(If the come from Expressions, I imagine they'd be traced, if not then not.)

TiagoCavalcante:
@rocky yes, this makes more sense as it is not as Wolfram Alpha's "show steps".

mmatera (approved):
LGTM

PR #85, opened 2021-12-02
Improve make clean

TiagoCavalcante:


Makefile
```diff
@@ -68,6 +68,9 @@ clean: clean-cython
 	   ($(MAKE) -C "$$dir" clean); \
 	done; \
 	rm -f factorials || true
+	find mathics -name *.py[co] -type f -delete
+	find mathics -name __pycache__ -type d -delete
+	rm -rf build
```
rocky, line 73:
```
rm -fr build || true
```
so that if build doesn't exist this isn't considered a failure.

rocky: Just as we have a `clean-cython` I'd put those `find`s inside its own target, e.g. `clean-pyc` so that this can be done independent. And remember to add a `#:` comment beforehand so that people can more easily know what is up. Thanks.

TiagoCavalcante: @rocky done now.

mmatera:
This seems to fix the issue with mathics-bench, so LGTM

Makefile
rocky: @TiagoCavalcante Thanks! I only have small observations. Removing build isn't part of the cache. I could be part of the general clean target or its own separate target. 

However this is small if you want to fix do so. If not, leave and possibly in the future it will be addressed.

TiagoCavalcante: @rocky I'll move that.

PR #88, opened 2021-12-03
Covert is_symbol functions to class attributes

TiagoCavalcante:


mmatera:
@TiagoCavalcante, can you add some benchmarks to see the impact of this change?

TiagoCavalcante:
@mmatera this was mainly about readability, I forget to push the commit which changed `is_atom` (doing this now).

Anyway, the benchmarks:
```python
>>> from timeit import timeit
>>> class X:
...     is_x = True
... 
>>> class Y:
...     @property
...     def is_x(self):
...             return True
... 
>>> class Z:
...     def is_x(self):
...             return True
... 
>>> timeit(lambda: X().is_x); timeit(lambda: Y().is_x); timeit(lambda: Z().is_x())
0.15901221499916574
0.2110749980001856
0.2190195160001167
```

mmatera (commented):
With this change, maybe it also worth to change all the occurrences of `isinstance(obj, Symbol)` to `obj.is_symbol` and the ones of  `isinstance(obj, Atom)` to `obj.is_atom`, isn´t it?

mmatera:
@TiagoCavalcante, this example rephrases yours doing a mimic of the Mathics class structure:

```
from timeit import timeit

class basebaseQ(object):
    def __init__(self):
        pass
    
class baseQ(basebaseQ):
    def __init__(self):
        pass
    

class X(baseQ):
    is_x = True

class Y(X):
    @property
    def is_x(self):
        return True

class Z(Y):
    def is_x(self):
        return True

z=Z()
x=X()
y=Y()
    
    
(timeit(lambda: x.is_x), timeit(lambda: y.is_x), timeit(lambda: z.is_x()), timeit(lambda: isinstance(z,X)),timeit(lambda: isinstance(z,Z)))
```
I get the following result:

0.16340709396172315,  -> class attribute
 0.23979647096712142,  -> @property
 0.24807486997451633,  -> explicit method
 0.1908878959948197,    -> `isinstance` for X
 0.19201096100732684  ->  `isinstance`  for Z

TiagoCavalcante:
Ok, lets keep it as is by now.

mmatera:
> Ok, lets keep it as is by now.

why?

rocky:
I will give my reasons why I would say not:  basically all of this in my opinion is a distraction from the large picture of addressing the main performance problems. 

Ok so we've e done timings in 2021 . Will that be true in 2026?  Maybe then we'll go through the same amount of work to undo this. 

To me all of this is like remodeling a house and starting out with painting the walls, rather than doing that at the end.

PR #90, opened 2021-12-03
Improve association

TiagoCavalcante:
@rocky you said about storing the associations as Python dicts, I agree, but I can't see how it would be done.  
What I have been thinking is that associations should be primitives.

What do you think?


mmatera:
> @rocky you said about storing the associations as Python dicts, I agree, but I can't see how it would be done. What I have been thinking is that associations should be primitives.
> 
> What do you think?

Maybe this could be done by making Associations to be InstanceableBuilitin or AtomBuiltin. Looks the implementation of CompiledFunction or ImageAtom

mathics/builtin/list/associations.py
```diff
@@ -16,13 +16,16 @@
...
 from mathics.core.atoms import Integer
-from mathics.core.symbols import Symbol, SymbolList
+from mathics.core.symbols import Symbol, SymbolList, SymbolTrue
 from mathics.core.systemsymbols import (
     SymbolAssociation,
     SymbolMakeBoxes,
 )
 
 
+SymbolRowBox = Symbol("RowBox")
```
mmatera, line 26: I think this should be in mathics.core.systemsymbos

mmatera (approved):
LGTM

mathics/builtin/list/associations.py
TiagoCavalcante: Done.

mathics/builtin/list/associations.py
```diff
@@ -95,8 +98,8 @@ def validate(exprs):
             for expr in exprs:
                 if expr.has_form(("Rule", "RuleDelayed"), 2):
                     pass
-                elif expr.has_form("List", None) or expr.has_form("Association", None):
-                    if validate(expr.leaves) is not True:
+                elif expr.has_form(("List", "Assocition"), None):
```
axkr, line 99: Probably a typo?

TiagoCavalcante: Yes.

TiagoCavalcante: Fixed now.

TiagoCavalcante:
@mmatera @rocky the advantage from making it a primitive type is that we could add `from_python` and `to_python`.

axkr:
> What I have been thinking is that associations should be primitives.

I've also not implemented `Associaton` in this way in Symja, but other programming languages (Clojure, Scala, Frege...) seem to use a "Hash array mapped trie":
- https://en.wikipedia.org/wiki/Hash_array_mapped_trie

For example Clojure:
- https://github.com/clojure/clojure/blob/master/src/jvm/clojure/lang/PersistentHashMap.java

or JavaScript:
- https://github.com/mattbierner/hamt

TiagoCavalcante:
@axkr seems they are hash tables (see [CPython's dictobject.c](https://github.com/python/cpython/blob/main/Objects/dictobject.c)).

I think that if Mathics be written in a lower-level language we should try other types of dictionaries, but writing a faster algorithm for dictionaries in Python would still be slower than the algorithm Python uses.  
Supposing HAMTs be twice as fast as hash tables, C is 10 times as fast as Python, and Python's dict is written in C, so a HAMT would be 5 times slower than a Python's dict.

TiagoCavalcante:
@mmatera I think that lists should be a primitive type too.

rocky:
@TiagoCavalcante I greatly apologize for not seeing this sooner and getting this in sooner. (And I recall you reminding me to look at, at some point and not realizing possibly that this is what you were referring to.)

I just tried to adjust to master and I can't. If you would redo this I promise this won't slip through the cracks. I realize this is a lot of work that wasn't necessary if I had just been more attentive. Again, I feel really bad about this.

rocky:
LGTM - this code has been rebased in #160 to address the above merge conflicts. Again, my apologies for this.

TiagoCavalcante:
@rocky no problem for the delay. But to say the truth, I still can't answer the question "How to map Associtions to Python dict?". If you could help there...

rocky:
Ok. I'll open an issue and assign myself so I don't forget.

PR #92, opened 2021-12-06
fixing attributes for True and False

mmatera:
This PR adds the `Locked` attribute to ```System`True``` and ```System`False```  according to the WL standard.

### DIFF
--- a/mathics/builtin/logic.py
+++ b/mathics/builtin/logic.py
@@ -10,7 +10,14 @@ from mathics.core.symbols import (
     SymbolFalse,
 )
 
-from mathics.core.attributes import flat, hold_all, one_identity, orderless, protected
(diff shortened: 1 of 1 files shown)


TiagoCavalcante:
@mmatera now I'm thinking, shoudln't all operators be locked?

mmatera:
According to what I see in WL, no. `List`, `True` and `False` are `Locked`. Other symbols (like `Pi`, `Sin` or `N`) are just protected.

TiagoCavalcante:
Ok, so LGTM.

@mmatera I think we should start doing rebase merges to have a clean log in the master branch. The rebase merges don't have a commit like "Merge #92".

On the side of the merge button there is an arrow, when you click you can select "rebase and merge".

PR #95, opened 2021-12-11
Allow using Sympy 1.9

rocky:
Sympy 1.9 is out and it seems to work without problems. Allow (but don't require) it.

### DIFF
--- a/mathics/builtin/numbers/calculus.py
+++ b/mathics/builtin/numbers/calculus.py
@@ -12,7 +12,6 @@ from mathics.core.expression import Expression
 from mathics.core.atoms import (
     String,
     Integer,
-    Integer0,
     Integer1,
     Number,
     Rational,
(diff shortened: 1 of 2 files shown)


TiagoCavalcante:
Seems jut 1 test fails, I'll try to fix that tommorrow.

rocky:
> Seems just 1 test fails, I'll try to fix that tommorrow.

@TiagoCavalcante  Thanks!

TiagoCavalcante:
@rocky what is happening here is that Sympy isn't returning `Indeterminate` for the expression `Integrate[-Infinity, {x, 0, Infinity}]`, but that is keeping the expression as is. Then the apply function converts that expression to a Mathics Expression, which again is `Integrate[-Infinity, {x, 0, Infinity}]`, then the apply method is called again...
A way to fix this is the atom `Integrate` instead of the expression integrate.

rocky:
@TiagoCavalcante  Thanks for doing this. Looks good to me!

Issue #96, opened 2021-12-11
Reference to non-existing file?


rebcabin:
This issue is something I found just by reading the code. In examples/symbolic_logic/gries_schneider/GS1.m, there is a command to "<<GS0.m", which (according to the prose), brings into existence the symbol "expect" for the internal test framework. 

I'm not equipped to run the examples now (on my iPad), but I don't see how they could run. I'll upload a copy of `GS0.m` from my archives later, if it would be helfpul to do.

TiagoCavalcante:
@rebcabin yes, that file is missing, could you put that file in a PR?

rebcabin:
@TiagoCavalcante Will do when i get back to my computer, possibly today but might not be till Monday.

rebcabin:
PR should show up as issue #97.

PR #97, opened 2021-12-11
Issue #96: missing file

rebcabin:
last tested on legacy Mathics, before Mathics was split up. Uploading for completeness.


TiagoCavalcante:
@rebcabin thanks for doing this and sorry for the delay!

Issue #103, opened 2022-01-10
Overflow in computation with large integers


oscarbenjamin:
This example is from the mathematica docs:
https://reference.wolfram.com/language/tutorial/DiophantineReduce.html

This calculation produces a large integer and then numerical evaluation overflows somehow:
```python
In[1]:= poly = x^100000 + 1234 x^77777 - 2121 x^12345 + 7890 x^999 - x^11
Out[1]:= -x ^ 11 + 7890 x ^ 999 - 2121 x ^ 12345 + 1234 x ^ 77777 + x ^ 100000

In[2]:= freeterm = poly /. x -> 1234567;
Out[2]:= None

In[3]:= N[freeterm]
General::ovfl: Overflow occurred in computation.
```
Also displaying `freeterm` is very slow.

By contrast with SymPy (not sure if SymPy is actually involved here):
```python
In [1]: poly = x**100000 + 1234* x**77777 - 2121* x**12345 + 7890* x**999 - x**11

In [2]: freeterm = poly.subs(x, 1234567)

In [3]: N(freeterm)
Out[3]: 2.92690499812734e+609151
```

rocky:
If you set `$TraceEvaluation = True` you will the steps that Mathics is taking and get some sense of why it is slow. 

The variable "poly" is stored symbolically as a Mathics Expression, but freeterm isn't - all of those calculations in large precision are being done which is why freeterm is slow.  Of course change the `=` (Set) to `:=` (SetDelayed) but when you go to evaluate all of this with `N`, all of the expansion is done inside Mathics using calls I think to sympy.

mmatera:
Actually, the evaluation is "fast":
![imagen](https://user-images.githubusercontent.com/2862288/148833194-583f51cc-2ebe-4015-a073-e3fd02d64db5.png)

what is very slow in Mathics  is to show the 609152 digits of the number.  WMA does it pretty fast, while the experiment in Sympy converts it to a floating point number. 

In Mathics, this last conversion produces the overflow, because the number is too large to be converted to a MachineReal number. I think this part is easy to fix (later I will send a PR to fix this).

rocky:
See also https://github.com/Mathics3/mathics-core/discussions/111#discussioncomment-1960660

PR #106, opened 2022-01-10
fix N (Issue #103)

mmatera:
The evaluation is still slow, but this avoids the overflow. There are two things that could be improved:
* to check the size of the number before the conversion, to avoid using an exception.
* the default value of `d` could be taken from `mathics.core.number`  instead of hardcoding it to 15.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,8 +35,8 @@ Internals
 Bugs
 ++++
 
-* ``First``, ``Rest`` and  ``Last`` now handle invalid arguments.
-  
+* ``First``, ``Rest`` and  ``Last`` now handle invalid arguments (issue #104).
+* ``N`` now handles properly the conversion of large integers to PrecisionReal (issue #103).
   
 4.0.1
 -----
--- a/mathics/core/atoms.py
+++ b/mathics/core/atoms.py
@@ -165,9 +165,11 @@ class Integer(Number):
 
     def round(self, d=None) -> typing.Union["MachineReal", "PrecisionReal"]:
         if d is None:
-            return MachineReal(float(self.value))
-        else:
-            return PrecisionReal(sympy.Float(self.value, d))
+            try:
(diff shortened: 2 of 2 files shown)


rocky:
> The evaluation is still slow, but this avoids the overflow. 

Great! 

> There are two things that could be improved:
> 
> * to check the size of the number before the conversion, to avoid using an exception.

For this, [`bitlength()`](/src/external-vcs/github/Mathics3/mathics-core/mathics/core/number.py)  is what you want. 

> * the default value of `d` could be taken from `mathics.core.number`  instead of hardcoding it to 15.

https://github.com/Mathics3/mathics-core/blob/master/mathics/core/number.py#L28-L35 seems to suggest that the default value would be 17, no?  @mmatera where does 15 come into this? 

See #107 for a suggestion as to how to use. Feel free to commit over or adapt here otherwise.

mmatera:
> > The evaluation is still slow, but this avoids the overflow.
> 
> Great!
> 
> > There are two things that could be improved:
> > 
> > * to check the size of the number before the conversion, to avoid using an exception.
> 
> For this, [`bitlength()`](/src/external-vcs/github/Mathics3/mathics-core/mathics/core/number.py) is what you want.
> 
> > * the default value of `d` could be taken from `mathics.core.number`  instead of hardcoding it to 15.
> 
> https://github.com/Mathics3/mathics-core/blob/master/mathics/core/number.py#L28-L35 seems to suggest that the default value would be 17, no? @mmatera where does 15 come into this?
> 
@rocky: I used 15, in order to reproduce what we get using Sympy, but you are right, WMA uses 17. 


> See #107 for a suggestion as to how to use. Feel free to commit over or adapt here otherwise.

mmatera:
@rocky, trying to merge your branch into this one, I merge by accident this branch into master. I tried to revert, but it seems it does not work. To avoid more noise, I continue this in #109.

rocky:
np

PR #107, opened 2022-01-10
Use bit_length and machine precision

rocky:
Use `bit_length()` and `machine_precision`
    
These two items suggested by mmatera. 

I am not certain though if this fully addresses the concerns, so double check.


TiagoCavalcante (commented):
Seems your code is correct to me.

PR #114, opened 2022-01-12
Add .gitpod.yml

TiagoCavalcante:
To test this:
https://gitpod.io/#https://github.com/TiagoCavalcante/mathics-core

We should add an easy-to-find mention to this, I recommend a badge in README.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,34 @@
+github:
+  prebuilds:
+    addComment: false
+    addBadge: true
+
+ports:
+  - port: 8080-8081
+    onOpen: open-browser
+
+tasks:
+  - init: |
+      make
+      make install
+
+      cd /workspace
+      git clone https://github.com/Mathics3/mathicsscript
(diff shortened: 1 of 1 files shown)


.gitpod.yml
```diff
@@ -0,0 +1,34 @@
+github:
+  prebuilds:
+    addComment: false
+    addBadge: true
```
TiagoCavalcante, line 4: This adds a "Review in GitPod" badge to PRs.

mmatera:
I like this!

rocky (approved):
LGTM - I am not sure if I trust gitpod right now with my permissions, but that is up to an individual to decide.

rocky:
After mulling this over a bit and after looking over gitpod, okay, let's try it! I will mention though that I personally do not want to expose myself to allowing gitpod write access from my account.  It is not that I don't trust gitpod, but its more about being concerned about those who may use gitpod as a surface to attack. 

So what that said, let's try it.

PR #116, opened 2022-01-17
Reduce unnecessary expansion & untangle imports...

rocky:
A number of different kinds of atoms that can't change value unnecessarily go through an extra expansion for fixpoint check.

In particular many constants values: strings, numbers, fixed symbols e.g. True, False, N, etc.

The one place where we *do* need to check though is Symbol, because a value like "x" may change. So we separate Symbols from Constants.

Also we remove "import" of a name where that name isn't used directly in the module. This seems to happen so that some other module can find the name instead of where it was defined. This isn't helpful. In fact, it is the opposite.

### DIFF
--- a/mathics/algorithm/parts.py
+++ b/mathics/algorithm/parts.py
@@ -6,7 +6,7 @@ Algorithms to access and manipulate elements in nested lists / expressions
 
 
 from mathics.core.expression import Expression
-from mathics.core.symbols import Symbol
+from mathics.core.symbols import Constant, Symbol
 from mathics.core.atoms import Integer
 from mathics.core.systemsymbols import SymbolInfinity
 from mathics.core.subexpression import SubExpression
@@ -18,7 +18,7 @@ from mathics.builtin.exceptions import (
     PartRangeError,
 )
 
-SymbolNothing = Symbol("Nothing")
+SymbolNothing = Constant("Nothing")
 
 # TODO: delete me
 # def join_lists(lists):
--- a/mathics/builtin/assignments/internals.py
+++ b/mathics/builtin/assignments/internals.py
@@ -153,7 +153,7 @@ def find_tag_and_check(lhs, tags, evaluation):
 
 
 def unroll_patterns(lhs, rhs, evaluation):
-    if type(lhs) is Symbol:
+    if isinstance(lhs, Symbol):
         return lhs, rhs
     name = lhs.get_head_name()
     lhsleaves = lhs._leaves
@@ -171,7 +171,7 @@ def unroll_patterns(lhs, rhs, evaluation):
 
 def unroll_conditions(lhs):
     condition = None
-    if type(lhs) is Symbol:
+    if isinstance(lhs, Symbol):
         return lhs, None
     else:
         name, lhs_leaves = lhs.get_head_name(), lhs._leaves
@@ -651,7 +651,7 @@ class _SetOperator(object):
     }
 
     def assign_elementary(self, lhs, rhs, evaluation, tags=None, upset=False):
-        if type(lhs) is Symbol:
+        if isinstance(lhs, Symbol):
             name = lhs.name
         else:
             name = lhs.get_head_name()
--- a/mathics/builtin/box/compilation.py
+++ b/mathics/builtin/box/compilation.py
@@ -1,9 +1,10 @@
 # -*- coding: utf-8 -*-
 
 from mathics.builtin.base import BoxConstruct
+from mathics.core.symbols import ExpandOnce
 
 
-class CompiledCodeBox(BoxConstruct):
+class CompiledCodeBox(BoxConstruct, ExpandOnce):
     """Routines which get called when Boxing (adding formatting and bounding-box information)
     to CompiledCode.
     """
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -15,6 +15,7 @@ from mathics.core.convert import sympy_symbol_prefix, SympyExpression
 from mathics.core.symbols import (
     Atom,
     BaseExpression,
+    ExpandOnce,
     Monomial,
     Symbol,
     SymbolList,
@@ -22,9 +23,7 @@ from mathics.core.symbols import (
     SymbolSequence,
     system_symbols,
     ensure_context,
-    strip_context,
 )
-from mathics.core.systemsymbols import SymbolSequence
 
 from mathics.core.attributes import (
     flat,
@@ -807,7 +806,6 @@ class Expression(BaseExpression):
         return expr
 
     def evaluate_next(self, evaluation) -> typing.Tuple["Expression", bool]:
-        from mathics.builtin.base import BoxConstruct
 
         head = self._head.evaluate(evaluation)
         attributes = head.get_attributes(evaluation.definitions)
@@ -909,7 +907,7 @@ class Expression(BaseExpression):
         for rule in rules():
             result = rule.apply(new, evaluation, fully=False)
             if result is not None:
-                if isinstance(result, BoxConstruct):
+                if isinstance(result, ExpandOnce):
                     return result, False
                 if result.sameQ(new):
                     new._timestamp_cache(evaluation)
--- a/mathics/core/subexpression.py
+++ b/mathics/core/subexpression.py
@@ -8,7 +8,7 @@ from mathics.core.atoms import Integer
 from mathics.builtin.base import MessageException
 
 """
-This module provides some infraestructure to deal with SubExpressions. 
+This module provides some infraestructure to deal with SubExpressions.
 
 """
 
@@ -143,7 +143,7 @@ class ExpressionPointer(object):
         parent = self.parent
         p = self.position
         if p == 0:
-            if type(parent) is Symbol:
+            if isinstance(parent, Symbol):
                 return parent
             else:
                 return parent.head.copy()
--- a/mathics/core/symbols.py
+++ b/mathics/core/symbols.py
@@ -14,6 +14,15 @@ from mathics.core.attributes import nothing
 sympy_symbol_prefix = "_Mathics_User_"
 sympy_slot_prefix = "_Mathics_Slot_"
 
+# This feels like a hack. It generalizes the method expression used in expression regarding
+# BoxConstruct.
+class ExpandOnce:
+    """Inherit from this class when you have an object that you never want expanded after
+    the initial try. BoxConstructs are like this
+    """
+
+    pass
+
 
 # system_symbols('A', 'B', ...) -> [Symbol('System`A'), Symbol('System`B'), ...]
 def system_symbols(*symbols) -> typing.FrozenSet[str]:
@@ -807,26 +816,40 @@ class Symbol(Atom):
         return (self.name, self.sympy_dummy)
 
 
+class Constant(Symbol, ExpandOnce):
+    """For Symbols that are cannot be changed"""
+
+    def __init__(self, name, sympy_dummy=None):
+        self = super().__new__(self, name, sympy_dummy)
+
+    pass
+
+
 # Symbols used in this module.
 
-SymbolFalse = Symbol("System`False")
-SymbolGraphics = Symbol("System`Graphics")
-SymbolGraphics3D = Symbol("System`Graphics3D")
-SymbolHoldForm = Symbol("System`HoldForm")
-SymbolList = Symbol("System`List")
-SymbolMachinePrecision = Symbol("MachinePrecision")
-SymbolMakeBoxes = Symbol("System`MakeBoxes")
-SymbolMaxPrecision = Symbol("$MaxPrecision")
-SymbolMinPrecision = Symbol("$MinPrecision")
-SymbolN = Symbol("System`N")
-SymbolNull = Symbol("System`Null")
-SymbolNumberForm = Symbol("System`NumberForm")
-SymbolPostfix = Symbol("System`Postfix")
-SymbolRepeated = Symbol("System`Repeated")
-SymbolRepeatedNull = Symbol("System`RepeatedNull")
-SymbolSequence = Symbol("System`Sequence")
-SymbolTrue = Symbol("System`True")
+SymbolFalse = Constant("System`False")
+SymbolGraphics = Constant("System`Graphics")
+SymbolGraphics3D = Constant("System`Graphics3D")
+SymbolHoldForm = Constant("System`HoldForm")
+SymbolList = Constant("System`List")
+SymbolMachinePrecision = Constant("MachinePrecision")
+SymbolMakeBoxes = Constant("System`MakeBoxes")
+SymbolMaxPrecision = Constant("$MaxPrecision")
+SymbolMinPrecision = Constant("$MinPrecision")
+SymbolN = Constant("System`N")
 
+# FIXME: LinearModelFit[{{2, 1}, {3, 4}, {5, 3}, {7, 6}}, x, x] ;
+# fails if we use a Constant below. Fix whatever is making this fail.
+# SymbolNull = Constant("System`Null")
+SymbolNull = Symbol("System`Null")
+
+SymbolNumberForm = Constant("System`NumberForm")
+
+SymbolPostfix = Constant("System`Postfix")
+SymbolRepeated = Constant("System`Repeated")
+SymbolRepeatedNull = Constant("System`RepeatedNull")
+SymbolSequence = Constant("System`Sequence")
+SymbolTrue = Constant("System`True")
 
 # The available formats.
 
@@ -841,13 +864,13 @@ format_symbols = system_symbols(
 )
 
 
-SymbolInputForm = Symbol("InputForm")
-SymbolOutputForm = Symbol("OutputForm")
-SymbolStandardForm = Symbol("StandardForm")
-SymbolFullForm = Symbol("FullForm")
-SymbolTraditionalForm = Symbol("TraditionalForm")
-SymbolTeXForm = Symbol("TeXForm")
-SymbolMathMLForm = Symbol("MathMLForm")
+SymbolInputForm = Constant("InputForm")
+SymbolOutputForm = Constant("OutputForm")
+SymbolStandardForm = Constant("StandardForm")
+SymbolFullForm = Constant("FullForm")
+SymbolTraditionalForm = Constant("TraditionalForm")
+SymbolTeXForm = Constant("TeXForm")
+SymbolMathMLForm = Constant("MathMLForm")
 
 
 # Used to check if a symbol is `Numeric` without evaluation.
--- a/mathics/builtin/arithfns/basic.py
+++ b/mathics/builtin/arithfns/basic.py
@@ -30,7 +30,13 @@ from mathics.core.atoms import (
     String,
     from_mpmath,
 )
-from mathics.core.symbols import Symbol, SymbolList, SymbolNull, SymbolHoldForm
+from mathics.core.symbols import (
+    Constant,
+    Symbol,
+    SymbolList,
+    SymbolNull,
+    SymbolHoldForm,
+)
 from mathics.core.systemsymbols import (
     SymbolBlank,
     SymbolComplexInfinity,
@@ -63,7 +69,7 @@ from mathics.core.attributes import (
 )
 
 
-SymbolLeft = Symbol("Left")
+SymbolLeft = Constant("Left")
 
 
 class CubeRoot(Builtin):
--- a/mathics/builtin/assignments/assignment.py
+++ b/mathics/builtin/assignments/assignment.py
@@ -5,7 +5,7 @@ Forms of Assignment
 
 
 from mathics.builtin.base import Builtin, BinaryOperator
-from mathics.core.symbols import Symbol
+from mathics.core.symbols import Symbol, SymbolNull
 
 from mathics.core.systemsymbols import (
     SymbolFailed,
@@ -159,7 +159,7 @@ class SetDelayed(Set):
         "lhs_ := rhs_"
 
         if self.assign(lhs, rhs, evaluation):
-            return Symbol("Null")
+            return SymbolNull
         else:
             return SymbolFailed
 
--- a/mathics/builtin/base.py
+++ b/mathics/builtin/base.py
@@ -12,11 +12,7 @@ from typing import Any, cast
 
 from mathics.builtin.exceptions import (
     BoxConstructError,
-    InvalidLevelspecError,
     MessageException,
-    PartError,
-    PartDepthError,
-    PartRangeError,
 )
 from mathics.core.convert import from_sympy
 from mathics.core.definitions import Definition
@@ -29,6 +25,7 @@ from mathics.core.symbols import (
     strip_context,
 )
 from mathics.core.atoms import (
+    ExpandOnce,
     Integer,
     MachineReal,
     PrecisionReal,
@@ -41,7 +38,7 @@ from mathics.core.symbols import (
     SymbolTrue,
 )
 
-from mathics.core.attributes import nothing, protected
+from mathics.core.attributes import protected
 
 
 def get_option(options, name, evaluation, pop=False, evaluate=True):
@@ -590,7 +587,7 @@ class SympyFunction(SympyObject):
         return sympy_expr
 
 
-class BoxConstruct(InstanceableBuiltin):
+class BoxConstruct(InstanceableBuiltin, ExpandOnce):
     def __new__(cls, *leaves, **kwargs):
         instance = super().__new__(cls, *leaves, **kwargs)
         instance._leaves = leaves
--- a/mathics/builtin/box/graphics.py
+++ b/mathics/builtin/box/graphics.py
@@ -42,7 +42,7 @@ from mathics.core.atoms import (
     Real,
     String,
 )
-from mathics.core.symbols import SymbolList
+from mathics.core.symbols import ExpandOnce, SymbolList
 
 from mathics.core.attributes import hold_all, protected, read_protected
 
@@ -352,7 +352,7 @@ class DiskBox(_ArcBox):
     face_element = True
 
 
-class GraphicsBox(BoxConstruct):
+class GraphicsBox(BoxConstruct, ExpandOnce):
     """Boxing method which get called when Boxing (adding formatting and bounding-box information)
     Graphics.
     """
--- a/mathics/builtin/box/image.py
+++ b/mathics/builtin/box/image.py
@@ -1,9 +1,10 @@
 # -*- coding: utf-8 -*-
 
 from mathics.builtin.base import BoxConstruct
+from mathics.core.symbols import ExpandOnce
 
 
-class ImageBox(BoxConstruct):
+class ImageBox(BoxConstruct, ExpandOnce):
     """Routines which get called when Boxing (adding formatting and bounding-box information)
     an Image object.
     """
--- a/mathics/builtin/inout.py
+++ b/mathics/builtin/inout.py
@@ -26,7 +26,14 @@ from mathics.builtin.lists import list_boxes
 from mathics.builtin.options import options_to_rules
 
 from mathics.core.expression import Expression, BoxError
-from mathics.core.symbols import Symbol, SymbolList, SymbolTrue, SymbolFalse, SymbolNull
+from mathics.core.symbols import (
+    ExpandOnce,
+    Symbol,
+    SymbolList,
+    SymbolTrue,
+    SymbolFalse,
+    SymbolNull,
+)
 
 from mathics.core.atoms import (
     String,
@@ -752,7 +759,7 @@ def is_constant_list(list):
 
 # TODO: Inheritance of options["ColumnAlignments"] prevents us from
 # putting this in mathics.builtin.box. Figure out what's up here.
-class GridBox(BoxConstruct):
+class GridBox(BoxConstruct, ExpandOnce):
     r"""
     <dl>
     <dt>'GridBox[{{...}, {...}}]'
--- a/mathics/builtin/list/eol.py
+++ b/mathics/builtin/list/eol.py
@@ -12,9 +12,9 @@ from itertools import chain
 from mathics.builtin.base import (
     BinaryOperator,
     Builtin,
-    InvalidLevelspecError,
-    PartError,
 )
+
+from mathics.builtin.exceptions import InvalidLevelspecError, PartError
 
 from mathics.builtin.lists import list_boxes
 from mathics.algorithm.parts import (
--- a/mathics/builtin/lists.py
+++ b/mathics/builtin/lists.py
@@ -29,15 +29,18 @@ from mathics.algorithm.clusters import (
 from mathics.builtin.base import (
     Builtin,
     CountableInteger,
-    InvalidLevelspecError,
     MessageException,
     NegativeIntegerException,
-    PartDepthError,
-    PartError,
-    PartRangeError,
     Predefined,
     SympyFunction,
     Test,
+)
+
+from mathics.builtin.exceptions import (
+    InvalidLevelspecError,
+    PartDepthError,
+    PartError,
+    PartRangeError,
 )
 
 from mathics.builtin.numbers.algebra import cancel
--- a/mathics/builtin/strings.py
+++ b/mathics/builtin/strings.py
@@ -759,11 +759,11 @@ class String_(Builtin):
 class ToString(Builtin):
     """
     <dl>
-    <dt>'ToString[$expr$]'
-        <dd>returns a string representation of $expr$.
-    <dt>'ToString[$expr$, $form$]'
-        <dd>returns a string representation of $expr$ in the form
-          $form$.
+      <dt>'ToString[$expr$]'
+      <dd>returns a string representation of $expr$.
+
+      <dt>'ToString[$expr$, $form$]'
+      <dd>returns a string representation of $expr$ in the form $form$.
     </dl>
 
     >> ToString[2]
--- a/mathics/builtin/structure.py
+++ b/mathics/builtin/structure.py
@@ -10,8 +10,8 @@ from mathics.builtin.base import (
     BinaryOperator,
     Test,
     MessageException,
-    PartRangeError,
 )
+from mathics.builtin.exceptions import InvalidLevelspecError, PartRangeError
 from mathics.core.expression import Expression
 from mathics.core.symbols import (
     Symbol,
@@ -34,7 +34,6 @@ from mathics.core.rules import Pattern
 from mathics.builtin.lists import (
     python_levelspec,
     walk_levels,
-    InvalidLevelspecError,
     List,
 )
 
--- a/mathics/core/atoms.py
+++ b/mathics/core/atoms.py
@@ -70,7 +70,17 @@ def from_mpmath(value, prec=None):
         raise TypeError(type(value))
 
 
-class Number(Atom):
+# This feels like a hack. It generalizes the method expression used in expression regarding
+# BoxConstruct.
+class ExpandOnce:
+    """Inherit from this class when you have an object that you never want expanded after
+    the initial try. BoxConstructs are like this
+    """
+
+    pass
+
+
+class Number(Atom, ExpandOnce):
     def __str__(self) -> str:
         return str(self.value)
 
@@ -125,7 +135,7 @@ _number_form_options = {
 }
 
 
-class Integer(Number):
+class Integer(Number, ExpandOnce):
     value: int
     class_head_name = "System`Integer"
 
@@ -220,7 +230,7 @@ Integer0 = Integer(0)
 Integer1 = Integer(1)
 
 
-class Rational(Number):
+class Rational(Number, ExpandOnce):
     class_head_name = "System`Rational"
 
     @lru_cache()
@@ -324,7 +334,7 @@ class Rational(Number):
 RationalOneHalf = Rational(1, 2)
 
 
-class Real(Number):
+class Real(Number, ExpandOnce):
     class_head_name = "System`Real"
 
     def __new__(cls, value, p=None) -> "Real":
@@ -715,7 +725,7 @@ class Complex(Number):
         return real_zero and imag_zero
 
 
-class String(Atom):
+class String(Atom, ExpandOnce):
     value: str
     class_head_name = "System`String"
 
--- a/mathics/main.py
+++ b/mathics/main.py
@@ -15,9 +15,9 @@ from mathics import version_string, license_string, __version__
 from mathics.builtin.trace import TraceBuiltins, traced_do_replace
 from mathics.core.definitions import autoload_files, Definitions, Symbol
 from mathics.core.evaluation import Evaluation, Output
-from mathics.core.expression import strip_context
 from mathics.core.parser import MathicsFileLineFeeder, MathicsLineFeeder
 from mathics.core.rules import BuiltinRule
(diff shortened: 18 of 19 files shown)


mathics/core/expression.py
```diff
@@ -909,7 +907,7 @@ def rules():
         for rule in rules():
             result = rule.apply(new, evaluation, fully=False)
             if result is not None:
-                if isinstance(result, BoxConstruct):
```
rocky, line 912: When this was added an alarm should have gone off - Expression needs knows that there is something called BoxConstruct? The fact that it needed to be imported in that funky way should have raised its own alarm. 

The code is trying to figure out whether we need to consider expanding an expression further, and there should be a way to do that without knowing anything more that this. 

There might be a better way to do this than it check `ExpandOnce` instead. And perhaps there is a better name for this as well.  Some version of "never try to expand",  or "result is final"?

mathics/builtin/assignments/internals.py
```diff
@@ -153,7 +153,7 @@ def find_tag_and_check(lhs, tags, evaluation):
 
 
 def unroll_patterns(lhs, rhs, evaluation):
-    if type(lhs) is Symbol:
```
rocky, line 156: Even if we hadn't added Constant, this idiom is deprecated. Probably for exactly the reason it needs to change here. When Constant was added, this code is now wrong as is, but had `isinstance()` been used in stead no change would have been needed.

mathics/builtin/box/compilation.py
```diff
@@ -1,9 +1,10 @@
 # -*- coding: utf-8 -*-
 
 from mathics.builtin.base import BoxConstruct
+from mathics.core.symbols import ExpandOnce
 
 
-class CompiledCodeBox(BoxConstruct):
+class CompiledCodeBox(BoxConstruct, ExpandOnce):
```
rocky, line 7: I don't understand why, but ExpandOnce has to be included here, even though it is mentioned in BoxConstruct. I googled and see one might try `issubclass()` instead of `isinstance()`, but I actually see a benefit in listing it ere even if it were unnecessary so as to make the intent explicit.

mathics/core/symbols.py
```diff
@@ -807,26 +816,40 @@ def __getnewargs__(self):
...
+SymbolMachinePrecision = Constant("MachinePrecision")
+SymbolMakeBoxes = Constant("System`MakeBoxes")
+SymbolMaxPrecision = Constant("$MaxPrecision")
+SymbolMinPrecision = Constant("$MinPrecision")
+SymbolN = Constant("System`N")
+
+# FIXME: LinearModelFit[{{2, 1}, {3, 4}, {5, 3}, {7, 6}}, x, x] ;
+# fails if we use a Constant below. Fix whatever is making this fail.
+# SymbolNull = Constant("System`Null")
 SymbolNull = Symbol("System`Null")
```
rocky, line 844: Probably the most prevalent place where we do go through an unneeded recheck of a value looking for a fixed point is when the value is Null, because that is probably the most common value returned. 

I don't understand what causes a failure here, but if this were fixed it would probably have a small impact on the performance.

rocky:
I wanted to make progress on exposing SymPy objects that they can be manipulated more directly by the user and directed to be used in specific computations. 

I see there is already a SymPyObject. But it seems to contain the entire collection of SymPy kinds of things rather than information about a single expression.   I think all of this is very promising. However it is sad that too often when I try to do something I feel held back by the confusion and mess that we have here. 

It looks like we probably have most of what we need, but the organization the parts seems a little weird.

mathics/core/symbols.py
rocky: It would be nice to see how the effect these changes have on the performance. Print tests, seem to indicate on startup we are making additional trips quite a bit.

TiagoCavalcante: There are some `is SymbolNull` in the code, and `Constant("Symbol`Null")` is different than `Symbol("Symbol`Null")` because expressions are first evaluated to symbols, not to a specific class. A straight `is SymbolNull` doesn't convert the symbol `Null` to an instance of `Contant`. I don't believe this is a bug, but a defined behavior.

mathics/algorithm/parts.py
```diff
@@ -18,7 +18,7 @@
     PartRangeError,
 )
 
-SymbolNothing = Symbol("Nothing")
+SymbolNothing = Constant("Nothing")
```
TiagoCavalcante, line 21: Here and in other places `Symbol` is fine. `Constant` is not a singleton, but `Symbol` is.

mathics/core/expression.py
TiagoCavalcante: What about `FinalForm`?

TiagoCavalcante (changes requested):
@rocky except for the comments, LGTM.
I'm just not sure if this should be an attribute or a class.

mathics/core/expression.py
rocky: I don't have a strong feeling one way or another. Looking that the wolfram docs Form often refers to the way something is presented. However in the code we use Form as a way of matching structure. And I don't think that's quite right either. When Symbol "x" gets evaluated to 5 its shape stays the same, it is the _evaluation_ here that we are avoiding. So it si more like FinalEvaluation.

mmatera:
@rocky, this is interesting, but I want to check if there are side effects that break the compatibility with wl. Give me a couple of days to analize these changes.

mathics/algorithm/parts.py
rocky: We need to distinguish when an object can change its final value or not. In many languages like lisp, Symbol is intended to be something whose value isn't inspected. The item is tested in totality.  

In Mathematica though Symbol is often used for this other idea of something that holds an unknown value which _can_ expected to change over time. It is that property we want to separate. Whether you call it Symbol("Nothing") or Constant("Nothing") it is value doesn't change. 

In fact though there are 3 distinct kinds of objects: those that have a single name an there is no more information and this doesn't change. True and False are like that. Then there are things that have a value but it is expected it shouldn't be altered after its _initial_ value. Numbers and string are like that. And then there are these placeholders for values that can have values, usually these start out unknown and somethings the value gets resolved later, and it might even get changed to a different value later.  From the evaluation standpoint that's what we are trying to distinguish here. 

Singleton is along a different dimension and has to do with implementation efficiency - whether all uses of a value share the same location or now. Since it seems preferable that Constant should be implemented a singleton kind of object that should be done.

rocky:
> @rocky except for the comments, LGTM. I'm just not sure if this should be an attribute or a class.

I thought about an attribute on a class, but the thing that struck me was that `isinstance()` has to be a pretty fast and simple test. As a class variable, I guess you'd have to see the  attribute of the class exists or not, right? And would you also check on the value?   So exactly how would this work? 

And I suppose adding a decorator on the class would be Pythonic.

rocky:
> @rocky, this is interesting, but I want to check if there are side effects that break the compatibility with wl. Give me a couple of days to analize these changes.

No problem. It looks like this code needs revision anyway. I'll mark it as a draft then.

rocky:
> > @rocky, this is interesting, but I want to check if there are side effects that break the compatibility with wl. Give me a couple of days to analize these changes.
> 
> No problem. It looks like this code needs revision anyway. I'll mark it as a draft then.

But I will probably add a PR for the funky imports in the wrong place. There shouldn't be a reason for that.

mathics/core/symbols.py
rocky: Ok. I think I understand and this make sense.  I noticed in parsing Symbol("System`Null") was used as it should be. And it makes sense that initially values are their own distinct placeholders of Null values. 

Then we just need to make note of this information in a comment here.

mathics/core/symbols.py
```diff
@@ -14,6 +14,15 @@
 sympy_symbol_prefix = "_Mathics_User_"
 sympy_slot_prefix = "_Mathics_Slot_"
 
+# This feels like a hack. It generalizes the method expression used in expression regarding
```
mmatera, line 17: This looks more like a property than a class, isn't it?

mathics/core/symbols.py
```diff
@@ -807,26 +816,40 @@ def __getnewargs__(self):
         return (self.name, self.sympy_dummy)
 
 
+class Constant(Symbol, ExpandOnce):
```
mmatera, line 819: About this, my feeling is again that to be "Constant" should be a property of a `Symbol` instead of a daughter class.

mathics/core/symbols.py
mmatera: My idea about "singletonize" symbols was to make faster the "sameQ" comparisons: to check that a leaf in an expression is `System`Null` requires just to check its memory position. In principle, with this new implementation that also works (just changes the class). However, this would require an extra step of parsing/evaluating `Symbol(Null)->Constant(Null)`.

mathics/core/subexpression.py
```diff
@@ -143,7 +143,7 @@ def to_expression(self):
         parent = self.parent
         p = self.position
         if p == 0:
-            if type(parent) is Symbol:
+            if isinstance(parent, Symbol):
```
mmatera, line 146: I used here `type` instead of `isinstance` because it is a little bit faster: `isinstance` checks all the inheritance hierarchy. For this, I exploded that `Symbol` didn't have daughter classes.

mathics/core/expression.py
mmatera: `BoxConstruct` was more than just a flag over an `Expression` to avoid reevaluation: It was something that shares a part of its interface with `Expression`, but 1. stores information in a different way, and 2. has an extra interface to produce certain kind of outputs (like Graphics, Graphics3d, or another kind of still not implemented objects, like waveform sounds or manipulable expressions). This class has members of the form `boxes_to_*` that produces the representation of the Box structure.    What maybe makes sense is to tag (by adding a property or by inheritance) `BoxConstruct` as `ExpandOnce`

mathics/builtin/box/compilation.py
mmatera: Again, because of the hackiness needed to check it, and since `ExpandOnce` does not define a different interface, shouldn't be simpler to add a property to the class?

mathics/builtin/assignments/internals.py
mmatera: I just used this because the check is faster if `lhs` is not a `Symbol`.

mathics/algorithm/parts.py
mmatera: @rocky, in WL, a `Symbol` is an atomic expression that can have associated rules and attributes. There are just a few of symbols (like `True`, `False` or `List`) (those which have the attribute `Locked`) which can not be assigned under any circumstance. All the other symbols can be unproctected and assigned with a value. For example
``
In[1]:=Sin[Pi]
Out[1] = 0
In[2]:=Unprotect[Pi]; Pi=4; Sin[Pi]
Out[2] = Sin[4]
``
Notice also that the attribute `Constant` does not mean that a symbol has always the same value, but that in expressions like ``Dt[F[Pi,x]]`` the derivative regarding  `Pi` always vanishes.

mathics/core/symbols.py
rocky: Yes, we should create a class decorator for this.

mathics/core/subexpression.py
rocky: Please don't do this. Lint checkers will tell you it is wrong. Better again would be to use a property on the type. What is really being asked for is a property like does this class have a leaf?

mathics/algorithm/parts.py
rocky: Ok thanks for chekcing. But we still have things like Numbers, Integers, Reals, Complex, Strings with constant or literal values.

mathics/core/symbols.py
rocky: > My idea about "singletonize" symbols was to make faster the "sameQ" comparisons: to check that a leaf in an expression is `System`Null`requires just to check its memory position. In principle, with this new implementation that also works (just changes the class). However, this would require an extra step of parsing/evaluating`Symbol(Null)->Constant(Null)`.

This idea is really part of a larger principle found in Lisp and other programing languages (Python has "intern" for example),  that strings and symbols are different.  A string can be thought of as composition of characters that can be selected and mutated. For example I can form a new string by adding letters or extracting letters. 

A symbol has the property name, and you generally don't want to work on the name other than check that it is the same as some other name. That is why in Lisp and other languages that SameQ which is modelled on Lisp's Eq works that way and is effient. It is more about the concept that affords an efficient implementation rather than the other way around. 

It is unfortunate and possibly confusing that Symbol has two meanings the Lisp meaning described above and the meaning in WL where it is a symbolic placeholder for a value. 

We should be making this distinction clearer in docstrings and comments, especially as it has caused confusion among us. 

Also what is needed is a reevaluation of what properties are needed for Symbols of the Mathematica kind and Symbols of the Lisp kind (just a placeholder for a name of a thing). We don't need all all the methods in the Lisp Symbol that Mathics Symbol currently has. 

And then there is concept of a Literal value (which also is being overloaded by the word Constant). That too has certain operations that make sense on it and some that don't make sense on it. And the ones that don't make sense but are a part of Symbol should be separated. 

This is a mess.

mathics/algorithm/parts.py
TiagoCavalcante: @rocky three problem is: with the Constant attribute we can easily use non constant and constant symbols. If we singletonize Constant we would need to check for symbols in 4 dictionaries (we want it to be just 1).

TiagoCavalcante:
> thought about an attribute on a class, but the thing that struck me was that isinstance() has to be a pretty fast and simple test. As a class variable, I guess you'd have to see the attribute of the class exists or not, right? And would you also check on the value? So exactly how would this work?

The attributes-bitset idea was implemented to make attributes as fast as they can be, to check if a class has an attribute you just do a bitwise and, I wouldn't be surprised if this is faster than `isinstance`.

mathics/algorithm/parts.py
mmatera: In WL, `Symbol`s do not have a "value" (or any other attribute, apart its name). This are atoms (like numbers and strings) that can compose an `Expression`. The specific aspect of a `Symbol` is that they have associated `Definition`s. 

 In WMA, each `Symbol` has a unique `Definition`, with its attributes and replace rules ("values"). In Mathics, a `Symbol` could have at the same time several `Definition`s (because we handle different "sessions" in the same process). 

Indeed, you can erase a `Definition` at some point, but you can not erase a `Symbol`, since it can be a part of existent `Expression`s. 

Comming back to the motivation of this PR: `Symbol`s are not `Constants` or `Variable` by itself. If a `Symbol` represents a `Constant` or not, depends on its `Definition`, which can be different in different sessions.

One way to reduce the overhead of the evaluation sequence  Expression-> lookup_name->definition->rules->apply rule  would be then to change the way in which we implement different sessions (which is important for Mathics-Django) in a similar way in which WMA works: having kernels as an independent process. Then, on each instance, each `Symbol` would have just one `Definition`, and hence, a `Symbol` can be tagged as `Constant`.

mathics/core/subexpression.py
mmatera: @rocky, I am sure you are right about this. I just was exposing the motivation I had to use this syntax. On the other hand, the "fragility" was also a way to alert about a change in the object model: In the current OM, it does not make any sense to specialize the `Symbol` object. Symbols are just references to `Definition`s.

mathics/algorithm/parts.py
rocky: What is becoming clear is that there are three separate concepts that Mathics uses and confuses some of these. 

`Symbol` can be Mathics' class object for a WL Symbol. But in implementing this and in _implementing_ expressions and the system in general we have the concept of a Lisp kind of Symbol which I suggest calling `SymbolName`. Think of it at as an object where the only thing you can do is ask to bring it up and see if it is the same as some other name. And there is the concept of something that holds a unchangeable literal value, let's call that `LiteralValue`. 

Literal values never need reevaluation. The value that comes out when you evaluate Integer(5) or String("Foo") or True are literal values.

SymbolNames never need reevaluation when definitions haven't changed which is most of the time. 

There should be distinct implementations of each of these and we should go over the classes so that the properties of each is distinct.

mathics/core/subexpression.py
rocky: @mmatera I think you are still confused. A WL symbol is composed of an atomic or non decomposable name and a value that is listed under than name. Once the name is defined it can't be changed. Think if it as analogous in Python as defining a function. That function has a name that can't be changed. You can decide never to use that name, or assign have that function name hidden, or assign a variable name initially to that function name but then change it. But the function name for a code object never changes.  And it is in that sense that when we write Symbol("Pi") we what we really mean is the name of the function that was created.

mmatera: > @mmatera I think you are still confused. A WL symbol is composed of an atomic or non decomposable name and a value that is listed under than name.
`Symbol`s do not have a value.  A `Symbol` has a `Definition`, which you access by its name.


> Once the name is defined it can't be changed. Think if it as analogous in Python as defining a function, e.g `def Pi(...)`. That function has a name that can't be changed. You can decide never to use that name, or never use the function, or assign a variable name initially to that function name but then change it. But the function name for a code object never changes. 

A fully agree about this: this is why I have insisted in deal with symbols as singletons.  

> And it is in that sense that when we write Symbol("Pi") we what we really mean is the name of the function that was created. The way a value is associated with that name of course can change.
> 
> And that is why changing `Symbol(Pi)` to `Constant(Pi)` has no effect when function return values of the function change. In both cases you asked using the name `Pi`. There is no way to use a different name but have that _name_ get swapped under the covers to `Pi`.

OK, so the idea is: the difference between `Symbol(Pi)` and `Constant(Pi)` is in the implementation of the `evaluate` method: the second just tries to evaluate if the associated definition has changed, isn't it?

mmatera:
> > thought about an attribute on a class, but the thing that struck me was that isinstance() has to be a pretty fast and simple test. As a class variable, I guess you'd have to see the attribute of the class exists or not, right? And would you also check on the value? So exactly how would this work?
> 
> The attributes-bitset idea was implemented to make attributes as fast as they can be, to check if a class has an attribute you just do a bitwise and, I wouldn't be surprised if this is faster than `isinstance`.

The problem with this is that you can use `SymbolPi` in many expressions, and is always the same symbol. So, if you set a property (bitset or whatever) you change all the instances in all the expressions at the same time.

mathics/core/subexpression.py
rocky: > A fully agree about this: this is why I have insisted in deal with symbols as singletons.

What you seem to have done a number of times, is rediscover a familiar concept in programming languages and come up with a possible solution, but not as good as a solution as what has been done before where you to have understood the familiar concept. Woflram modeled parts of Mathematica on Lisp. 

Recall that I started the proliferation of `Symbol(foo)`.  It was my confusion with Lisp Symbol and not realizing how different WL Symbol is and not realizing at the thing that the two should should not be confused. 

What I am suggesting is changing `Constant()` to `SymbolName()`.  While `SymbolName()` is a concept that users see implicitly, and could be a way to describe behavior, it definitely has an aspect and impact about the way we implement expressions.

This kind of thing is analogous to the concept of a "Literal".  For example google "Python literal" and you'll get the idea of some other analogous piece of a programming language that appears and is a concept in the implementation of some language. However in Python there is no "literal" type and you can declare or define something as literal other than by writing the literal value in some way, e.g. 5 or [1, 2, 3],  {"foo":  1}, {"bar"}.

rocky: Also there should be a separate LiteralValue class which is a parent class for things like Integer, Real, String.

TiagoCavalcante:
> The problem with this is that you can use SymbolPi in many expressions, and is always the same symbol. So, if you set a property (bitset or whatever) you change all the instances in all the expressions at the same time.
@mmatera I don't quite understand, if SymbolPi calls the class Pi then it is an instance of `ExpandOnce`, and in this case, we could change the parent class `ExpandOnce` to an attribute. If SymbolPi doesn't call the class Pi then it is not an instance of `ExpandOnce`, so changing `ExpandOnce` to an attribute would work.
In both ways seems having an attribute would work.

rocky:
> > The problem with this is that you can use SymbolPi in many expressions, and is always the same symbol. So, if you set a property (bitset or whatever) you change all the instances in all the expressions at the same time.
> 
> @mmatera I don't quite understand, if SymbolPi calls the class Pi then it is an instance of `ExpandOnce`, and in this case, we could change the parent class `ExpandOnce` to an attribute. If SymbolPi doesn't call the class Pi then it is not an instance of `ExpandOnce`, so changing `ExpandOnce` to an attribute would work. In both ways seems having an attribute would work.

It would help me if you'd implement whatever solution you want so we could try it and talk about it in concrete terms. 

It doesn't have to be a complete implementation changing everything, just enough to get across the idea and since everyone is concerned about performance, benchmark the various implementation methods. Thanks.

mathics/core/subexpression.py
mmatera: > > A fully agree about this: this is why I have insisted in deal with symbols as singletons.
> 
> What you seem to have done a number of times, is rediscover a familiar concept in programming languages and come up with a possible solution, but not as good as a solution as what has been done before were you to have understood the familiar concept. 

Indeed! I am a physicist, not a computer scientist! Sometimes, I realize where to look for implementing a concept, but most of the time, I have to rediscover the wheel to see where is the tireshop :)

> Woflram modeled parts of Mathematica on Lisp.

I know that (barely), but I am not an expert in Lisp. 

> 
> Recall that I started the proliferation of `Symbol(foo)`. It was my confusion with Lisp Symbol and not realizing how different WL Symbol is and not realizing at the time that the two should not be confused, and that was happening with this proliferation.
> 

What is exactly the problem with the `Symbol(foo)` definitions?
 


> What I am suggesting is changing `Constant()` to `SymbolName()`. While `SymbolName()` is a concept that users see implicitly, and could be a way to describe behavior, it definitely has an aspect and impact about the way we implement expressions. A `Symbol` has a `SymbolName` and a `Definition`.
> 

OK, that could work...


> This kind of thing is analogous to the concept of a "Literal". For example google "Python literal" and you'll get the idea of some other analogous piece of a programming language that appears and is a concept in the implementation of some language. However in Python there is no "literal" type and you can declare or define something as literal other than by writing the literal value in some way, e.g. 5 or [1, 2, 3], {"foo": 1}, {"bar"}.

not sure to follow you, but let's try it!

rocky: > What is exactly the problem with the Symbol(foo) definitions?

One problem of using `Symbol()` when you mean `SymbolName` or Lisp Symbol we have already seen: we now have a circular import, because as a result of pulling in more stuff than is needed in an object.

SymbolNames have a more limited function and use that Symbols.  When you create a large object the encompases everything it leads to undue complexity and it prohibits splitting up the code into smaller more easily testable parts that are more easily understood.  

What would the world be like if we always had to talk about quantum interactions when we refer to biology rather than dividing things into atoms, elements, chemical elements, biology and so on.

mathics/core/symbols.py
mmatera: @rocky, is this the notion of a Lisp symbol that you are talking about? https://www.cs.cmu.edu/Groups/AI/html/cltl/clm/node107.html#SECTION001400000000000000000

TiagoCavalcante:
> It would help me if you'd implement whatever solution you want so we could try it and talk about it in concrete terms.
> It doesn't have to be a complete implementation changing everything, just enough to get across the idea and since everyone is concerned about performance, benchmark the various implementation methods. Thanks.

Attributes are 1/100 of a second slower:
```python
>>> import timeit
>>> class x:
...     attributes = 0b0101001010
...
>>> attribute = 0b000001
>>> class y:
...     pass
... 
>>> class z(y):
...     pass
...
>>> obj = z()
>>> timeit.timeit(lambda: isinstance(z, y))
0.1337610839982517
>>> timeit.timeit(lambda: x.attributes & attribute)
0.14236413198523223
```

Anyway, making it a class seems a bit odd, like "oh, there's one more way to do something", while an attribute sounds like "oh, I can mark my functions to make them faster".

Also, @rocky maybe it is better to split SymbolName and ExpandOnce in different PRs.

mmatera:
> > It would help me if you'd implement whatever solution you want so we could try it and talk about it in concrete terms.
> > It doesn't have to be a complete implementation changing everything, just enough to get across the idea and since everyone is concerned about performance, benchmark the various implementation methods. Thanks.
> 
> Attributes are 1/100 of a second slower:
> 
> ```python
> >>> import timeit
> >>> class x:
> ...     attributes = 0b0101001010
> ...
> >>> attribute = 0b000001
> >>> class y:
> ...     pass
> ... 
> >>> class z(y):
> ...     pass
> ...
> >>> obj = z()
> >>> timeit.timeit(lambda: isinstance(z, y))
> 0.1337610839982517
> >>> timeit.timeit(lambda: x.attributes & attribute)
> 0.14236413198523223
> ```
> 
> Anyway, making it a class seems a bit odd, like "oh, there's one more way to do something", while an attribute sounds like "oh, I can mark my functions to make them faster".
> 
> Also, @rocky maybe it is better to split SymbolName and ExpandOnce in different PRs.

This depends on how complicated is the class hierarchy. In any case, I think @rocky is right: the best we can do would be to split the basic implementation of  `Symbol` (let's say, `SymbolName`) from something like `DefinedSymbol` which aggregates both `SymbolName` and the corresponding definition, and the methods for evaluation.

rocky:
> > It would help me if you'd implement whatever solution you want so we could try it and talk about it in concrete terms.
> > It doesn't have to be a complete implementation changing everything, just enough to get across the idea and since everyone is concerned about performance, benchmark the various implementation methods. Thanks.
> 
> Attributes are 1/100 of a second slower:
> 
> ```python
> >>> import timeit
> >>> class x:
> ...     attributes = 0b0101001010
> ...
> >>> attribute = 0b000001
> >>> class y:
> ...     pass
> ... 
> >>> class z(y):
> ...     pass
> ...
> >>> obj = z()
> >>> timeit.timeit(lambda: isinstance(z, y))
> 0.1337610839982517
> >>> timeit.timeit(lambda: x.attributes & attribute)
> 0.14236413198523223
> ```

Ok but what you are timing are mechanisms. I was thinking more along the line of actually changing the code base. From what it is in this PR to some other mechanism for maybe one class. 

> 
> Anyway, making it a class seems a bit odd, like "oh, there's one more way to do something", while an attribute sounds like "oh, I can mark my functions to make them faster".
> 

What bothers me about all of this discussion is that we are worried about this and whether `type(x) == Name` is faster than `isinstance()` but missing the bigger stuff like not making performing an evaluation for literals which is 10 x 100 times slower.

And in the process of the micro efficiencies we've made the code more fragile, no better tested or document. 

> Also, @rocky maybe it is better to split SymbolName and ExpandOnce in different PRs.

As far as I am concerned I brought up the issues here, the silly imports thing is fixed in another PR. I think the way we mark classes that don't need further expanding can go in another PR. It doesn't have to be part of this one. 

Cleaning up Symbol, SymbolName, and Literal classes are also  separate PRs and I hope that gets done.

mmatera:
> What bothers me about all of this discussion is that we are worried about this and whether `type(x) == Name` is faster than `isinstance()` but missing the bigger stuff like not making performing an evaluation for literals which is 10 to 100 times slower.
> 

@rocky, to be super clear: I used that "idiom"  not because I thought that was the panacea of the performance, but 1. if I have several options, and I do not have a good reason, I take what seems clear and faster. 2. Again, in the current logic, `Symbol` is a class that shouldn't have inheritance. I saw that "fragility" as an "alarm". But please, if you consider that the way using `isinstance` is clearer, go ahead and change it!





> And in the process of the micro efficiencies we've made the code more fragile, no better tested or document.
> 
> > Also, @rocky maybe it is better to split SymbolName and ExpandOnce in different PRs.
> 
> As far as I am concerned I brought up the issues here, the silly imports thing is fixed in another PR. I think the way we mark classes that don't need further expanding can go in another PR. It doesn't have to be part of this one.
> 
> Cleaning up Symbol, SymbolName, and Literal classes are also separate PRs and I hope that gets done.

TiagoCavalcante:
> Ok but what you are timing are mechanisms. I was thinking more along the line of actually changing the code base. From what it is in this PR to some other mechanism for maybe one class.

@rocky if I benchmarked the real code you won't see the difference (aka. there is no real difference).

> What bothers me about all of this discussion is that we are worried about this and whether type(x) == Name is faster than isinstance() but missing the bigger stuff like not making performing an evaluation for literals which is 10 to 100 times slower.
> And in the process of the micro efficiencies we've made the code more fragile, no better tested or document.

I'm worried about runtime performance here, but I'm worried about development performance.  
I think that making `ExpandOnce` a class is bad because this is the type of thing that can take the time of someone in the future. I think a newcomer won't guess that he/she can use inheritance in everything but `Symbol` and `ExpandOnce`.

Also, with attributes, a user can mark Mathics code to make it faster.

Anyway, I guess we already discussed too much for such a small thing, so do whatever you and @mmatera want.

rocky:
> > What bothers me about all of this discussion is that we are worried about this and whether `type(x) == Name` is faster than `isinstance()` but missing the bigger stuff like not making performing an evaluation for literals which is 10 to 100 times slower.
> 
> @rocky, to be super clear: I used that "idiom" not because I thought that was the panacea of the performance, but 1. if I have several options, and I do not have a good reason, I take what seems clear and faster. 2. Again, in the current logic, `Symbol` is a class that shouldn't have inheritance. I saw that "fragility" as an "alarm". But please, if you consider that the way using `isinstance` is clearer, go ahead and change it!
> 

I am now of the view now the more important thing I think is to come up with a clearer and simpler and more modular design. From that we can work on a how best to implement that or determine what the bottlenecks are.

You were going to finish up the work you had started and then you were going to let it sit. There was a number of other things that you are better suited for and that are unfinished. 

After coming to a stopping, I'd like the same amount of time undisturbed to look things over, review and discuss.

rocky:
I've looked over https://reference.wolfram.com/language/ref/Symbol.html?q=Symbol and that does indeed match the Lisp Symbol idea.  When I compare it to Mathic's Symbol class: 

````
class Symbol(Atom):
    def __new__(cls, name, sympy_dummy=None):
    def __str__(self) -> str:
    def do_copy(self) -> "Symbol":
    def get_head(self) -> "Symbol":
    def get_head_name(self):
    def boxes_to_text(self, **options) -> str:
    def to_sympy(self, **kwargs):
    def to_python(self, *args, **kwargs):
    def default_format(self, evaluation, form) -> str:
    def get_attributes(self, definitions):
    def get_name(self) -> str:
    def is_symbol(self) -> bool:
    def get_sort_key(self, pattern_sort=False):
    def equal2(self, rhs: Any) -> Optional[bool]:
    def sameQ(self, rhs: Any) -> bool:
    def __eq__(self, other) -> bool:
    def __ne__(self, other) -> bool:
    def replace_vars(self, vars, options={}, in_scoping=True):
    def has_symbol(self, symbol_name) -> bool:
    def is_true(self) -> bool:
    def is_numeric(self, evaluation=None) -> bool:
    def __hash__(self):
    def user_hash(self, update) -> None:
    def __getnewargs__(self):
 ```
I am seeing a lot of extra stuff that is not related to just the string to name part or testing symbol equality. In some cases it seems to about working with symbol-like _defnitions_ like is `is_xxx`.  A number of these functions just don't make sense at all like the get_head or boxes to text. 

Also, I see that `SymbolName` is a WL function which turns the name into a string. So I guess we shouldn't use `SymbolName`.  

At some point, I'd like to just go over this code and try to rationalize and explain what is there.

mmatera:
> I've looked over https://reference.wolfram.com/language/ref/Symbol.html?q=Symbol and that does indeed match the Lisp Symbol idea. When I compare it to Mathic's Symbol class:
> 
> ```
> class Symbol(Atom):
>     def __new__(cls, name, sympy_dummy=None):
>     def __str__(self) -> str:
>     def do_copy(self) -> "Symbol":
>     def get_head(self) -> "Symbol":
>     def get_head_name(self):
>     def boxes_to_text(self, **options) -> str:
>     def atom_to_boxes(self, f, evaluation) -> "String":
>     def to_sympy(self, **kwargs):
>     def to_python(self, *args, **kwargs):
>     def default_format(self, evaluation, form) -> str:
>     def get_attributes(self, definitions):
>     def get_name(self) -> str:
>     def is_symbol(self) -> bool:
>     def get_sort_key(self, pattern_sort=False):
>     def equal2(self, rhs: Any) -> Optional[bool]:
>     def sameQ(self, rhs: Any) -> bool:
>     def __eq__(self, other) -> bool:
>     def __ne__(self, other) -> bool:
>     def replace_vars(self, vars, options={}, in_scoping=True):
>     def has_symbol(self, symbol_name) -> bool:
>     def evaluate(self, evaluation):
>     def is_true(self) -> bool:
>     def is_numeric(self, evaluation=None) -> bool:
>     def __hash__(self):
>     def user_hash(self, update) -> None:
>     def __getnewargs__(self):
> ```
> 
> I am seeing a lot of extra stuff that is not related to just the string to name part or testing symbol equality. In some cases it seems to about working with symbol-like _defnitions_ like is `is_xxx`. A number of these functions just don't make sense at all like the get_head or boxes to text.
> 
> Also, I see that `SymbolName` is a WL function which turns the name into a string. So I guess we shouldn't use `SymbolName`.
> 
> At some point, I'd like to just go over this code and try to rationalize and explain what is there.

So, the plan would be to keep in BaseExpression all the methods related to handle the data structure, and move all the evaluation stuff (all what depends on the Definition) to non-member functions or other classes. This would mean a drastic change in the interface, but I think is the right direction.

rocky:
> > At some point, I'd like to just go over this code and try to rationalize and explain what is there.
> 
> So, the plan would be to keep in BaseExpression all the methods related to handle the data structure, and move all the evaluation stuff (all what depends on the Definition) to non-member functions or other classes. This would mean a drastic change in the interface, but I think is the right direction.

Write such a new Symbol function say SymbolWL or SymbolNew. Get that working the really simple way that it seems to be done in WL. See if we can get Symbol to use that. 

Now the big but largely mechanical task... Rename Symbol to SymbolMathics (I was going to write SymbolMess), and then rename SymbolWL to Symbol.  Then iterate on SymbolMathics to figure out how to get it to match concepts in WL better. Or split it up into logical pieces. At this point though it wouldn't have the pure Symbol parts.

TiagoCavalcante:
> > Ok but what you are timing are mechanisms. I was thinking more along the line of actually changing the code base. From what it is in this PR to some other mechanism for maybe one class.
> 
> @rocky if I benchmarked the real code you won't see the difference (aka. there is no real difference).
> 
> > What bothers me about all of this discussion is that we are worried about this and whether type(x) == Name is faster than isinstance() but missing the bigger stuff like not making performing an evaluation for literals which is 10 to 100 times slower.
> > And in the process of the micro efficiencies we've made the code more fragile, no better tested or document.
> 
> I'm worried about runtime performance here, but I'm worried about development performance. I think that making `ExpandOnce` a class is bad because this is the type of thing that can take the time of someone in the future. I think a newcomer won't guess that he/she can use inheritance in everything but `Symbol` and `ExpandOnce`.
> 
> Also, with attributes, a user can mark Mathics code to make it faster.
> 
> Anyway, I guess we already discussed too much for such a small thing, so do whatever you and @mmatera want.

So, @rocky and @mmatera, what is decided?

rocky:
My take: the Symbol class should look more like the Lisp and WL Symbol definition. It just encapsulates a name, no more. 

The Symbol class in Mathics may be encapsulating some other idea, I don't know what that is. But if it has a WL equivalent it should be renamed to whatever that is. And split up so that it uses the simpler WL Symbol class.

Doing this is going to eliminate the circular dependencies because many times what is needed is WL/Lisp Symbol that the heavier more complex class. 

More broadly, the code base should be gone over at the lowest levels to try to align it better with the publicly available information on how WL works. A number of the terms used in Mathics feel awkward to me. The name Instanceable is horrible. 
I don't have very clear ideas as to me what Atom (which is now an experimental WL concept), Builtin, and Instanceable, and Symbol are and how those make sense or make good building block.  

In my view we have some fundamental rot at the lowest levels. And bit twiddling like we've been doing has been making it worse, not better.

mmatera:
> I wanted to make progress on exposing SymPy objects that they can be manipulated more directly by the user and directed to be used in specific computations.
> 
> I see there is already a SymPyObject. But it seems to contain the entire collection of SymPy kinds of things rather than information about a single expression.
> 
> I think all of the prospect of saving results as sympy objects and being able to convert to Mathics and back on demand is a promising way to improve performance.
> 
> However it is sad that too often when I try to do something I feel held back by the confusion and mess that we have here.
> 
> It looks like we probably have most of what we need, but the organization the parts seems a little weird.

`SympyObject` is an specialization of `Builtin` to generate definitions for Symbols that uses `Sympy`. Notice that `Builtin` classes are just a way to produce (`builtin`) definitions of symbols, and are instantiated just once, at the moment of creating the `Definitions` object.

I think what you are looking for is something similar to `ListExpression` (#121) which stores expressions as a Sympy object, providing an interface to bring back head and leaves.

rocky:
It looks like I wrote the `SymPyObject` class so this mess is mine. It looks to me there is much that is wrong about this. 

This looks like it has information about SymPy _functions_. SymPy can represent numbers and compositions of functions (which is analogous to an Expression). And we see that in some cases there can be more than one sympy function that can be used to implement a Mathics function.  

So there is some very narrow and thus probably wrong view of what a Sympy object is.





When I get a chance, I will try to figure out what i was thinking here.

mmatera:
> It looks like I wrote the `SymPyObject` class so this mess is mine. It looks to me there is much that is wrong about this.
> 
> This looks like it has information about SymPy _functions_. SymPy can represent numbers and compositions of functions (which is analogous to an Expression). And we see that in some cases there can be more than one sympy function that can be used to implement a Mathics function.
> 
> So there is some very narrow and thus probably wrong view of what a Sympy object is.
> 
> When I get a chance, I will try to figure out what i was thinking here.

I think this is a second order problem. SympyExpressions could be also problematic, because there are things that sympy can not evaluate (let's say, `Sin[NIntegrate[Sin[x],{x,0,2}]]`). Moreover, if you convert  `Sin[a]` to a sympy object, and afterward you set the value of `a`, sympy does not have a way to update that value in a further evaluation. This is why the current implementation requires to bring back the sympy expressions to mathics on each evaluation step.

rocky:
> > So there is some very narrow and thus probably wrong view of what a Sympy object is.
> > When I get a chance, I will try to figure out what i was thinking here.
> 
> I think this is a second order problem.

I don't know what "second order" means here and how that relates to the below. I guess you are suggesting that the below is a fundamental problem and this is a consequence of that? If so, I don't see the relation. 

 > SympyExpressions could be also problematic, because there are things that sympy can not evaluate (let's say, `Sin[NIntegrate[Sin[x],{x,0,2}]]`).

You asked how you can help? One way is when you give an example ask, is this the _simplest_ example that demonstrates a _single_ problem? If not work on reducing problems to the simplest example that demonstrates the problem. 

I think one of the problems in that example can be described without the enclosing `Sin[]` - the iteration is over the `Nintegrate` is probably good enough. 

> Moreover, if you convert `Sin[a]` to a sympy object, and afterward you set the value of `a`, sympy does not have a way to update that value in a further evaluation. This is why the current implementation requires to bring back the sympy expressions to mathics on each evaluation step.

If that is the case, then it should be fixed and I suspect and there are simple ways to do so. 

But this does explain why we do very badly, like 10 time worse or more on anything that iterates. The good news is that I think all of this is solvable in a somewhat straightforward way.

PR #117, opened 2022-01-17
Improving series (rebase #56)

mmatera:
For some reason, #56  and other PRs developed a lot of conflicts against master. There is the squashed merge. @rocky, @TiagoCavalcante do you have any idea about how to recover the other PRs without checking by hand each of the 170+ files?


rocky:
> For some reason, #56 and other PRs developed a lot of conflicts against master. There is the squashed merge. @rocky, @TiagoCavalcante do you have any idea about how to recover the other PRs without checking by hand each of the 170+ files?

I am sure there is but I don't understand exactly what you are asking. Maybe a specific example would help.

TiagoCavalcante:
@mmatera I think this is because of the new master history. Maybe `git cherry-pick` helps.

mmatera:
> I am sure there is but I don't understand exactly what you are asking. Maybe a specific example would help.

I mean: for some reason, the original branch (which involved changes in seven files), at some point, started to have conflicts with 176 files in master. I fixed #56 in this branch, by merging by hand all the files. On the other hand, I would expect that there would be a way to do the merge in a less painful way (let's say, without diving into each file of the project and pick the lines with the new bitset Attribute instead of the old lines with `set` attributes, which are mostly unrelated to the PR)

Issue #67, opened 2021-11-14
failure of Nintegrate for Debye function


recohen:
Attached is an example.  When run under wolframscript:
wolframscript -file t.wls 210 0.0325 55.85
410.8394779818933
Under mathics:
mathics t.wls 210 0.0325 55.85
Part::partw: Part 2 of {} does not exist.
Part::partw: Part 3 of {} does not exist.
Part::partw: Part 4 of {} does not exist.
Part::partw: Part 2 of {} 
...
FindRoot::nnum: The function value is not a number at TD = 500..
ReplaceAll::reps: FindRoot[msA - const phi[t / TD] / (t / TD) / (M TD), {TD, 500.}] is not a valid replacement rule.
TD /. FindRoot[msA - const phi[t / TD] / (t / TD) / (M TD), {TD, 500.}]

Ron Cohen
[t.wls.zip](https://github.com/Mathics3/mathics-core/files/7534115/t.wls.zip)

TiagoCavalcante:
@recohen `$ScriptCommandLine` is `{}` (even when calling Mathics with arguments). @rocky is that a bug?

mmatera:
First, let's fix some issues in the example.

* The command line to call a script with parameters would be, in mathics,
```
mathics -script t.wls -- 210 0.0325 55.85 410.8394779818933
```
* The script `t.wls` should have semicolons at the end of each line:
```
phi[x_?NumberQ]:=(1/x)*NIntegrate[y/(Exp[y]-1),{y,0,x}];
HBARSI=1.054571800 *10^(-34);KBSI=1.38064852*10^(-23);MPSI=1.672621898*10^(-27);
const=10^(20)*9.*HBARSI^2/KBSI/MPSI;
t=ToExpression[$ScriptCommandLine[[2]]];
msA=ToExpression[$ScriptCommandLine[[3]]];
M=ToExpression[$ScriptCommandLine[[4]]];
td=FindRoot[msA==const/M*(phi[t/TD]/(t/TD))/TD,{TD,500.00000}];
Print[TD/.td]
```

With this, `$ScriptCommandLine` works as it should.

The problem now is that `FindRoot` is not working properly with numeric expressions. This is what we need to solve.

mmatera:
Another problem I just realized is about the `ScriptCommandLine` index. This is another issue to fix.

rocky:
If there is more here, please open a new issue with narrowed information.

PR #119, opened 2022-01-18
Fix issue67

mmatera:
From #67, some issues related both to ``FindRoot`` and ``NIntegrate`` become apparent:  
* `D[NIntegrate[...]]` - like were not evaluated.
* ``FindRoot`` didn't handle functions whose derivatives can not be explicitly evaluated.
* ``$ScriptCommandLine`` now has the name of the script as a first element (like in WMA).


mathics/builtin/numbers/calculus.py
```diff
@@ -1287,24 +1320,104 @@ def find_root_secant(f, x0, x, opts, evaluation) -> (Number, bool):
...
+        Check if val is zero upto the precision and accuracy goals
+        """
+        if not val.is_numeric():
+            return False
+        if val.is_zero:
+            return True
+        if acc_goal:
+            if prec_goal:
+                eps = apply_N(
+                    Expression(
```
TiagoCavalcante, line 1355: Maybe create a `SymbolLog`?

TiagoCavalcante (approved):
LGTM

mathics/builtin/system.py
```diff
@@ -345,10 +345,9 @@ def evaluate(self, evaluation):
         except ValueError:
             # not run in script mode
             return Expression(SymbolList)
-
-        return Expression(
-            SymbolList, *(String(arg) for arg in sys.argv[dash_index + 1 :])
-        )
+        scriptname = "" if dash_index == 0 else sys.argv[dash_index - 1]
```
rocky, line 348: This seems totally independent. What's up here? 

This probably should go into a separate PR.

CHANGES.rst
```diff
@@ -4,7 +4,11 @@ CHANGES
 New Builtins
 ============
 * ``TraceEvaluation[]`` shows expression name calls and return values of it argument. The variable ``$TraceEvalution`` when set True will show all expression evaluations.
-
+* Now, ``D`` can act over ``NIntegrate``.
+* ``FindRoot`` was improved for supporting numerical derivatives.
+* ``FindRoot`` (for the ``newton`` method) partially supports ``EvaluationMonitor`` and ``StepMonitor`` options. 
+* ``$ScriptCommandLine`` now has the filename of the script as its first element.
```
rocky, line 10: See below for `$ScriptCommandLine` change.

mathics/builtin/assignments/internals.py
```diff
@@ -153,7 +154,7 @@ def find_tag_and_check(lhs, tags, evaluation):
 
 
 def unroll_patterns(lhs, rhs, evaluation):
-    if type(lhs) is Symbol:
+    if isinstance(lhs, Atom):
```
rocky, line 157: Is the idea that unrolling works if we have number as well and we weren't catching this?

mathics/builtin/numbers/calculus.py
```diff
@@ -58,6 +66,33 @@
...
+def get_accuracy_and_prec(opts: dict, evaluation: "Evaluation"):
+    """
+    Looks at an opts dictionary and tries to determine the numeric values of
+    Accuracy and Precision goals. If not available, returns None.
+    """
+    acc_goal = opts.get("System`AccuracyGoal", None)
+    if acc_goal:
+        acc_goal = apply_N(acc_goal, evaluation)
+        if acc_goal is SymbolAutomatic:
+            acc_goal = from_python(9.0)
```
rocky, line 78: Where does the number 9.0 come from?

mathics/builtin/numbers/calculus.py
```diff
@@ -1407,10 +1535,30 @@ class FindRoot(Builtin):
...
     methods = {
+        "Automatic": find_root_newton,
         "Newton": find_root_newton,
         "Secant": find_root_secant,
     }
 
+    @staticmethod
+    def determine_epsilon(x0, options, evaluation):
+        """Determine epsilon  from a reference value, and from the accuracy and the precision goals"""
+        acc_goal, prec_goal = get_accuracy_and_prec(options, evaluation)
```
rocky, line 1546: My own personal opinion this is the kind of unnecessary coding is that I see a lot of which I find penny wise and pound foolish.

There are two independent parameters, acc_goal and prec_goal and each contributes to the returned value. We have this two if/else's embedded into in another if/else each with some sort of repeated expression given in Mathics expression - like format (i.e. is a little harder to understand than say a Python expression). 

A straightforward and easy-to-understand way to code this is simply not nest and duplicate expressions 

For example: 

```
# from_python(1e-10) is computed once outside and given a name and comment that describes its meaning 

if not (pres_goal or acc_goal): 
     return WHATEVER_FROM_1E_MINUS10_MEANS

eps = abs(x0) *  Integer10 ** (-prec_goal) if pres_goal else Integer0
if acc_goal:
   eps += Integer10 ** (-acc_goal)
return = apply_N(eps)
```

mathics/builtin/numbers/calculus.py
```diff
@@ -1315,12 +1430,23 @@ def sub(evaluation):
...
             evaluation.message("FindRoot", "nnum", x, x0)
             return x0, False
-        # TODO: use Precision goal...
-        if x1 == x0:
-            break
-        x0 = apply_N(x1, evaluation)
-        # N required due to bug in sympy arithmetic
-        count += 1
+
+        # Check convergency:
```
rocky, line 1434: convergency -> convergence

CHANGES.rst
```diff
@@ -4,7 +4,11 @@ CHANGES
 New Builtins
 ============
 * ``TraceEvaluation[]`` shows expression name calls and return values of it argument. The variable ``$TraceEvalution`` when set True will show all expression evaluations.
-
+* Now, ``D`` can act over ``NIntegrate``.
```
rocky, line 7: Let's please separate the ``D`` change in the PR from the ``FindRoot`` change.

rocky:
A thought I keep having after looking over stuff like this. Although I think we should merge this, I keep thinking that what we should be focusing on instead is how to hook into Sympy to do Newton's (or someone else's method) for finding Roots and Minima.

mathics/builtin/system.py
mmatera: @rocky,
the issue contained several aspects. 
- How `FindRoot` handles expressions which can not have an analytic derivative
- How to evaluate derivatives of `NIntegral`
- How we are processing command line parameters.

So I put here all together to see if it solves the test case proposed.

In this particular module, I proposed a fixed for the last one.

mmatera:
> A thought I keep having after looking over stuff like this. Although I think we should merge this, I keep thinking that what we should be focusing on instead is how to hook into Sympy to do Newton's (or someone else's method) for finding Roots and Minima.

Rocky, to use libraries to implement these calculus is  work in progress. Now I am implementing an interface for Scipy, but I wanted to have some feedback from this PR to polish and publish it.

mathics/builtin/numbers/calculus.py
mmatera: @rocky,  OK, I like it! And no, it was not about performance, it was just about having a simplest routine that came to my mind that provides the output I needed. This one is much better!

mathics/builtin/numbers/calculus.py
mmatera: In WMA, `Chop` by default round numbers to 10 significant digits. Here I just look for a number around 10 that provides results accurate enough to pass the tests. In any case, for a final version, should let to the different methods that define the defaults.

mathics/builtin/assignments/internals.py
mmatera: This is something that found by chance: suppose that by mistake, you evaluate
`7/; F[a]:=7`
in master, it raises an unhandled exception. With this change, just an error message.

mathics/builtin/system.py
rocky: Ok. But let's me clear. Users and issues will sometimes be something like: please fix everything that I just encountered.

We should try to split up the issues so that each covers one thing so that thing can be tracked. And similarly in PRs it would be nice to split separate problems into separate PRs for the same reason. They will get through faster. 

And in 6 months when we are looking over looking at a particular line of code and are wondering why it is there and then look at the commit or merge associated with it, it reflects one idea, and not several.

mmatera:
split and enlarged in  #129, #131, #133

PR #121, opened 2022-01-20
List as atom

mmatera:
In this PR, a basic proposal about how to handle `List` evaluations is implemented. Same ideas could be implemented to other frequently used symbols, or certain direct `sympy` expressions.

### DIFF
--- a/mathics/builtin/base.py
+++ b/mathics/builtin/base.py
@@ -611,6 +611,8 @@ class BoxConstruct(InstanceableBuiltin):
 
     def sameQ(self, expr) -> bool:
         """Mathics SameQ"""
+        if isinstance(expr, BoxConstruct):
+            return expr is self
         return expr.sameQ(self)
 
     def is_atom(self):
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,7 +30,7 @@ Internals
 * A bug was fixed relating to the order in which ``mathics.core.definitions`` stores the rules
 * Improved support for ``Series`` Issue #46
 * ``Cylinder`` rendering is implemented in Asymptote
-
+* `List` expressions are stored as `ListExpression` objects, with a simpler `evaluate` method.
 
 Bugs
 ++++
--- a/mathics/builtin/lists.py
+++ b/mathics/builtin/lists.py
@@ -49,6 +49,7 @@ from mathics.core.convert import from_sympy
 from mathics.core.evaluation import BreakInterrupt, ContinueInterrupt, ReturnInterrupt
 
 from mathics.core.expression import Expression, structure
+from mathics.core.lists import ListExpression
 from mathics.core.atoms import (
     ByteArrayAtom,
     Integer,
@@ -75,7 +76,6 @@ from mathics.core.systemsymbols import (
     SymbolFailed,
     SymbolMakeBoxes,
     SymbolRule,
-    SymbolSequence,
 )
 
 from mathics.core.attributes import (
@@ -546,6 +546,13 @@ class List(Builtin):
     """
 
     attributes = locked | protected
+
+    def apply(self, items, evaluation):
+        """{items___}"""
+        if items.get_head() is SymbolSequence:
+            return ListExpression(items._leaves)
+        else:
+            return ListExpression(tuple((items,)))
 
     def apply_makeboxes(self, items, f, evaluation):
         """MakeBoxes[{items___},
--- a/mathics/core/lists.py
+++ b/mathics/core/lists.py
@@ -1,0 +1,80 @@
+# cython: language_level=3
+# -*- coding: utf-8 -*-
+
+from mathics.core.symbols import (
+    Atom,
+)
+from mathics.core.expression import Expression
+from mathics.core.symbols import SymbolList
+
+
+class ListExpression(Atom):
+    """
+    This class implements `List` expressions
+    as a simplified kind of  `Expression` that does
+    not requires to check for evaluation rules.
+
+    """
+
+    class_head_name = "System`List"
+
+    def __new__(cls, leaves):
+        # Here we could check if all the leaves are machinereal,
+        # and specialize to something like ListMachineReals
+        # or ListMachineComplex  or ListNumpy or whatever...
+        self = super().__new__(cls)
+        self._leaves = leaves
+        return self
+
+    def __str__(self) -> str:
+        return "{" + ",".join(leaf.__str__() for leaf in self._leaves) + "}"
+
+    def get_head_name(self):
+        return "System`List"
+
+    def boxes_to_text(self, **options) -> str:
+        # TODO: take into account the MakeBoxes rules...
+        raise NotImplementedError
+
+    def boxes_to_mathml(self, **options) -> str:
+        # TODO: take into account the MakeBoxes rules...
+        raise NotImplementedError
+
+    def boxes_to_tex(self, **options) -> str:
+        # TODO: take into account the MakeBoxes rules...
+        raise NotImplementedError
+
+    def to_python(self, *args, **kwargs):
+        return [leaf.to_python(*args, **kwargs) for leaf in self._leaves]
+
+    def sameQ(self, other):
+        if other is self:
+            return True
+        if isinstance(other, ListExpression):
+            if len(other._leaves) != len(self._leaves):
+                return False
+            return all(a.sameQ(b) for (a, b) in zip(self._leaves, other._leaves))
+        if isinstance(other, Expression) and other._head is SymbolList:
+            return all(a.sameQ(b) for (a, b) in zip(self._leaves, other._leaves))
+
+        return False
+
+    def evaluate(self, evaluation):
+        # TODO: handle upvalues
+        new_leaves = tuple(leaf.evaluate(evaluation) for leaf in self._leaves)
+        if all(leaf is None for leaf in new_leaves):
+            return None
+        return ListExpression(
+            tuple(
+                oldleaf if newleaf is None else newleaf
+                for newleaf, oldleaf in zip(new_leaves, self._leaves)
+            )
+        )
+
+    @property
+    def leaves(self):
+        return self._leaves
+
+    @leaves.setter
+    def leaves(self, value):
+        raise ValueError("Expression.leaves is write protected.")


TiagoCavalcante:
`ExpressionList` is something that would benefit from `FinalForm`.

mathics/builtin/base.py
```diff
@@ -611,6 +611,8 @@ def get_string_value(self):
 
     def sameQ(self, expr) -> bool:
         """Mathics SameQ"""
+        if isinstance(expr, BoxConstruct):
```
rocky, line 614: When a general method like `SameQ` is asking about a specific kind of thing like `BoxConstruct` that should be a harbinger that there is a design problem and that a general concept is missing. 

At a minimum, a lot of explaining is needed as to why sameQ which is testing whether something is identical needs to understand why something is a Boxing construct. 

Sometimes instead there is a fundamental concept that needs to be made explicit.

mmatera:
> `ExpressionList` is something that would benefit from `ExpandOnce`.

Actually, the idea of this experiment is to see if we can avoid the need to tag leaves in an expression as `ExpandOnce`. When the evaluator finds an expression of the form  `Expression(SymbolList, ...)`, the expression is replaced by something with a an evaluation logic that depends on the kind of leaves. If all the leaves are numbers, (or maybe, undefined symbols) no more further evaluations will be carried on. Otherwise, at least we avoid looking at definitions and attributes for ``Symbol`List``.

rocky:
> > `ExpressionList` is something that would benefit from `ExpandOnce`.
> 
> Actually, the idea of this experiment is to see if we can avoid the need to tag leaves in an expression as `ExpandOnce`. When the evaluator finds an expression of the form `Expression(SymbolList, ...)`, the expression is replaced by something with a an evaluation logic that depends on the kind of leaves. If all the leaves are numbers, (or maybe, undefined symbols) no more further evaluations will be carried on. Otherwise, at least we avoid looking at definitions and attributes for `` Symbol`List ``.

Personally, I'd like to see a bigger picture description of the overall evaluation process and in that it would be clearer to discuss under what conditions we need pattern matching or reevaluation or not. 

In that it is also becoming clear to me that we should have means for alternate representations of expressions. In some cases for example there could be a SymPy object that is equivalent. Possibly others like numpy, mpmath, scipy, compiled form - I just don't know the details of these well. 

I do know however that this kind of incremental change might not be the best approach. It hasn't served us that well so far.

mathics/builtin/base.py
mmatera: This is because if `BoxConstruct` compares against other kind of expression, it can delegate the implementation to the other object. But if `expr` is also a `BoxConstruct`, this would produce an infinite loop. However, in this case, we just need to compare if `self` and `expr` are the same Python object.

mmatera:
> > > `ExpressionList` is something that would benefit from `ExpandOnce`.
> > 
> > 
> > Actually, the idea of this experiment is to see if we can avoid the need to tag leaves in an expression as `ExpandOnce`. When the evaluator finds an expression of the form `Expression(SymbolList, ...)`, the expression is replaced by something with a an evaluation logic that depends on the kind of leaves. If all the leaves are numbers, (or maybe, undefined symbols) no more further evaluations will be carried on. Otherwise, at least we avoid looking at definitions and attributes for `` Symbol`List ``.
> 
> Personally, I'd like to see a bigger picture description of the overall evaluation process and in that it would be clearer to discuss under what conditions we need pattern matching or reevaluation or not.
> 

I think that, at this point, I would be in a position to describe the (current) evaluation algorithm with a certain detail. If you think it would be useful, I can try to write a draft describing the steps, or just by adding comments to the code in a new branch. 



> In that it is also becoming clear to me that we should have means for alternate representations of expressions. In some cases for example there could be a SymPy object that is equivalent. Possibly others like numpy, mpmath, scipy, compiled form - I just don't know the details of these well.
> 
> I do know however that this kind of incremental change might not be the best approach. It hasn't served us that well so far.

Incremental changes are not the solution but can provide some idea of what we need to change, and what would be the side effects.

rocky:
> > > > `ExpressionList` is something that would benefit from `ExpandOnce`.
> > > 
> > > 
> > > Actually, the idea of this experiment is to see if we can avoid the need to tag leaves in an expression as `ExpandOnce`. When the evaluator finds an expression of the form `Expression(SymbolList, ...)`, the expression is replaced by something with a an evaluation logic that depends on the kind of leaves. If all the leaves are numbers, (or maybe, undefined symbols) no more further evaluations will be carried on. Otherwise, at least we avoid looking at definitions and attributes for `` Symbol`List ``.
> > 
> > 
> > Personally, I'd like to see a bigger picture description of the overall evaluation process and in that it would be clearer to discuss under what conditions we need pattern matching or reevaluation or not.
> 
> I think that, at this point, I would be in a position to describe the (current) evaluation algorithm with a certain detail. If you think it would be useful, I can try to write a draft describing the steps, or just by adding comments to the code in a new branch.
> 
> > In that it is also becoming clear to me that we should have means for alternate representations of expressions. In some cases for example there could be a SymPy object that is equivalent. Possibly others like numpy, mpmath, scipy, compiled form - I just don't know the details of these well.
> > I do know however that this kind of incremental change might not be the best approach. It hasn't served us that well so far.
> 
> Incremental changes are not the solution but can provide some idea of what we need to change, and what would be the side effects.

Sure, in many cases this kind of thing can work. Here, in this specific case can I ask please can we have a moratorium on the little tweaks and suggested changes here and there. 

I'd like some quiet time to take a bigger more holistic view of what's going on and all of this is noise and distraction. Others have had time and free reign to try things out. It hasn't been working out. 

In taking a more holistic view, in some cases things may even get slower micro-improvement wise. But that is okay if in we have something that is more modular, more understandable, more scalable, and will most likely be given how far off we are from how fast this _should_ run much faster too.

rocky:
> I think that, at this point, I would be in a position to describe the (current) evaluation algorithm with a certain detail. If you think it would be useful, I can try to write a draft describing the steps, or just by adding comments to the code in a new branch.
> 

@mmatera yes, it would be very helpful to describe in as high a level as we can, the evaluation process. We started this in [Code Overview](https://mathics-development-guide.readthedocs.io/en/latest/extending/code-overview.html)

If that needs revising or expanding, please do that.

mmatera:
> > > > > `ExpressionList` is something that would benefit from `ExpandOnce`.
> > > > 
> > > > 
> > > > Actually, the idea of this experiment is to see if we can avoid the need to tag leaves in an expression as `ExpandOnce`. When the evaluator finds an expression of the form `Expression(SymbolList, ...)`, the expression is replaced by something with a an evaluation logic that depends on the kind of leaves. If all the leaves are numbers, (or maybe, undefined symbols) no more further evaluations will be carried on. Otherwise, at least we avoid looking at definitions and attributes for `` Symbol`List ``.
> > > 
> > > 
> > > Personally, I'd like to see a bigger picture description of the overall evaluation process and in that it would be clearer to discuss under what conditions we need pattern matching or reevaluation or not.
> > 
> > 
> > I think that, at this point, I would be in a position to describe the (current) evaluation algorithm with a certain detail. If you think it would be useful, I can try to write a draft describing the steps, or just by adding comments to the code in a new branch.
> > > In that it is also becoming clear to me that we should have means for alternate representations of expressions. In some cases for example there could be a SymPy object that is equivalent. Possibly others like numpy, mpmath, scipy, compiled form - I just don't know the details of these well.
> > > I do know however that this kind of incremental change might not be the best approach. It hasn't served us that well so far.
> > 
> > 
> > Incremental changes are not the solution but can provide some idea of what we need to change, and what would be the side effects.
> 
> Sure, in many cases this kind of thing can work. Here, in this specific case can I ask please can we have a moratorium on the little tweaks and suggested changes here and there for purposes of performance?
> 
> I'd like some quiet time to take a bigger more holistic view of what's going on and all of this is noise and distraction. Others have had time and free reign to try things out. It hasn't been working out.
> 

OK, so @rocky, a few questions:
1. Should I close this PR? The only reason to keep this open is just to collect/discuss ideas.
2. Would be useful to have a PR with more comments on the part of the code related to the `evaluate` method?
3. Apart to do not introducing more tweaks, is there anything else we can do for helping you with the redesign?

> In taking a more holistic view, in some cases things may even get slower micro-improvement wise. But that is okay if in we wind up in having something that is more modular, more understandable, more scalable, and will most likely be — given how far off we are from how fast this _should_ run — much faster too.

rocky:
> 1. Should I close this PR? The only reason to keep this open is just to collect/discuss ideas.

The ideas and motivation behind this are good. So we should forget the general idea. The specific code I don't think it likely we will  modify or incrementally change this code. So in that sense, yes, I guess close it. 

 
> 2. Would be useful to have a PR with more comments on the part of the code related to the `evaluate` method?

It would be good to start annotating the existing code. If you have comments, concerns, or thoughts about how to change things, then yes, that is great. 

> 3. Apart to do not introducing more tweaks, is there anything else we can do for helping you with the redesign?

I expect this to be a group process, with everyone involved all the way along. But first I think we all need to take stock of what we have, how this works, how it might work and so on. 

So right now I would say focus on elaborating and explaining what's there. 

The other aspect that I think we are _extremely_ week on is in testing little parts. for example the simple Lisp/WL Symbol should be testable independent of the rest of the code. The tests would bhat when we try to create two Symbols, they map to the same thing. And that we can extract a string back from the Symbol.

When I look at https://reference.wolfram.com/language/ref/Symbol.html there are basic things described their, like Unique and Context that would seem to help support whether Symbol has the right properties, but we don't seem to have those. 

The same things would go for pattern matching. 

I don't understand precisely Atom, Instanceable the current definition of Mathics Symbol and the differences between these and Builtin and how that corresponds to published parts of Mathematica.  


> 
> > In taking a more holistic view, in some cases things may even get slower micro-improvement wise. But that is okay if in we wind up in having something that is more modular, more understandable, more scalable, and will most likely be — given how far off we are from how fast this _should_ run — much faster too.

mmatera:
> The ideas and motivation behind this are good. So we should forget the general idea. The specific code I don't think it likely we will modify or incrementally change this code. So in that sense, yes, I guess close it.
> 

OK

> > 2. Would be useful to have a PR with more comments on the part of the code related to the `evaluate` method?
> 
> It would be good to start annotating the existing code. If you have comments, concerns, or thoughts about how to change things, then yes, that is great.
> 

I am doing this a little. I will create a new branch with just the docstrings / comments.

> > 3. Apart to do not introducing more tweaks, is there anything else we can do for helping you with the redesign?
> 
> I expect this to be a group process, with everyone involved all the way along. But first I think we all need to take stock of what we have, how this works, how it might work and so on.
> 
> So right now I would say focus on elaborating and explaining what's there.
> 

OK, then let's go in that direction.

mmatera:
@rocky this is the branch
 commented-evaluation-process

rocky:
Ok, but I think this misses the point then or is looking at things in a too narrow way that is missing the bigger picture which I'll try to describe elsewhere.  The Notions from programming languages like literal values, unbound and bound expressions are the higher level concepts we should consider. 

Also, this is a suggestion of something new not  describing what is currently in the code base or a discussion of how the evaluation process works or might work.

rocky:
> @rocky this is the branch [commented-evaluation-process](https://github.com/Mathics3/mathics-core/tree/commented-evaluation-process)

@mmatera My apologies of repeating more or less what I wrote before. I didn't understand that this comment was referring to _another_ branch. I mistakenly thought "this" branch meant this closed PR.  I've updated this to add the link and I'll look at this branch over the weekend. 

Again, my apologies.

PR #124, opened 2022-01-21
Commented evaluation process

rocky:
@mmatera thanks - this is the kind of thing we need more of.


mathics/core/expression.py
```diff
@@ -743,7 +755,13 @@ def flatten(
         else:
             return self
 
-    def evaluate(self, evaluation) -> typing.Union["Expression", "Symbol"]:
+    def evaluate(self, evaluation) -> "BaseExpression":
+        """
+        Evaluate until reach a fixed point.
+        """
+        # comment @mmatera: Exceptions (like ReturnInterrupt) should be in a different module
```
rocky, line 762: Yes, I agree. And at this stage it is important to me because we want to reevaluate the overall structure and starting off with something that is unnecessarily complicated doesn't hellp. 

So I have done this in  https://github.com/Mathics3/mathics-core/pull/134

mathics/core/expression.py
```diff
@@ -743,7 +755,13 @@ def flatten(
         else:
             return self
 
-    def evaluate(self, evaluation) -> typing.Union["Expression", "Symbol"]:
+    def evaluate(self, evaluation) -> "BaseExpression":
```
rocky, line 758: This is wrong. What is returned is some subclass of "Expression". I corrected this now in #134 




Furthermore, this I believe is something to be important.   We start

mathics/core/expression.py
```diff
@@ -743,7 +755,13 @@ def flatten(
...
+        """
+        Evaluate until reach a fixed point.
```
rocky, line 760: Personally, I prefer a more expansive style - describing what a fixed point means here. 

But if you don't mind I'll do this at a later date rather than do something now.  

What is here serves is good for now.

mathics/core/expression.py
mmatera: @rocky, ``evaluate`` can return a  ``String`` or a `PrecisionReal` number, not necessarily an  ``Expression``.

mathics/core/expression.py
mmatera: @rocky, I didn't expand that more because I was not sure how much to say:  Something more verbose could be " `evaluate` applies rules associated to the initial expression until the expression failts to be evaluated again or the resulting expression coincides with the previous one. "
But again, I was not sure how much to put in the docstring and how much in the comments...

rocky: Sure, I understand. I have made my attempt not in #134. 

The guide I go for in a docstring: the information that a programmer calling the class or method needs to understand.  So here it is what you put in and what you expect to get out from what you put in and how the two are related. 

Specific implementation details like we use a cache to do this, or opinions like I think this code should be rewritten this way are probably better as a comment.

rocky:
FYI I just came across this https://library.wolfram.com/infocenter/Conferences/4683/ via https://mathematica.stackexchange.com/questions/29353/how-does-return-work which is mentioned in the code

rocky:
@mmatera & @TiagoCavalcante  the next weekend or so, I plan on going over `evaluate_next()`  especially incorporating the comments here.   

One thing, small as it is, has been bothering me for a while. I would like to rename `evaluate_next()` to `evaluate_step()`.

The "next" part never made sense, in that next can refer to the first time, or only time. (Python's `next()` function is like that too, but I think this too isn't ideal)

"step", makes more sense to me: the top level calls a number of evaluation steps.  Even better would be "transform_and_evaluate_step" since each step isn't just about evaluation but it includes a transformation step beforehand.

Your thoughts?

mmatera:
``evaluation_step`` seems good to me. I would avoid very long names...

rocky:
> `evaluation_step` seems good to me. I would avoid very long names...

I generally agree with one exception which occurs here. When the function is used from basically one place, as it is here, then a long name also serves to emphasize that. 

There is _another_ method in class `BaseExpression()`  called `evaluate_next`  which calles `evaluate()` and it's `evaluate()` method calls `check_stopped()`.  Of course there are no docstrings on either and I find this pretty confusing. I wonder what's up here. Is this intentional or a mistake?

TiagoCavalcante:
@rocky I have no problem with long names, but I think that just the prefix "evaluate_" is better because of `evaluate_fast`.

rocky:
> @rocky I have no problem with long names, but I think that just the prefix "evaluate_" is better because of `evaluate_fast`.

Right now this doesn't exist. As I understand it though, there is first a process by which transformation occurs _before_ the evaluation process with Python class method resolution.  So there might still be  other methods called from the transform-and-evaluate step. And caching the last method that was used might get integrated into the pure evaluation part. 

I expect the Expression object to change a bit.as well as the Symbol (which you might think of as a "variable" object). The expression object might have alternate forms like the expression a form that is more suitable for sympy, and the Symbol might have a slot for the last Python class method that was called along with some sort of signature of the arguments for that call.

All of this is speculative though. And I am not opposed to changing the function name more than once. 

The two things though that are misleading about `evaluate_next` is that not only is "next" a weird choice, but so is "evaluate" because generally expression transformation generally as part of evaluation. That is generally thought of as a separate phase and indeed it looks like it is a separate phase here as well.

mmatera:
> I generally agree with one exception which occurs here. When the function is used from basically _one_ place, as it is here, then a long name also serves to emphasize that.

Maybe better would be to have a hidden method (let's say ``_evaluation_step``).

> 
> There is _another_ method in class `BaseExpression()` called `evaluate_next` which calls `evaluate()` and it's `evaluate()` method calls `check_stopped()`. Of course there are no docstrings on either and I find this pretty confusing. What's up here - is this intentional or a mistake?

I just add some comments about this. Currently, each time the result of `evaluate_next` is an Atom, ``evaluate`` checks that the evaluation was not stopped. Maybe we gain something if this check is just done for `Symbols`.

mathics/core/expression.py
```diff
@@ -776,25 +788,43 @@ def evaluate(
                 "  " * evaluation.recursion_depth + "Evaluating: %s" % expr
             )
         try:
+            # Evaluation loop:
             while reevaluate:
```
rocky, line 792: What is becoming clear is that we are doing a series of _applications_ rather than _evaluations_. 

The word "application" in functional languages is a general term that is used with the form of an expression is modified such as happens in term rewriting. And "application" is also used when you have a list and you read pick out the first element of the list and use that as a function and the remaining leaves as the function parameters. 

See for example: https://stackoverflow.com/a/8869553/546218

rocky:
@mmatera  - This has been very helpful. In gratitude I will be adding comments or notes for things here for discussion and feedback. 

But let's not incorporate that here. Next weekend I would like to use this as a base for rewriting parts of what is called `evaluate_next()` here and in the code. 

One goal in that effort is to be able to test parts of the application, rewrite and evaluation process in isolation and so that we can have a better handle on benchmarking.

mathics/core/expression.py
```diff
@@ -856,16 +904,40 @@ def eval_range(indices):
             eval_range(range(len(leaves)))
             # rest_range(range(0, 0))
 
+        # Step 2: Build a new expression. Notice that leaves are given
```
rocky, line 907: I extracted these steps and condensed them a little bit and put this [here](https://stackoverflow.com/a/8869553/546218)  in the developer's guide. 

Steps 2 and 3 are combined there because both really are about building a new expression.

mathics/core/expression.py
```diff
@@ -881,17 +953,30 @@ def eval_range(indices):
...
         if flat & attributes:
             new = new.flatten(new._head, callback=flatten_callback)
+
+        # If the attribute `Orderless` is set, sort the leaves, according to the
+        # `get_sort` criteria.
+        # the most expensive part of this is to build the sort key.
         if orderless & attributes:
             new.sort()
 
+        # Step 4:  Now, rebuilds the ExpressionCache, which tracks which symbols
```
rocky, line 972: This seems to _create_ a new cache, not rebuild an existing one.

That said, right now I am not that concerned with the behavior because this is more a "nice to have" kind of thing rather than a necessary part of the operational logic.

mathics/core/expression.py
```diff
@@ -901,6 +986,33 @@ def flatten_callback(new_leaves, old):
                 else:
                     return threaded, True
 
+        # Step 6: Now,the next step is to look at the rules associated to
```
rocky, line 989: In https://github.com/Mathics3/mathics-development-guide/blob/3a82527c3461e4db838bd0ded4d86fe304a01df6/docs/extending/code-overview/evaluation.rst#search-for-a-rule-in-head-to-apply I greatly condensed the description of step 6.

A lot of what is being described here is how symbols and parameters get bound. All of this really should be its own method that can be independently tested.

mathics/core/expression.py
```diff
@@ -943,11 +1060,27 @@ def rules():
             new = Expression(head)
             new._leaves = tuple(dirty_leaves)
 
+        # copy the "unformatted" form of the original expression,
         new.unformatted = self.unformatted
+        # Step 8:updates the cache and returns the new form, with the reevaluate flag to false.
         new._timestamp_cache(evaluation)
```
rocky, line 1066: I am not sure why we would need to create a new timestamp rather than keep a copy of the old one and restore that.

The same is true for the code above.  

Some care is needed that when we create new timestamps we aren't sharing information from old once, but I think that's already covered.  That said, at least for the cache, I really am not all that interested in resolving what's up here until we have the rest properly isolated made modular and have tests for these.

mathics/core/expression.py
mmatera: In WL, Apply takes wo arguments : a head an an expression, and the result is a new expression with the same leaves of the original, but changing the head to be tee first argument.


In mathics, apply means to apply a replacement rule. Evaluate, on the other hand, is a process in which rules associated to symbols involved in a certain expression are applied iteratively until a fixed point is reached.

mathics/core/expression.py
mmatera: If the cache is rebuilt or created or updated depends on which changes the expression has suffered on the previous steps.

rocky:
@mmatera is this ready for merge?

mmatera:
> @mmatera is this ready for merge?

I think so. We can iterate over this later.

Issue #128, opened 2022-01-23
unhandled exception in Assignment


mmatera:
```
 7/; F[a]:=7
```
raises a unhandled exception instead of aborting the evaluation.

rocky:
This doesn't happen anymore. 

```
In[2]:= 7/; F[a]:=7
SetDelayed::setraw: Cannot assign to raw object 7.
Out[2]= $Failed
```

WMA gives: 

```
In[16]:= 7/; F[a]:=7                                                              

SetDelayed::setrpt: Cannot assign to raw object 7 in pattern 7 /; F[a].

Out[16]= $Failed
```

The error tag is different. But that's a different and less significant issue.

PR #129, opened 2022-01-23
fix issue #128

mmatera:


CHANGES.rst
```diff
@@ -36,8 +37,10 @@ Bugs
 ++++
 
 * ``First``, ``Rest`` and  ``Last`` now handle invalid arguments.
-* ``N`` now handles arbitrary precision numbers when the number of digits is not specified.  
-  
+* ``N`` now handles arbitrary precision numbers when the number of digits is not specified.
```
rocky, line 40: Do the changes below do this? And the `SetTag` thing too?

mmatera: no, this is older. I think that the update is due a blank line...

mmatera: and yes, `Set*` is affected.

PR #133, opened 2022-01-23
fix #132

### LINKED ISSUE #132: ScriptCommandLine compatibility with WMA
From #67 it becomes apparent that the behavior of ``ScriptCommandLine`` in Mathics differs from WMA in the sense that the first element of the list in WMA is the script filename instead of the first paramenter.

mmatera:
and part of #67

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -32,6 +32,12 @@ Internals
 * ``Cylinder`` rendering is implemented in Asymptote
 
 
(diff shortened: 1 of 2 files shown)


rocky (commented):
LGTM - thanks for splitting this off.

rocky (approved):
LGTM - thanks for splitting this off.

PR #134, opened 2022-01-23
clean imports, reduce expression module, ...

rocky:
* Interrupt exceptions moved ot of evaluation.py and into expression.py
* Correct return type of Expression.evaluate()
* remove unused imports including those that are imported improperly
  elsewhere
* import from a module only once

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -742,11 +742,29 @@ class Expression(BaseExpression):
         else:
             return self
 
-    # When we allow 3.9 only, the return type chould become type[Expression]
+    # When we allow 3.9 only, the return type could become type[Expression]
     # See https://adamj.eu/tech/2021/05/16/python-type-hints-return-class-not-instance/
-    def evaluate(self, evaluation) -> typing.Type["Expression"]:
-        """
-        Evaluate until reach a fixed point.
+    # Note that the return type is some subclass of BaseExpression, it could be
+    # a Real, an Expression, etc. It probably will *not* be a BaseExpression since
+    # the point of evaluation when there is not an error is to produce a concrete result.
+    def evaluate(
+        self, evaluation: typing.Type["BaseExpression"]
+    ) -> typing.Type["BaseExpression"]:
+        """Pattern match and replace subexpressions evaluating the resulting
+        expression until either we are told not to go further, or
+        until we determine that the result of an evaluation cannot
+        produce a resulting expression that is different from the one
+        we started out with.
+
+        Some ways that we might be told not to do further are:
+        * a  limit of some sort is hit: timeout or iteration count
+        * some other exception occurs
+        * the result of evaluation indicates not to go futher
+
+        Some ways that a we determine the result of another evaluation cannot produce a different result:
+        * The result was exactly the same as the last time, and there were no definition changes
+        * the result of an evaluation indicates that we have the final value (sort of the same thing
+          as the last case in the list item above).
         """
         if evaluation.timeout:
             return
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -10,9 +10,11 @@ from itertools import chain
 from bisect import bisect_left
 
 from mathics.core.atoms import from_python, Number, Integer
+from mathics.core.convert import sympy_symbol_prefix, SympyExpression
+from mathics.core.evaluation import Evaluation
+
 from mathics.core.interrupt import ReturnInterrupt
 from mathics.core.number import dps
-from mathics.core.convert import sympy_symbol_prefix, SympyExpression
 from mathics.core.symbols import (
     Atom,
     BaseExpression,
@@ -748,23 +750,14 @@ class Expression(BaseExpression):
     # a Real, an Expression, etc. It probably will *not* be a BaseExpression since
     # the point of evaluation when there is not an error is to produce a concrete result.
     def evaluate(
-        self, evaluation: typing.Type["BaseExpression"]
+        self,
+        evaluation: Evaluation,
     ) -> typing.Type["BaseExpression"]:
-        """Pattern match and replace subexpressions evaluating the resulting
-        expression until either we are told not to go further, or
-        until we determine that the result of an evaluation cannot
-        produce a resulting expression that is different from the one
-        we started out with.
+        """Apply transformation rules and expression evaluation to `evaluation` via
+        `evaluate_next()` until it tells us to stop or we hit some limit.
+        `evaluate_next()` may call us recusively.
 
-        Some ways that we might be told not to do further are:
-        * a  limit of some sort is hit: timeout or iteration count
-        * some other exception occurs
-        * the result of evaluation indicates not to go futher
-
-        Some ways that a we determine the result of another evaluation cannot produce a different result:
-        * The result was exactly the same as the last time, and there were no definition changes
-        * the result of an evaluation indicates that we have the final value (sort of the same thing
-          as the last case in the list item above).
+        Limits are either an evaluation iteration count or a timeout value.
         """
         if evaluation.timeout:
             return
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -10,6 +10,7 @@ from itertools import chain
 from bisect import bisect_left
 
 from mathics.core.atoms import from_python, Number, Integer
+from mathics.core.interrupt import ReturnInterrupt
 from mathics.core.number import dps
 from mathics.core.convert import sympy_symbol_prefix, SympyExpression
 from mathics.core.symbols import (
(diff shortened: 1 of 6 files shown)


mathics/core/expression.py
```diff
@@ -743,9 +742,12 @@ def flatten(
         else:
             return self
 
-    def evaluate(self, evaluation) -> typing.Union["Expression", "Symbol"]:
-        from mathics.core.evaluation import ReturnInterrupt
-
+    # When we allow 3.9 only, the return type chould become type[Expression]
+    # See https://adamj.eu/tech/2021/05/16/python-type-hints-return-class-not-instance/
+    def evaluate(self, evaluation) -> typing.Type["Expression"]:
```
mmatera, line 747: OK, but ``evaluate`` can return a `Symbol`, or a `String` or a `Real` number among other things. All of them are instances of the class  ``BaseExpression``.

mmatera:
apart from the comment, LGTM.

mathics/core/expression.py
rocky: Yes, I have this backwards. I mean `typing.Type["BaseExpression"]`. I will change that soon.

rocky: BTW, this also suggests why the there might be a `LiteralExpression` class separate from `BaseExpression`.

In the case of `LiteralExpression` that is something that should not be subclassed, it is an end kind of thing of itself.

mmatera: BTW, in the reference, the example is:
```
def make_animal(animal_class: Type[Animal], name: str) -> Animal:
    return animal_class(name=name)
```
so `Type` seems to be used in arguments, but not in the return type. (weird, isn't it?)

mmatera: > BTW, this also suggests why the there might be a `LiteralExpression` class separate from `BaseExpression`.
> 
> In the case of `LiteralExpression` that is something that should not be subclassed, it is an end kind of thing of itself.

I am not fully sure to follow you. Maybe later a UML graph could be helpful.

rocky: I think things will become clearer over time. Please be patient. 

Also, it is going to take a while to fully understand both this code and how WL and how that relates to conventional practice.

BTW, I have now updated the code and the docstring for Expression to be more clear about what is meant by "fixed point".  Please look over to make sure I have this correct.

mathics/core/expression.py
```diff
@@ -743,9 +742,30 @@ def flatten(
...
-    def evaluate(self, evaluation) -> typing.Union["Expression", "Symbol"]:
-        from mathics.core.evaluation import ReturnInterrupt
-
+    # When we allow 3.9 only, the return type could become type[Expression]
+    # See https://adamj.eu/tech/2021/05/16/python-type-hints-return-class-not-instance/
+    # Note that the return type is some subclass of BaseExpression, it could be
+    # a Real, an Expression, etc. It probably will *not* be a BaseExpression since
+    # the point of evaluation when there is not an error is to produce a concrete result.
+    def evaluate(
+        self, evaluation: typing.Type["BaseExpression"]
```
mmatera, line 751: ``evaluation: typing.Type["Evaluation"]``

TiagoCavalcante:
LGTM, except for the typing part, I agree with @mmatera.

mathics/core/expression.py
```diff
@@ -743,9 +742,30 @@ def flatten(
...
+
+        Some ways that we might be told not to do further are:
+        * a  limit of some sort is hit: timeout or iteration count
+        * some other exception occurs
+        * the result of evaluation indicates not to go futher
+
+        Some ways that a we determine the result of another evaluation cannot produce a different result:
+        * The result was exactly the same as the last time, and there were no definition changes
+        * the result of an evaluation indicates that we have the final value (sort of the same thing
+          as the last case in the list item above).
```
mmatera, line 767: Also if the expression was cached, and the symbols involved didn't change since then.

mmatera: Notice that line 788 has a similar effect than replacing the expression by a "Literal".

mathics/core/expression.py
rocky: Actually, I think it is exactly an "Evaluation", right?

mmatera: Yes

mathics/core/expression.py
rocky: After reading the description by David Withhoff, I want to back off a more detailed description of everything  and narrow the top-level description to just what _it_ does, not everything underneath that it calls. 

It is the next level `evaluate_next` which i would like renamed  where there is a lot more complexity. 

But I think we also need just a document on this in the readthedocs for this aspect alone. 

As for caching and all of that. You would do the design a favor if you forgot it exists. A simple and sane description of behavior or how to think about how things work shouldn't need  to involve caches. 

As an analogy, you might fit a bunch of pieces together in a jig-saw puzzle to form a scene, but you wouldn't describe the scene as how the jigsaw puzzles fit together. 

Reading what David Withoff writes, I am sorry to report that the more common word variable isn't mentioned but that's in fact what is going on. Given this, I am leaning towards calling this new Literal-value symbol something other than Symbol, e.g. LispSymbol, or SymbolLiteral, and make explicit that Symbol is akin to variable in programming languages.

mmatera: @rocky, I also have read  Withhoff's paper, and it describes something very similar to what we have in Mathics. A difference I found is that evaluations are not carried on over Raw (Atoms) expressions, (except for symbols). 

Regarding "caches", in `mathics.core.expression`, `ExpressionCache` just tracks which symbols are used in an Expression, and when they were modified. Is a way to avoid reevaluating a leaf that has already reached its fixed point. I think we need to take that part into account if we want to think about a redesign.

rocky: Sure, I understand what the expression cache is supposed to do. That is also mentioned in Withoff's paper. It may come about in the implementation at some point, the part I strongly object to is using it in describing a _behavior_. 

Caches should be transparent in terms of behavior, other than to speeding up operations. If you need a cache to ensure the correct _behavior_ then something is wrong. 

Finally, I want to make it clear that it is not simply the fact that Symbol means variable (which often is the case even outside of Mathematica), but what I mean is that inside WL docs it is used in two different ways. They say that there is a differentiation Symbol. And in Mathics (and probably WL) this gets put in an expression as that symbol at least initially. It just then gets transformed at some point to this other kind of symbol which feels acts like a variable in that it has changeable dynamic properties associated with it. That doesn't happen when we talk about the a zodiac symbol as being one of the many symbols that WL has. 

And we need to distinguish these to things.

mmatera: > Sure, I understand what the expression cache is supposed to do. 


Good!, because it took to me several months to understand what this object was caching.

> That is also mentioned in Withoff's paper. It may come about in the implementation at some point. I strongly object to is using it in describing a _behavior_. And this is what I am more concerned about.

The behavior is: "if certain Expression comes from evaluating a previous Expression, and the symbols that are involved have not changed since its creation, the Expression is not going to be evaluated again", isn't it?

> 
> Caches should be transparent in terms of behavior, other than to speed up the execution time of operations. If you need a cache to ensure the _correct_ behavior then something is wrong.
> 

hum, OK: if you remove cache in this case, you will get the same result, but only after a lot of extra steps.


> Finally, I want to make it clear that it is not simply the fact that Symbol means variable (which often is the case even outside of Mathematica), but what I mean is that inside WL docs it is used in two different ways. They say that there is a differentiation Symbol. And in Mathics (and probably WL) this gets put in an expression as that symbol at least initially. It just then gets transformed at some point to this other kind of symbol which feels acts like a variable in that it has changeable dynamic properties associated with it. That doesn't happen when we talk about the a zodiac symbol as being one of the many symbols that WL has.
> 
> And we need to distinguish these two things.


OK, but the problem here is that just a very limited set of Symbols (`List`, `Sequence`, `Symbol`, and maybe a couple more) have rules that can not be modified. For example, something funny you can do is:
```
In[1]:= Unprotect[Plus]; Plus=4; 4+2
Out[1]=4[4,2]
```
I think that this is what makes it tough to distinguish the two notions.

rocky: I was lamenting that the terminology is confused and I was trying to figure out a good approach to use to disambiguate the two. Right now given the way the code is named, probably we should add a SymbolicLiteral class which has the properties of a Lisp Symbol. 

I wasn't implying that the way WL or what's going on  in Mathics is confusing, Just that the lack of more precise naming makes it easy for people to get confused in our code, to the point of introducing bugs.

There are Lisp-like symbols and there are variables tied to a scope. A variable (called a Symbol in Mathics and often in WL) consists of a symbol (name) along with properties and bindings. 

This isn't hard to understand; Lisp novices manage it.

PR #120, opened 2022-01-19
FindMinimum

mmatera:
First working attempt. Probably, `_BaseFinder ` and friends should be moved to another module.

PR #135, opened 2022-01-25
FindMinimum

mmatera:
Squash of  #120

Apart from extending the support for  `FindRoot`, `FindMinimum` and `FindMaximum` tries to establish a basic framework for using `scipy` and other optional libraries.


mathics/builtin/numbers/calculus.py
```diff
@@ -1947,82 +1744,38 @@ class NIntegrate(Builtin):
...
 
-        self.messages["bdmtd"] = (
-            "The Method option should be a "
-            + "built-in method name in {`"
-            + "`, `".join(list(self.methods))
-            + "`}. Using `Automatic`"
+    try:
+        from mathics.builtin.scipy_utils.integrators import (
+            scipy_nintegrate_methods,
+            scipy_nintegrate_messages,
```
mmatera, line 1761: Here I have a doubt: to have the examples in the doctests, we need to install scipy. One possibility is to modify the docstring of `NIntegrate` according to the scipy availability. Another possibility is to move all of this to a pymathics module that requires scipy, and then load it optionally. Also, I can remove the examples or add the dependency to the Github actions. Thoughts?

rocky:
@mmatera It occurs to me that if adding scipy ias a dependency s a big concern (and I am not sure I think it is),

The scipy-specific code could be split off into a pymathics module.

mmatera:
> @mmatera It occurs to me that if adding scipy ias a dependency s a big concern (and I am not sure I think it is),
> 
> The scipy-specific code could be split off into a pymathics module.

Yes, for this reason, I didn't do that.

Issue #137, opened 2022-01-27
``N[F[Pi],prec,Method->method]`` does evaluates to ``F[N[Pi,prec,Method->method]]``


mmatera:
The use of the `Method` option for numeric evaluation breaks the behavior or `N`, that should propagate to its leaves.
For example, ``N[F[Pi],10]`` does evaluates to ``F[N[Pi,10]]`` produces the (``TraceEvaluation``):
```
    Evaluating: System`N[Global`F[System`Pi], 10]
      Evaluating: System`N
      Evaluating: Global`F[System`Pi]
        Evaluating: Global`F
        Evaluating: System`Pi
           [...]
      Evaluating: System`N[10]
         [...]
      -> 10.0
      Evaluating: System`N[Global`F, 10]
        [...]
      -> Global`F
      Evaluating: System`N[System`Pi, 10]
        [...]
      -> 3.141592654
    -> Global`F[3.141592654]
      Evaluating: Global`F
F[3.141592654]
```
But, ``N[F[Pi],10, Method->"numpy"]`` does evaluates to ``F[N[Pi,10]]``  produces (``TraceEvaluation``):

```
```
    Evaluating: System`N[Global`F[System`Pi], 10, System`Rule[System`Method, "numpy"]]
      Evaluating: System`N
      Evaluating: Global`F[System`Pi]
        Evaluating: Global`F
        Evaluating: System`Pi
      Evaluating: System`Rule[System`Method, "numpy"]
        Evaluating: System`Rule
        Evaluating: System`Method
N[F[Pi], 10, Method -> numpy]
```


```

mmatera:
There are different possibilities to handle ``N[expr, prec, Method->mthd]``:

* Just add a rule to discard the option, so 
`N[F[Pi],10, Method->"numpy"]`->`N[F[Pi],10]`->`F[N[Pi,10]]`->`F[N[Pi,10,Method->Automatic]`
* N[expr, prec, Method->mthd] could be evaluated to  N[expr, prec], but setting an internal variable in `Evaluation`, that can be inspected for `N[const,prec]` when `method` is supported.
* Evaluate  `N[F[leaf1, ...], prec, Method->mthd]` to  ``N[F, prec, Method->mthd][N[leaf1, prec, Method->mthd],...]

I guess the second option would be optimal.

rocky:
I spent the last 10 minutes going over this trying to figure out what you are talking about. For now I have to give up. 

[...]

PR #138, opened 2022-01-28
Fix n method (issue #137)

mmatera:
In this PR a method to propagate the choice of the method used to compute constants is implemented. Also, this should clarify the interface for defining built-in `NValues` rules:  Now, all the builtin rules for ``N``  has the form ``N[expr_,prec_]``.
The only exception is  the generic ``N[expr_,prec_,OptionsValue[]]`` associated to the Builtin ``N``, which are necessarily not a `NValues` rule.
``N[expr, prec, Method->method]`` establishes the preferred method, by storing it in the queue property `_preferred_n_method` of the `Evaluation` object.

### DIFF
--- a/mathics/builtin/numbers/constants.py
+++ b/mathics/builtin/numbers/constants.py
@@ -73,23 +73,14 @@ class _Constant_Common(Predefined):
     nargs = 0
     options = {"Method": "Automatic"}
 
-    def apply_N(self, precision, evaluation, options={}):
-        "N[%(name)s, precision_?NumericQ, OptionsPattern[%(name)s]]"
-
-        preference = self.get_option(options, "Method", evaluation).get_string_value()
-        if preference == "Automatic":
-            return self.get_constant(precision, evaluation)
-        else:
-            return self.get_constant(precision, evaluation, preference)
-
-    def apply_N2(self, evaluation, options={}):
-        "N[%(name)s, OptionsPattern[%(name)s]]"
-        return self.apply_N(None, evaluation, options)
+    def apply_N(self, precision, evaluation):
+        "N[%(name)s, precision_?NumericQ]"
+        return self.get_constant(precision, evaluation)
 
     def is_constant(self) -> bool:
         return True
 
-    def get_constant(self, precision, evaluation, preference=None):
+    def get_constant(self, precision, evaluation):
         # first, determine the precision
         machine_d = int(0.30103 * machine_precision)
         d = None
@@ -104,15 +95,20 @@ class _Constant_Common(Predefined):
 
         # If preference not especified, determine it
         # from the precision.
+        preference = None
+        preflist = evaluation._preferred_n_method.copy()
+        while preflist:
+            pref_method = preflist.pop()
+            if pref_method in ("numpy", "mpmath", "sympy"):
+                preference = pref_method
+                break
+
         if preference is None:
             if d <= machine_d:
                 preference = "numpy"
             else:
                 preference = "mpmath"
-        # If preference is not valid, send a message and return.
-        if not (preference in ("sympy", "numpy", "mpmath")):
-            evaluation.message(f'{preference} not in ("sympy", "numpy", "mpmath")')
-            return
+
         # Try to determine the numeric value
         value = None
         if preference == "mpmath" and not hasattr(self, "mpmath_name"):
@@ -207,7 +203,7 @@ class Catalan(_MPMathConstant, _SympyConstant):
     </dl>
 
     >> Catalan // N
-     = 0.915965594177219
+     = 0.915966
 
     >> N[Catalan, 20]
      = 0.91596559417721901505
@@ -286,8 +282,8 @@ class Degree(_MPMathConstant, _NumpyConstant, _SympyConstant):
             # return mpmath.degree
             return numpy.pi / 180
 
-    def apply_N(self, precision, evaluation, options={}):
-        "N[Degree, precision_, OptionsPattern[%(name)s]]"
+    def apply_N(self, precision, evaluation):
+        "N[Degree, precision_]"
         try:
             if precision:
                 d = get_precision(precision, evaluation)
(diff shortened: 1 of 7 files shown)


mathics/builtin/numbers/constants.py
```diff
@@ -207,7 +203,7 @@ class Catalan(_MPMathConstant, _SympyConstant):
     </dl>
 
     >> Catalan // N
-     = 0.915965594177219
+     = 0.915966
```
rocky, line 206: Does this change in precision match Mathematica more closely? 

I am curious what causes the change in precision here?

mathics/core/number.py
```diff
@@ -56,7 +56,12 @@ def _get_float_inf(value, evaluation) -> typing.Optional[float]:
     return value.round_to_float(evaluation)
 
 
-def get_precision(value, evaluation) -> typing.Optional[int]:
+def get_precision(value, evaluation, show_messages=True) -> typing.Optional[int]:
```
rocky, line 59: If this returns float, then the signature `Optional[int]` needs adjusting.

mathics/builtin/numeric.py
```diff
@@ -863,24 +878,52 @@ class N(Builtin):
...
     summary_text = "numerical evaluation to specified precision and accuracy"
 
-    def apply_with_prec(self, expr, prec, evaluation):
-        "N[expr_, prec_]"
+    def apply_with_prec(self, expr, prec, evaluation, options=None):
+        "N[expr_, prec_, OptionsPattern[%(name)s]]"
+
+        # If options are passed, set the preference in evaluation, and call again
+        # without options set.
+        # This also prevents to store this as an nvalue (nvalues always have two leaves).
```
rocky, line 888: Where in either WL docs or in this code are nvalues described?

mathics/builtin/numbers/constants.py
mmatera: Yes: In WMA,
```
In[1]:= Catalan //N                                                             

Out[1]= 0.915966
```

rocky: Great - thanks! Which "method" is getting used here. And what was it before?

mathics/core/number.py
mmatera: You are right! I just left it as I found it. What I am not sure about is if we have to change the signature or the returned value...

rocky: Line 67, the assignment to the return value is:   `d = value.round_to_float(evaluation)`

mathics/builtin/numeric.py
mmatera: image


For some reason, the documentation of this symbol was removed from WR.

rocky: Maybe for stuff like this we should start a "compatibility" package for Mathematica 5 so that if this is what you want you need to ask for it explicitly.

rocky: Okay it's there. But we say that it has been removed then when it isn't!

mmatera: https://reference.wolfram.com/language/tutorial/TransformationRulesAndDefinitions.html  section "Defining Numerical Values":

"The Wolfram Language treats numerical values essentially like upvalues. When you define a numerical value for f, the Wolfram Language effectively enters your definition as an upvalue for f with respect to the numerical evaluation operation N. "

mmatera: > Maybe for stuff like this we should start a "compatibility" package for Mathematica 5 so that if this is what you want, you need to ask for it explicitly.

OK, but I am not sure that is was "removed" and kept just for compatibility.  I think it becomes something "internal". Still, if you evaluate
```
N[f[x]]:=3
```
the rule is not stored as an ``Upvalue`` but as a ``NValue``. What is not exposed in the documentation is the keyword ``NValues``.

mathics/core/number.py
mmatera: OK, but the function, should return an integer or a floating-point number?

mathics/builtin/numbers/constants.py
mmatera: "sympy" is the default, but this is decided inside the `get_constant` method.

mathics/core/number.py
rocky: Good point! The other places where `get_precision()` appears it returns an int. And semantically it should be an `int`. 

We don' thave a `value.round_to_int(evaluation)` so I believe that should be added and used here.

mathics/builtin/numeric.py
rocky: > https://reference.wolfram.com/language/tutorial/TransformationRulesAndDefinitions.html section "Defining Numerical Values":
> 
> "The Wolfram Language treats numerical values essentially like upvalues. When you define a numerical value for f, the Wolfram Language effectively enters your definition as an upvalue for f with respect to the numerical evaluation operation N. "

Interesting. When we write a description of the evaluation process we need to make sure to add links to things like this. Notice they didn't call this an NValue but "define a numerical value", and I am not certain what that means.

mathics/core/number.py
mmatera: Looking at it more carefully, it seems that for compatibility with WMA, it should be a Float: WMA uses real values for precision, to take into account the internal representation of numbers. This is why $MachinePrecision is not 16, but 15.9546`

mathics/builtin/numeric.py
mmatera: I think that what currently is not "published" by the kernel (documentation) is the access to these "NValues". However, it makes a lot of sense to store them in a different place, because are rules that are not supposed to be applied in a general evaluation.

mathics/core/number.py
rocky: Again - good work! 

What I'd ask though is emphasize that this is a float and not an int. If nothing else it will prevent future people inspecting the code from taking the time to wonder about it as we just did.

And then this calls into question whether the other corresponding `get_precision()` methods shouldn't be changed to return a float as well.

rocky:
@mmatera and @TiagoCavalcante ok - so if you don't mind let's merge this into master and continue....

PR #141, opened 2022-01-30
Segregate atomic elements the way WL does

rocky:
Also, expand the AtomQ examples.

The  was originally motivated by the desire simply to improve the examples for `AtomQ`.

However in doing this I realized that WL have an entire section for [Atomic Elements](https://reference.wolfram.com/language/guide/AtomicElementsOfExpressions.html ) which is very useful for someone trying to understand WL/Mathics. 

It would have been a lot less work had this been done earlier. And in general it is a reminder to me that stuff is better done first before macro expanding functions like apply_N everywhere or else you'll have a lot of changes going on.

I should also mention that parts of the moved sections for number and strings don't totally match the organization of the mathematica docs. That isn't intentional, but just that this was such as heavy lift and massive set of changes that for the rest of the tidying and reorganization to make this more like Mathematica docs (which I think make sense here), I'll do that some other time.


mathics/builtin/atomic/numbers.py
```diff
@@ -5,9 +5,9 @@
 # printed output, so be carful not to add then mid-sentence.
 
 """
-Numeric Evaluation and Precision
```
mmatera, line 8: In mathics.builtin.numeric we had things like `NIntegrate`. Should them be here now?

mmatera:
@rocky, in general, I like this a lot. Two observations:
* In `mathics.core.numeric` we had things like `NIntegrate`. Probably this kind of symbols should be in a (strongly stripped) `mathics.builtiin.numeric` module.
* what about moving `apply_N` to a specific module inside `mathics.core`? This function involves more than just working with certain specific kinds of atoms (numbers) and seems to be something more related to methods like `evaluate` and `evaluate_next`, which takes an `Expression` object and produces a new `Expression` object using ruled defined in a `Definitions` object.

mathics/builtin/atomic/numbers.py
rocky: I would like to follow the Mathematica docs organization. However this was taking too much time and I wanted to finish getting a start here. 

Feel free to adjust this along the organization of the Mathematica docs. 

What I wanted really to make more headway this weekend towards trying to just come up with something that we could start to discuss and document the current evaluation process.  (And of course this is a prerequisite for discussing the current design and what proposed changes we might want to make. 

But my take is that in order to make forward progress spending a little bit of time trying to organize what we have will be helpful for everyone to be on the same page.

rocky:
> @rocky, in general, I like this a lot. Two observations:
> 
> * In `mathics.core.numeric` we had things like `NIntegrate`. Probably this kind of symbols should be in a (strongly stripped) `mathics.builtiin.numeric` module.
> * what about moving `apply_N` to a specific module inside `mathics.core`? This function involves more than just working with certain specific kinds of atoms (numbers) and seems to be something more related to methods like `evaluate` and `evaluate_next`, which takes an `Expression` object and produces a new `Expression` object using ruled defined in a `Definitions` object.

Sure, move `apply_N` out of `mathics.builtin.atomic` and into `core`.  

Overall I am not a fan of the current organization where we have some very low level implementation code sitting right next to this high-level Built in Function class stuff.  The two things feel very different to me. 

As for redoing the Class organization so that it isn't as hoaky and homegrown similar to the way the doctest stuff is, we'll probably have to live with this for a while like we do with the doctest stuff. 

So in sum, sure, this moving low-level fundamental concepts out of the implementation of specific built-in functions will help a lot.

rocky:
I do however like `apply` methods starting out with `apply_`.   Or rather, more specifically having method that is an entry point in a Mathics function call, follow some naming convention.

mmatera:
> I do however like `apply` methods starting out with `apply_`. Or rather, more specifically having any method that is an entry point in a Mathics built-in function call, follow some naming convention.

Actually, they should start with  `apply` if you want that method to be considered a built-in rule.  
The scheme I was thinking about is something similar to #139: low-level generic methods line `apply_N` or `apply_NValues` or `numerify` that are used in several builtins and in the basic evaluation process, are moved to a separate module inside core.

rocky:
> > I do however like `apply` methods starting out with `apply_`. Or rather, more specifically having any method that is an entry point in a Mathics built-in function call, follow some naming convention.
> 
> Actually, they should start with `apply` if you want that method to be considered a built-in rule. 

If you are talking about in the class, then of course. That's the (wonky) way the code was written,. 

I meant the name when they are separated from some builtin class and are standalone methods as is the case for `apply_N`. And here it is the Mathics builtin name that is appended after `apply_`. 

There, I don't think there is anything _enforcing_ this, it is just a convention we use.  And not that often because we don't have that great a need right now.

 I was just saying that keeping that convention I find helpful, whether or not we use the specific prefix is `apply_`.  That part I am ambivalent about, but I guess as a name it seems appropriate, or at least, not inappropriate. 

> The scheme I was thinking about is something similar to #139: low-level generic methods line `apply_N` or `apply_NValues` or `numerify` that are used in several builtins and in the basic evaluation process, are moved to a separate module inside core.

I looked again briefly at that PR and don't understand what you are talking about. But it doesn't matter. Just make the changes here in whatever style that is consistent the current `apply_N()` convention.

rocky:
I just realized why I am so lukewarm about using the prefix `apply_` everywhere. In functional programming which is what this is modelling, the "apply" step is the part that comes immediately _before_ calling this function. Technically the right prefix is `eval_`.

But the change to too great and pervasive to contemplate making right now.

mathics/builtin/atomic/atomic.py
```diff
@@ -0,0 +1,73 @@
+# -*- coding: utf-8 -*-
+"""Atomic Primitives
```
TiagoCavalcante, line 2: The convention is to put `Atomic Primitives` in the next line.

mathics/builtin/atomic/atomic.py
```diff
@@ -0,0 +1,73 @@
...
+    So are symbolic constants:
+    >> AtomQ[2 + I]
+     = True
+
+    A 'Symbol' not bound to a value is an atom too:
+    >> AtomQ[x]
+     = True
+
+    On the other hand, expressions aren't atoms:
+    >> AtomQ[2 + Pi]
```
TiagoCavalcante, line 33: The same as my comment in the documentation PR.

TiagoCavalcante (approved):
LGTM

mmatera:
> > > I do however like `apply` methods starting out with `apply_`. Or rather, more specifically having any method that is an entry point in a Mathics built-in function call, follow some naming convention.
> > 
> > 
> > Actually, they should start with `apply` if you want that method to be considered a built-in rule.
> 
> If you are talking about in the class, then of course. That's the (wonky) way the code was written,.
> 
> I meant the name when they are separated from some builtin class and are standalone methods as is the case for `apply_N`. And here it is the Mathics builtin name that is appended after `apply_`.
> 
> There, I don't think there is anything _enforcing_ this, it is just a convention we use. And not that often because we don't have that great a need right now.
> 
> I was just saying that keeping that convention I find helpful, whether or not we use the specific prefix `apply_`. That part I am ambivalent about, but I guess as a name it seems appropriate, or at least, not inappropriate.
> 
> > The scheme I was thinking about is something similar to #139: low-level generic methods line `apply_N` or `apply_NValues` or `numerify` that are used in several builtins and in the basic evaluation process, are moved to a separate module inside core.
> 
> I looked again briefly at that PR and don't understand what you are talking about. But it doesn't matter. Just make the changes here in whatever style that is consistent the current `apply_N()` convention.

I am working on the patch for this branch addressing my comments. Later today I will publish it.

rocky:
> > > > I do however like `apply` methods starting out with `apply_`. Or rather, more specifically having any method that is an entry point in a Mathics built-in function call, follow some naming convention.
> > > 
> > > 
> > > Actually, they should start with `apply` if you want that method to be considered a built-in rule.
> > 
> > 
> > If you are talking about in the class, then of course. That's the (wonky) way the code was written,.
> > I meant the name when they are separated from some builtin class and are standalone methods as is the case for `apply_N`. And here it is the Mathics builtin name that is appended after `apply_`.
> > There, I don't think there is anything _enforcing_ this, it is just a convention we use. And not that often because we don't have that great a need right now.
> > I was just saying that keeping that convention I find helpful, whether or not we use the specific prefix `apply_`. That part I am ambivalent about, but I guess as a name it seems appropriate, or at least, not inappropriate.
> > > The scheme I was thinking about is something similar to #139: low-level generic methods line `apply_N` or `apply_NValues` or `numerify` that are used in several builtins and in the basic evaluation process, are moved to a separate module inside core.
> > 
> > 
> > I looked again briefly at that PR and don't understand what you are talking about. But it doesn't matter. Just make the changes here in whatever style that is consistent the current `apply_N()` convention.
> 
> I am working on the patch for this branch addressing my comments. Later today I will publish it.

@mmatera Please hold off working on the faster napply and numerify branch. Right now this is not helpful. It just adds chaos to an area that is under review. Afterwards, or as part of the apply/evaluate discussion is done, then we can go back and suggest performance improvements.  Thanks.

mathics/builtin/atomic/atomic.py
rocky: If I understand correctly, this should be addressed in 2a5ffcbe

BTW GNU Emacs Python mode prefers not to add a \n after a leading `"""`.

mathics/builtin/atomic/atomic.py
rocky: Should be fixed now the same way it was done in the documentation PR. But please double check.

mmatera: 2 + Pi is clearly not an atom. Consider the following example:

```
a= 2+Pi;
Unprotect[Pi]; Pi=3;
a
```
Then, ``a`` evaluates to 5. 

Again, the main property of an atomic expression is that if you evaluate it, you get always the same expression, independently of what are the state of the ``Definitions`` object.

mmatera (approved):

mathics/builtin/atomic/atomic.py
rocky: From https://reference.wolfram.com/language/ref/AtomQ.html:

> An expression which cannot be divided into subexpression [after evaluation]

I'll make this more explicit that there is more than one part after evaluation.

rocky:
I think everyone is on the same page, and probably understands this at least implicitly, but for just in case or of there are any onlookers: the reason all of this is important to nail down is that we are going over details of the semantics of evaluation and will eventually be revising this so as to speed things up. 

In that, understanding what is meant in Mathics classes is important.  `Atom` happens to be the class this is really getting used when the function `AtomQ[]` is called. And it is pretty much at the bottom of the class hierarchy when it comes to the kinds of objects that can appear in an expression.

rocky:
Note: there will be more builtin function moving around and documentation cleanup/additions down the line.  This is just a small step.

mathics/builtin/atomic/atomic.py
TiagoCavalcante: It is as the other files now.

PR #143, opened 2022-01-30
suggested changes

mmatera:
* `NIntegrate` and auxiliary functions are moved from `mathics.core.atomic.numbers` to `mathics.core.numbers.calculus`
* the non-member ``apply_N`` was moved to the new module ``mathics.core.evaluators``. The logic of ``mathics.builtin.atomic.numeric.N.n_apply_with_prec``
  was also moved to that module.
* most of flake8 errors on the changed modules were fixed.


rocky (approved):
LGTM - let's merge in and continue

There is always more to do do. NIntegrate could use a summary description (along with Reals and Root) 

But let's just merge and do in another PR.

PR #144, opened 2022-01-31
Make a pass separating atomic.numbers from numeric

rocky:
Here I've separated out all of the functions from `atomic.numbers` (Representation of Numbers) that aren't listed in https://reference.wolfram.com/language/guide/RepresentationOfNumbers.html

As always, there is more work that could be done, but this is all the time I have for this for now. As as is custom, my take is with the exception of whatever smallish changes noticed, we merge and continue.

There are probably more functions elsewhere that belong here. And a number of the inappropriate ones have been shuffled back to `numeric` (Numerical Functions) or https://reference.wolfram.com/language/guide/NumericalFunctions.html


mmatera:
LGTM

mathics/builtin/numbers/exptrig.py
```diff
@@ -21,11 +25,140 @@
 from mathics.core.symbols import Symbol
 
 from mathics.builtin.arithmetic import _MPMathFunction
-from mathics.builtin.atomic.numbers import Fold
 
 from mathics.core.attributes import listable, numeric_function, protected
 
 
+class Fold(object):
```
TiagoCavalcante, line 32: In Python3 the `(object)` part isn't needed, and usually is hidden.

mathics/builtin/numbers/exptrig.py
```diff
@@ -21,11 +25,140 @@
...
+        FLOAT: ComputationFunctions(
+            cos=math.cos,
+            sin=math.sin,
+        ),
+        MPMATH: ComputationFunctions(
+            cos=mpmath.cos,
+            sin=mpmath.sin,
+        ),
+        SYMBOLIC: ComputationFunctions(
+            cos=lambda x: Expression("Cos", x),
```
TiagoCavalcante, line 52: SymbolCos

mathics/builtin/numbers/exptrig.py
```diff
@@ -21,11 +25,140 @@
...
+            sin=lambda x: Expression("Sin", x),
```
TiagoCavalcante, line 53: SymbolSin

mathics/builtin/string/operations.py
```diff
@@ -45,6 +47,96 @@
...
+        h = hash_func()
+        user_hash(h.update)
+        res = h.hexdigest()
+        if py_format in ("HexString", "HexStringLittleEndian"):
+            return String(res)
+        res = int(res, 16)
+        if py_format == "DecimalString":
+            return String(str(res))
+        elif py_format == "ByteArray":
+            return from_python(bytearray(res))
```
TiagoCavalcante, line 130: from_python can be replaced by an Expression.

TiagoCavalcante (approved):
@rocky I know that the changes you've made are related to move the code from one place to another, but I believe it is good to use the standard on every added line.

mathics/builtin/numbers/exptrig.py
mmatera: BTW, `Fold` is also a Builtin. Maybe this is an opportunity to rename it to `_Fold` or something else...

mathics/builtin/string/operations.py
rocky: @TiagoCavalcante Please just go ahead and make the changes you suggest. Thanks.

Issue #146, opened 2022-02-01
SameQ for PrecisionReal


mmatera:
When `SameQ` is applied to two `PrecisionReal` numbers that differs in a quantity under the precision, WMA returs `True`, while Mathics returns  `False`. The issue does not affect to `Equal`. For example, in WMA, 
```
In[1]:= a=N[2/9,3]; b=0.22222`5; {a,b,a==b, a===b}
```
returns 
```
Out[1]= {0.222,0.22222,True, True}
```
while in Mathics,
```
Out[1]:=  {0.222,0.22222,True, False}
```
This issue comes up when we try to implement low-level tests (see for example #145), and seems to be related to how the comparison operator is implemented in Sympy. 

A similar issue comes up if we evaluate `a-b`. In WMA, 
```
In[2]:={a-b, Precision[a-b]}
```
results in 
```
Out[2]={0.`, 0.}
```
while in Mathics,
```
Out[2]={0.0000153, 3.}
```

A possible fix would be to check the precision in  `PrecisionReal.sameQ` and in the constructors of `Real` and `PrecisionReal`.

rocky:
> When `SameQ` is applied to two `PrecisionReal` numbers that differs in a quantity under the precision, WMA returs `True`, while Mathics returns `False`. The issue does not affect to `Equal`. For example, in WMA,
> 
> ```
> In[1]:= a=N[2/9,3]; b=0.22222`5; {a,b,a==b, a===b}
> ```
> 
> returns
> 
> ```
> Out[1]= {0.222,0.22222,True, True}
> ```
> 
> while in Mathics,
> 
> ```
> Out[1]:=  {0.222,0.22222,True, False}
> ```

Currently, we get:

```
Out[1]= {0.222,0.22222,True, True}
```

I think this issue should be revised for the part that is still relevant so it is easier to track.

PR #147, opened 2022-02-02
SameQ testing against Precision Real and adding SameQ unit tests

mmatera:
This PR fixes the issue #146:

* Improved `sameQ` routine for `PrecisioReal` and `MachineReal` to behave like in WMA.
* Adding tests.

One alternative would be to ask to the sympy developer team to implement a similar fix.


mathics/core/atoms.py
```diff
@@ -448,9 +448,13 @@ def sameQ(self, other) -> bool:
...
         if isinstance(other, MachineReal):
             return self.value == other.value
-        elif isinstance(other, PrecisionReal):
-            return self.to_sympy() == other.value
-        return False
+        if isinstance(other, PrecisionReal):
+            other_value = other.value
+        else:
+            return False
+        value = self.value
```
rocky, line 455: I think this logic could be simplified by moving it in the "then" branch above. 

Also, I think the docstring here could be more detailed. 

I also see that none of the Comparison operators appear in the docs.

I will put in a PR based on this to start to address this.

rocky:
Let's merge and continue....

PR #148, opened 2022-02-03
SameQ tweaks -

rocky:
* Simplify SameQ logic for MachineReal
* Hook comparison operators into documentation
* Fill out description and examples for SameQ

Note: comparison.py probably should be split out into its own subsection

As always, with regards to documentation, and sectioning there is a lot more there to  do.


mmatera (approved):
LGTM

PR #149, opened 2022-02-03
Improving NumericQ

mmatera:
This PR improves the compatibility for `NumericQ` evaluations in several ways:
* Allowing to mark symbols as Numeric as a special case of assignment (`NumericQ[a]=True`).
* Adding the symbol `NumericFunction` that allows working with the corresponding attribute.
* Improving the implementation of `Symbol.is_numeric` and `Expression.is_numeric`


rocky:
This is probably needed for what will come next. We should probably merge in when you are satisfied.

PR #150, opened 2022-02-05
Pytest evaluators2

mmatera:
the same as #145 but without changing the name `apply_N->eval_N`


mathics/core/evaluators.py
```diff
@@ -37,20 +43,32 @@ def apply_N(expression, evaluation, prec=SymbolMachinePrecision):
     return result.evaluate(evaluation)
 
 
-def apply_nvalues(expr, prec, evaluation):
+def apply_nvalues(
```
rocky, line 46: I appreciate and like the comments added below. I feel that somehow there is some higher-level idea that we might be missing.  So I'd just suggest a comment above this function to look over again at some future point. 

For example: 

```
# TODO:  Revisit - can this be simplified? Is some broader framework this fits into?

mmatera: Done.

rocky:
@mmatera I apologize for not having gotten this merged in before my leaves -> elements disruption. 

I tried to resolve the merge conflicts, but I couldn't this is to large and complex and I just don't understand it all that well.

Would you please resolve the merge conflicts so we could merge this. It might be useful in this regard to split it up into smaller PRs.

Possibly next week (not this week), I'd like to continue with replacing "leaves" with "elements" when that is what we mean. And in particular replacing the `@leaves` property on Expression with `@elements` would probably be the next thing. (And then after that we probably could replace `.get_elements()` with `.elements` and things would be simpler and tidier. 

So it would be good to fix this and any outstanding PRs that may cause conflicts. 

Many thanks.

mmatera:
OK, I think I can handle it.

mmatera:
> Do you have an example, then this rule will apply? In the best case `b^a` is returned, so no change?

``0^a`` can not be evaluated in general, however, ``Simplify[0^a,Assumptions-> a>0]`` evaluates to ``0``. 
The second rule, for `a^b` tries to ensure that  ``a^0`` simplifies to 1 if ``a!=0``. However, in WMA (and in Mathics)  ``a^0`` evaluates to ``1`` even if ``a`` could be 0, so in the current state, that rule is irrelevant.

mmatera:
@rocky, I merge this before merging your new PR to avoid further file conflicts.

PR #155, opened 2022-02-12
First step in evaluation refactor...

rocky:
evaluate_next() -> rewrite_apply_eval()

Start commenting important classes Atom, BaseExpression

Not sure if this should be a draft of whether we want to get changes little bits and pieces at a time.

One thing that leaps out in just this much so far is that we don't need the rewrite/apply aspect for most (all) Atoms with the exception of possibly Symbols that would be better called SymbolVariable were this not so ingrained. (There are other Symbols however like True and False and CompiledCode that also probablyl shouldn't go through the rewrite/apply aspect either.)


mmatera:
As far this is just comment and rename a few things, I am OK to merge it, but to go further, it would be nice to have some plan and a diagram of the new model before start changing the core.

rocky:
> As far this is just comment and rename a few things, I am OK to merge it, but to go further, it would be nice to have some plan and a diagram of the new model before start changing the core.

The plan is to understand the code first, which is being done by a combination of working the docstrings and the lager-picture read-thedocs description.

If small stuff like this falls out, I have a slight preference, given our working habits and how things have worked out in the past, to make PRs for small uncontroversial stuff with the possibly small additional changes here and there (e.g. `class Foo` instead of `class Foo(object)` is something of this kind of variety that has come up). 

Larger-scale stuff I'll probably do in a branch privately and make experiments before opening discussion. As we have seen in the past, I have have often made draft PRs that suggest a direction and are either abandoned or are totally reworked and improved. That might happen here, but that would happen a bit later.  Things are still at a very early stage. 

But even here, it feels already that this heavy-weight `evaluate-next()` or `rewrite-apply-eval()` process is going on for many Atoms when it shouldn't. Note: I am not suggesting dropping everything and fixing that. But right now just making note of it. 
If someone want's to benchmark that to see what effect this has that'd be great. I'm not going to do that now though.

Trying to get any more detailed plan  let alone a diagram at this stage when I am trying to get a grip on what is here, and what is going on is premature. In the past there have been more myopic suggestions and detailed proposals that after the fact felt like they didn't really address some of the systemic design decisions precisely because the design was neither explored or understood very well.

rocky:
@TiagoCavalcante I have been meaning to describe how this relates (or not) to the earlier work with that had `evaluate_fast()` which allowed an argument to speed up Python method lookup. 

Now that I understand a bit more, the short answer is that this is still a valid thing to consider at some point, but as the name suggest -- that we do an rewrite step first , _then_ followed by an apply step to get to the eval stage. This kind of speedup makes sense at a level lower down, not here. 

At least to me the combination of unhelpful name and lack of basic documentation is the kind of thing that leads to this kind of confusion. And as we may find out there are more fundamental things wrong than this. I can think of several right now that have been alluded to in the past, but best not to go down that rabbit hole until we get some basics established.

TiagoCavalcante:
@rocky I agree with this change, `evaluate_next` seems a thing that should be used in loops, but I already needed to throw some code out because I knew that fixing it would take more time than rewriting it from scratch, and my mistake was I assumed evaluate_next/evaluate_fast worked in ~10 cases, so it should work in the other ones, but it didn't. This worked because the tests (as we already know) are incomplete, and I believe it is impossible (well, at least would take wayyy more time than it should) to write tests without understanding the code.

About writing diagrams, I believe the "right" thing is to plan a diagram before writing the code, but [at least for me] drawing diagrams take more time than writing the code, and I guess that you wouldn't draw just 1 diagram if you made it before writing the code, you'd draw a few diagrams, because every time you (at least I) look at the evaluation code, there is something new which we (I) didn't know/didn't understand/is just wrong.

I believe our objective should be to have just 1 evaluation method, not 3. Currently, we don't have [m]any Pymathics modules from other teams, but when it happens, I guess they won't try to understand why there are other evaluation methods, they are just going to pick `.evaluate`.  
And what I understand from the code tells that these other evaluation methods could be specializations of `.evaluate` or FinalForm (I don't remember which name was decided).

---

TLDR: LGTM

PR #157, opened 2022-02-20
Start converting leaves -> elements

rocky:
I realize when other change X to Y it seems silly. But here X (leaves) is just plain wrong and Y (elements) is correct.

"elements" is the term Mathematica uses and it  is a perfectly appropriate term. 

"leaves" is just plain wrong. I don't know why but its bone headedness annoys me. It may have to do with this being at the level we find ourselves having to document, and then rework the code in the near future so we can get reasonable performance. Or maybe it is because I thought we were really talking about leaves of a tree and I now find this isn't correct and I am annoyed about having been confused by such a thoughtless kind of thing.

We have in the existing code running AtomQ on a "leaf" so see if it has children. This is confused/confusing.

It may seem a small thing, but if we can't think clearly about the code I think it is going to be harder to discuss it and improve it.

### DIFF
--- a/mathics/algorithm/integrators.py
+++ b/mathics/algorithm/integrators.py
@@ -178,8 +178,8 @@ def apply_D_to_Integral(func, domain, var, evaluation, options, head):
     # if the integration is along several variables, take the integration of the inner
     # variables as func.
     if domain._head is SymbolSequence:
-        func = Expression(head, func, *(domain._leaves[:-1]), *options)
-        domain = domain._leaves[-1]
+        func = Expression(head, func, *(domain._elements[:-1]), *options)
+        domain = domain._elements[-1]
 
     terms = []
     # Evaluates the derivative regarding the integrand:
--- a/mathics/algorithm/parts.py
+++ b/mathics/algorithm/parts.py
@@ -42,7 +42,7 @@ def get_part(varlist, indices):
             if cur.is_atom():
                 raise PartDepthError(rest[0])
             pos = rest[0]
-            leaves = cur.get_leaves()
+            leaves = cur.get_elements()
             try:
                 if pos > 0:
                     part = leaves[pos - 1]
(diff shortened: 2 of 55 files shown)


rocky:
Although this commit is somewhat large, it is conservative, believe it or not. In other words I haven't been overly aggressive about finding _all_ of the opportunities. And in cases where something _might_ be leaves I've left it as is. For example if I create a list of Integers, integers being atoms would be leaves. 

There seems to be this `leaves` accessor function which may be inaccurate too. So those and things involving those I've left for some other time and PR. 

Finally it is kind of funny that we now sometimes see in graphical items `xxx.elements.elements`. The first element is an element of the graphic item, and the second element is some (possibly nested or structured) item in that. 

The way to address this is _not_ to call the second element a leaves (when it probably isn't), but rather to clarify that one thing is a xxx kind of element where the second thing is a yyy kind of element, e.g. possibly a generic graphics element.

mmatera:
@rocky, if it is only a change of names, and you think is very helpful to change the misleading name, go ahead. The drawbacks I find here are:
* we need to change this in all the other projects.
* I would have to check but, in WL, I think the head of an expression is also an element.

rocky:
> @rocky, if it is only a change of names, and you think is very helpful to change the misleading name, go ahead. The drawbacks I find here are:
> 
> * we need to change this in all the other projects.

I'll check the ones I know about. 

> * I would have to check but, in WL, I think the head of an expression is also an element.

Please check and let me know what you find.

mmatera:
> > @rocky, if it is only a change of names, and you think is very helpful to change the misleading name, go ahead. The drawbacks I find here are:
> > 
> > * we need to change this in all the other projects.
> 
> I'll check the ones I know about.
> 

OK

> > * I would have to check but, in WL, I think the head of an expression is also an element.
> 
> Please check and let me know what you find.

No, you are right: https://reference.wolfram.com/language/guide/ElementsOfLists.html
The head is a *Part* of a list. The other parts are their  *elements*.

mmatera (approved):
LGTM

rocky:
> @rocky, if it is only a change of names, and you think is very helpful to change the misleading name, go ahead. The drawbacks I find here are:
> 
> * we need to change this in all the other projects.
> * I would have to check but, in WL, I think the head of an expression is also an element.

I see one place in mathics django, and two places in mathicsscript.

So a very simple thing is to keep the get_leaves() function as an alias to get_elements() for a little bit.

But ideally if things were better designed in terms of how things get formatted we wouldn't have needed those to places to begin with .

rocky:
> LGTM

Great! Let's do and iterate. @TiagoCavalcante if you have comments feel free to make them and we'll add them in the next round.

TiagoCavalcante:
@rocky, ok. I agree that "leaves" was wrong.

PR #158, opened 2022-02-20
Fix153

### LINKED ISSUE #153: Double integration with `f(x,y)=1` fails, while any other constant succeeds
`Integrate[Integrate[1,{y,0,E^x}],{x,0,Log[13]}]` gives `Integrate[ConditionalExpression[E, {True}], {x, 0, Log[13]}]` when `Integrate[Integrate[2,{y,0,E^x}],{x,0,Log[13]}]` or `Integrate[Integrate[0,{y,0,E^x}],{x,0,Log[13]}]` or even `Integrate[Integrate[0.9999999,{y,0,E^x}],{x,0,Log[13]}]` works.

mmatera:
This PR fixes #153, which was due to a but in the `Simplify` implementation of the rule `Simplify[b_^a_, assumptions_]`

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -63,7 +63,7 @@ Bugs
 * ``N`` now handles arbitrary precision numbers when the number of digits is not specified.
 *  ``Set*``: fixed issue #128.
 *  ``SameQ``: comparison with MachinePrecision only needs to be exact within the last bit Issue #148.
-
+* Fix a bug in `Simplify` that produced expressions of the form ``ConditionalExpression[_,{True}]``.
 
 4.0.1
 -----
--- a/mathics/builtin/numbers/algebra.py
+++ b/mathics/builtin/numbers/algebra.py
@@ -416,8 +416,8 @@ class Simplify(Builtin):
         "Simplify[rule_Rule]": "Simplify /@ rule",
         "Simplify[list_List, assum_]": "Simplify[#1, assum]& /@ list",
         "Simplify[rule_Rule, assum_]": "Simplify[#1, assum]& /@ rule",
-        "Simplify[0^a_, assum_]": "ConditionalExpression[0,Simplify[a>0]]",
-        "Simplify[b_^a_, assum_]": "ConditionalExpression[b,Simplify[{Or[a>0, b!=0]}]]",
+        "Simplify[0^a_, assum_]": "ConditionalExpression[0,Simplify[a>0, assum]]",
+        "Simplify[b_^a_, assum_]": "ConditionalExpression[b,Simplify[Or[a>0, b!=0],assum]]",
     }
 
     def apply_assuming(self, expr, assumptions, evaluation):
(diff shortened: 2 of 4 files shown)


rocky:
This is good, but what I would ask is (and I don't know if this is relevant here) whether this is a workaround for a misfeature of the code or something you might reasonably find in _any_ WL implementation? 

I looked in Symja for something like this and don't see anything, but I could have missed something. 

If it should be something that we may want to revisit in the future, it would be great if this could be tagged in some way. Even better would be segregated this to a section which makes it easier to review such decisions should things change to make it not necessary. 

But, again, I don't know if this is relevant here.

mmatera:
> This is good, but what I would ask is (and I don't know if this is relevant here) whether this is a workaround for a misfeature of the code or something you might reasonably find in _any_ WL implementation?
> 
> I looked in Symja for something like this and don't see anything, but I could have missed something.
> 
> If it should be something that we may want to revisit in the future, it would be great if this could be tagged in some way. Even better would be segregated this to a section which makes it easier to review such decisions should things change to make it not necessary.
> 
> But, again, I don't know if this is relevant here.

No, this is just a fix for a bug in the implementation of `Simplify`, that has some consequences in `Integrate`. Maybe the rule shouldn't be there, or maybe Symja does not have support for the parameter ``assumptions``. 
For example, in Mathics (and WMA)
```
In[1]:=0^a
Out[1]=0^a
```
while
```
In[2]:=Simplify[0^a,a>0]
Out[2]=0
```
or
```
In[3]:=$Assumptions={a>0}; Simplify[0^a]
Out[3]=0
```
What is the output in Symja?

rocky:
>> But, again, I don't know if this is relevant here.
> 
> No, this is just a fix for a bug in the implementation of `Simplify`, that has some consequences in `Integrate`. Maybe the rule shouldn't be there, or maybe Symja does not have support for the parameter `assumptions`. For example, in Mathics (and WMA)
> 
> ```
> In[1]:=0^a
> Out[1]=0^a
> ```
> 
> while
> 
> ```
> In[2]:=Simplify[0^a,a>0]
> Out[2]=0
> ```
> 
> or
> 
> ```
> In[3]:=$Assumptions={a>0}; Simplify[0^a]
> Out[3]=0
> ```
> 
> What is the output in Symja?

symja-simplify

mmatera:
OK, so in that case, we are doing it better...

rocky:
Ok - so then let's merge and iterate.

TiagoCavalcante:
@rocky could you open an issue in Symja for this?

rocky:
I am not  the right person for this. I am not really a Symja user.

axkr:
Changed it in the code base:
- https://github.com/axkr/symja_android_library/issues/456

rocky:
@mmatera @axkr noticed something that is still amiss in our code: 

```
In[1]:= Simplify[0^x, x<0]
Out[1]= Undefined
```

It should be instead ComplexInfinity, right?

PR #159, opened 2022-02-20
sanity test to pass in a more aggressive sympy translation

mmatera:
One of the ideas about how to speed up certain evaluations in Mathics would be to try to send to sympy complete expressions instead of doing several steps of partial conversions in the evaluation loop. 
The risk of doing it is that it could break the expected evaluation logic in WL. Here I propose some tests to detect this kind of bugs.

### DIFF
--- a/test/test_calculus.py
+++ b/test/test_calculus.py
@@ -3,7 +3,7 @@
 Unit tests from builtins ... calculus.py
 """
 
-from .helper import check_evaluation
+from .helper import check_evaluation, session
 
 
 def test_calculus():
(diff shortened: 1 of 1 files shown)


rocky:
I definitely like the idea! Write the tests first!

mmatera:
> I definitely like the idea! Write the tests first!
> 

I am learning! 

> Things are failing though currently. :-(

Yes, because there was a previous definition of some of the symbols. Now it works.

rocky:
Github workflows CI testing on OSX is currently broken. Rather than wait, I think we should merge this in now. Since this passes on other OS's and since this just adds a few tests, I doubt this introduces a bug, and adjusting the single test should it prove to be a problem can easily be done later. On the other hand, our time to devote on this may be limited.

PR #160, opened 2022-02-20
Improve association - rebase PR#90

mmatera:
From original:  associations should be primitives

@TiagoCavalcante, @rocky, this rebases PR#90 over the new changes.


rocky:
Looks like github workflows is having troubles with OSX builds. Since a few of these work we'll presume the others will as well.

TiagoCavalcante:
@mmatera thanks for rebasing this!

PR #161, opened 2022-02-20
{leaf,leaves}->element{s}, part 2

rocky:
Also there are some S-Expression to M-expression changes and things of that kind.

But of particular interest were the changes of "leaf" to "element" in  to rewrite_apply_step which is where we will be focusing.

@mmatera @TiagoCavalcante I generally think of myself as a live and let live kind of person - if there is a mess in the code and I don't have to be confronted with it I am happy to ignore it. 

The problem is this leaves versus elements confusion is right at the place I am looking at now in rewrite_apply_eval_step. 

So then I ask myself do I want to just keep reminding myself, "black" really means "red "and "up" really means "left" or do we fix this so I don't  have to do this and get confused. 

**BEGIN RANT**

I think what annoys me most is that _every_ time I want go off and look at something, there is already a mess that has been left there from the beginning that we need to deal with before we can even consider getting to the part I might be interested in. And so anything I am interested in has generally put off indefinitely (so far). 

And the wrongness has been so pervasive and in such an advanced state that we can't even fix it in one go. Redoing documentation is like that.  Getting to Jupyter is like that too. 

**END RANT** 

So here we are at another PR. 

There are tons of _other_ conversions from leaf to element that probably should occur. In a sense I am not happy that we are in this hybrid state that sometimes we have to still refer to an element as a "leaf" or "leaves" when it is not. 

But on the other hand I'd rather be in a hybrid state pointing to the right direction rather than always wrong, and being complacent about it being wrong. 

BTW - Here is what I was doing that motivated this: I read in the Mathematica book for version 5 that describes the rewrite_apply_eval step that the first thing to do is to check to see if the element (or stupidly called "leaf" in that section of our code) is a literal value, it can just return the value and not have to do any fancy rewrite/apply. 

It doesn't look like we do that currently. @mmatera do you think this simple check is still correct?  I suspect it might result in a noticeable speed improvement if it is still accurate.


TiagoCavalcante (approved):
LGTM

mmatera (approved):
LGTM

PR #163, opened 2022-02-21
Create CODE_OF_CONDUCT.md

rocky:
Although this is a good thing to do in of itself. It will also be useful should we want to join NumFocus (which I am in favor of).

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,128 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, religion, or sexual identity
+and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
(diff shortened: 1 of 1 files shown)


mmatera:
LGTM

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,128 @@
...
+like social media. Violating these terms may lead to a temporary or
+permanent ban.
+
+### 3. Temporary Ban
+
+**Community Impact**: A serious violation of community standards, including
+sustained inappropriate behavior.
+
+**Consequence**: A temporary ban from any sort of interaction or public
+communication with the community for a specified period of time. No public or
```
TiagoCavalcante, line 101: Should we specify this?

TiagoCavalcante (approved):
LGTM

CODE_OF_CONDUCT.md
rocky: I selected the "Contributor Covenant" text unmodified. Personally, I don't want to get into the business of customizing this.

It's sort of how you don't like it when there isn't a newline after triple quotes in a docstring. 

But joking aside, if we start to customize this then it makes it harder for organizations like NumFocus to decide whether they want to do "business" with us.

mmatera: I think that since we do not have any experience with this kind of behavior issue, the standard is OK.

rocky:
@TiagoCavalcante would you do the same for mathics-threejs-backend and Django? 

I'll do the other repositories when I get a chance or anyone feel free to just do the same for the others. Thanks.

TiagoCavalcante:
@rocky yes, I can do that.

PR #164, opened 2022-02-21
Mostly better code documentation...

rocky:
some more lea{f,ves} -> element{,s} converted in places we are currently
trying to describe and don't want to have to explain that by X we really
mean Y.

### DIFF
--- a/mathics/core/parser/ast.py
+++ b/mathics/core/parser/ast.py
@@ -1,8 +1,18 @@
-#!/usr/bin/env python3
 # -*- coding: utf-8 -*-
+"""
+Classes and Objects that the parser uses to create an initial Expression (an M-Expression).
+
+The parser's AST is an M-Expression.
+"""
 
 
-class Node(object):
+class Node:
+    """
+    The base class for a node or "elements" of an M-Expression.
+    Really there are only two kinds of nodes: Atoms which are the
+    expression's leaves and a non-leaf nodes.
+    """
+
     def __init__(self, head, *children):
         if isinstance(head, Node):
             self.head = head
@@ -43,6 +53,11 @@ class Node(object):
 
 
 class Atom(Node):
+    """
+    Atoms form the leaves of an M-Expression and have no internal structure of
+    their own. You can however compare Atoms for equality.
+    """
+
     def __init__(self, value):
         self.head = Symbol(self.__class__.__name__)
         self.value = value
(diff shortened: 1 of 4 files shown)


mathics/core/parser/ast.py
```diff
@@ -105,10 +137,17 @@ def __repr__(self):
...
+    A string is is pretty much the same as in any other programming language, a sequence of characters.
+    Having this in a class is useful so that we can distinguish it from Symbols.
+    The display of a String is surrounded by double quotes.
+    """
+
     def __repr__(self):
         return '"' + self.value + '"'
 
 
+# rocky: I really don't get why Filename has to be special, but so be it.
```
TiagoCavalcante, line 150: I also don't get it.

TiagoCavalcante (approved):
LGTM

mmatera:
LGTM

mmatera:
Side comment: some time ago, I noticed that in the AST module we have classes with names that appears also as ``BaseExpression`` s. For example, we have two classes called ``Symbol``. Do we really need these two classes? Could be just one? Otherwise, shouldn't we rename the AST Symbol to avoid confussions?

mathics/core/parser/ast.py
mmatera: This is just because we choose different print representations: Strings are rounded by quotes, and filenames don't.

rocky:
> Side comment: some time ago, I noticed that in the AST module we have classes with names that appears also as `BaseExpression` s. For example, we have two classes called `Symbol`. Do we really need these two classes? Could be just one? Otherwise, shouldn't we rename the AST Symbol to avoid confusions?

Yeah, I wondered about this too. This may be yet more sloppiness, made harder to find due to a lack of positive description of intent. Or maybe it _was_ the intent. Like Fermat's last theorem, keeping quiet is sometimes be assumed to be brilliance.

Python's duck typing doesn't see this as an error. That the API interfaces are compatible is probably not be a coincidence. (Because if this was not the case, then Python would throw an exception.)

At the Python level, one class is parser.Atom and the other class s expression.Atom. And I could see having these two be the same because they are equivalent being intentional or being viewed as the right thing.

Renaming is a possibility and so is subclassing. 

But right now though I'd rather keep this as is and come back to this after we think about it some more. I'll add a comment to note the issue and think about and possibly take action.

mathics/core/parser/ast.py
rocky: Ok. I guess that makes sense, but then why isn't Filename distinct from Symbol? 

Is this how Mathematica does things too? I would have hoped they would say use another pair of delimiters like `<Filename>`.

I'll add a comment about this upon learning that `__repr__` really is the "right" thing to do.

PR #165, opened 2022-02-21
Surname for node classes in the AST

mmatera:
(See https://github.com/Mathics3/mathics-core/pull/164#issuecomment-1046807179)
This change affects just a small portion of the code and avoids confusion ( I face something like that sometime ago, while I was working on singletonize the symbol class).

### DIFF
--- a/mathics/core/parser/README.md
+++ b/mathics/core/parser/README.md
@@ -3,7 +3,8 @@
 ## Introduction
 
 The Mathics parser is an operator precedence parser that implements the
-precedence climbing method.
+precedence climbing method. The goal of this module is to built an Abstract Syntax Tree (AST)
+from a string containing a valid WL expression.
 
 The Wolfram language is surprisingly complex and has quite a few subtleties;
 this document attempts to cover the major ones. The language is documented on
--- a/mathics/core/parser/ast.py
+++ b/mathics/core/parser/ast.py
@@ -7,13 +7,13 @@ class Node(object):
         if isinstance(head, Node):
             self.head = head
         else:
-            self.head = Symbol(head)
+            self.head = SymbolNode(head)
         self.value = None
         self.children = list(children)
         self.parenthesised = False
 
     def get_head_name(self):
-        if isinstance(self.head, Symbol):
+        if isinstance(self.head, SymbolNode):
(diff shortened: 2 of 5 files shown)


rocky:
I am not 100% certain about this. The use of the same name base name might be intentional because the interfaces are intended to be upward compatible. 

So here what might be acceptable or better (we should investigate what prevailing practice is) is to just document the behavior and state that the similarity is intentional. Or have symbols.Atom subclass parser.Atom.

And _if_ we were to rename things, I would use for Symbol something like SymbolLiteral (but Symbol is in fact in my opinion the best name)

mmatera:
> And _if_ we were to rename things, I would use for Symbol something like SymbolLiteral (but Symbol is in fact in my opinion the best name)

In any case, what we learned from this is that the only place in which both hierarchies coexist is in the `mathics.core.convert` module.

rocky:
Hmm. You know I wonder if we could use mathics.core.parser.Symbol for names like True, False, Null,  and so on all instead of mathics.core.symbol.Symbol?

mmatera:
The question is: what is the difference between both classes? From my point of view, ``mathics.core.symbols.Symbol`` is just ``mathics.core.parser.Symbol`` with a lot of methods that shouldn't be there...

rocky:
> The question is: what is the difference between both classes? From my point of view, `mathics.core.symbols.Symbol` is just `mathics.core.parser.Symbol` with a lot of methods that shouldn't be there...

If that's the case then a simple migration path is to substitute `mathics.core.parser.Symbol`  and if everything works out remove `mathics.core.symbols.Symbol`

mmatera:
> Symbol

OK, maybe I was not clear: ``BaseExpression``  has a lot of methods, like `format`, `evaluate`, `numerify`, `is_true`, `is_symbol`, `is_atom` and a long etc. Some of these methods can be avoided just by using the standard `isinstance` keywords. The others, produce new expressions by applying some kind of re-writting rules. These ones, I guess, should be implemented out of the class, in a way that circular module reference be avoided. 

So, right now, the trivial change will fail.   However, this could be a guidance to reformulate the core class hierarchy.

Issue #154, opened 2022-02-10
Double Integration is very slow, while almost instant in maxima.


DUOLabs333:
`Integrate[Integrate[20000*E^y/(1-x/2),{x,-5,0}],{y,-3,0}]+Integrate[Integrate
[20000*E^y/(1+x/2),{x,0,5}],{y,-3,0}]` is much slower in Mathics than maxima.

DUOLabs333:
If you switch the x and y parts, you can do it (you have to do `2000.` to force Mathics to simplify the `Log`).

DUOLabs333:
It seems that Maxima keeps the intermediate results in symbolic form, while Mathics (sympy) tries to expand it.

rocky:
This is just one instance of an infinite number of examples of a long-standing problem regarding how valuation is done in Mathics and how it uses SymPy as a backend computation engine. 

It is being worked on currently, albeit slowly.

DUOLabs333:
So, is Mathics working on it's own engine or are you waiting for Sympy to fix it on their end?

DUOLabs333:
Yeah, I did that already -- it was roughly the same speed.

mmatera:
Notice that iterated integrals are slower than integrals over several variables:

In[1]:= Timing[Integrate[Integrate[1,{y,0,E^x}],{x,0,Log[13]}]]
Out[1]= {0.0838364, 12}

In[2]:= Timing[Integrate[1,{y,0,E^x},{x,0,Log[13]}]]
Out[2]= {0.0450555, 12}

This is due to the conversion back and forward to sympy is done two times. Something that could help is to define a rule

``Integrate[Integrate[f_,x_],y_]->Integrate[f,x,y]``
to reduce the conversion to one step.

rocky:
I am not necessarily opposed to this, but in the long term we need a way to have SymPy applied to larger expressions in one go _automatically_.  

Otherwise we'll be adding rules for every combination of composition of SymPy functions imaginable.  So lots of work, lots of Rules which could keep us busy for a long time and in the end cause lots of Rule bloat, when this is finally solved. 

So maybe this could be isolated into some startup code were this is mentioned should we find one day it is no longer needed.

rocky:
I see _one_ place in mathics django, and _two_ places in mathicsscript. 

So a very simple thing is to keep the `get_leaves()` function as an alias to `get_elements()` for a little bit.

[...]

PR #166, opened 2022-02-21
timing TraceEvaluation

mmatera:
Trying to understand the problem in #154, I realized that it would be useful to record how long each evaluation step takes, so here is a PR that implements that.


TiagoCavalcante (approved):
LGTM - good idea!

rocky:
This had occurred to me a while ago and I think it is a good idea. But can we just make add it as an option to the exisiting $TraceEvaluation?

PR #168, opened 2022-02-21
leaf -> element more places

rocky:
Side Note: Tracing Add[1, 2] seems to do a lot of unnecessary data motion and conversion back and forth.

We convert to from a tuple to a list, evaluate elements that are already in their reduced form, try to flatten a flattened list convert that into a Mathics Sequence only to convert it back to something like what it was before. 

The lack of comments/commentary about what is done and why and lack of modularity (which I really still don't know how to deal with) just is obscuring this convoluted process. Sure, in some cases it needs to be this complex. Just not _all_ of the time.

### DIFF
--- a/mathics/core/atoms.py
+++ b/mathics/core/atoms.py
@@ -75,11 +75,24 @@ def from_mpmath(value, prec=None):
 
 
 class Number(Atom):
+    """
+    Different kinds of Mathics Numbers, the main built-in subclasses
+    being: Integer, Rational, Real, Complex.
+    """
+
     def __str__(self) -> str:
         return str(self.value)
 
     def is_numeric(self, evaluation=None) -> bool:
         return True
+
+    def evaluate(self, evaluation) -> "Number":
+        """Evaluation of a Number is just itself"""
+        # Why bother checking for a time out? If there should be one,
+        # it will probably be caught at a higher level. And
+        # returning `self` is pretty fast anyway.
+        # evaluation.check_stopped()
+        return self
 
 
 def _ExponentFunction(value):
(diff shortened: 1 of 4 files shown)


mathics/core/atoms.py
```diff
@@ -75,12 +75,24 @@ def from_mpmath(value, prec=None):
...
+
     def __str__(self) -> str:
         return str(self.value)
 
     def is_numeric(self, evaluation=None) -> bool:
         return True
 
+    def evaluate(self, evaluation) -> "Number":
+        """Evaluation of a Number is just itself"""
+        # Why bother checking timout? If there is one
```
TiagoCavalcante, line 91: Also: evaluating a number is (should be) instantaneous, so it is probably not the best place to check for timeouts.

TiagoCavalcante (approved):
LGTM

TiagoCavalcante:
@rocky this remembers me that Plus should be FinalForm (I don't remember the name), but its algorithm has a flaw that doesn't put terms together in just one evaluation.

rocky:
> @rocky this remembers me that Plus should be FinalForm (I don't remember the name), but its algorithm has a flaw that doesn't put terms together in just one evaluation.

I don't understand exactly what you mean. Maybe though it would be better to open an issue and give a detailed example.

TiagoCavalcante:
> I don't understand exactly what you mean. Maybe though it would be better to open an issue and give a detailed example.

I'll make a PR this weekend fixing this. In the Plus evaluation, we put together multiples of symbols (e.g.: y + 2y + x = 3y + x). But the Plus code put together just some variables, so something like this (but with way more terms) happen: Plus[a, b, c, a, b] -> Plus[2a, b, c, b] -> 2a + 2b + c

PR #169, opened 2022-02-22
Clarify "threading" in Exprs. DRY Numeric classes

rocky:
Small stuff.

With respect to `get_head_name()` this is in the class hierarchy and does the right thing (use the class name) for all of the functions removed. This function seems to be used mostly in `has_form()` but right now doesn't seem to be used very much. 

I did check that the custom functions give the same result as when they are removed.

With respect th threading, note the fact this has to do with expression transformation and is _not_ optional (as it would be for hardware threading).  Alas, right now don't do any sort of Python/OS threading so which seems to be what a large part of the motivation in the language for adding this.

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -1014,18 +1014,20 @@ class Expression(BaseExpression):
 
         new._timestamp_cache(evaluation)
 
-        # Step 5: Would the operation benefit running in separate threads?
+        # Step 5: Must we need to thread-rewrite the expression?
         #
-        # We allow threading in when head has the ``Listable``
-        # Attribute.  Here ``Expression.thread`` changes:
-        #  ``F[{a,b,c,...}]`` to:
+        # Threading is needed when head has the ``Listable``
+        # Attribute.  ``Expression.thread`` rewrites the expression:
+        #  ``F[{a,b,c,...}]`` as:
         #  ``{F[a], F[b], F[c], ...}``.
 
-        # TODO: For a small number of arguments threading is just overhead.
-        # For zero or one argument this is always the case.
-        # For two arguments, note that many listable functions are binary operators
-        # so two arguments is really the same as one argument and threading is overhead.
+        # Note: Threading here is different from Python or OS threads,
+        # even though the intent of this attribute was to allow for
+        # hardware threading to make use of more cores.
         #
+        # Right now, we do not make use of Python thread or hardware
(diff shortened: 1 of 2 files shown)


TiagoCavalcante (approved):
LGTM

mmatera:
LGTM

mathics/core/expression.py
```diff
@@ -1014,18 +1014,20 @@ def flatten_callback(new_elements, old):
...
+        # Attribute.  ``Expression.thread`` rewrites the expression:
+        #  ``F[{a,b,c,...}]`` as:
         #  ``{F[a], F[b], F[c], ...}``.
 
-        # TODO: For a small number of arguments threading is just overhead.
-        # For zero or one argument this is always the case.
-        # For two arguments, note that many listable functions are binary operators
-        # so two arguments is really the same as one argument and threading is overhead.
+        # Note: Threading here is different from Python or OS threads,
+        # even though the intent of this attribute was to allow for
```
mmatera, line 1025: I would say that ``Thread`` in WL does not have anything to do with OS threads. In order to follow the semantics of the language, {F[a],F[b],...} -in general- must be evaluated sequentially.

rocky: This sentence from https://reference.wolfram.com/language/ref/Listable.html leads me to believe that parallelization is the motivation: 

> Arguments that are not lists are copied as many times as there are elements in the lists.

In particular, this means that the rewrite process can be done in parallel, and if the list is _very_ large there may be a benefit here. 

Also, from the Mathematica V5 book section 3.8.4 Operations on Scalars, Vectors and Matrices:

> Most mathematical functions in Mathematica are set up to apply themselves separately to each element in a list. This is true in particular of all functions that carry the attribute Listable. 

Another Language that Wolfram says he took inspiration from is APL, and this is definitely a design feature of that language.

mmatera: Interesting... that maybe works for atomic leaves. However, consider this:
```
In[6]:= Sin[{a=1,a=a+1,a=a+1}]                                                  

Out[6]= {Sin[1], Sin[2], Sin[3]}
```

rocky: The list copy is done in parallel. The _evaluation_ or _eval()_ part isn't.

axkr: See: https://reference.wolfram.com/language/ref/Parallelize.html
`Scope -> Listable functions`

PR #170, opened 2022-02-22
Extend `Simplify` to handle the `Assumptions` argument more correctly 

### LINKED ISSUE #167: `Simplify[0^x, x<0]` returns Undefined. Should it return ComplexInfinity?

mmatera:
This should fix #167 and improve the implementation of `Simplify`. Also, makes the semantics to be closed to the one in WMA:

Now, ``Simplify`` always takes into account ``$Assumptions``, except if the option ``Assumptions`` is passed.


mathics/builtin/numbers/algebra.py
```diff
@@ -405,44 +403,59 @@ class Simplify(Builtin):
...
-    # be restaured.
-    # >> $Assumptions={a <= 0};
-    # >> Simplify[ConditionalExpression[1, a > 0]]
-    # = Undefined
-    # >> Simplify[ConditionalExpression[1, a > 0], { a > 0 }]
-    # = 1
+    >> $Assumptions={a <= 0};
+    >> Simplify[ConditionalExpression[1, a > 0]]
+     = Undefined
+    >> Simplify[ConditionalExpression[1, a > 0], Assumptions -> { a > 0 }]
```
TiagoCavalcante, line 409: In Wolfram Language this is `SimplifyConditionalExpression[1, a > 0], { a > 0 }]`. But for me, having an option instead of a parameter is better.

mathics/builtin/numbers/algebra.py
```diff
@@ -405,44 +403,59 @@ class Simplify(Builtin):
...
             return expr
         # else, use sympy:
-        elements = [self.apply(element, evaluation) for element in expr._elements]
-        head = self.apply(expr.get_head(), evaluation)
+        # Here we want to pass through the full evaluation process to use all the defined rules...
+        elements = [
+            Expression("Simplify", element).evaluate(evaluation)
+            for element in expr._elements
+        ]
+        head = Expression("Simplify", expr.get_head()).evaluate(evaluation)
```
TiagoCavalcante, line 458: SymbolSimplify

mathics/builtin/numbers/algebra.py
```diff
@@ -470,12 +483,9 @@ class FullSimplify(Simplify):
...
 
     rules = {
-        "FullSimplify[list_List]": "FullSimplify /@ list",
-        "FullSimplify[rule_Rule]": "FullSimplify /@ rule",
-        "FullSimplify[eq_Equal]": "FullSimplify /@ eq",
-        "FullSimplify[list_List, assum_]": "FullSimplify[#1, assum]& /@ list",
-        "FullSimplify[rule_Rule, assum_]": "FullSimplify[#1, assum]& /@ rule",
-        "FullSimplify[eq_Equal, assum_]": "FullSimplify[#1, assum]& /@ eq",
+        "FullSimplify[list_List, OptionsPattern[]]": "FullSimplify /@ list",
+        "FullSimplify[rule_Rule, OptionsPattern[]]": "FullSimplify /@ rule",
```
TiagoCavalcante, line 487: Isn't it the behavior of a Listable function?

TiagoCavalcante (approved):
LGTM

mathics/builtin/numbers/algebra.py
mmatera: This is the documentation of `Simplify` in WMA 12:

```
Simplify[expr] performs a sequence of algebraic and other\
         
        >     transformations on expr
         
        >     and returns the simplest form it finds. Simplify[expr, assum]

         
        >      does simplification using assumptions. 


        Attributes[Simplify]={Protected}


        Options[Simplify]={Assumptions :> $Assumptions, 
         
        >    ComplexityFunction -> Automatic, ExcludedForms -> {}, 
         
        >    TimeConstraint -> 300, TransformationFunctions -> Automatic, 
         
        >    Trig -> True}
```

mmatera: However, it is true: ``Simplify[expr, assump]`` is equivalent to ``Simplify[expr, Assumptions->assump]``

rocky: I'd be grateful if you'd go over our Simplify doc that we have to expand the explanation for the expanded behavior. 

(That reminds me that I have been meaning to revise our doc for Expression to make it clear that in the user docs we are talking about the symbol rather than the data structure. I'll be doing that soon.)

mathics/builtin/numbers/algebra.py
mmatera: Better is ``self.get_name()`` to handle also ``FullSymplify``

mathics/builtin/numbers/algebra.py
mmatera: But is not listable...

PR #172, opened 2022-02-24
Add Reduce class

### LINKED ISSUE #171: `Reduce` not implemented
`Reduce[24.0-2x-x^2>0,x,Reals]` just returns `Reduce[24.0-2x-x^2>0,x,Reals]`.

DUOLabs333:
Should fix #171 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/172"></a>

### DIFF
--- a/mathics/builtin/numbers/calculus.py
+++ b/mathics/builtin/numbers/calculus.py
@@ -932,7 +932,8 @@ class Solve(Builtin):
 
     def apply(self, eqs, vars, evaluation):
         "Solve[eqs_, vars_]"
-
+        
+        InequalityOperators={"Less":"<","LessEqual":"<=","Greater":">","GreaterEqual":">="}
         vars_original = vars
         head_name = vars.get_head_name()
         if head_name == "System`List":
@@ -961,7 +962,15 @@ class Solve(Builtin):
             elif eq is SymbolFalse:
                 return Expression(SymbolList)
             elif not eq.has_form("Equal", 2):
-                return evaluation.message("Solve", "eqf", eqs_original)
+                for operator in InequalityOperators:
+                    if eq.has_form(operator,2):
+                        eq=eq.to_sympy()
+                        is_inequality=True
+                        break      
+                try:
+                    is_inequality
+                except NameError:      
+                    return evaluation.message("Solve", "eqf", eqs_original)
             else:
                 left, right = eq.leaves
                 left = left.to_sympy()
@@ -971,9 +980,9 @@ class Solve(Builtin):
                 eq = left - right
                 eq = sympy.together(eq)
                 eq = sympy.cancel(eq)
-                sympy_eqs.append(eq)
                 numer, denom = eq.as_numer_denom()
                 sympy_denoms.append(denom)
+            sympy_eqs.append(eq)
 
         vars_sympy = [var.to_sympy() for var in vars]
         if None in vars_sympy:
--- a/mathics/builtin/numbers/calculus.py
+++ b/mathics/builtin/numbers/calculus.py
@@ -879,7 +879,7 @@ class Root(SympyFunction):
                 return Expression(SymbolList)
             elif not eq.has_form("Equal", 2):
                 if any(eq.has_form(operator,2) for operator in ["Less","LessEqual","Greater","GreaterEqual"]):
-                    eq=eq.to_sympy()
+                    eq_sympy=eq.to_sympy()
                 else:     
                     return evaluation.message("Solve", "eqf", eqs_original)
             else:
@@ -888,12 +888,12 @@ class Root(SympyFunction):
                 right = right.to_sympy()
                 if left is None or right is None:
                     return
-                eq = left - right
-                eq = sympy.together(eq)
-                eq = sympy.cancel(eq)
-                numer, denom = eq.as_numer_denom()
+                eq_sympy = left - right
+                eq_sympy = sympy.together(eq_sympy)
+                eq_sympy = sympy.cancel(eq_sympy)
+                numer, denom = eq_sympyq.as_numer_denom()
                 sympy_denoms.append(denom)
-            sympy_eqs.append(eq)
+            sympy_eqs.append(eq_sympy)
 
         vars_sympy = [var.to_sympy() for var in vars]
         if None in vars_sympy:
--- a/mathics/builtin/numbers/calculus.py
+++ b/mathics/builtin/numbers/calculus.py
@@ -824,11 +824,11 @@ class Root(SympyFunction):
     >> Reduce[{x y == 0, y == 4},{x, y}]
      = y == 4 && x == 0
     #>> Reduce[{x y == 0, y > 4},{x, y}]
-     =  x == 0 && y > 4
+     #=  x == 0 && y > 4
     #>> Reduce[x*y>0,{x,y}]
-     = (x < 0 && y < 0) || (x > 0 && y > 0)
+     #= (x < 0 && y < 0) || (x > 0 && y > 0)
     #>> Reduce[{x*y > 0, y>0},{x,y}]
-     =  x > 0 && y > 0
+     #=  x > 0 && y > 0
     
     >> Reduce[a<b,a]
      =  b ∈ Reals && a < b
--- a/mathics/builtin/numbers/calculus.py
+++ b/mathics/builtin/numbers/calculus.py
@@ -812,6 +812,193 @@ class Root(SympyFunction):
         except Exception:
             return None
 
+ class Reduce(Builtin):
+    
+    """
+    <dl>
+       <dt>'Reduce[$eqs$, $vars$]'
+       <dd> Reduces the statements in $eqs$ by solving equations in $vars$ and eliminating quantifiers.
+    </dl>
+    >> Reduce[x y == 0,{x, y}]
+     = x == 0  ||   y == 0 
+    >> Reduce[{x y == 0, y == 4},{x, y}]
+     = y == 4 && x == 0
+    >> Reduce[{x y == 0, y > 4},{x, y}]
+     =  x == 0 && y > 4
+    >> Reduce[x*y>0,{x,y}]
+     = (x < 0 && y < 0) || (x > 0 && y > 0)
+    >> Reduce[{x*y > 0, y>0},{x,y}]
+     =  x > 0 && y > 0
+    
+    >> Reduce[a<b,a]
+     =  b ∈ Reals && a < b
+    """
+    
+    messages = {
+        "eqf": "`1` is not a well-formed equation.",
+        "svars": 'Equations may not give solutions for all "solve" variables.',
+    }
+
+    rules = {
+        "Reduce[eqs_, vars_, Complexes]": "Reduce[eqs, vars]",
+       
+    }
+
+    summary_text = "find generic solutions for variables, in logical expression form"
+    
+    
+    def apply(self, eqs, vars, evaluation):
+        "Reduce[eqs_, vars_]"
+        
+        vars_original = vars
+        head_name = vars.get_head_name()
+        if head_name == "System`List":
+            vars = vars.leaves
+        else:
+            vars = [vars]
+        for var in vars:
+            if (
+                (var.is_atom() and not var.is_symbol())
+                or head_name in ("System`Plus", "System`Times", "System`Power")  # noqa
+                or constant & var.get_attributes(evaluation.definitions)
+            ):
+
+                evaluation.message("Solve", "ivar", vars_original)
+                return
+        eqs_original = eqs
+        if eqs.get_head_name() in ("System`List", "System`And"):
+            eqs = eqs.leaves
+        else:
+            eqs = [eqs]
+        sympy_eqs = []
+        sympy_denoms = []
+        for eq in eqs:
+            if eq is SymbolTrue:
+                pass
+            elif eq is SymbolFalse:
+                return Expression(SymbolList)
+            elif not eq.has_form("Equal", 2):
+                if any(eq.has_form(operator,2) for operator in ["Less","LessEqual","Greater","GreaterEqual"]):
+                    eq=eq.to_sympy()
+                else:     
+                    return evaluation.message("Solve", "eqf", eqs_original)
+            else:
+                left, right = eq.leaves
+                left = left.to_sympy()
+                right = right.to_sympy()
+                if left is None or right is None:
+                    return
+                eq = left - right
+                eq = sympy.together(eq)
+                eq = sympy.cancel(eq)
+                numer, denom = eq.as_numer_denom()
+                sympy_denoms.append(denom)
+            sympy_eqs.append(eq)
+
+        vars_sympy = [var.to_sympy() for var in vars]
+        if None in vars_sympy:
+            return
+
+        # delete unused variables to avoid SymPy's
+        # PolynomialError: Not a zero-dimensional system
+        # in e.g. Solve[x^2==1&&z^2==-1,{x,y,z}]
+        all_vars = vars[:]
+        all_vars_sympy = vars_sympy[:]
+        vars = []
+        vars_sympy = []
+        for var, var_sympy in zip(all_vars, all_vars_sympy):
+            pattern = Pattern.create(var)
+            if not eqs_original.is_free(pattern, evaluation):
+                vars.append(var)
+                vars_sympy.append(var_sympy)
+
+        def transform_dict(sols):
+            if not sols:
+                yield sols
+            for var, sol in sols.items():
+                rest = sols.copy()
+                del rest[var]
+                rest = transform_dict(rest)
+                if not isinstance(sol, (tuple, list)):
+                    sol = [sol]
+                if not sol:
+                    for r in rest:
+                        yield r
+                else:
+                    for r in rest:
+                        for item in sol:
+                            new_sols = r.copy()
+                            new_sols[var] = item
+                            yield new_sols
+                break
+
+        def transform_solution(sol):
+            if not isinstance(sol, dict):
+                if not isinstance(sol, (list, tuple)):
+                    sol = [sol]
+                sol = dict(list(zip(vars_sympy, sol)))
+            return transform_dict(sol)
+
+        if not sympy_eqs:
+            sympy_eqs = True
+        elif len(sympy_eqs) == 1:
+            sympy_eqs = sympy_eqs[0]
+        try:
+            if isinstance(sympy_eqs, bool):
+                result = sympy_eqs
+            else:
+                result = sympy.solve(sympy_eqs, vars_sympy)
+            if not isinstance(result, list):
+                result = [result]
+            if isinstance(result, list) and len(result) == 1 and result[0] is True:
+                return Expression(SymbolList, Expression(SymbolList))
+            if result == [None]:
+                return Expression(SymbolList)
+            results = []
+            for sol in result:
+                results.extend(transform_solution(sol))
+            result = results
+            if any(
+                sol and any(var not in sol for var in all_vars_sympy) for sol in result
+            ):
+                evaluation.message("Solve", "svars")
+            
+            # Filter out results for which denominator is 0
+            # (SymPy should actually do that itself, but it doesn't!)
+            result = [
+                sol
+                for sol in result
+                if all(sympy.simplify(denom.subs(sol)) != 0 for denom in sympy_denoms)
+            ]
+            def get_result(var,sol,var_sympy):
+                if any( "System`"+operator in str(from_sympy(sol[var_sympy])) for operator in ["Less","LessEqual","Greater","GreaterEqual"]):
+                    return from_sympy(sol[var_sympy])
+                else:
+                    return Expression("Equal", var, from_sympy(sol[var_sympy]))
+            return Expression(
+                "Or",
+                *(
+                    Expression(
+                        "And",
+                        *(
+                            get_result(var,sol,var_sympy)
+                            
+                            for var, var_sympy in zip(vars, vars_sympy)
+                            if var_sympy in sol
+                        ),
+                    )
+                    for sol in result
+                ),
+            )
+        except sympy.PolynomialError:
+            # raised for e.g. Solve[x^2==1&&z^2==-1,{x,y,z}] when not deleting
+            # unused variables beforehand
+            pass
+        except NotImplementedError:
+            pass
+        except TypeError as exc:
+            if str(exc).startswith("expected Symbol, Function or Derivative"):
+                evaluation.message("Solve", "ivar", vars_original)
 
 class Solve(Builtin):
     """
@@ -929,7 +1116,8 @@ class Solve(Builtin):
     }
 
     summary_text = "find generic solutions for variables"
-
+    
+    
     def apply(self, eqs, vars, evaluation):
         "Solve[eqs_, vars_]"
         
@@ -961,10 +1149,7 @@ class Solve(Builtin):
             elif eq is SymbolFalse:
                 return Expression(SymbolList)
             elif not eq.has_form("Equal", 2):
-                if any(eq.has_form(operator,2) for operator in ["Less","LessEqual","Greater","GreaterEqual"]):
-                    eq=eq.to_sympy()
-                else:     
-                    return evaluation.message("Solve", "eqf", eqs_original)
+                return evaluation.message("Solve", "eqf", eqs_original)
             else:
                 left, right = eq.leaves
                 left = left.to_sympy()
@@ -976,7 +1161,7 @@ class Solve(Builtin):
                 eq = sympy.cancel(eq)
                 numer, denom = eq.as_numer_denom()
                 sympy_denoms.append(denom)
-            sympy_eqs.append(eq)
+                sympy_eqs.append(eq)
 
         vars_sympy = [var.to_sympy() for var in vars]
         if None in vars_sympy:
@@ -1078,7 +1263,6 @@ class Solve(Builtin):
         except TypeError as exc:
             if str(exc).startswith("expected Symbol, Function or Derivative"):
                 evaluation.message("Solve", "ivar", vars_original)
-
 
 class Integers(Builtin):
     """
--- a/mathics/builtin/numbers/calculus.py
+++ b/mathics/builtin/numbers/calculus.py
@@ -823,11 +823,11 @@ class Root(SympyFunction):
      = x == 0  ||   y == 0 
     >> Reduce[{x y == 0, y == 4},{x, y}]
      = y == 4 && x == 0
-    >> Reduce[{x y == 0, y > 4},{x, y}]
+    #>> Reduce[{x y == 0, y > 4},{x, y}]
      =  x == 0 && y > 4
-    >> Reduce[x*y>0,{x,y}]
+    #>> Reduce[x*y>0,{x,y}]
      = (x < 0 && y < 0) || (x > 0 && y > 0)
-    >> Reduce[{x*y > 0, y>0},{x,y}]
+    #>> Reduce[{x*y > 0, y>0},{x,y}]
      =  x > 0 && y > 0
     
     >> Reduce[a<b,a]
--- a/mathics/builtin/numbers/calculus.py
+++ b/mathics/builtin/numbers/calculus.py
@@ -933,7 +933,6 @@ class Solve(Builtin):
     def apply(self, eqs, vars, evaluation):
         "Solve[eqs_, vars_]"
         
-        InequalityOperators={"Less":"<","LessEqual":"<=","Greater":">","GreaterEqual":">="}
         vars_original = vars
         head_name = vars.get_head_name()
         if head_name == "System`List":
@@ -962,14 +961,9 @@ class Solve(Builtin):
             elif eq is SymbolFalse:
                 return Expression(SymbolList)
             elif not eq.has_form("Equal", 2):
-                for operator in InequalityOperators:
-                    if eq.has_form(operator,2):
-                        eq=eq.to_sympy()
-                        is_inequality=True
-                        break      
-                try:
-                    is_inequality
-                except NameError:      
+                if any(eq.has_form(operator,2) for operator in ["Less","LessEqual","Greater","GreaterEqual"]):
+                    eq=eq.to_sympy()
+                else:     
                     return evaluation.message("Solve", "eqf", eqs_original)
             else:
                 left, right = eq.leaves


mmatera:
Here https://github.com/Mathics3/mathics-core/tree/patch-1-mm there is a patch over this that solves most of the errors. However,
`Solve[a < b, a]` results in ` {{a -> a < b}}` which I think is not the expected behaviour.
In WMA we obtain:
```
In[1]:= Solve[a<b,a]                                                                                                                                                            

Solve::fdimc: When parameter values satisfy the condition b ∈ Reals, the solution set contains a full-dimensional component; use Reduce for complete solution information.

Out[1]= {}

```

mmatera:
Look also https://mathematica.stackexchange.com/questions/17127/what-is-the-difference-between-reduce-and-solve

DUOLabs333:
What I'm seeing is that Solve should only be for equations, and Reduce is for inequalities?

mmatera:
`Solve` produces replacement lists (like `{{x->2,y->3}}`) while `Reduce` produce logical expressions (like ' a>3 || a<-3').

DUOLabs333:
Given that `sympy.solve` can produce both, I'm not sure what's the best idea here.

mmatera:
> Unless you mean to just change the form of the output?

When you call `Solve`, you expect a certain kind of output.  If you call  `Resolve` or `Reduce` you expect others. Just copy the structure of the `Solve` class, and modify the name and the ``apply`` method, in a way that using sympy produce the right form of the corresponding output.

DUOLabs333:
Ok, got it.

DUOLabs333:
It actually would be best to just call Solve, and then modify the output. Is there a way to do that (something with Expression.evalute?)

mmatera:
> It actually would be best to just call Solve, and then modify the output. Is there a way to do that (something with Expression.evalute?)

No, because `Solve` does not return intervals or inequalities, just expressions that evaluate to numeric values. For example, what would you expect from  `Solve[{a<b},a]` ?

DUOLabs333:
Ok then -- like I said before, so Reduce should be for both equations and inequalities, while Solve is for equations only, right?

mmatera:
Yep

DUOLabs333:
Given that sympy.solve does most if it already, I just need to figure out how to deal with an equation output -- replacing commas with && and sets of answers with ||. I'm going to see if sympy has a function for that.

DUOLabs333:
Hmmm, I defined `class Reduce`, but I still can't use it in `mathicsscript`

DUOLabs333:
I basically copies and pasted the `Solve` class, but changed it to `class Reduce`:

<details>
<summary>`Reduce` code</summary>
class Reduce(Builtin):
    """
    <dl>
      <dt>'Solve[$equation$, $vars$]'
      <dd>attempts to solve $equation$ for the variables $vars$.

      <dt>'Solve[$equation$, $vars$, $domain$]'
      <dd>restricts variables to $domain$, which can be 'Complexes' or 'Reals' or 'Integers'.
    </dl>

    >> Solve[x ^ 2 - 3 x == 4, x]
     = {{x -> -1}, {x -> 4}}
    >> Solve[4 y - 8 == 0, y]
     = {{y -> 2}}

    Apply the solution:
    >> sol = Solve[2 x^2 - 10 x - 12 == 0, x]
     = {{x -> -1}, {x -> 6}}
    >> x /. sol
     = {-1, 6}

    Contradiction:
    >> Solve[x + 1 == x, x]
     = {}
    Tautology:
    >> Solve[x ^ 2 == x ^ 2, x]
     = {{}}

    Rational equations:
    >> Solve[x / (x ^ 2 + 1) == 1, x]
     = {{x -> 1 / 2 - I / 2 Sqrt[3]}, {x -> 1 / 2 + I / 2 Sqrt[3]}}
    >> Solve[(x^2 + 3 x + 2)/(4 x - 2) == 0, x]
     = {{x -> -2}, {x -> -1}}

    Transcendental equations:
    >> Solve[Cos[x] == 0, x]
     = {{x -> Pi / 2}, {x -> 3 Pi / 2}}

    Solve can only solve equations with respect to symbols or functions:
    >> Solve[f[x + y] == 3, f[x + y]]
     = {{f[x + y] -> 3}}
    >> Solve[a + b == 2, a + b]
     : a + b is not a valid variable.
     = Solve[a + b == 2, a + b]
    This happens when solving with respect to an assigned symbol:
    >> x = 3;
    >> Solve[x == 2, x]
     : 3 is not a valid variable.
     = Solve[False, 3]
    >> Clear[x]
    >> Solve[a < b, a]
     : a < b is not a well-formed equation.
     = Solve[a < b, a]

    Solve a system of equations:
    >> eqs = {3 x ^ 2 - 3 y == 0, 3 y ^ 2 - 3 x == 0};
    >> sol = Solve[eqs, {x, y}] // Simplify
     = {{x -> 0, y -> 0}, {x -> 1, y -> 1}, {x -> -1 / 2 + I / 2 Sqrt[3], y -> -1 / 2 - I / 2 Sqrt[3]}, {x -> (1 - I Sqrt[3]) ^ 2 / 4, y -> -1 / 2 + I / 2 Sqrt[3]}}
    >> eqs /. sol // Simplify
     = {{True, True}, {True, True}, {True, True}, {True, True}}

    An underdetermined system:
    >> Solve[x^2 == 1 && z^2 == -1, {x, y, z}]
     : Equations may not give solutions for all "solve" variables.
     = {{x -> -1, z -> -I}, {x -> -1, z -> I}, {x -> 1, z -> -I}, {x -> 1, z -> I}}

    Domain specification:
    >> Solve[x^2 == -1, x, Reals]
     = {}
    >> Solve[x^2 == 1, x, Reals]
     = {{x -> -1}, {x -> 1}}
    >> Solve[x^2 == -1, x, Complexes]
     = {{x -> -I}, {x -> I}}
    >> Solve[4 - 4 * x^2 - x^4 + x^6 == 0, x, Integers]
     = {{x -> -1}, {x -> 1}}

    #> Solve[x^2 +1 == 0, x] // FullForm
     = {{Rule[x, Complex[0, -1]]},{Rule[x, Complex[0, 1]]}}

    #> Solve[x^5==x,x]
     = {{x -> -1}, {x -> 0}, {x -> 1}, {x -> -I}, {x -> I}}

    #> Solve[g[x] == 0, x]
     = Solve[g[x] == 0, x]
    ## (should use inverse functions, actually!)
    #> Solve[g[x] + h[x] == 0, x]
     = Solve[g[x] + h[x] == 0, x]

    #> Solve[Sin(x) == 1, x]
     = {{x -> 1 / Sin}}

    #> Solve[E == 1, E]
     : E is not a valid variable.
     = Solve[False, E]
    #> Solve[False, Pi]
     : Pi is not a valid variable.
     = Solve[False, Pi]

    """

    messages = {
        "eqf": "`1` is not a well-formed equation.",
        "svars": 'Equations may not give solutions for all "solve" variables.',
    }

    rules = {
        "Solve[eqs_, vars_, Complexes]": "Solve[eqs, vars]",
        "Solve[eqs_, vars_, Reals]": (
            "Cases[Solve[eqs, vars], {Rule[x_,y_?RealNumberQ]}]"
        ),
        "Solve[eqs_, vars_, Integers]": (
            "Cases[Solve[eqs, vars], {Rule[x_,y_Integer]}]"
        ),
    }

    summary_text = "find generic solutions for variables"
    
    
    def apply(self, eqs, vars, evaluation):
        "Solve[eqs_, vars_]"
        
        vars_original = vars
        head_name = vars.get_head_name()
        if head_name == "System`List":
            vars = vars.leaves
        else:
            vars = [vars]
        for var in vars:
            if (
                (var.is_atom() and not var.is_symbol())
                or head_name in ("System`Plus", "System`Times", "System`Power")  # noqa
                or constant & var.get_attributes(evaluation.definitions)
            ):

                evaluation.message("Solve", "ivar", vars_original)
                return
        eqs_original = eqs
        if eqs.get_head_name() in ("System`List", "System`And"):
            eqs = eqs.leaves
        else:
            eqs = [eqs]
        sympy_eqs = []
        sympy_denoms = []
        for eq in eqs:
            if eq is SymbolTrue:
                pass
            elif eq is SymbolFalse:
                return Expression(SymbolList)
            elif not eq.has_form("Equal", 2):
                if any(eq.has_form(operator,2) for operator in ["Less","LessEqual","Greater","GreaterEqual"]):
                    eq=eq.to_sympy()
                else:     
                    return evaluation.message("Solve", "eqf", eqs_original)
            else:
                left, right = eq.leaves
                left = left.to_sympy()
                right = right.to_sympy()
                if left is None or right is None:
                    return
                eq = left - right
                eq = sympy.together(eq)
                eq = sympy.cancel(eq)
                numer, denom = eq.as_numer_denom()
                sympy_denoms.append(denom)
            sympy_eqs.append(eq)

        vars_sympy = [var.to_sympy() for var in vars]
        if None in vars_sympy:
            return

        # delete unused variables to avoid SymPy's
        # PolynomialError: Not a zero-dimensional system
        # in e.g. Solve[x^2==1&&z^2==-1,{x,y,z}]
        all_vars = vars[:]
        all_vars_sympy = vars_sympy[:]
        vars = []
        vars_sympy = []
        for var, var_sympy in zip(all_vars, all_vars_sympy):
            pattern = Pattern.create(var)
            if not eqs_original.is_free(pattern, evaluation):
                vars.append(var)
                vars_sympy.append(var_sympy)

        def transform_dict(sols):
            if not sols:
                yield sols
            for var, sol in sols.items():
                rest = sols.copy()
                del rest[var]
                rest = transform_dict(rest)
                if not isinstance(sol, (tuple, list)):
                    sol = [sol]
                if not sol:
                    for r in rest:
                        yield r
                else:
                    for r in rest:
                        for item in sol:
                            new_sols = r.copy()
                            new_sols[var] = item
                            yield new_sols
                break

        def transform_solution(sol):
            if not isinstance(sol, dict):
                if not isinstance(sol, (list, tuple)):
                    sol = [sol]
                sol = dict(list(zip(vars_sympy, sol)))
            return transform_dict(sol)

        if not sympy_eqs:
            sympy_eqs = True
        elif len(sympy_eqs) == 1:
            sympy_eqs = sympy_eqs[0]

        try:
            if isinstance(sympy_eqs, bool):
                result = sympy_eqs
            else:
                result = sympy.solve(sympy_eqs, vars_sympy)
            if not isinstance(result, list):
                result = [result]
            if isinstance(result, list) and len(result) == 1 and result[0] is True:
                return Expression(SymbolList, Expression(SymbolList))
            if result == [None]:
                return Expression(SymbolList)
            results = []
            for sol in result:
                results.extend(transform_solution(sol))
            result = results
            if any(
                sol and any(var not in sol for var in all_vars_sympy) for sol in result
            ):
                evaluation.message("Solve", "svars")

            # Filter out results for which denominator is 0
            # (SymPy should actually do that itself, but it doesn't!)
            result = [
                sol
                for sol in result
                if all(sympy.simplify(denom.subs(sol)) != 0 for denom in sympy_denoms)
            ]

            return Expression(
                "List",
                *(
                    Expression(
                        "List",
                        *(
                            Expression(SymbolRule, var, from_sympy(sol[var_sympy]))
                            for var, var_sympy in zip(vars, vars_sympy)
                            if var_sympy in sol
                        ),
                    )
                    for sol in result
                ),
            )
        except sympy.PolynomialError:
            # raised for e.g. Solve[x^2==1&&z^2==-1,{x,y,z}] when not deleting
            # unused variables beforehand
            pass
        except NotImplementedError:
            pass
        except TypeError as exc:
            if str(exc).startswith("expected Symbol, Function or Derivative"):
                evaluation.message("Solve", "ivar", vars_original)
</details>

TiagoCavalcante:
@DUOLabs333 I believe you aren't re-installing mathics-core after making your changes.

That can be done with `make install`.
(it takes a bit of time).

DUOLabs333:
I'm editing the files directly in my `__pypackages__`

DUOLabs333:
You needed `"Reduce[eqs_, vars_]"` in `apply`

TiagoCavalcante:
> You needed `"Reduce[eqs_, vars_]"` in `apply`

Yes, and in the rules too.

DUOLabs333:
Should `Reduce[a<b,a]` be `{{a -> a < b}}` or `{{a < b}}`?

TiagoCavalcante:
@DUOLabs333 

Symja returns `a<b`, while WolframCloud returns `b ∈ ℝ && a<b`. My take would be to follow Symja in this, but any of these results are good.

DUOLabs333:
How would remove the variable assignments from the result. `sympy.solve` doesn't have them.

mmatera:
> Should `Reduce[a<b,a]` be `{{a -> a < b}}` or `{{a < b}}`?
This is the output in WL:

```
In[1]:= Reduce[a<b,a]                                                           

Out[1]= b ∈ Reals && a < b
```

mmatera:
For ``Solve`` one possibility (but is not what WMA does) would be
``Solve[a<b,a]->Interval[{-Infinity,b}]``

DUOLabs333:
Is there something similar to SymbolRule that uses equal signs instead? Or is SymbolRule OK for `Reduce` for Equations?

mmatera:
> Is there something similar to SymbolRule that uses equal signs instead? Or is SymbolRule OK for `Reduce` for Equations?

You can use  ``Expression("Equal", lhs, rhs)``

mmatera:
> Equal simplifies `x->x` to `True`. Is there a way to not simplify it?

Do you want to keep `x==x` ? You can put it like ``Expression("HoldForm", Expression("Equal",x,x))``. However, in which context do you want that?

DUOLabs333:
Well, for example, `Reduce[x*y==0,{x,y}]` should return `{{x==x && y==0} || {x==0 && y==y}`

mmatera:
> Well, for example, `Reduce[x*y==0,{x,y}]` should return `{{x==x && y==0} || {x==0 && y==y}`



> 

```
In[1]:= Reduce[x*y==0,{x,y}]                                                    

Out[1]= x == 0 || y == 0
```

mmatera:
> > Well, for example, `Reduce[x*y==0,{x,y}]` should return `{{x==x && y==0} || {x==0 && y==y}`
> 
> > 
> 
> In WMA,
> 
> ```
> In[1]:= Reduce[x*y==0,{x,y}]                                                    
> 
> Out[1]= x == 0 || y == 0
> ```
> 
> So it is OK that `x==x && y==0` evaluates to `True && y==0` and afterwards to `y==0`

what are wrong are the curly brackets.

DUOLabs333:
Is there a way of removing `True`s from `Expression?

mmatera:
> Is there a way of removing `True`s from `Expression?

You do not need it. The evaluator is going to handle them.

DUOLabs333:
If I do `Expression("Equal", var, from_sympy(sol[var_sympy]))`, I get `{{x == 0, True}, {True, y == 0}}`.

TiagoCavalcante:
@DUOLabs333 you can flatten the list:
```py
expr = Expression("Flatten", expr).evaluate(evaluation)
```
Then filter the `True`s:
```py
expr.elements = [element for element in expr.elements if not element.sameQ(SymbolTrue)]
```
And then apply the `Or` function:
```py
expr = Expression("Apply", "Or", expr)
```

*warning*: the code above is pseudo-code.

mmatera:
> If I do `Expression("Equal", var, from_sympy(sol[var_sympy]))`, I get `{{x == 0, True}, {True, y == 0}}`.
What you need to do now is to convert that expression in something like
```Or[And[x==0, True],And[True, y==0]]```

A way to do that would be

```
#This produces your output and store it in result1
result1 = Expression("Equal", var, from_sympy(sol[var_sympy])).evaluate(evaluation)
result = Expression("Or",*[Expression("And", *branch.elements)    for branch in result1.elements]).evaluate(evaluation)
```

**Comment by DUOLabs333** [comment]
> #This produces your output and store it in result1 
Don't you mean it would store it in `result`?

**Comment by mmatera** [comment]
> > #This produces your output and store it in result1
> 
> Don't you mean it would store it in `result`?

The first line produces the output that you reported. The second one produces the output that you are looking for

**Comment by DUOLabs333** [comment]
How would I put that on one line: `Expression("Or",*[Expression("And", *branch.elements)    for branch in (Expression("Equal", var, from_sympy(sol[var_sympy])).evaluate(evaluation)).elements])` fails with `'Expression' object has no attribute 'elements'`

**Comment by mmatera** [comment]
> How would I put that on one line: `Expression("Or",*[Expression("And", *branch.elements) for branch in (Expression("Equal", var, from_sympy(sol[var_sympy])).evaluate(evaluation)).elements])` fails with `'Expression' object has no attribute 'elements'`

Sorry, we are making changes over the interface of the core. Use `leaves` or `_elements` instead `elements`.

**Comment by DUOLabs333** [comment]
Now I get `'Symbol' object has no attribute 'elements'`

**Comment by mmatera** [comment]
> Now I get `'Symbol' object has no attribute 'elements'`

Can you upload the branch you are working on to your github repo in a way I can test and analyze your code?

**Comment by DUOLabs333** [comment]
Using `Expression("Or",*[Expression("And", *branch.get_elements())    for branch in (Expression("Equal", var, from_sympy(sol[var_sympy])).evaluate(evaluation)).get_elements()])` gives me `{{True, False}, {False, True}}`

**Comment by mmatera** [comment]
> Oh, I'm working on it directly, then just paste the code to GitHub when I'm done.

Ok, so paste here the code of the class to try it locally and see what is going wrong...

**Comment by DUOLabs333** [comment]
At this point, I have:
<details>
<summary>`Reduce` code</summary>
class Reduce(Builtin):
    
    messages = {
        "eqf": "`1` is not a well-formed equation.",
        "svars": 'Equations may not give solutions for all "solve" variables.',
    }

    rules = {
        "Reduce[eqs_, vars_, Complexes]": "Reduce[eqs, vars]",
        "Reduce[eqs_, vars_, Reals]": (
            "Cases[Reduce[eqs, vars], {Rule[x_,y_?RealNumberQ]}]"
        ),
        "Reduce[eqs_, vars_, Integers]": (
            "Cases[Reduce[eqs, vars], {Rule[x_,y_Integer]}]"
        ),
    }

    summary_text = "find generic solutions for variables"
    
    
    def apply(self, eqs, vars, evaluation):
        "Reduce[eqs_, vars_]"
        
        vars_original = vars
        head_name = vars.get_head_name()
        if head_name == "System`List":
            vars = vars.leaves
        else:
            vars = [vars]
        for var in vars:
            if (
                (var.is_atom() and not var.is_symbol())
                or head_name in ("System`Plus", "System`Times", "System`Power")  # noqa
                or constant & var.get_attributes(evaluation.definitions)
            ):

                evaluation.message("Solve", "ivar", vars_original)
                return
        eqs_original = eqs
        if eqs.get_head_name() in ("System`List", "System`And"):
            eqs = eqs.leaves
        else:
            eqs = [eqs]
        sympy_eqs = []
        sympy_denoms = []
        for eq in eqs:
            if eq is SymbolTrue:
                pass
            elif eq is SymbolFalse:
                return Expression(SymbolList)
            elif not eq.has_form("Equal", 2):
                if any(eq.has_form(operator,2) for operator in ["Less","LessEqual","Greater","GreaterEqual"]):
                    eq=eq.to_sympy()
                else:     
                    return evaluation.message("Solve", "eqf", eqs_original)
            else:
                left, right = eq.leaves
                left = left.to_sympy()
                right = right.to_sympy()
                if left is None or right is None:
                    return
                eq = left - right
                eq = sympy.together(eq)
                eq = sympy.cancel(eq)
                numer, denom = eq.as_numer_denom()
                sympy_denoms.append(denom)
            sympy_eqs.append(eq)

        vars_sympy = [var.to_sympy() for var in vars]
        if None in vars_sympy:
            return

        # delete unused variables to avoid SymPy's
        # PolynomialError: Not a zero-dimensional system
        # in e.g. Solve[x^2==1&&z^2==-1,{x,y,z}]
        all_vars = vars[:]
        all_vars_sympy = vars_sympy[:]
        vars = []
        vars_sympy = []
        for var, var_sympy in zip(all_vars, all_vars_sympy):
            pattern = Pattern.create(var)
            if not eqs_original.is_free(pattern, evaluation):
                vars.append(var)
                vars_sympy.append(var_sympy)

        def transform_dict(sols):
            if not sols:
                yield sols
            for var, sol in sols.items():
                rest = sols.copy()
                del rest[var]
                rest = transform_dict(rest)
                if not isinstance(sol, (tuple, list)):
                    sol = [sol]
                if not sol:
                    for r in rest:
                        yield r
                else:
                    for r in rest:
                        for item in sol:
                            new_sols = r.copy()
                            new_sols[var] = item
                            yield new_sols
                break

        def transform_solution(sol):
            if not isinstance(sol, dict):
                if not isinstance(sol, (list, tuple)):
                    sol = [sol]
                sol = dict(list(zip(vars_sympy, sol)))
            return transform_dict(sol)

        if not sympy_eqs:
            sympy_eqs = True
        elif len(sympy_eqs) == 1:
            sympy_eqs = sympy_eqs[0]

        try:
            if isinstance(sympy_eqs, bool):
                result = sympy_eqs
            else:
                result = sympy.solve(sympy_eqs, vars_sympy)
            if not isinstance(result, list):
                result = [result]
            if isinstance(result, list) and len(result) == 1 and result[0] is True:
                return Expression(SymbolList, Expression(SymbolList))
            if result == [None]:
                return Expression(SymbolList)
            results = []
            for sol in result:
                results.extend(transform_solution(sol))
            result = results
            if any(
                sol and any(var not in sol for var in all_vars_sympy) for sol in result
            ):
                evaluation.message("Solve", "svars")

            # Filter out results for which denominator is 0
            # (SymPy should actually do that itself, but it doesn't!)
            result = [
                sol
                for sol in result
                if all(sympy.simplify(denom.subs(sol)) != 0 for denom in sympy_denoms)
            ]

            return Expression(
                "List",
                *(
                    Expression(
                        "List",
                        *(
                            Expression("Or",*[Expression("And", *branch.get_elements())    for branch in (Expression("Equal", var, from_sympy(sol[var_sympy])).evaluate(evaluation)).get_elements()])
                            
                            for var, var_sympy in zip(vars, vars_sympy)
                            if var_sympy in sol
                        ),
                    )
                    for sol in result
                ),
            )
        except sympy.PolynomialError:
            # raised for e.g. Solve[x^2==1&&z^2==-1,{x,y,z}] when not deleting
            # unused variables beforehand
            pass
        except NotImplementedError:
            pass
        except TypeError as exc:
            if str(exc).startswith("expected Symbol, Function or Derivative"):
                evaluation.message("Solve", "ivar", vars_original)
</details>

**Comment by mmatera** [comment]
OK, lets start by adding a doctring to the class, and some tests:

    """
    <dl>
       <dt>'Reduce[$eqs$, $vars$]'
       <dd> Reduces the statements in $eqs$ by solving equations in $vars$ and eliminating quantifiers.
    </dl>
    >> Reduce[x y == 0,{x, y}]
     = x == 0  ||   y == 0 
    >> Reduce[{x y == 0, y == 4},{x, y}]
     = y == 4 && x == 0
    >> Reduce[{x y == 0, y > 4},{x, y}]
     =  x == 0 && y > 4
    >> Reduce[x*y>0,{x,y}]
     = (x < 0 && y < 0) || (x > 0 && y > 0)
    >> Reduce[{x*y > 0, y>0},{x,y}]
     =  x > 0 && y > 0

    >> Reduce[a<b,a]
     =  b ∈ Reals && a < b
    """

**Comment by mmatera** [comment]
Then,  in the ``rules`` property, the first rule is OK, but the other two no:
```Reduce[eqs_, vars_, Reals]": ("Cases[Reduce[eqs, vars], {Rule[x_,y_?RealNumberQ]}]"),```
``Cases`` requires a ``pattern`` as a second rule, not a ``Rule``. Lets deal with the domain afterwards.

**Comment by mmatera** [comment]
I think ``transform_dict`` and ``transform_solution`` are not useful here.

**Comment by DUOLabs333** [comment]
A problem: `Reduce[{x y == 0},{x, y}]` gives `{{x == 0, True}, {True, y == 0}}` -- thjat is `x==0 || y==0`. However, `Reduce[24-2x-x^2==0,x]` gives `{{x == -6}, {x == 4}}` -- that is, `x==6 && x==-4`. How do I distinguish an AND or OR?

**Comment by TiagoCavalcante** [comment]
@DUOLabs333 you could check if both elements of the list are the same, if they are, you use an OR (as a number can't be both 1 and 2 at the same time), otherwise use AND (as we are dealing with different variables).

**Comment by DUOLabs333** [comment]
I'm thinking about just completely writing the class from scratch -- Do I need to have `apply`, `transform_dict` and `transform_solution`?

**Comment by TiagoCavalcante** [comment]
@DUOLabs333 every (with exceptions...) must have an apply method, the other functions are optional.

**Comment by DUOLabs333** [comment]
> @DUOLabs333 you could check if both elements of the list are the same, if they are, you use an OR (as a number can't be both 1 and 2 at the same time), otherwise use AND (as we are dealing with different variables).

I was able to kill two birds with one stone -- Switching `List` for `Or` and `And` dealt with the formatting issue and with the operator issue. Now, I just have to run through the tests...

**Comment by DUOLabs333** [comment]
However, `Reduce[{x y == 0, y==0},{x, y}]` gives `x == 0`, probably because it assigned `y=0`, so `Equal` would make that True`.

**Comment by TiagoCavalcante** [comment]
@DUOLabs333 for "un-assign" a variable you can run `ClearAll[variableName]`.

**Comment by DUOLabs333** [comment]
How would I run that in Mathics -- `ClearAll(y)`?

**Comment by DUOLabs333** [comment]
Ok, this is weird -- `Reduce[{x y == 0, x==0},{x, y}]` gives `x == 0`, but `Reduce[{x y == 0, x==0},{y,x}]` gives `y==0`.

**Comment by TiagoCavalcante** [comment]
@DUOLabs333 you shouldn't clear the variables from a built-in, that is a side effect.

> This is weird -- Reduce[{x y == 0, x==0},{x, y}] gives x == 0, but Reduce[{x y == 0, x==0},{y,x}] gives y==0. So, it seems that orders messes with the evaluation.

Yes, that is weird.

**Comment by DUOLabs333** [comment]
Changing the order of the equations themselves also switches the evaluation.

**Comment by DUOLabs333** [comment]
Ok, I understand now -- it's because even though `x` should be treated as `x=0`, it isn't by `solve`.

**Comment by DUOLabs333** [comment]
Why does `Reduce[x*y>0,{x,y}]` return `NotImplementedError: 
inequality has more than one symbol of interest.`

**Comment by DUOLabs333** [comment]
Sympy can not currently deal with multi-variate inequalities in `solve`.

**Comment by mmatera** [comment]
OK, in any case, to handle equalities and inequalities for a single variable is a progress!

**Comment by DUOLabs333** [comment]
Yes, I'm just going to finish that up, and push it out.

**Comment by DUOLabs333** [comment]
How do you check if "System`Less" is in an Expression?

**Comment by DUOLabs333** [comment]
Ok, pushed my changes.

File: mathics/builtin/numbers/calculus.py [hunk 1]
```diff
@@ -812,6 +812,193 @@ def to_sympy(self, expr, **kwargs):
...
+       
+    }
+
+    summary_text = "find generic solutions for variables, in logical expression form"
+    
+    
+    def apply(self, eqs, vars, evaluation):
+        "Reduce[eqs_, vars_]"
+        
+        vars_original = vars
```
mmatera, line 853: Maybe it would be useful to add comments on what each line does.

mathics/builtin/numbers/calculus.py
```diff
@@ -812,6 +812,193 @@ def to_sympy(self, expr, **kwargs):
...
+        head_name = vars.get_head_name()
```
mmatera, line 854: The simplest thing to do here would be to check if ``vars`` is an Atom, and if it is not an atom, that all its elements be Atoms.

DUOLabs333: Why would that be needed?

DUOLabs333:
> Sympy can not currently deal with multi-variate inequalities in solve. 

[Here](https://github.com/sympy/sympy/issues/9479#issuecomment-109721680) the Sympy mention that the functionality will be added to solveset. The problem is that solveset has an output form completely different to sympy. I don't know if they are going to add a "Legacy" compatibility layer on top, though.

mathics/builtin/numbers/calculus.py
mmatera: First, to raise an error if the parameter does not fit that form. Also, because it simplifies the code that follows this line.

mmatera:
> > Sympy can not currently deal with multi-variate inequalities in solve.
> 
> [Here](https://github.com/sympy/sympy/issues/9479#issuecomment-109721680) the Sympy mention that the functionality will be added to solveset. The problem is that solveset has an output form completely different to solve. I don't know if they are going to add a "Legacy" compatibility layer on top, though.

I think that the best would be to split the code in a way that the specific function used in sympy be encapsulated. Then, if in the future we want to use the new interface, we just need to change this part of the code. 
Again, I think it is important to comment each code block in the ` `apply`` method.

mathics/builtin/numbers/calculus.py
```diff
@@ -812,6 +812,193 @@ def to_sympy(self, expr, **kwargs):
...
+    >> Reduce[{x y == 0, y == 4},{x, y}]
+     = y == 4 && x == 0
+    >> Reduce[{x y == 0, y > 4},{x, y}]
+     =  x == 0 && y > 4
+    >> Reduce[x*y>0,{x,y}]
+     = (x < 0 && y < 0) || (x > 0 && y > 0)
+    >> Reduce[{x*y > 0, y>0},{x,y}]
+     =  x > 0 && y > 0
+    
+    >> Reduce[a<b,a]
```
mmatera, line 833: You can comment out the tests that do not pass by adding a ``# `` behind  ``>>``  and ``=``.

mathics/builtin/numbers/calculus.py
```diff
@@ -812,6 +812,193 @@ def to_sympy(self, expr, **kwargs):
...
+        "Reduce[eqs_, vars_]"
+        
+        vars_original = vars
+        head_name = vars.get_head_name()
+        if head_name == "System`List":
+            vars = vars.leaves
+        else:
+            vars = [vars]
+        for var in vars:
+            if (
```
mmatera, line 860: In particular, the code that follows that line could be eliminated.

mathics/builtin/numbers/calculus.py
```diff
@@ -812,6 +812,193 @@ def to_sympy(self, expr, **kwargs):
...
+                evaluation.message("Solve", "ivar", vars_original)
+                return
+        eqs_original = eqs
+        if eqs.get_head_name() in ("System`List", "System`And"):
+            eqs = eqs.leaves
+        else:
+            eqs = [eqs]
+        sympy_eqs = []
+        sympy_denoms = []
+        for eq in eqs:
```
mmatera, line 875: Notice that ``eq`` is a Mathics  ``Expression``, but in line 882 it becomes into a sympy ``Expression``.

mathics/builtin/numbers/calculus.py
```diff
@@ -812,6 +812,193 @@ def to_sympy(self, expr, **kwargs):
...
+            if eq is SymbolTrue:
+                pass
```
mmatera, line 877: ``continue`` instead ``pass``

mathics/builtin/numbers/calculus.py
```diff
@@ -812,6 +812,193 @@ def to_sympy(self, expr, **kwargs):
...
+            elif eq is SymbolFalse:
+                return Expression(SymbolList)
+            elif not eq.has_form("Equal", 2):
+                if any(eq.has_form(operator,2) for operator in ["Less","LessEqual","Greater","GreaterEqual"]):
+                    eq=eq.to_sympy()
```
mmatera, line 882: ``eq_sympy = eq.to_sympy()``

mathics/builtin/numbers/calculus.py
```diff
@@ -812,6 +812,193 @@ def to_sympy(self, expr, **kwargs):
...
+                left = left.to_sympy()
+                right = right.to_sympy()
+                if left is None or right is None:
+                    return
+                eq = left - right
+                eq = sympy.together(eq)
+                eq = sympy.cancel(eq)
+                numer, denom = eq.as_numer_denom()
+                sympy_denoms.append(denom)
+            sympy_eqs.append(eq)
```
mmatera, line 896: ``sympy_eqs.append(eq_sympy)``. This would prevent to add by accident Mathics ``Expression``(s) to this list.

mathics/builtin/numbers/calculus.py
```diff
@@ -812,6 +812,193 @@ def to_sympy(self, expr, **kwargs):
...
+                    eq=eq.to_sympy()
+                else:     
+                    return evaluation.message("Solve", "eqf", eqs_original)
+            else:
+                left, right = eq.leaves
+                left = left.to_sympy()
+                right = right.to_sympy()
+                if left is None or right is None:
+                    return
+                eq = left - right
```
mmatera, line 891: ``eq_sympy = left-rignt``

mathics/builtin/numbers/calculus.py
```diff
@@ -812,6 +812,193 @@ def to_sympy(self, expr, **kwargs):
...
+            ):
+
+                evaluation.message("Solve", "ivar", vars_original)
+                return
+        eqs_original = eqs
+        if eqs.get_head_name() in ("System`List", "System`And"):
+            eqs = eqs.leaves
+        else:
+            eqs = [eqs]
+        sympy_eqs = []
```
mmatera, line 873: Here it would be nice to add a comment like:
```
# This chunk of code converts the equations to a list of sympy expressions.
```

mathics/builtin/numbers/calculus.py
```diff
@@ -812,6 +812,193 @@ def to_sympy(self, expr, **kwargs):
...
+            return
+
+        # delete unused variables to avoid SymPy's
+        # PolynomialError: Not a zero-dimensional system
+        # in e.g. Solve[x^2==1&&z^2==-1,{x,y,z}]
+        all_vars = vars[:]
+        all_vars_sympy = vars_sympy[:]
+        vars = []
+        vars_sympy = []
+        for var, var_sympy in zip(all_vars, all_vars_sympy):
```
mmatera, line 909: It would be nice to understand and explain in a comment what we are doing in this chunk of code.

mathics/builtin/numbers/calculus.py
```diff
@@ -812,6 +812,193 @@ def to_sympy(self, expr, **kwargs):
...
+            pattern = Pattern.create(var)
+            if not eqs_original.is_free(pattern, evaluation):
+                vars.append(var)
+                vars_sympy.append(var_sympy)
+
+        def transform_dict(sols):
```
mmatera, line 915: This internal function makes sense in  ``Solve`` but not here: we need a different function that brings the output from sympy to the output that we want (a logical expression)

mathics/builtin/numbers/calculus.py
```diff
@@ -812,6 +812,193 @@ def to_sympy(self, expr, **kwargs):
...
+                        yield r
+                else:
+                    for r in rest:
+                        for item in sol:
+                            new_sols = r.copy()
+                            new_sols[var] = item
+                            yield new_sols
+                break
+
+        def transform_solution(sol):
```
mmatera, line 935: same here.

mathics/builtin/numbers/calculus.py
```diff
@@ -812,6 +812,193 @@ def to_sympy(self, expr, **kwargs):
...
+                    sol = [sol]
+                sol = dict(list(zip(vars_sympy, sol)))
+            return transform_dict(sol)
+
+        if not sympy_eqs:
+            sympy_eqs = True
+        elif len(sympy_eqs) == 1:
+            sympy_eqs = sympy_eqs[0]
+        try:
+            if isinstance(sympy_eqs, bool):
```
mmatera, line 947: rework after fixing the inline functions

mathics/builtin/numbers/calculus.py
DUOLabs333: Done.

mathics/builtin/numbers/calculus.py
DUOLabs333: Why?

mathics/builtin/numbers/calculus.py
DUOLabs333: So? This is exactly what is supposed to happen.

mathics/builtin/numbers/calculus.py
DUOLabs333: That's not my code.

mathics/builtin/numbers/calculus.py
DUOLabs333: This is not done in Solve.

mathics/builtin/numbers/calculus.py
DUOLabs333: This is not from me.

mathics/builtin/numbers/calculus.py
mmatera: Yes, but for clarity, it would be useful the change of name.

mathics/builtin/numbers/calculus.py
DUOLabs333: It's fine -- you can just convert it in the output.

mathics/builtin/numbers/calculus.py
DUOLabs333: For me, the more variables you add, the more confusing it gets. This comes from Solve, though.

mathics/builtin/numbers/calculus.py
mmatera: I mean, if you handle the ``vars`` before, you do not need to do this sanity check...

mathics/builtin/numbers/calculus.py
mmatera: In the first ``if``, there was a ``pass`` instead a ``continue`` instruction. That produced that a (mathics) expression get into the list of simpy expressions, which at the end raised a weir error. If we use another name for the variable, it would be simpler to track that what really happened was that the variable was never set.

mathics/builtin/numbers/calculus.py
mmatera: Yes, is code coming from  ``Solve``.  The idea is to get a clear code for ``Reduce``.

DUOLabs333: Shouldn't it get merged first, then later do a clean up of both classes?

mathics/builtin/numbers/calculus.py
DUOLabs333: Is it, though?

mathics/builtin/numbers/calculus.py
mmatera: again, this is something that I discovered because you started working on this. It was wrong also in ``Solve``. The idea is not to criticize you, but the code, independently of who wrote it.

mathics/builtin/numbers/calculus.py
mmatera: If you want, we can do that. However, I think the best is to merge the cleaner possible version.

mathics/builtin/numbers/calculus.py
DUOLabs333: Done.

mathics/builtin/numbers/calculus.py
DUOLabs333: Done.

mathics/builtin/numbers/calculus.py
mmatera: Yes, but it is a double conversion and does not help to follow the code or to make the code maintainable.

mathics/builtin/numbers/calculus.py
DUOLabs333: Done.

DUOLabs333:
Note for the future: If/when `solve` is replaced with `solveset`, you can do `as_relational`, so the work is done for you. For example -- `solveset(x**2-6*x+8,x).as_relational(x)` gives you `x==2 | x==4`. However, both `solveset` and `as_relational` only accept one variable. This will probably change in the future.

DUOLabs333:
`diofant`, a sympy fork can do it, but it is not drop in for `calculus`, it probably has enough incompatibilities that it may require a change across the entire codebase.

DUOLabs333:
I rebased, and now I get `AttributeError: 'Symbol' object has no attribute 'is_atom'`

rocky:
> 'Symbol' object has no attribute 'is_atom'

`x.is_atom() `was replaced with `isinstance(x, Atom)` some time ago. 

Basic bug reporting 101. Novices will report some error message out of context and expect someone else will fill in the rest. 

If you want people like me to pay more attention, given not just the error message but the input and surrounding output that gave the error report. From 

https://www.osadl.org/Eric-Raymond-How-to-ask-questions-the-s.questions-the-smart-way.0.html#beprecise : 

The clues most useful in figuring out something that went wrong often lie in the events immediately prior. So, your account should describe precisely what you did, and what the machine and software did, leading up to the blowup. In the case of command-line processes, having a session log (e.g., using the script utility) and quoting the relevant twenty or so lines is very useful.

PR #173, opened 2022-02-24
Expression refactor

rocky:


mathics/core/symbols.py
```diff
@@ -128,202 +128,34 @@ def __ne__(self, other) -> bool:
...
 
-    options: Any
-    pattern_sequence: bool
-    unformatted: Any
     last_evaluated: Any
     # this variable holds a function defined in mathics.core.expression that creates an expression
     create_expression: Any
 
+    # __new__ seems to be used because this object references itself.
+    # In particular:
```
mmatera, line 146: @rocky, Notice that some descendants of BaseExpression decide the class of the object according to the input. For example, ``Real`` receives a float or a string and a precision parameter. According to the parameter, a ``MachineReal`` or a ``PrecisionReal`` object is created. In the case of ``Symbol``, we use ``__new__`` to return always the same instance of symbols with the same name.

rocky: @mmatera I had considered that, but experiments show that this might not be the case. 

Consider this code: 

```
class Foo:
    def __new__(cls, *args, **kwargs):
        self = object.__new__(cls)
        print("XXX", id(self))
        return self


x = Foo()
y = Foo()

print(id(x))
print(id(y))
print(x == y)
```

If you run it you will see that `x` and `y` have different memory locations. And that x is not the same as y. 

So what leads you to believe that "new" is doing this? 

Also, I should say that if you wanted a single memory instance there is a design pattern to do this, but this isn't the way.

mmatera: @rocky, no: I am talking about ``Symbol``. I just mention this because I found that sometimes problems appear if you redefine ``__new__`` in a subclass with a different pattern of arguments than the default ``__new__`` in the parent class has.

rocky: A link to a line(s) of code would be nice.

rocky: Also, when there is non-obvious stuff going on, comments in code are a good place to describe this. If that gets too long an involved, we also have a developer's guide. 

Github comments in PRs will probably get lost. 

Think of coding like doing school homework in certain classes: you don't get credit for getting the right answer, you also need to show you work and calculcations.

rocky: One other thing. I am drafting a discussion for this whole Expression mess. So let's hold off on detailed discussions until I post that, and we can carry on the discussions there.

mmatera:
https://github.com/Mathics3/mathics-core/blob/d5c4ecfe350a33859de75601271f085a91aed1c9/mathics/core/symbols.py#L826
https://github.com/Mathics3/mathics-core/blob/d5c4ecfe350a33859de75601271f085a91aed1c9/mathics/core/atoms.py#L336

I do not have the code that produces the error (was a failed experiment), but probably you will detect it faster than I if it comes up.

rocky:
> https://github.com/Mathics3/mathics-core/blob/d5c4ecfe350a33859de75601271f085a91aed1c9/mathics/core/symbols.py#L826
> 
> 
> https://github.com/Mathics3/mathics-core/blob/d5c4ecfe350a33859de75601271f085a91aed1c9/mathics/core/atoms.py#L336
> 
> I do not have the code that produces the error (was a failed experiment), but probably you will detect it faster than I if it comes up.

Good - thanks. I will document this weird behavior then in this branch and update the comment.

mathics/core/symbols.py
```diff
@@ -128,202 +128,34 @@ def __ne__(self, other) -> bool:
...
     def __new__(cls, *args, **kwargs):
         self = object.__new__(cls)
         self.options = None
         self.pattern_sequence = False
-        self.unformatted = self
+        self.unformatted = self  # This may be a garbage-collection nightmare.
         self._cache = None
         return self
 
-    def clear_cache(self):
```
mmatera, line 154: Why is useful to sort the method by alphabetic order?

rocky: When we have very long lists of things it helps me to follow things. I know one can always search for stuff, but I'll just tend to scroll when things are in alphabetic order.

There are a couple of caveats though. Some things _need_ to be in a certain order. And so when the default is alphabetic, the break in order suggests that this kind of thing is going on.

Also, when there is  _small_ list of methods, ordering by logical group is okay.

mathics/core/symbols.py
TiagoCavalcante: Here is my [small] opinion: maybe making every expression have a different memory allocation is *good* (makes the things "easier" in other places). But Python has exited in the last decades, if they have chosen to do it this way, probably there is a *why* (not saying that Python makes the best design choices, it doesn't). So we *should* (I'm not asking anyone to do this, just saying my opinion) do it the Pythonic way.

mathics/core/symbols.py
TiagoCavalcante: Here I agree, that is why all code editors (which I know) sort the files alphabetically.

(This may not be necessary after drying BaseElement, but as we [I] already saw in the past, it is better to do the things step by step)

mathics/core/symbols.py
```diff
@@ -468,30 +382,57 @@ def format(self, evaluation, form, **kwargs) -> "BaseExpression":
...
+        # Comment @mmatera:
+        # This function is used just in Graphics.apply_makeboxes
+        # to check if a graphics expression is composed just by
+        # real numbers (or integer) or graphics symbols.
+        # Probably, there is a simpler way to implement it without using
+        # this method.
+        return []
 
-        return item_is_free(self, form, evaluation)
+    def get_attributes(self, definitions):
```
TiagoCavalcante, line 397: Here and in other places I think that `@attribute` should be used.

mathics/core/symbols.py
```diff
@@ -468,30 +382,57 @@ def format(self, evaluation, form, **kwargs) -> "BaseExpression":
...
+    # Compatibily with old code. Deprecated, but remove after a little bit.
+    get_leaves = get_elements
+
+    def get_float_value(self, permit_complex=False):
+        return None
+
+    def get_int_value(self):
         return None
 
+    def get_lookup_name(self):
```
TiagoCavalcante, line 422: This function also doesn't smell quite right for me.

TiagoCavalcante (approved):
My take would be to merge this and continue in other PRs.

This Looks Great For Me.

mathics/builtin/numbers/constants.py
```diff
@@ -158,7 +159,7 @@ def to_mpmath(self, args):
         return getattr(mpmath, self.mpmath_name)
 
 
-class _NumpyConstant(_Constant_Common):
+class _NumpyConstant(_Constant_Common, NumericOperators):
```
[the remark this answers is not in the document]
rocky: Don't know but will check and let you know...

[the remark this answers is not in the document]
mmatera: This `*Constant` are superclasses of some Builtins. Builtins are not `Element`s so, I think 
* They shouldn't be a subclass of `Number` that are Elements.
* In general, subclasses of `Builtin` shouldn't  implement `NumericOperators`.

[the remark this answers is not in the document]
TiagoCavalcante: @mmatera, I don't understand this enough to give my opinion about the first, but I agree with the second.

mathics/core/symbols.py
```diff
@@ -72,6 +72,49 @@ def strip_context(name) -> str:
     return name
 
 
+# FIXME: move to new module element.py
+class NumericOperators:
```
mmatera, line 76: I like this, as a first step. However, notice that ``Plus[3,"a"]`` or ``Times[x,Graphics[{}]]``  are  perfectly valid ``Expression`` s in WL.

mmatera: Nevertheless, from the Python side, maybe it also makes sense to define specialized behaviors for expressions like
```(Python)
Integer(1)+Integer(2)
```
or 
```(Python)
Integer(3)+MachineReal(.12)
```

rocky: The issue here is more about what we need to express in the Python code. While I can sort of see and tolerate writing
`abs(Integer(-8)`  I am not convinced it is a good idea over writing `Expression(SymbolAbs, 8)`.

And I would not be surprised if down the line we come up with a better way of doing such things. 

So the question is whether in mathics Python code we would ever want to write `Integer(3) + String("a")` or the other thing.  

My take is that when the _need_ arises we can reconsider this. But I really doubt the right thing will ever be to put this in the base class. 

And right now clarity and simplicity is way more important that being able to handle hypothetical edge cases like the ones described above. 

Sure, inside Mathics, you should be able to do that. Inside Python code implementing Mathics, I'd need convincing though.

rocky: > Nevertheless, from the Python side, maybe it also makes sense to define specialized behaviors for expressions like
> 
> ```
> Integer(1)+Integer(2)
> ```
> 
> or
> 
> ```
> Integer(3)+MachineReal(.12)
> ```

Ok. It is not clear that this is broken in the current code. None of the tests show it. If this isn't used right now in the code base let's leave it for later or feel free to at the mixin in the right place.

mmatera: My point is: in the Python side, usually you do not need to write ``abs(Integer(-8))`` but something like
```
def apply(expr, evaluation):
    '''F[expr_]'''
     ....
     new_expr = abs(expr)

```

in a way that ``element`` eventually is ``Integer(-8)`` or  ``Symbol("Pi")`` or ``Expression("F",37)``  or ``"Hello world"``. With this implementation, the last option would raise an exception. But OK, probably you would need to check that ``expr`` is something that makes sense to use as an argument of ``abs``.

mathics/builtin/numbers/calculus.py
```diff
@@ -1077,7 +1078,7 @@ def transform_solution(sol):
                 evaluation.message("Solve", "ivar", vars_original)
 
 
-class Integers(Builtin):
+class Integers(Builtin, NumericOperators):
```
mmatera, line 1081: Again, a `Builtin` shouldn't be at the same time  ``NumericOperators``

mmatera:
If we keep from here just the changes in the core side and revert the changes in builtins,  I would merge it now.

mathics/core/symbols.py
mmatera: > > Nevertheless, from the Python side, maybe it also makes sense to define specialized behaviors for expressions like
> > ```
> > Integer(1)+Integer(2)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > or
> > ```
> > Integer(3)+MachineReal(.12)
> > ```
> 
> Ok. It is not clear that this is broken in the current code. None of the tests show it. If this isn't used right now in the code base let's leave it for later or feel free to add the mixin in the right place.

No, I do not mean that your changes break something. I'm OK with them.  What I am talking about is just to consider possible extensions, just to avoid repeating mistakes. I mean, maybe in the future we realize that makes sense to specialize __add__   for integers, in a way that instead of return an expression ``Expression("Plus",Integer(1),Integer(2))`` returns directly  ``Integer(3)`` (for example).

mathics/builtin/numbers/constants.py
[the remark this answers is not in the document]
rocky: @mmatera Ok, it took me a while to understand what you mean, but yes, I believe you are correct and 1edd13d1 should address this. 

@TiagoCavalcante In effect what mmatera is saying is that inside Python you can't write say `Pi() + 3` (with or without the parenthesis) like you could do say `Integer(5) + 3`. This is because `Pi` acts as a built-in symbol which needs to be evaluated. You could presumably write in some code to do the evaluation, e.g. call the `evaluate()` function, but the the whole convenience of using infix `+` is kind of lost. 

So in summary, this kind of shorthand just doesn't apply for Mathics built-in variables/symbols like this.

mathics/core/symbols.py
```diff
@@ -456,6 +334,42 @@ def format_expr(expr):
         finally:
             evaluation.dec_recursion_depth()
 
+    def equal2(self, rhs: Any) -> Optional[bool]:
```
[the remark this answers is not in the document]
rocky: As far as I can tell `SameQ` and `Equal` are different, both operators are pervasive over all kinds of expression types. And the note that the specific implementation here is not the same as SameQ. 

That said, possibly it could be attached in some other class and just be made something that is mandatory to define. 

Your thoughts @mmatera ?

mathics/core/symbols.py
```diff
@@ -456,6 +334,42 @@ def format_expr(expr):
...
+        return self
+
+    def flatten(self, head, pattern_only=False, callback=None) -> "BaseExpression":
+        return self
+
+    def flatten_sequence(self, evaluation) -> "BaseExpression":
+        return self
+
+    def flatten_pattern_sequence(self, evaluation) -> "BaseExpression":
+        return self
```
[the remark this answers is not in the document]
rocky: I agree, I think it should be moved to `Expression`. Your thoughts @mmatera

mathics/core/symbols.py
rocky: I sort of agree but this is a pervasive change and not something to be put into this PR now. 

I had also considered this for `get_head` and `get_elements`. 

I have opened https://github.com/Mathics3/mathics-core/issues/174 to track this. 
Feel free to add to that or correct what is there.

TiagoCavalcante (approved):
LGTM

mathics/core/symbols.py
```diff
@@ -468,30 +382,57 @@ def format(self, evaluation, form, **kwargs) -> "BaseExpression":
...
+        # this method.
+        return []
 
-        return item_is_free(self, form, evaluation)
+    def get_attributes(self, definitions):
+        return nothing
 
-    def is_inexact(self) -> bool:
-        return self.get_precision() is not None
+    # Probably, this method shouldn't be here.
```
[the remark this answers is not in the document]
rocky: Noted in #174 for another PR

mathics/core/symbols.py
```diff
@@ -468,30 +382,57 @@ def format(self, evaluation, form, **kwargs) -> "BaseExpression":
...
 
-        This function is called by method `is_inexact()`.
-        """
+    def get_head_name(self):
+        raise NotImplementedError
+
+    # Compatibily with old code. Deprecated, but remove after a little bit.
+    get_leaves = get_elements
+
+    def get_float_value(self, permit_complex=False):
```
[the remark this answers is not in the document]
rocky: Noted in #174 for another PR.

mathics/core/symbols.py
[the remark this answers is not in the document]
mmatera: Yep, al these ``flatten_*`` should be specific for  ``Expression``, not for general ``elements``

mathics/core/symbols.py
rocky: I am guessing this should be an `Expression` thing, not something here. @mmatera ?

mathics/core/symbols.py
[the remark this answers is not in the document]
mmatera: ``SameQ`` is something that all elements should handle.  ``equal2`` I do not think so, since in general require  an  ``Evaluation`` object to be determined.

mathics/core/symbols.py
rocky: Yeah, let's revisit this later.

rocky: > 
> in a way that `element` eventually is `Integer(-8)` or `Symbol("Pi")` or `Expression("F",37)` or `"Hello world"`. With this implementation, the last option would raise an exception. But OK, probably you would need to check that `expr` is something that makes sense to use as an argument of `abs`.

Yes, I'd rather raise an exception here than east and execute on some mystery meat that might come back. (A friend of mine asked a Vietnamese person if they really ate dogs in Viet Nam; the response was "in the North only").

PR #175, opened 2022-02-25
BaseExpression -> BaseElement, move some methods out of BaseExpression (into Atom)

rocky:
And lastly reorder methods in the Expression class alphabetically. 

@mmatera since you asked about this, here is how this was useful in the PR alone. 

I wanted to know which methods of BaseElement are also listed in Expression. Those that are candidates for moving into Atom.

Having these two sets of methods sorted alphabetically made it possible for me to detet this.

### DIFF
--- a/mathics/algorithm/optimizers.py
+++ b/mathics/algorithm/optimizers.py
@@ -16,7 +16,7 @@ from mathics.core.atoms import (
     from_python,
 )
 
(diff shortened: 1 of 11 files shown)


TiagoCavalcante (approved):
LGTM

mmatera:
LGTM

PR #176, opened 2022-02-25
Reduce BaseElement interface

rocky:
* move round_to_float() into NumericOperations
* replace x.is_atom() with isinstance(x, Atom)
* replace x.is_symbol() with isinstance(x, Symbol)

As before, these mechanical changes touches lots of files.  (Not something I am fond of doing, gratuitously, were it not for the fact it was holding me up from more important things - see below).

Next up - splitting off BaseElement from mathics.core.symbols into a new module matthics.core.element

I had originally wanted to do this first, but because of the modularity violations in BaseElement I couldn't and had to do all of these others _first_. (I think the one that was crucial was the work to NumericOperations. But I couldn't have know that previously)

### DIFF
--- a/mathics/core/symbols.py
+++ b/mathics/core/symbols.py
@@ -75,12 +75,20 @@ def strip_context(name) -> str:
 # FIXME: move to new module element.py
 class NumericOperators:
     """
-    This is a mixin class that adds methods to the class to facilite building
-    ``Expression``s in Mathics code using Python syntax. For example,
-    instead of writing:
+    This is a mixin class for Element-like objects that might have numeric values.
+    It adds or "mixes in" numeric functions for these objects like round_to_float().
+
+    It also adds methods to the class to facilite building
+    ``Expression``s in the Mathics Python code using Python syntax.
+
+    So for example, instead of writing in Python:
+
         Expression("Abs", -8)
-    write:
+        Expression("Plus", 1, 2)
+
+    you can instead have:
         abs(Integer(-8))
+        Integer(1) + Integer(2)
     """
 
     def __abs__(self) -> "BaseElement":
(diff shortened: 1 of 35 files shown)


mathics/core/symbols.py
```diff
@@ -123,7 +146,7 @@ def __pow__(self, other) -> "BaseElement":
 # to be intercombined here.
 
 
-class KeyComparable(object):
+class KeyComparable:
```
rocky, line 149: @TiagoCavalcante a historical note: the reason `object `was there originally is that way back when this code was written (but the change to Python went in around 2.4 or so) there was skill a notion of the the kind of Python lame "class" that Python originally had.  

It always amazed me how long it too Python to fix things that started out doing brokenly but not were broken in other contemporaneous languages.

Back then, in order to get the current style class you had to use this `X(object)` form. 
https://stackoverflow.com/questions/54867/what-is-the-difference-between-old-style-and-new-style-classes-in-python

At some point, probably Python 3 the non-broken from of class became the default. But I think in Python 2.7 that wasn't the default. 

These are just stories you can tell your grandchildren.

rocky: Oh - more stories you can tell your grandchildren about the good old days. 

Back then there was no notion of a "folder" or "directory" so all programs had to put in one "file" which was called "a deck of cards".  So Originally Mathics was one file. 

And this was before graph theory was invented and they were still trying to figure out whether all the nodes in a tree should be called "leaves"

TiagoCavalcante (approved):
LGTM

mathics/core/symbols.py
TiagoCavalcante: > @TiagoCavalcante a historical note: the reason `object `was there originally is that way back when this code was written (but the change to Python went in around 2.4 or so) there was still a notion of the the kind of Python lame "class" that Python originally had.
> 
> It always amazed me how long it took Python to fix things that started out doing brokenly but not were broken in other contemporaneous languages.
> 
> Back then, in order to get the current style class you had to use this `X(object)` form. https://stackoverflow.com/questions/54867/what-is-the-difference-between-old-style-and-new-style-classes-in-python
> 
> At some point, probably Python 3 the non-broken from of class became the default. But I think in Python 2.7 that wasn't the default.
> 
> These are just stories you can tell your grandchildren.

Oh, I didn't know it, I thought that in Python 2 every class needed to extend something.

mmatera (approved):
LGTM

PR #177, opened 2022-02-25
Move BaseElement into mathics.core.element

rocky:
As stated before, this is most important part of what I wanted to do in the refactor. Although I wanted to do it first, alas it had to come after all of that other more disruptive stuff. 

So with this, I am done doing major refactoring and when I pick this up again will focus on adding properties to an the elements part of an Expression. Which was a digression of I forget what, improving interpreter evaluation performance? 

But even though I am more or less stopping the refactoring part here, there is still a lot more to do, by far. If others are interested please keep going. 

In particular, there are a number of methods in BaseElement that really shouldn't be there. They should either be moved into the subclasses that import BaseElement, or put in a Mixin class like NumericOperators, or the calling code should be rewritten to not call such methods. 

And then there is the format mess which we've never fully addressed. `do_format` and `format` is probably another candidate for a mixin like NumericOperations.

### DIFF
--- a/mathics/core/definitions.py
+++ b/mathics/core/definitions.py
@@ -11,16 +11,17 @@ import bisect
 from collections import defaultdict
 
 import typing
+
+from mathics.core.atoms import String
+from mathics.core.attributes import nothing
+from mathics.core.element import fully_qualified_symbol_name
+from mathics.core.expression import Expression
 from mathics.core.symbols import (
     Atom,
     Symbol,
-    fully_qualified_symbol_name,
     strip_context,
 )
-from mathics.core.expression import Expression
-from mathics.core.atoms import String
 from mathics_scanner.tokeniser import full_names_pattern
-from mathics.core.attributes import nothing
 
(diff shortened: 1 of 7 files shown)


TiagoCavalcante (approved):
LGTM - symbols.py seems to be a really bad place to keep BaseElement.

rocky:
> LGTM - symbols.py seems to be a really bad place to keep BaseElement.

I was looking at the history of this code.  It looks like the first and second maintainers each did about half the work in bulk of the code that we had when we it was largely abandoned in 2016.

When everything was in Expression in 2011 this wasn't so bad logically - Symbol, Expression, and BaseElement were in one file. That file was at 1.7K lines.  However all the wrong directions were there. It just was smaller. By 2016 this grew to 2.3K lines.  So when we got this we have been faced with the problem of basically double maintenance since  2011. 

We split off stuff into Symbol. And that was a big effort too. This is just yet another part of I guess this effort. But even here we are a bit short cleaning this up fully.

PR #178, opened 2022-02-25
Move mpmath implementation out of BaseElement...

rocky:
and put in alphabetical order more of the Expression methods. 

I had actually tried doing more but this is taking up too much time and I have other things that I need to do.

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -175,6 +175,12 @@ class Expression(BaseElement, NumericOperators):
         self._format_cache = None
         return self
 
+    def __getnewargs__(self):
+        return (self._head, self._elements)
+
+    def __hash__(self):
+        return hash(("Expression", self._head) + tuple(self._elements))
+
     def __repr__(self) -> str:
         return "<Expression: %s>" % self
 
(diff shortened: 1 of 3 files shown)


mathics/core/expression.py
```diff
@@ -430,6 +613,19 @@ def filter(self, head, cond, evaluation):
         # faster equivalent to: Expression(head, [element in self.leaves if cond(element)])
         return structure(head, self, evaluation).filter(self, cond)
 
+    def filter_elements(self, head_name):
```
TiagoCavalcante, line 616: This function is never used.

mathics/core/expression.py
```diff
@@ -1165,6 +1250,132 @@ def sameQ(self, other: BaseElement) -> bool:
...
+    def sequences(self):
+        cache = self._cache
+        if cache:
+            seq = cache.sequences
+            if seq is not None:
+                return seq
+
+        return self._rebuild_cache().sequences
+
+    def set_head(self, head):
```
TiagoCavalcante, line 1262: This is used only in 3 places:
- algorithms/parts.py
- builtin/structure.py
- core/subexpression.py

mathics/core/expression.py
```diff
@@ -1165,6 +1250,132 @@ def sameQ(self, other: BaseElement) -> bool:
...
+        self._head = head
+        self._cache = None
+
+    def set_element(self, index: int, value):
```
TiagoCavalcante, line 1266: And this is used in 2 places:
- algorithms/parts.py
- core/subexpression.py

mathics/core/expression.py
```diff
@@ -1165,6 +1250,132 @@ def sameQ(self, other: BaseElement) -> bool:
...
+        """
+        Update element[i] with value
+        """
+        elements = list(self._elements)
+        elements[index] = value
+        self._elements = tuple(elements)
+        self._cache = None
+
+    def set_reordered_elements(
```
TiagoCavalcante, line 1275: This is used only in sort.

TiagoCavalcante (approved):
LGTM - just see the 1st comment.

mathics/core/expression.py
rocky: Ok - good. It is been removed then. 

It is interesting in that when I looked at the v0.5 released code where the evaluate() method was very different, that code was in there. I especially remember he a TODO in there about how element sorting needed to be done. 

I guess that when the 2nd maintainer added that this was orphaned and didn't get removed.

mathics/core/expression.py
rocky: This and the other comment are all got, but I need to stop working on this. 

@TiagoCavalcante @mmatera I am happy to leave open this PR if someone wants to move it forward. Or we could comment and continue later. 

Discuss and decide.

TiagoCavalcante: @mmatera do you know whether we really need to invalidate the cache? If that isn't necessary, it would be just set a property to another value, so it could be inlined, otherwise, I think it is better to keep it where it is by now.

rocky: @TiagoCavalcante @mmatera please stop with the performance improvement suggestions for evaluation. Or make note of that somewhere and we can come back to it. Thanks.

TiagoCavalcante: @rocky that isn't about performance. My thoughts [rewritten]: if that function does 2 things, then keep it where it is, if the function does 1 thing, inline it in the 3 places it is called.

rocky: @TiagoCavalcante ok - sorry. 

But a couple of things here. As we've seen too many times in the past if something seems a little bit more complicated, there probably is a reason for this and best to assume that complications were put there for a reason. 

Sadly, these reasons haven't been spelled out so we can't tell if the reason still holds or had a fallacy in it. But to start out we should assume it is true. 

Second, here in the refactor part I am concerned about clarity.  And a-priori I don't see how inlining this is going to clarify something. Adding a docstring to this code would though. But that would mean understanding something about the code more than just where is it used. 

And in general there has been in my opinion too much discussion about things at the level of  inlining pieces of code or the order of if statements versus what is it we are trying to accomplish in the bigger picture. And this keeps us from moving forward towards understanding how to do what is necessary clearly, effectively (and efficiently).

TiagoCavalcante: @rocky ok, I think it is better to keep it as is then. From my part, we can merge now.

mathics/core/expression.py
rocky: Here I have "inlined" this. But here's the rationale and this is kind of important to me. 

My first aim was to reduce the method exposure of an Expression object since right now this is so large that I find it is a 3 line and `sort` was about 5 lines. With such small custom stuff, adding a comment before the 3 lines seems sufficient for clarity. 

And for the fans of squeezing the last ounce of performance, it occured to me that since this code is a bit heavy weight we should check to see if there is something to sort. That is there is more than 1 element. Otherwise we do the extra work to convert a tuple into a list, set up a list sorting key, do the sort, convert the list back into a tuple, and then possibly do something with a cache. 

Testing shows indeed there are a number of functions with zero and 1 argument.

mathics/core/expression.py
mmatera: Maybe it would be useful to add this as comments in the code.

TiagoCavalcante: @mmatera Rocky already fixed this.

mathics/core/expression.py
TiagoCavalcante: @rocky @mmatera maybe the sorting could be made after the evaluation has finished?

mathics/core/expression.py
rocky: Someone feel free to do. In a sense I am always for having comments. But I am trying to understand what why this would be helpful. Set element is setting an element so I would imagine something involving subexpressions and parts to want to modify this. 

If the only purpose is nothing this fact which I hope some sort of automated system would handle then we have the problem with this being wrong when the code changes. Because when someone adds another location or removes one of those locations who is going to know to update the comment. And then it will be wrong. 

So again what are we going to do with this information? You'll notice that in the PR I did add a docstring.

mathics/core/expression.py
rocky: @mmatera @TiagoCavalcante 

The focus here was on _clarity_ not figuring out how to improve it. 

What got us here was trying to describe in a high-level what is involved when we evaluate. And there was a bit of a mess that I couldn't do the first proposal without making things possibly more complicated.

Discussing this proposal in that big picture discussion is to my mind preferable to looking at a PR and discussing that here which is tangential.

So @mmatera I'd suggest looking at the overall steps and in that describe how you think things should change. Then if you think this still works, let's discuss.

mmatera (approved):
LGTM

mmatera (approved):
LGTM

mmatera:
@rocky, merge this when you feel it is ready.

PR #180, opened 2022-02-26
Remove unformatted property from BaseElement

TiagoCavalcante:
unformatted was used only in 1 place, and seems it works without it there.

I think that now the `__new__` method can be converted to a `__init__`, fixing the garbage collection problem.

### DIFF
--- a/mathics/builtin/inout.py
+++ b/mathics/builtin/inout.py
@@ -1002,7 +1002,7 @@ class TableForm(Builtin):
         f:StandardForm|TraditionalForm|OutputForm]"""
 
         dims = len(get_dimensions(table, head=SymbolList))
-        depth = self.get_option(options, "TableDepth", evaluation).unformatted
+        depth = self.get_option(options, "TableDepth", evaluation)
         depth = expr_min((Integer(dims), depth))
         depth = depth.get_int_value()
         if depth is None:
--- a/mathics/core/element.py
+++ b/mathics/core/element.py
@@ -97,18 +97,10 @@ class BaseElement(KeyComparable):
     # this variable holds a function defined in mathics.core.expression that creates an expression
     create_expression: Any
 
-    # __new__ seems to be used because this object references itself.
-    # In particular:
-    #    self.unformatted = self
-    #
-    # See if there's a way to get rid of this, or ensure that this isn't causing
-    # a garbage collection problem.
-    def __new__(cls, *args, **kwargs):
-        self = object.__new__(cls)
+    def __init__(self, *args, **kwargs):
         self.options = None
         self.pattern_sequence = False
         self._cache = None
-        return self
 
     def apply_rules(
         self, rules, evaluation, level=0, options=None
--- a/mathics/core/element.py
+++ b/mathics/core/element.py
@@ -107,7 +107,6 @@ class BaseElement(KeyComparable):
         self = object.__new__(cls)
         self.options = None
         self.pattern_sequence = False
-        self.unformatted = self  # This may be a garbage-collection nightmare.
         self._cache = None
         return self
 
@@ -178,12 +177,10 @@ class BaseElement(KeyComparable):
                 if not (form is SymbolOutputForm and head is SymbolStandardForm):
                     form = head
                     include_form = True
-            unformatted = expr
             # If form is Fullform, return it without changes
             if form is SymbolFullForm:
                 if include_form:
                     expr = self.create_expression(form, expr)
-                    expr.unformatted = unformatted
                 return expr
 
             # Repeated and RepeatedNull confuse the formatter,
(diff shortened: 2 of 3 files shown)


mmatera:
In general, LGTM. 

*side comments*:

I think that the ``unformatted`` parameter was introduced for implementing a feature that was never complete: Suppose you have certain input Expression ``iexpr``. Then you evaluate it and you get a resulting expression ``rexpr``. Then, ``Evaluation`` is going to format it, to a new expression ``fexpr``.    ``fexpr`` is composed of boxes and strings, so lose the semantical information from the expression from which it came from, so if you want to reuse the output as a new input expression, in general, you can not parse it.

I think a more clever way to do this is to have special objects to handle boxes expressions. For these specialized classes, maybe it makes sense to have a reference to the original expression, and also all the ``boxes_to_*`` methods.

TiagoCavalcante:
> In general, LGTM.
> 
> _side comments_:
> 
> I think that the `unformatted` parameter was introduced for implementing a feature that was never complete: Suppose you have certain input Expression `iexpr`. Then you evaluate it and you get a resulting expression `rexpr`. Then, `Evaluation` is going to format it, to a new expression `fexpr`. `fexpr` is composed of boxes and strings, so lose the semantical information from the expression from which it came from, so if you want to reuse the output as a new input expression, in general, you can not parse it.
> 
> I think a more clever way to do this is to have special objects to handle boxes expressions. For these specialized classes, maybe it makes sense to have a reference to the original expression, and also all the `boxes_to_*` methods.

@mmatera I think that in the cases we need the original expression, it is better to just place it somewhere else, e.g.:
```py
original_expr = expr
new_expr = expr.do_format(...)
```

And I think that would be no harm compared to the current garbage collection problem.

I'm going to do some benchmarks here to see if this improves the performance by a perceptible amount. But even if it doesn't, unused code is dead code, and we can re-add this if we think it is necessary.

mmatera:
> @mmatera I think that in the cases we need the original expression, it is better to just place it somewhere else, e.g.:
> 
> ```python
> original_expr = expr
> new_expr = expr.do_format(...)
> ```
>

The point is that you could want to do this in a different place:  suppose that in a method you get the formatted expression object as a parameter, , and  you need to get its fullform to reformat it, or reevaluate it. 



> And I think that would be no harm compared to the current garbage collection problem.
> 
> I'm going to do some benchmarks here to see if this improves the performance by a perceptible amount. But even if it doesn't, unused code is dead code, and we can re-add this if we think it is necessary.

Sure. In any case, what would be important is that you do not get something worst.

TiagoCavalcante:
> The point is that you could want to do this in a different place: suppose that in a method you get the formatted expression object as a parameter, , and you need to get its fullform to reformat it, or reevaluate it.

In that case, we could just get the FullForm of the expression. Of course, it would be an "unnecessary" evaluation, but this "unnecessary" evaluation never happened until today, so I think it is worth removing it.

TiagoCavalcante:
@rocky @mmatera this has no significant changes in performance.

mmatera:
In master:
```
In[1]:=Timing[Integrate[Integrate[20000*E^y/(1-x/2),{x,-5,0}],{y,-3,0}]+Integrate[Integrate [20000*E^y/(1+x/2),{x,0,5}],{y,-3,0}]]

Out[1]={42.5735, 2 Log[2 ^ (-40000 + 40000 / E ^ 3) 7 ^ (40000 - 40000 / E ^ 3)]}
```
in this branch:
```
In[1]:= Timing[Integrate[Integrate[20000*E^y/(1-x/2),{x,-5,0}],{y,-3,0}]+Integrate[Integrate [20000*E^y/(1+x/2),{x,0,5}],{y,-3,0}]]

Out[1]={41.1603, 2 Log[2 ^ (-40000 + 40000 / E ^ 3) 7 ^ (40000 - 40000 / E ^ 3)]}
```

We have an improvement here, dealing with a complicate expression. Also, this pass the tests so, let's merge it!

rocky (approved):
LGTM

rocky:
> In master:
> 
> ```
> In[1]:=Timing[Integrate[Integrate[20000*E^y/(1-x/2),{x,-5,0}],{y,-3,0}]+Integrate[Integrate [20000*E^y/(1+x/2),{x,0,5}],{y,-3,0}]]
> 
> Out[1]={42.5735, 2 Log[2 ^ (-40000 + 40000 / E ^ 3) 7 ^ (40000 - 40000 / E ^ 3)]}
> ```
> 
> in this branch:
> 
> ```
> In[1]:= Timing[Integrate[Integrate[20000*E^y/(1-x/2),{x,-5,0}],{y,-3,0}]+Integrate[Integrate [20000*E^y/(1+x/2),{x,0,5}],{y,-3,0}]]
> 
> Out[1]={41.1603, 2 Log[2 ^ (-40000 + 40000 / E ^ 3) 7 ^ (40000 - 40000 / E ^ 3)]}
> ```
> 
> We have an improvement here, dealing with a complicate expression. Also, this pass the tests so, let's merge it!

Note that this branch also had the stuff improving performance for expressions with 0 or 1 argument and that looks like this is going on above too. 

So do we know how much of the improvement was obtained by each in isolation rather than in combination?

PR #181, opened 2022-02-28
Rework boxconstruct

mmatera:
This is a work in progress. The plan is to use a specialized kind of ``Expression`` for the output of ``MakeBoxes``. This specialized class should not implement ``evaluate`` or ``evaluate_next`` but methods like ``boxes_to_text``, which shouldn´t be in an evaluable ``Expression``. 

Again, this is something that can be done in a single, large PR, or doing step by step. By now, I started to add comments on the parts to be reworked.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/181"></a>

### DIFF
--- a/mathics/builtin/box/boxexpression.py
+++ b/mathics/builtin/box/boxexpression.py
@@ -1,0 +1,334 @@
+# cython: language_level=3
+# -*- coding: utf-8 -*-
+
+import typing
+from typing import Any, Optional
+
+from mathics.core.element import ensure_context
+
+from mathics.core.symbols import (
+    BaseElement,
+    Symbol,
+)
+
+# from mathics.core.systemsymbols import (
+#    SymbolSymbol,
+# )
+
+
+class BoxError(Exception):
+    def __init__(self, box, form) -> None:
+        super().__init__("Box %s cannot be formatted as %s" % (box, form))
+        self.box = box
+        self.form = form
+
+
+class BoxExpression(BaseElement):
+    head: "Element"
+    leaves: typing.List[Any]
+    _sequences: Any
+
+    def __init__(self, *args, **kwargs):
+        if len(args) == 0:
+            return
+        head = args[0]
+        if isinstance(head, str):
+            head = Symbol(head)
+        self._head = head
+        self._elements = args[1:]
+
+    def __repr__(self) -> str:
+        return "<BoxExpression: %s>" % self
+
+    def __str__(self) -> str:
+        if not hasattr(self, "_head"):
+            return "<BoxExpression uninitialized>"
+        return "%s[%s]" % (
+            self._head,
+            ", ".join([element.__str__() for element in self._elements]),
+        )
+
+    @property
+    def elements(self):
+        return self._elements
+
+    def evaluate(self, evaluation):
+        print("Deprecated!")
+        return self
+
+    def flatten_pattern_sequence(self, evaluation):
+        """
+        called from mathics.core.pattern
+        """
+        print("Deprecated!")
+        return self
+
+    def get_attributes(self, definitions):
+        return nothing
+
+    def get_elements(self):
+        return self._elements
+
+    # Compatibily with old code. Deprecated, but remove after a little bit
+    get_leaves = get_elements
+
+    def get_head(self):
+        return self._head
+
+    def get_head_name(self):
+        return self._head.name if isinstance(self._head, Symbol) else ""
+
+    def get_mutable_elements(self) -> list:
+        """
+        Return a shallow mutable copy of the elements
+        """
+        return list(self._elements)
+
+    def has_form(self, heads, *element_counts):
+        """
+        element_counts:
+            (,):        no elements allowed
+            (None,):    no constraint on number of elements
+            (n, None):  leaf count >= n
+            (n1, n2, ...):    leaf count in {n1, n2, ...}
+        """
+
+        head_name = self._head.get_name()
+        if isinstance(heads, (tuple, list, set)):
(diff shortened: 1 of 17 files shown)


mathics/builtin/box/boxexpression.py
```diff
@@ -0,0 +1,334 @@
...
+
+    def evaluate(self, evaluation):
+        print("Deprecated!")
+        return self
+
+    def flatten_pattern_sequence(self, evaluation):
+        """
+        called from mathics.core.pattern
+        """
+        print("Deprecated!")
```
TiagoCavalcante, line 63: With this we can't see what is deprecated just reading the logs, maybe it could print "flatten_pattern_sequence is deprecated".

But this is WIP, so I probably shouldn't write this comment.

mmatera: Yes @TiagoCavalcante, I just put this to debug the code.

mmatera:
@TiagoCavalcante, @rocky: here is a first working proposal. As I promised, I didn't touch anything in the code of ``mathics.core``, except by adding comments. However, several of the ``boxes_to_`` functions in ``Element`` are not being called anymore.

By now ``mathics.builtin.box.boxexpression`` was not used in any place. It is just a draft of what I think should be, but probably it will change after the Rocky's refactor.

rocky:
The test I think should be split off as a separate PR and where there are bugs, fixed and merge in.

The concept of specializing Boxing Expressions is the right idea. But personally, am don't even want to go down considering this since evaluation is under upheaval. I briefly looked at a number of comments and agree that this should be done is normally done. But going down the specifics of how is something I don't want to go down right now.

mmatera:
@TiagoCavalcante, @rocky, I think this is almost ready, at least for a first round. My question now is if I should drop 'mathics/builtin/box/boxexpression.py' and rename `BoxConstruct` to  `BoxExpression`, or if it is good to keep the prototype and then see if `BoxConstruct` should be dedicated from  `BoxExpression`.

mmatera:
Another concern I have is if we should implement classes like ``_BoxedString``  for the other atoms, (Integers, Rational, Reals, Complex) to remove `boxes_to_*` methods for their classes in  `mathics.core`, or if we should keep these methods in the `Atom` subclasses. In any case, I guess @rocky will have a better idea about how to make this splitting cleaner.

mmatera:
> The test I think should be split off as a separate PR and where there are bugs, fixed and merge in.

The tests are in #184. However, to fix the bugs, I should work over `mathics.core`, which is something we agree not to modify until you finish the refactor. Bugs are not critical, so the fixes can wait until that.

> 
> The concept of specializing Boxing Expressions is the right idea. But personally, am don't even want to go down considering this since evaluation is under upheaval. I briefly looked at a number of comments and agree that this should be done is normally done. But going down the specifics of how is something I don't want to go down right now.

OK, in that case, I leave this PR here until the refactor be finished. Changes in core are just comments, and the changes in Builtin are localized enough to be easy to rebase.

rocky:
> > The test I think should be split off as a separate PR and where there are bugs, fixed and merge in.
> 
> The tests are in #184. However, to fix the bugs, I should work over `mathics.core`, which is something we agree not to modify until you finish the refactor. Bugs are not critical, so the fixes can wait until that.

We can put this in with a mark skipped in pytest. 

> 
> > The concept of specializing Boxing Expressions is the right idea. But personally, am don't even want to go down considering this since evaluation is under upheaval. I briefly looked at a number of comments and agree that this should be done is normally done. But going down the specifics of how is something I don't want to go down right now.
> 
> OK, in that case, I leave this PR here until the refactor be finished. Changes in core are just comments, and the changes in Builtin are localized enough to be easy to rebase.

Thanks for your understanding.  If you want to add comments, please put those in a separate PR and we should merge that in.

Having worked on getting the more clarity in the code feels like the right thing for now.

rocky:
BTW I was thinking about whether Boxing functions should be an Expression class or a mixin to existing classes that need boxing (which in some cases feels more like _unboxing_ since it is expanding what used to be an Atom into lots of format-specific code). 

And it might be both. 

But for these things, I think the best way to go is to start a discussion rather than work off of a particular PR. 

If the PR helps with understanding what the problems and issues are then that's great. But personally, while this has been a todo item, I need to stay somewhat focused on finishing a pass at the evaluation loop. And this has only just begun.

mmatera:
OK, so I close this and split it in more ore less independent parts:

* Tests alone are in #184
* All the comments about the core and ``builtin.base` plus the cleaning of  ``BoxConstruct`` are in #185. 
* #186 has the changes in ``BoxConstruct`` plus the work of decoupling ``boxes_to_*`` usind derivated classes, plus the fixes.  Here there are no changes or comments in ``mathics.core``.

mmatera:
> BTW I was thinking about whether Boxing functions should be an Expression class or a mixin to existing classes that need boxing (which in some cases feels more like _unboxing_ since it is expanding what used to be an Atom into lots of format-specific code).
> 
> And it might be both.
> 
> But for these things, I think the best way to go is to start a discussion rather than work off of a particular PR.
> 
> If the PR helps with understanding what the problems and issues are then that's great. But personally, while this has been a todo item, I need to stay somewhat focused on finishing a pass at the evaluation loop. And this has only just begun.

I actually was working on this as a proof of concept. I wanted to know how feasible was to decouple the formatting step from the ``Expression`` class, without touching the core. This allows for instance to put all this PR in a pymathics module, and if some context requires it, to write a different module supporting other `boxes_to_*` methods.

PR #183, opened 2022-03-06
Make a stab reducing BaseElement...

rocky:
* comment why we need _new__() where we do.
* comment has_changed
* Try to specialize some has_changed functions. It appears that `has_changed()` is used as some sort of proxy for deciding when to cache as well as whether to re-evaluate an expression. Start to  straighten this out.
* put in alphabetic order methods in the Atom class.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/183"></a>

### DIFF
--- a/mathics/core/atoms.py
+++ b/mathics/core/atoms.py
@@ -84,6 +84,13 @@ class Number(Atom, NumericOperators):
     def __str__(self) -> str:
         return str(self.value)
 
+    def has_changed(self, definitions) -> bool:
+        """
+        Used in Expression.evaluate() to determine if we need to reevaluation
+        an expression. Numbers never change.
+        """
+        return False
+
     def is_numeric(self, evaluation=None) -> bool:
         return True
 
@@ -139,6 +146,8 @@ class Integer(Number):
     value: int
     class_head_name = "System`Integer"
 
+    # We use __new__ here to unsure that two String's that have the same value
+    # return the same object.
     def __new__(cls, value) -> "Integer":
         n = int(value)
         self = super(Integer, cls).__new__(cls)
@@ -888,6 +897,14 @@ class String(Atom):
         else:
             return [0, 1, self.value, 0, 1]
 
+    def has_changed(self, definitions) -> bool:
+        """
+        Used in Expression.evaluate() to determine if we need to reevaluation
+        an expression.
+        """
+        # Should be false - investigate why we need to set this True.
+        return True
+
     def sameQ(self, other) -> bool:
         """Mathics SameQ"""
         return isinstance(other, String) and self.value == other.value
@@ -920,6 +937,8 @@ class ByteArrayAtom(Atom):
     value: str
     class_head_name = "System`ByteArrayAtom"
 
+    # We use __new__ here to unsure that two ByteArrayAtom's that have the same value
+    # return the same object.
     def __new__(cls, value):
         self = super().__new__(cls)
         if type(value) in (bytes, bytearray):
--- a/mathics/core/atoms.py
+++ b/mathics/core/atoms.py
@@ -146,7 +146,7 @@ class Integer(Number):
     value: int
     class_head_name = "System`Integer"
 
-    # We use __new__ here to unsure that two String's that have the same value
+    # We use __new__ here to unsure that two Integer's that have the same value
     # return the same object.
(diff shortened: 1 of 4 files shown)


mathics/core/atoms.py
```diff
@@ -139,6 +146,8 @@ class Integer(Number):
     value: int
     class_head_name = "System`Integer"
 
+    # We use __new__ here to unsure that two String's that have the same value
```
TiagoCavalcante, line 149: Isn't it Integer?

mathics/core/expression.py
```diff
@@ -861,6 +864,11 @@ def get_sort_key(self, pattern_sort=False):
                 return [1 if self.is_numeric() else 2, 3, head, self._elements, 1]
 
     def has_changed(self, definitions):
+
+        # Some Atoms just don't have a cache.
+        if not hasattr(self, "_cache"):
+            return
```
TiagoCavalcante, line 870: It don't need to be a boolean?

mathics/core/symbols.py
```diff
@@ -288,6 +288,15 @@ def get_sort_key(self, pattern_sort=False):
         else:
             raise NotImplementedError
 
+    def has_changed(self, definitions) -> bool:
+        """
+        Used in Expression.evaluate() to determine if we need to reevaluation
+        an expression.
+        """
+        # It would be nice to make the default "False" here and have Symbols
```
TiagoCavalcante, line 296: I think it should be false.

mathics/core/symbols.py
```diff
@@ -490,8 +430,21 @@ def evaluate(self, evaluation):
...
+    def get_head(self) -> "Symbol":
+        return Symbol("Symbol")
+
+    def get_head_name(self):
+        return "System`Symbol"
+
+    def has_changed(self, definitions):
+        """
+        Used in Expression.evaluate() to determine if we need to reevaluation
+        an expression.
```
TiagoCavalcante, line 442: Maybe false here too?

TiagoCavalcante (approved):
See the comments, anyway LGTM.

mmatera (approved):
LGTM

mathics/core/atoms.py
rocky: Thanks  - should be fixed now.

mathics/core/expression.py
rocky: Yes, it would be better as a boolean so that is fixed now. Python is loose in that `return` is `return None` and `None` behaves like  `False`.  But better to be explicit about the intent.

mathics/core/symbols.py
rocky: In one situation, testing if an expression or atom has_changed` is used in rebuilding a cache.  I addressed that particular place, but I think there are other places that need to be fixed before we can make this False.

I am kind of disgusted at how the caching which should have been limited to perhaps compound expressions has invaded everything so everything has to coddle to the cache. (And if the cache were _that_ good we wouldn't be having the performance problems we have). 

It should not be mandatory for an Atom to provide caching when it is nonsensical. Rather it should be up to the evaluator to see whether there is a cache. I have only looked at and changed one place. 

As too often is the case there is too much mess and I have too little time for me to try to address this in one go.

This is basically what I have for this week. And I _still_ haven't gotten to adding a property on the elements to indicate that all of them never need reevaluation.

mathics/core/symbols.py
rocky: Symbol is probably one of the very few places where we might have to worry about. You can update a symbol's definition during evaluation. 

But the thing that is a mystery (and therefore may be bad coding) is why this definition kind of check isn't done inside this method. I am guessing that in those cases where things change it is noticed in the caller. 

So has_changed may really mean "could possibly have changed".

mathics/core/symbols.py
rocky: @TiagoCavalcante @mmatera looking over the single use of `has_changed()` it looks like a more appropriate name would be `needs_evaluation` (in order to set some sort of "value" field?).

For certain numbers like Integers and Rational this could be done initially. I think that's why we have `needs_evaluation` set False for a number of things like String. 

If this is the case, I'd like (next week) to rename this method. 

Comments?

PR #186, opened 2022-03-07
Rework boxconstruct stripped

mmatera:
This PR is the part of #181 that does the actual rework, without the comments in ``mathics.core``.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/186"></a>


mathics/builtin/__init__.py
```diff
@@ -216,3 +216,12 @@ def is_builtin(var):
 _builtins = {}
 
 add_builtins(new_builtins)
+
+
+display_operators_set = set()
```
TiagoCavalcante, line 221: I didn't find where it is used.

mmatera: This is used in `_BoxedString`

mathics/builtin/box/inout.py
```diff
@@ -1,11 +1,179 @@
...
+                return render("<ms>%s</ms>", text[1:-1])
+            else:
+                outtext = ""
+                for line in text[1:-1].split("\n"):
+                    outtext += render("<mtext>%s</mtext>", line)
+                return outtext
+        elif text and ("0" <= text[0] <= "9" or text[0] == "."):
+            return render("<mn>%s</mn>", text)
+        else:
+            if text in operators or text in extra_operators:
```
mmatera, line 100: Here...

mathics/builtin/__init__.py
mmatera: In ``mathics.core.atom.String.boxes_to_mathml`` this set is built each time the method is called. I just move it to ``mathics.builtin.base`` to do it just once.

PR #187, opened 2022-03-07
Reduce atom has changed

rocky:
Following up on @TiagoCavalcante ' s comment about (most) Atoms, especially Strings not being needing re-evaluation (or a complex cache) I have reduced this at the slight expense of an extra check in one of the caching routines.

That could be removed by fixing up BaseElement not to set up a cache in the first place (which I think is the right thing), but there is some other interaction with Boxing that is causing infinite recursion. Go figure. 

Until we get that problem fixed (and will that lead to some other unrelated place that needs to change as well?) in my opinion the best thing is just to make that extra lightweight test in the cache update routine. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/187"></a>

### DIFF
--- a/mathics/core/atoms.py
+++ b/mathics/core/atoms.py
@@ -84,13 +84,6 @@ class Number(Atom, NumericOperators):
     def __str__(self) -> str:
         return str(self.value)
 
-    def has_changed(self, definitions) -> bool:
-        """
-        Used in Expression.evaluate() to determine if we need to reevaluation
-        an expression. Numbers never change.
-        """
-        return False
-
     def is_numeric(self, evaluation=None) -> bool:
         return True
 
@@ -896,13 +889,6 @@ class String(Atom):
             return super().get_sort_key(True)
         else:
             return [0, 1, self.value, 0, 1]
-
-    def has_changed(self, definitions) -> bool:
-        """
-        Used in Expression.evaluate() to determine if we need to reevaluation
-        an expression.
-        """
-        return False
 
     def sameQ(self, other) -> bool:
         """Mathics SameQ"""
--- a/mathics/core/atoms.py
+++ b/mathics/core/atoms.py
@@ -902,8 +902,7 @@ class String(Atom):
         Used in Expression.evaluate() to determine if we need to reevaluation
         an expression.
         """
-        # Should be false - investigate why we need to set this True.
-        return True
+        return False
 
     def sameQ(self, other) -> bool:
         """Mathics SameQ"""
--- a/mathics/core/element.py
+++ b/mathics/core/element.py
@@ -100,6 +100,7 @@ class BaseElement(KeyComparable):
     def __init__(self, *args, **kwargs):
         self.options = None
         self.pattern_sequence = False
+        # FIXME: this should be removed
         self._cache = None
 
     def apply_rules(
@@ -123,6 +124,7 @@ class BaseElement(KeyComparable):
                 return result, True
         return self, False
 
+    # FIXME: this should be removed
     def clear_cache(self):
         self._cache = None
 
--- a/mathics/core/element.py
+++ b/mathics/core/element.py
@@ -432,7 +432,7 @@ class BaseElement(KeyComparable):
 
     def has_changed(self, definitions) -> bool:
         """
-        Used in Expression.evaluate() to determine if we need to reevaluation
+        Used in Expression.evaluate() to determine if we need to reevaluate
         an expression. Each subclass should decide what is right here.
         """
         raise NotImplementedError
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -149,7 +149,7 @@ class ExpressionCache:
             if not hasattr(expr, "_cache") or expr.has_changed(definitions):
                 return None
 
-        # FIXME: this is workaround that some like String have a cache
+        # FIXME: this is workaround that some Atoms, like String, have a cache
         # even though they don't need it, by virtue of this getting set up
         # in BaseElement.__init__. Removing the self._cache in there the causes Boxing
         # to mess up. Untangle this mess.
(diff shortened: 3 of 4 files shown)


mmatera:
> Following up on @TiagoCavalcante ' s comment about (most) Atoms, especially Strings not being needing re-evaluation (or a complex cache) I have reduced this at the slight expense of an extra check in one of the caching routines.
> 
> That could be removed by fixing up BaseElement not to set up a cache in the first place (which I think is the right thing), but there is some other interaction with Boxing that is causing infinite recursion. Go figure.

I am going to check if together with #154, the infinite recursion disapears.

> 
> Until we get that problem fixed (and will that lead to some other unrelated place that needs to change as well?) in my opinion the best thing is just to make that extra lightweight test in the cache update routine.
> 
> [](https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/187)

mmatera (approved):
LGTM

mmatera:
> > Following up on @TiagoCavalcante ' s comment about (most) Atoms, especially Strings not being needing re-evaluation (or a complex cache) I have reduced this at the slight expense of an extra check in one of the caching routines.
> > That could be removed by fixing up BaseElement not to set up a cache in the first place (which I think is the right thing), but there is some other interaction with Boxing that is causing infinite recursion. Go figure.
> 
> I am going to check if together with #154, the infinite recursion disapears.
> 
> > Until we get that problem fixed (and will that lead to some other unrelated place that needs to change as well?) in my opinion the best thing is just to make that extra lightweight test in the cache update routine.
> > [](https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/187)

Indeed, all the test passes: See https://github.com/Mathics3/mathics-core/tree/reduce-atom-has_changed_and_boxconstruct

rocky:
Ok - thanks for the information. I am merging this in now, and let's continue. 

If in another PR we could get the part in where we remove the cache that is not wanted in the classes where it doesn't make sense that would be great. I will look at this next weekend if it hasn't been resolved.

rocky:
@mmatera @TiagoCavalcante One other change I am thinking about for next weekend is to rename "has_changed" to "continue_evaluation". Or suggest something along those lines. Other names were may_need_evaluation, or a name that somehow means does not need further evaluation.

And the docstring would be something like:

    Used in .. [same as before] 
    Returns True if evaluation is needed or could potentially change the current value.
    Returns False if we have a value and no further evaluation is needed.

The thing that seems wrong about `has_changed` is that it is not about what has happened in the past, but what might happen as a result of a _future_ evaluation. Do I have this correct?

rocky:
BTW - by looking at previous releases and git history you can learn a lot about the code. ExpressionCache was added relatively late in the era of the second dynasty in version 1.1.

rocky:
> > I am going to check if together with #154, the infinite recursion disapears.
(https://camo.githubusercontent.com/76e60919474807718793857d8eb615e7a50b18b04050577e5a35c19421f260a3/68747470733a2f2f676974706f642e696f2f627574746f6e2f6f70656e2d696e2d676974706f642e737667)](https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/187)
> 
> Indeed, (with very few changes) all the tests pass: See https://github.com/Mathics3/mathics-core/tree/reduce-atom-has_changed_and_boxconstruct

This is good news. I think we are in agreement, but to make clear about the cache. The issue per se is not about whether Boxing should cache, but more that it _should not_ be set up in BaseElement. If Boxing would benefit from a cache, the Boxing `__init__` or`__new__` should set that up and decide _what_ kind of cache it wants. Maybe it is ExpressionCache, but maybe it is something different or nothing at all is desired.

The other thing here that seems to happen all the time is that I start to work on something, then you go off in a related area in a different direction which from my side is just detracting from the primary action. And then all sorts of features get added to this side thing. 

It is good to know that the problem is solved by the Box class involvement with caching and that probably it doesn't need any sort of caching.  This will be fixed in core, not added only through code in some tangential or parallel feature.

mmatera:
> > > I am going to check if together with #154, the infinite recursion disapears.
> > > ([https://camo.githubusercontent.com/76e60919474807718793857d8eb615e7a50b18b04050577e5a35c19421f260a3/68747470733a2f2f676974706f642e696f2f627574746f6e2f6f70656e2d696e2d676974706f642e737667)](https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/187](https://camo.githubusercontent.com/76e60919474807718793857d8eb615e7a50b18b04050577e5a35c19421f260a3/68747470733a2f2f676974706f642e696f2f627574746f6e2f6f70656e2d696e2d676974706f642e737667)%5D(https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/187))
> > 
> > 
> > Indeed, (with very few changes) all the tests pass: See https://github.com/Mathics3/mathics-core/tree/reduce-atom-has_changed_and_boxconstruct
> 
> This is good news. I think we are in agreement, but to make clear about the cache. The issue per se is not about whether Boxing should cache, but more that it _should not_ be set up in BaseElement. If Boxing would benefit from a cache, the Boxing `__init__` or`__new__` should set that up and decide _what_ kind of cache it wants. Maybe it is ExpressionCache, but maybe it is something different or nothing at all is desired.
> 

Yep.


> The other thing here that seems to happen all the time is that I start to work on something, then you go off in a related area in a different direction which from my side is just detracting from the primary action. And then all sorts of features get added to this side thing.
> 

This is not the case this time, or at least it was not my intention. My idea with #181 was to decouple as much as possible the formatting from the core, without introducing anything new, but in a way that you can focus on the problem of evaluation without crashing with artifacts from the format step. Even without merging, I can check some hypotheses, like the one over the problem of removing  ``_cache`` from  ``BaseElement``.


> It is good to know that the problem is solved by removing the Box class involvement with caching and that probably it doesn't need any sort of caching. This will be fixed in core, not added only through code in some tangential or parallel feature.

Again, yes, but it took to me one day to disentangle the code and check that doing this didn't screw the expected behavior.

PR #189, opened 2022-03-08
Complexityfunction

### LINKED ISSUE #188: ``ComplexityFunction`` option in ``Simplify`` and ``FullSimplify``
> [Here's](https://github.com/sympy/sympy/issues/23210) the relevant issue.

This is interesting because it allows the implementation of the `Simplify` option `ComplexityFunction` which is available in  WMA. In any case, it seems that the default complexity function is not good for this context.

_Originally posted by @mmatera in https://github.com/Mathics3/mathics-core/issues/154#issuecomment-1060070468_

mmatera:
This fixes #188.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/189"></a>

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@ Enhancements
 * numeric overflows now do not affect the full evaluation, but instead just the element which produce it.
 * ``SameQ`` (``===``) handles chaining, e.g. ``a == b == c`` or ``SameQ[a, b, c]``
 * ``Simplify`` handles properly expressions of the form ``Simplify[0^a]`` (issue #167)
+* ``Simplify`` and ``FullSimplify`` support optional parameters ``Assumptions`` and ``ComplexityFunction``
 * The order of the context name resolution (and ``$ContextPath``) switched putting ``"System`"`` before ``"Global`"`.
 * In assignment to messages associated with symbols, the attribute ``Protected`` is not having into account, folliwing the standard in WMA. With this and the above change, Combinatorical 2.0 works as written.
 * ``Share[]`` performs an explicit call to the Python garbage collection and returns the amount of memory free.
--- a/mathics/algorithm/simplify.py
+++ b/mathics/algorithm/simplify.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+
+"""
+Algorithms for simplifying expressions.
+"""
+
+from mathics.core.symbols import Atom
+from mathics.core.atoms import Number
+from mathics.core.expression import Expression
+
+
+def default_complexity_function(x) -> int:
+    """
+    Evaluates the complexity of an expression. Each atom
+    counts 1, except for numbers that counts 1 each 5 characters
+    in its decimal expansion.
+    """
+    if isinstance(x, Number):
+        result = 1 + int(len(str(x)) / 2)
+    elif isinstance(x, Expression):
+        result = default_complexity_function(x.get_head())
+        result = result + sum(default_complexity_function(e) for e in x.elements)
+    else:
+        result = 1
+    #    print(x, result)
+    return result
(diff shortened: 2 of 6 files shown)


rocky:
@mmatera This one  I think should have been merged earlier. I tried to resolve the conflicts (not totally successfully). And the Python files I don't dare touch. If you could do the resolution and commit I'd be grateful.

mmatera:
I can see at this tomorrow

mmatera:
@rocky, now the merge is done, but for some reason, one test fails to pass in Python 3.6.

rocky:
> > @rocky, now the merge is done, but for some reason, one test fails to pass in Python 3.6.
> 
> SymPy 1.9 vs SymPy 1.10?

It definitely is SymPy < 1.10. SymPy 1.10 apparently isn't available in Python 3.6, so that's why Sympy 1.9 is used.

The `test_fullsimplify()` test of `test_algebra.py` test that fails is new even though the place where an exception is raised is very old.  

Something simple to do would be to skip that test if SymPy < 1.10 is installed. The fact that this is on Python 3.6 is secondary - 3.6 forces Sympy < 1.1.0. 

Alternatively we could require SymPy 1.10 or greater but that would also force us to use Python 3.7 or greater. 

Suggestions? 

We should probably check if there was a bug in SymPy for this kind of thing.

mmatera:
> > > @rocky, now the merge is done, but for some reason, one test fails to pass in Python 3.6.
> > 
> > 
> > SymPy 1.9 vs SymPy 1.10?
> 
> It definitely is SymPy < 1.10. SymPy 1.10 apparently isn't available in Python 3.6, so that's why Sympy 1.9 is used.
> 
> The `test_fullsimplify()` test of `test_algebra.py` that fails is new even though the place where an exception is raised is very old.
> 
> Something simple to do would be to skip that test if SymPy < 1.10 is installed. The fact that this is on Python 3.6 is secondary - 3.6 forces Sympy < 1.10.
> 
> Alternatively we could require SymPy 1.10 or greater but that would also force us to use Python 3.7 or greater.
> 
> Suggestions?
> 
> We should probably check if there was a bug in SymPy for this kind of thing.

Probably the best is to put a @xfail decorator to the test. I am going to do that later.

rocky:
> 
> Probably the best is to put a @xfail decorator to the test. I am going to do that later. And thank you for the research!

In my view, the best is to _skip_ the test, not expect a failure of some rather bizarre behavior. To me, an expected failure is used  when you want to check an error condition and that it is caught properly or errored properly.  I don't think we want to check that SymPy of a certain version may have a particular bug. Maybe SymPy will fix this in a future 1.9 backport release and then our expectations are wrong and this will have to be fixed before we can do commits.

On the other hand, skipping implies there is something about this test that we can't get it to work. Maybe the operating system doesn't support a feature, we can't install or we don't have a particular Python package installed, or here, there is a possibly a bug in a package.  By marking this skipped instead of failed, if SymPy changes behavior of 1.9 so this now starts working, okay, we'll still skip it but that's better than failing because we expected SymPy to have a bug that we don't really depend on. 

Also, if you know a test is going to give a traceback that isn't helpful from an user or operations side (i.e. our code isn't trapping the exception and making use of it), then what's the point in doing the work to cause the exception?

rocky:
LGTM - thanks for skipping the test rather than expecting a failure.  I was surprised you didn't use [@skipif](https://docs.pytest.org/en/7.1.x/how-to/skipping.html) as in https://github.com/Mathics3/mathics-core/blob/master/test/test_main.py#L15.

I'll say that personally I have a predilection for simple and generic mechanisms over those that are custom to a particular test framework for some particular versions of Python; pytest didn't always exist, and possibly someday another one will replace it which is equally quirky in using pytest lingo as pytest was with the test framework before it. And this will cause users of pytest to relearn what wasn't broken, as happened when pytest came into existence.

PR #192, opened 2022-03-12
Clarify function previously called "has_changed()"

rocky:
has_changed() is now either uncertain_final_value(), called in do_format()
or uncertain_final_definitions(), called in evaluate().

Document what these do, and motivate why specific constant values are
returned in specific classes.

Having the same name for two slightly different things was confusing.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/192"></a>

### DIFF
--- a/mathics/core/definitions.py
+++ b/mathics/core/definitions.py
@@ -259,7 +259,7 @@ class Definitions(object):
     def is_uncertain_final_value(self, last_evaluated_time: int, symbols: set) -> bool:
         """
         Used in Evaluate_do_format() to
-        determine if we should to (re)evaluate an expression.
+        determine if we should (re)evaluate an expression.
 
         Here, for a definitions object, we check if any symbol in the
         symbols has changed. `last_evaluated_time` indicates when the
--- a/mathics/core/definitions.py
+++ b/mathics/core/definitions.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 
 import pickle
@@ -10,7 +9,7 @@ import bisect
 
 from collections import defaultdict
 
-import typing
+from typing import List, Optional
 
 from mathics.core.atoms import String
 from mathics.core.attributes import nothing
@@ -257,19 +256,30 @@ class Definitions(object):
         for k in self.proxy.pop(tail, []):
             definitions_cache.pop(k, None)
 
-    def has_changed(self, maximum, symbols):
-        # timestamp for the most recently changed part of a given expression.
+    def uncertain_final_value(self, last_evaluated_time: int, symbols: set) -> bool:
+        """
+        Used in Evaluate_do_format() to
+        determine if we should to (re)evaluate an expression.
+
+        Here, for a definitions object, we check if any symbol in the
+        symbols has changed. `last_evaluated_time` indicates when the
+        evaluation started. If a symbol has a time greater than
+        that, then things have changed since the evaluation started
+        and evaluation may lead to a different result.
+        """
         for name in symbols:
-            symb = self.get_definition(name, only_if_exists=True)
-            if symb is None:
-                # symbol doesn't exist so it was never changed
+            symbol = self.get_definition(name, only_if_exists=True)
+            if symbol is None:
+                # "symbol" doesn't exist, so it was never changed.
                 pass
             else:
-                changed = getattr(symb, "changed", None)
-                if changed is None:
-                    # must be system symbol
-                    symb.changed = 0
-                elif changed > maximum:
+                # Get timestamp for the most-recently changed part of the given expression.
+                symbol_change_time = getattr(symbol, "changed", None)
+                if symbol_change_time is None:
+                    # Must be a system symbol that never changes.
+                    # FIXME: couldn't this initially start out 0 so no test is needed?
+                    symbol.change_timestamp = 0
+                elif symbol_change_time > last_evaluated_time:
                     return True
 
         return False
@@ -318,7 +328,7 @@ class Definitions(object):
         accessible_ctxts.add(self.current_context)
         return accessible_ctxts
 
-    def get_matching_names(self, pattern) -> typing.List[str]:
+    def get_matching_names(self, pattern) -> List[str]:
         """
         Return a list of the symbol names matching a string pattern.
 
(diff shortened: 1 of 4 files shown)


rocky:
@TiagoCavalcante @mmatera here is another example of something that on the surface seems like a "what's the big deal in changing names?" kind of thing. 

Actually, I do think it important. As with the confusion at least two of us had with parse.Symbol versus symbol.Symbol, we had to method called "has_changed()" which were slightly different, worked off of different kinds of objects and took different parameters.

At least for me, clarifying what is going on here such as by renaming the functions, make me better understand the need for example for a SymbolConstant class (A Symbol class for symbols that never change like True, False, Null) so we don't have to constantly check if their value or definition can possibly change.

rocky:
And I should mention that down the line, yes, probably the two kinds of Atoms, Symbols, and so on in the parse versus symbol module should be disambiguated since that to has caused confusion in at least two people.

But that is for another time. At least there, we have now documented this kind of thing.

mathics/core/definitions.py
```diff
@@ -257,19 +256,30 @@ def clear_definitions_cache(self, name) -> None:
         for k in self.proxy.pop(tail, []):
             definitions_cache.pop(k, None)
 
-    def has_changed(self, maximum, symbols):
-        # timestamp for the most recently changed part of a given expression.
+    def uncertain_final_value(self, last_evaluated_time: int, symbols: set) -> bool:
```
TiagoCavalcante, line 259: That returns a bool, so maybe `is_uncertain_final_value` be clearer.

mathics/core/symbols.py
```diff
@@ -288,11 +288,16 @@ def get_sort_key(self, pattern_sort=False):
...
-    def has_changed(self, definitions) -> bool:
+    def uncertain_final_definitions(self, definitions) -> bool:
         """
-        Used in Expression.evaluate() to determine if we need to reevaluate
-        an expression. No Atoms need reevaluation. And if this is wrong for a
-        subclass, the subclass should override this method.
+        Used in Expression.do_format() to determine if we may need to
+        (re)evaluate an expression.
+
+        Most Atoms, e.g. Numbers and Strings, need do not need
```
TiagoCavalcante, line 296: need do not need -> do not need

TiagoCavalcante (approved):
Except for the comments, LGTM.

mathics/core/definitions.py
rocky: @TiagoCavalcante  thanks - I will change this in the next commit.

mathics/core/definitions.py
```diff
@@ -257,19 +256,30 @@ def clear_definitions_cache(self, name) -> None:
...
             definitions_cache.pop(k, None)
 
-    def has_changed(self, maximum, symbols):
-        # timestamp for the most recently changed part of a given expression.
+    def is_uncertain_final_value(self, last_evaluated_time: int, symbols: set) -> bool:
+        """
+        Used in Evaluate_do_format() to
+        determine if we should to (re)evaluate an expression.
+
+        Here, for a definitions object, we check if any symbol in the
```
mmatera, line 264: Notice that the definition of a symbol could have changed during a previous evaluation. Let's say
```
session.evaluate("a=4")
session.evaluate("F[a]")
 ```
in the second call, the timestap for the definition of ``a`` was stored during the first call of ``evaluate``.

File: mathics/core/symbols.py [hunk 4]
```diff
@@ -435,16 +440,24 @@ def get_head(self) -> "Symbol":
...
-        Used in Expression.evaluate() to determine if we need to reevaluate
-        an expression.
+        Used in Expression.do_format() to determine if we need to
+        (re)evaluate an expression.
+
+        Here, we have to be pessimistic and return True.
+
+           InputForm[Context[]] == "Global`"
+
+        is an example where we need to evaluate "Global`" and it doesn't start out as a
```
mmatera, line 452: @rocky, do you mean the lhs or the rhs?

mmatera (commented):
LGTM

mathics/core/definitions.py
rocky: I don't understand what you'd like to see done to that comment. It states that there is may be either an evaluation for the first time, evaluate or a reevaluation.

Is the suggestion to add an example in the comment?

mathics/core/symbols.py
rocky: I just checked again, and I see what I have is wrong. The problem is that in evaluating ``$System`Context`` we need another evaluation. 

I will revise the comment in the next iteration.

rocky: @mmatera Ok - revised. See if this is correct now.

mathics/core/symbols.py
```diff
@@ -474,8 +470,21 @@ def is_numeric(self, evaluation=None) -> bool:
...
+        Here, we have to be pessimistic and return True. For example,
+        in:
+
+           Context[]
+
+        this routine will get called where "self" is $System`Context. We
+        can't stop here, but must continue evaluation to get the function's value,
+        such as "Global`".
+        """
+        return True
```
mmatera, line 487: What about checking if ``definitions`` have a definition for  ``self.get_name()`` ?

rocky: At this point in the code we don't know if the Mathics uses typed `Context[]` or `Context`. So I don't know if this routine has enough information to know whether to continue with an evaluation or not.

mmatera: Mathics uses `$Context` (or `Context[]`) for the current context,   and `Context[sym]` for the context of a symbol.

mmatera: But OK, if you want to start a `Definitions` object without the builtins, the best is your implementation: Symbols always should return True. Otherwise, we could have problems.

mmatera: So, I think this PR is ready to merge.

rocky: Ok - we should keep this in mind. In this PR, the main focus was on clarity or describing what the situation currently is as clearly as I can. (And as you see, this is kind of a challenge).

When it comes to improving the code we should keep all of these ideas in mind and discuss them. It is possible in the interim of going over the rest of the code other changes will cause changes in how we approach improvements. 

On the other hand, I _have_ been making little improvements here and there. However hopefully these are not far reaching decisions. They have more been of the type: here we don't return False instead of True, or don't try to sort a list with less than 2 items.

Merging now, then.

PR #193, opened 2022-03-12
Add PredefinedSymbol class...

rocky:
This is like Symbol but we know that the name should never change and therefore we can set `is_uncertain_final_value()` to return False unconditionally.

The rest of the code base might be gone over looking for more PredefinedSymbol possibilities.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/193"></a>

### DIFF
--- a/mathics/builtin/assignments/internals.py
+++ b/mathics/builtin/assignments/internals.py
@@ -174,7 +174,7 @@ def unroll_patterns(lhs, rhs, evaluation):
 
 def unroll_conditions(lhs):
     condition = None
-    if type(lhs) is Symbol:
+    if isinstance(lhs, Symbol):
         return lhs, None
     else:
         name, lhs_elements = lhs.get_head_name(), lhs._elements
@@ -686,7 +686,7 @@ class _SetOperator(object):
     }
 
     def assign_elementary(self, lhs, rhs, evaluation, tags=None, upset=False):
-        if type(lhs) is Symbol:
+        if isinstance(lhs, Symbol):
             name = lhs.name
         else:
             name = lhs.get_head_name()
--- a/mathics/builtin/scoping.py
+++ b/mathics/builtin/scoping.py
@@ -35,7 +35,7 @@ def get_scoping_vars(var_list, msg_symbol="", evaluation=None):
             new_def = var.leaves[1]
             if evaluation:
                 new_def = new_def.evaluate(evaluation)
-        elif var.has_form("Symbol"):
+        elif isinstance(var, Symbol):
             var_name = var.get_name()
             new_def = None
         if not var_name:
--- a/mathics/core/subexpression.py
+++ b/mathics/core/subexpression.py
@@ -143,7 +143,7 @@ class ExpressionPointer(object):
         parent = self.parent
         p = self.position
         if p == 0:
-            if type(parent) is Symbol:
+            if isinstance(parent, Symbol):
                 return parent
             else:
                 return parent.head.copy()
@@ -219,7 +219,7 @@ class SubExpression(object):
         if type(pos) is Integer:
             pos = pos.get_int_value()
         # pos == `System`All`
-        elif type(pos) is Symbol and pos.get_name() == "System`All":
+        elif isinstance(pos, Symbol) and pos.get_name() == "System`All":
             pos = None
         elif type(pos) is Expression:
             if pos.has_form("System`List", None):
--- a/mathics/core/symbols.py
+++ b/mathics/core/symbols.py
@@ -462,16 +462,15 @@ class Symbol(Atom, NumericOperators):
         Used in Expression.do_format() to determine if we need to
         (re)evaluate an expression.
 
-        Here, we have to be pessimistic and return True.
+        Here, we have to be pessimistic and return True. For example,
+        in:
 
-           InputForm[Context[]] == "Global`"
+           Context[]
 
-        is an example where we need to evaluate "Global`" and it doesn't start out as a
-        final value.
+        this routine will get called where "self" is $System`Context. We
+        can't stop here, but must continue evaluation to get the function's value,
+        such as "Global`".
         """
-
-        # FIXME: can we arrange things so that a string like "Global`"
-        # starts out as an evaluated symbol?
         return True
 
     def get_attributes(self, definitions):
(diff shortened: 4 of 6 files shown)


mathics/builtin/assignments/internals.py
```diff
@@ -174,7 +174,7 @@ def unroll_patterns(lhs, rhs, evaluation):
 
 def unroll_conditions(lhs):
     condition = None
-    if type(lhs) is Symbol:
```
rocky, line 177: I believe it is for exactly this kind of change that Python lint checkers tell you not to use 
this idom.

mmatera:
@rocky, actually, just the symbols with attribute ``Locked``  (``True``, ``False``, ``List``, etc) have the behavior of ``Predefined``. Look at the following example:
``
In[1]:={a+b,  Unprotect[Plus];Plus=Q;p+q, Clear[Plus]; r+s} 
Out[1]={Q[a, b], Q[p, q], r + s}
``

rocky:
> @rocky, actually, just the symbols with attribute `Locked` (`True`, `False`, `List`, etc) have the behavior of `Predefined`. Look at the following example: 

> ```
> In[1]:={a+b, Unprotect[Plus];Plus=Q;p+q, Clear[Plus]; r+s} 
> Out[1]={Q[a, b], Q[p, q], r + s}
> ```

I gather that the above is Mathematica's output, Right? 

In Mathics whether in this branch or in master, I get: 

```
{a + b, Q[p, q], Q[r, s]}
```

So could you elaborate on things here? 

We could change PredefinedSymbol to LockedSymbol and defined only LockedSymbols, but since it appears to me that the behavior is the same this doesn't seem reasonable.

mmatera:
> > @rocky, actually, just the symbols with attribute `Locked` (`True`, `False`, `List`, etc) have the behavior of `Predefined`. Look at the following example:
> 
> > ```
> > In[1]:={a+b, Unprotect[Plus];Plus=Q;p+q, Clear[Plus]; r+s} 
> > Out[1]={Q[a, b], Q[p, q], r + s}
> > ```
> 
> I gather that the above is Mathematica's output, Right?
> 

Yes, this is WMA.

> In Mathics whether in this branch or in master, I get:
> 
> ```
> {a + b, Q[p, q], Q[r, s]}
> ```
> 

Interesting. In master, I think the problem is that Clear is not doing its job. For example, if instead of `Clear[Plus]`
 I write ```Plus=.``` I get the expected behaviour:

(In Mathics)
```
In[1]:= {a+b, Unprotect[Plus];Plus=Q;p+q,Plus=.; r+s} 
Out[1]= {a + b, Q[p, q], r + s}
```

> So could you elaborate on what is bothersome or wrong here? And suggest what needs to be done?
> 

What is bothering me is that if something like ``Plus`` is set as ``PredefinedSymbol``, you want to assume that it does not need to be evaluated "again". 

A  similar (and maybe simpler) example would be
```
In[1]:= {Pi, Unprotect[Pi];Pi=4; Pi, Pi=.; Pi }                                 
```
which returns  (both in Mathics and WMA)
``{Pi, Unprotect[Pi];Pi=4; Pi, Pi=.; Pi }``

What happens if ``Pi`` would be set as a ``PredefinedSymbol`` ?


> We could change PredefinedSymbol to LockedSymbol and defined only LockedSymbols, but since it appears to me that the behavior is the same this doesn't seem reasonable.

So, that is exactly the question: is the behavior of a   ``PredefinedSymbol`` the one of a ``LockedSymbol``?

rocky:
> A similar (and maybe simpler) example would be
> 
> ```
> In[1]:= {Pi, Unprotect[Pi];Pi=4; Pi, Pi=.; Pi }                                 
> ```

YAY! You _are_ getting the hang of this. I greatly appreciate the simpler example.

> 
> which returns (both in Mathics and WMA) `{Pi, Unprotect[Pi];Pi=4; Pi, Pi=.; Pi }`
> 
> What happens if `Pi` would be set as a `PredefinedSymbol` ?
> 
> > We could change PredefinedSymbol to LockedSymbol and defined only LockedSymbols, but since it appears to me that the behavior is the same this doesn't seem reasonable.
> 
> So, that is exactly the question: is the behavior of a `PredefinedSymbol` the one of a `LockedSymbol`?

Ok, so I think I now understand. I rely on you however to decide what we should do. This kind of stuff is not something I understand well, and so I don't want to make decisions where I am not proficient and others are.

mmatera:
@rocky, Let's do this: just keep as "Predefined" (it is, not re-evaluable") just the symbols that we are completely sure we shouldn't evaluate: 

- `True` and `False`
- `List` and `Sequence`
- `Symbol`  (I mean, ``Symbol("Symbol")``)

and whatever is tagged with the attribute ``Locked``. We can go over this and iterate as we understand it better.

rocky:
> @rocky, Let's do this: just keep as "Predefined" (it is, not re-evaluable") just the symbols that we are completely sure we shouldn't evaluate:
> 
> * `True` and `False`,
> * `I`,
> * `List`, `Locked`, `Remove`, `Removed`, `Symbol`, `TooBig`,
> * `$Aborted`
> * `$Off`
> 
> and whatever is tagged with the attribute `Locked` and does not have values. We can go over this and iterate as we understand it better.

Ok. Sounds good. For what it is worth, locally I set Pi to be in the PredefinedSymbol class and get the same output:

```
Out[1]= {Pi, 4, Pi}
```

But I don't really care that much about whether we are trying to squeeze out the last little bit of performance here.  What I _am_ interested in doing is clarify, to the extent that we can now, what is going on and what might be the issues.

So we are in agreement here: 

1. Just do a small little bit now. 
1.  See how it goes. 
1.  Think about it and the broader context, 
1. make experiments, and 
1. adjust things down the line after we are in a better position of understanding the bigger picture and the code better.

mmatera (approved):
LGTM

rocky:
Great - let's get this in and iterate. 

I am pleased to report that I think the next series will finally be about one of the things I intended to do - Mark element lists which do not need further evaluation, e.g. "is_final_..." is always true. This happens for example on lists of number in an arithmetic operation (which happens a lot in graphs can plots). 

However I need to take a break and have some other work I promised to get to first.

PR #198, opened 2022-03-14
Reset sessions

mmatera:
In this PR I added a method to ``MathicsSession`` that resets the ``Definitions`` object. This is useful to ensure that different pytests uses a "fresh" Definitions object.   

Also, I have improved the interface and the documentation of ``test.helper.check_evaluation`` to handle tests that compare warning messages and printed outs during the evaluation. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/198"></a>

### DIFF
--- a/test/test_assignment.py
+++ b/test/test_assignment.py
@@ -138,11 +138,8 @@ def test_setdelayed_oneidentity():
 )
 def test_set_and_clear(str_expr, str_expected, msg):
     if str_expr is None:
-        print("Reset session")
         reset_session()
         return
-    session.evaluate("ClearAll[{H, Pi, F, Q, Plus}]")
-    print(session.evaluate("G[a,b]", ""))
     result = session.evaluate(str_expr, "")
     check_evaluation(
         str_expr,
--- a/test/test_assignment.py
+++ b/test/test_assignment.py
@@ -137,11 +137,6 @@ def test_set_and_clear(str_expr, str_expected, msg):
     in a way that the next test run over a fresh
     environment.
     """
-    if str_expr is None:
-        reset_session()
-        return
-    print(session.definitions)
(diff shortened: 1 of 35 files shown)


test/test_assignment.py
```diff
@@ -83,17 +83,42 @@ def test_setdelayed_oneidentity():
             "{F[a, b], H[a, b], F[a, b]}",
             None,
         ),
+        (
```
mmatera, line 86: I added this "idiom" to tell to the test routine that a refresh is needed between tests. Maybe there is a more canonical way to to this, but for now it works. @rocky, @TiagoCavalcante, do you have an idea about how to do this better?

test/test_combinatorica.py
```diff
@@ -1,11 +1,15 @@
 # -*- coding: utf-8 -*-
-from .helper import evaluate, check_evaluation
+from .helper import evaluate, check_evaluation, reset_session
+
+
+def test_00_initialize():
```
mmatera, line 5: This is another thing that I am not sure about how to do. What I want here is that an initializing routine being run just before the tests. In the way that was implemented before, the `evaluate(...)` call was done before all the tests in `mathics-core/test/` be run, so a call to ``session.reset`` unloaded the required package.

rocky: > This is another thing that I am not sure about how to do. What I want here is that an initializing routine being run just before the tests. In the way that was implemented before, the `evaluate(...)` call was done before all the tests in `mathics-core/test/` be run, so a call to `session.reset` unloaded the required package.

https://stackoverflow.com/questions/22627659/run-code-before-and-after-each-test-in-py-test may help

mmatera:
@rocky @TiagoCavalcante I think now this is ready for a final review.

PR #199, opened 2022-03-14
Improving Clear and ClearAll

mmatera:
In this PR I have
* improved the compatibility with WMA in Clear and ClearAll
* added tests for these symbols.
* improved the support for testing warnings and error messages.
* Removing rules for ``Clear[All]`` and ``ClearAll[All]`` that are not standard in WL.
* Raise a warning against attempts to clear  `$Context` or `$ContextPath`.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/199"></a>

### DIFF
--- a/test/helper.py
+++ b/test/helper.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import time
 from mathics.session import MathicsSession
+from typing import Optional
 
 session = MathicsSession()
 
@@ -26,7 +27,7 @@ def check_evaluation(
     to_string_expr: bool = True,
     to_string_expected: bool = True,
     to_python_expected: bool = False,
-    expected_messages=None,
+    expected_messages: Optional[tuple[str]] = None,
 ):
     """
     Helper function to test Mathics expression against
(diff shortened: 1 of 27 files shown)


test/test_combinatorica.py
```diff
@@ -1,11 +1,15 @@
 # -*- coding: utf-8 -*-
-from .helper import evaluate, check_evaluation
+from .helper import evaluate, check_evaluation, reset_session
+
+
+def test_00_initialize():
```
rocky, line 5: If you are expecting this to get run first either because of the position of the test or the name of the test, that is fragile and a bad idea.

mmatera: @rocky, I know that, but was the "less imaginative" way I found to do this. I am going to investigate how to do it properly as soon as I have some time for doing it.

TiagoCavalcante (approved):
LGTM

test/helper.py
```diff
@@ -16,26 +21,67 @@ def evaluate(str_expr: str):
...
-    message="",
-    to_string_expr=True,
-    to_string_expected=True,
-    to_python_expected=False,
+    message: str = "",
+    hold_expected: bool = False,
+    to_string_expr: bool = True,
+    to_string_expected: bool = True,
+    to_python_expected: bool = False,
+    msgs=None,
```
rocky, line 29: Adding `msgs` s a much welcome and needed improvement. 

It would be nice to change the name to something like `messages_expected` make it more distinct from `message`. And maybe `message` should be something like `test_failure_message`.

test/helper.py
```diff
@@ -16,31 +22,83 @@ def evaluate(str_expr: str):
...
-    if message:
+    if failure_message:
         print((result, expected))
-        assert result == expected, message
+        assert result == expected, failure_message
     else:
         print((result, expected))
         assert result == expected
+
+    if expected_messages is not None:
```
mmatera, line 96: here :)

rocky (approved):
LGTM

Issue #205, opened 2022-03-16
Failed inversion of matrix with expression elements


hrueter:
### Observed behavior
Mathics crashes after evaluating the following input for a few minutes
```
In[1]:= Inverse[{{g[0, 0], g[0, 1]}, {g[0, 1], g[1, 1]}}]
```
### Expected behavior
```
In[1]:= Inverse[{{g[0, 0], g[0, 1]}, {g[0, 1], g[1, 1]}}]
Out[1]= {{g[1, 1] / (g[0, 0] g[1, 1] - g[0, 1] ^ 2), -g[0, 1] / (g[0, 0] g[1, 1] - g[0, 1] ^ 2)}, {-g[0, 1] / (g[0, 0] g[1, 1] - g[0, 1] ^ 2), g[0, 0] / (g[0, 0] g[1, 1] - g[0, 1] ^ 2)}}
```
### Background
The program fails within SymPy, which seems to have problems dealing with `SympyExpression` elements. So I don't know if this is a bug in SymPy or if Mathics must provide a different input to the SymPy API.

The following cases work as expected.
Triangular matrix:
```In[1]:= Inverse[{{g[0,0], g[0,1]}, {0, g[1,1]}}]```
Inverting a matrix of symbols:
```In[1]:= Inverse[{{g00, g01}, {g01, g11}}]```

### Setup
Linux Mint 20.3
Mathics 4.0.0 on CPython 3.8.10 (default, Nov 26 2021, 20:14:08)  using SymPy 1.8, mpmath 1.2.1, numpy 1.17.4

mmatera:
Fixed in #206

PR #206, opened 2022-03-16
fixing matrix inverse  (Issue #205)

mmatera:
Sympy  has troubles inverting matrices whose elements are functions. This PR fix the issue.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/206"></a>

### DIFF
--- a/mathics/builtin/numbers/linalg.py
+++ b/mathics/builtin/numbers/linalg.py
@@ -194,8 +194,6 @@ class Inverse(Builtin):
      : The matrix {{1, 0}, {0, 0}} is singular.
      = Inverse[{{1, 0}, {0, 0}}]
 
-    >> Inverse[{{1, 0, 0}, {0, Sqrt[3]/2, 1/2}, {0,-1 / 2, Sqrt[3]/2}}]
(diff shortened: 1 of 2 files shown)


TiagoCavalcante (approved):
Edit: seems this isn't working, I wrote my comment before testing (only look at the code), and seems I can't delete my message.

TiagoCavalcante:
@mmatera what I tested and didn't work also doesn't on Wolfram Cloud, so I guess this shouldn't be a dreaft.

mmatera:
@TiagoCavalcante , what is exactly the issue? Could propose some test cases to check?

TiagoCavalcante:
@mmatera this doesn't work for non-square matrices, although this is [defined](https://en.wikipedia.org/wiki/Generalized_inverse), Wolfram Language doesn't define inverse for non-square matrices, so that isn't a problem.

mmatera:
Yes, WL has also builtins to deal with pseudoinverses and lateral inverses, but ``Inverse`` requires invertible square matrices.

mmatera:
What is still missing here are a couple of pytest lines checking certain simple cases and error messages.

TiagoCavalcante:
@mmatera ok, I'll add that tommorrow.

TiagoCavalcante:
@mmatera I see your tests, I guess they are enough.

Issue #201, opened 2022-03-15
OrderedQ implementation different from Wolfram Language


hrueter:
### Observed behavior
```
In[1]:= OrderedQ[{2,1}]
Out[1]= OrderedQ[{2,1}]
```
### Expected behavior
```
In[1]:= OrderedQ[{2,1}]
Out[1]= False
```
### Background
In WL `OrderedQ` takes a list of elements as input. Instead Mathics implementation of `OrderedQ` expects two arguments, whose order is checked against each other.
### Setup
Linux Mint 20.3
Mathics 4.0.0 on CPython 3.8.10 (default, Nov 26 2021, 20:14:08)  using SymPy 1.8, mpmath 1.2.1, numpy 1.17.4
### Possible Fix
I rewrote the `OrderedQ` class to obtain the behavior expected from above. However this still compares only the first two elements in the list, where Mathematica would check all of them.
https://github.com/Mathics3/mathics-core/blob/9eb8b1a47e69eea40c4ecf9532525649c0163e53/mathics/builtin/structure.py#L291

```
class OrderedQ(Builtin):
    """
    <dl>
    <dt>'OrderedQ[{$a$, $b$}]'
        <dd>is 'True' if $a$ sorts before $b$ according to canonical
        ordering.
    </dl>

    >> OrderedQ[{a, b}]
     = True
    >> OrderedQ[{b, a}]
     = False
    """

    def apply(self, expr, evaluation):
        "OrderedQ[expr_]"

        if expr.leaves[0] <= expr.leaves[1]:
            return SymbolTrue
        else:
            return SymbolFalse
```

adamantinum:
@hrueter adding a loop to your fix should have implemented the function correctly

TiagoCavalcante:
@hrueter this was fixed in #207 (thanks @adamantinum!)

PR #207, opened 2022-03-16
Fix behaviour of OrderedQ 

adamantinum:
Fix issue https://github.com/Mathics3/mathics-core/issues/201 changing argument of the function to list

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/207"></a>

### DIFF
--- a/mathics/builtin/structure.py
+++ b/mathics/builtin/structure.py
@@ -291,24 +291,26 @@ class PatternsOrderedQ(Builtin):
 class OrderedQ(Builtin):
     """
     <dl>
-    <dt>'OrderedQ[$a$, $b$]'
+    <dt>'OrderedQ[{$a$, $b$}]'
         <dd>is 'True' if $a$ sorts before $b$ according to canonical
         ordering.
     </dl>
 
-    >> OrderedQ[a, b]
+    >> OrderedQ[{a, b}]
      = True
-    >> OrderedQ[b, a]
+    >> OrderedQ[{b, a}]
      = False
     """
 
-    def apply(self, e1, e2, evaluation):
-        "OrderedQ[e1_, e2_]"
+    def apply(self, expr, evaluation):
+        "OrderedQ[expr_]"
 
-        if e1 <= e2:
-            return SymbolTrue
-        else:
-            return SymbolFalse
+        for index, value in enumerate(expr.leaves[:-1]):
+            if expr.leaves[index] <= expr.leaves[index + 1]:
+                continue
+            else:
+                return SymbolFalse
+        return SymbolTrue
 
 
 class Order(Builtin):


mathics/builtin/structure.py
```diff
@@ -291,24 +291,27 @@ def apply(self, p1, p2, evaluation):
...
-    def apply(self, e1, e2, evaluation):
-        "OrderedQ[e1_, e2_]"
+    def apply(self, expr, evaluation):
+        "OrderedQ[expr_]"
 
-        if e1 <= e2:
-            return SymbolTrue
-        else:
-            return SymbolFalse
+        for i in range(1, len(expr.leaves)):
```
TiagoCavalcante, line 308: This idiom is not recommended in Python, could you replace it with [enumerate](https://realpython.com/python-enumerate/)?

TiagoCavalcante (changes requested):
LGTM, but this should be backward compatible. Could you add an "if there is only 1 leave and that is a list then [new behavior] else [old behavior]"?

adamantinum:
> LGTM, but this should be backward compatible. Could you add an "if there is only 1 leave and that is a list then [new behavior] else [old behavior]"?

Rewritten with enumerate.
Regarding the other point I have a doubt: the old behaviour is not a correct statement in mathematica (i.e.: `Nonatomic expression expected at position 1 in OrderedQ[a, b].`), is it ok?

TiagoCavalcante:
> > LGTM, but this should be backward compatible. Could you add an "if there is only 1 leave and that is a list then [new behavior] else [old behavior]"?
> 
> Rewritten with enumerate. Regarding the other point I have a doubt: the old behaviour is not a correct statement in mathematica (i.e.: `Nonatomic expression expected at position 1 in OrderedQ[a, b].`), is it ok?

Yes, I'll merge this.

TiagoCavalcante (approved):

adamantinum:
> > > LGTM, but this should be backward compatible. Could you add an "if there is only 1 leave and that is a list then [new behavior] else [old behavior]"?
> > 
> > 
> > Rewritten with enumerate. Regarding the other point I have a doubt: the old behaviour is not a correct statement in mathematica (i.e.: `Nonatomic expression expected at position 1 in OrderedQ[a, b].`), is it ok?
> 
> Yes, I'll merge this.

Thus no need to maintain the old behaviour?

TiagoCavalcante:
@adamantinum no, that is very weird and wrong, that receives 2 parameters, and as its name isn't Ordered**Pair** that is just wrong, you should replace that with a LessEqual sign in those cases.

adamantinum:
> @adamantinum no, that is very weird and wrong, that receives 2 parameters, and as its name isn't Ordered**Pair** that is just wrong, you should replace that with a LessEqual sign in those cases.
I agree!

PR #211, opened 2022-03-17
Docs/tests for Equal, Unequal, Less, Greater

rocky:
Move unit tests out of class docstring into test_compare. Go over documentation for Equal, Less, Greater

I am pleased to see that we have now advanced to the level where we can fully remove those non-document tests fully out of docs where it never belonged since we now have the technology in place to this the right and normal way.

Now what remains is a lot of cleanup work (partly made worse because we pushed using incorrect technology rather than fix the technology)

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/211"></a>

### DIFF
--- a/test/test_compare.py
+++ b/test/test_compare.py
@@ -6,13 +6,7 @@ import pytest
 @pytest.mark.parametrize(
     ("str_expr", "str_expected", "fail_msg"),
     [
-        #        (None, None, None,),
         # Equal (==)
-        (
-            "ClearAll[g,a,b];",
-            "Null",
-            None,
-        ),
         ("g[2]==g[3]", "g[2] == g[3]", "not comparable expressions, Issue #200"),
         ("g[a]==g[3]", "g[a] == g[3]", "not comparable expressions"),
         ("g[2]==g[a]", "g[2] == g[a]", "not comparable expressions"),
--- a/test/test_compare.py
+++ b/test/test_compare.py
@@ -7,6 +7,11 @@ import pytest
     ("str_expr", "str_expected", "fail_msg"),
     [
         # Equal (==)
+        (
+            "ClearAll[g,a,b];",
+            "Null",
+            None,
+        ),
         ("g[2]==g[3]", "g[2] == g[3]", "not comparable expressions, Issue #200"),
         ("g[a]==g[3]", "g[a] == g[3]", "not comparable expressions"),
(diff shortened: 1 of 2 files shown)


test/test_compare.py
```diff
@@ -7,7 +7,7 @@
     ("str_expr", "str_expected", "fail_msg"),
     [
         #        (None, None, None,),
```
mmatera, line 9: Now we can use this line instead of line 11

mmatera (approved):
LGTM

test/test_compare.py
```diff
@@ -19,27 +18,58 @@
...
+        ("{1, 2, 3} < {1, 2, 3}", "{1, 2, 3} < {1, 2, 3}", "Less on a list"),
         ("g[2]<g[3]", "g[2] < g[3]", "not comparable expressions, Issue #200"),
         ("g[a]<g[3]", "g[a] < g[3]", "not comparable expressions"),
         ("g[2]<g[a]", "g[2] < g[a]", "not comparable expressions"),
         ("g[a]<g[b]", "g[a] < g[b]", "not comparable expressions"),
         ("g[a]<g[a]", "g[a] < g[a]", "not comparable expressions (like in WMA)"),
         ("g[1]<g[1]", "g[1] < g[1]", "not comparable expressions (like in WMA)"),
+        #
+        # chained compare
+        ("a != a != b", "False", "Strange MMA behavior"),
```
TiagoCavalcante, line 71: Not sure if this is strange behaviour, Python kinda works the same way (but without undefined symbol comparison):
```wl
a != a != b
a != a && a != b
False && a != b
False
```

TiagoCavalcante (approved):
LGTM - I like the documentation changes.

test/test_compare.py
```diff
@@ -7,6 +7,11 @@
     ("str_expr", "str_expected", "fail_msg"),
     [
         # Equal (==)
+        (
+            "ClearAll[g,a,b];",
```
TiagoCavalcante, line 11: I have looked at your old commit and seems you missed a (None, None, None) I guess that is why it wasn't working.

test/test_compare.py
```diff
@@ -6,13 +6,7 @@
 @pytest.mark.parametrize(
     ("str_expr", "str_expected", "fail_msg"),
     [
-        #        (None, None, None,),
```
TiagoCavalcante, line 9: This should be uncommented, not deleted.

Issue #115, opened 2022-01-15
add Diagonal


Kreijstal:
This returns a list of the diagonal entries in a square matrix.

rocky:
I wonder if this couldn't be written in WL/Mathics using the existing language? If so, if you write this, we could arrange to have it loaded in on startup.

Kreijstal:
I am not familiar with Mathics, where is the startup file?
I am not familiar enough with Mathematica error handling, so this is my attempt at trying to do this on WL
```mathematica
Diagonal[matrix_] := 
 If[Head[matrix] == List, 
  Map[matrix[[#]][[#]] &, 
   Range[Max[Length[matrix], Length[matrix[[1]]]]]], 
  "I do not know how to do Error handling"]
  ```

rocky:
https://mathics-development-guide.readthedocs.io/en/latest/extending/autoload.html as information on how setting get read in. I see though that currently we don't allow defining global functions this way. 

https://mathics-development-guide.readthedocs.io/en/latest/extending/packages.html has information about packages. 

I am not well versed on the Wolfram Language either, so I'd have to google or use one of many resources available for learning how to do this. But I'd rather you teach me than the other way around.

rocky:
Fixed in #215  - thanks @adamantium

PR #215, opened 2022-03-18
Add builtin Diagonal

adamantinum:
I tried to implement Diagonal and it seems working.
In https://github.com/Mathics3/mathics-core/issues/115 is proposed to use native Wolfram Language, but I hope this it will be sufficient, at least for the moment.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/215"></a>

### DIFF
--- a/mathics/builtin/matrices/partmatrix.py
+++ b/mathics/builtin/matrices/partmatrix.py
@@ -1,0 +1,71 @@
+# -*- coding: utf-8 -*-
+
+"""
+Parts of Matrices
+
+Set of methods for manipulating Matrices.
+"""
+
+
+from mathics.builtin.base import Builtin
+from mathics.core.expression import Expression
+from mathics.core.symbols import SymbolList
+
+
+class Diagonal(Builtin):
+    """
+    <dl>
+        <dt>'Diagonal[$m$]'
+        <dd>gives a list with the values in the diagonal of the matrix $m$.
+        <dt>'Diagonal[$m$, $k$]'
+        <dd>gives a list with the values in the $k$ diagonal of the matrix $m$.
+    </dl>
+
+    >> Diagonal[{{1, 2, 3}, {4, 5, 6}, {7, 8, 9}}]
+     = {1, 5, 9}
+    >> Diagonal[{{1, 2, 3}, {4, 5, 6}, {7, 8, 9}}, 1]
+     = {2, 6}
+    """
+
+    rules = {
+        "Diagonal[expr_]": "Diagonal[expr, 0]",
+    }
+
+    summary_text = "Return a list with the diagonal elements of a given matrix"
+
+    def apply(self, expr, diag, evaluation):
+        "Diagonal[expr_List, diag_Integer]"
+
+        result = []
+
+        for count, value in enumerate(expr.elements, start=diag.value):
+            if not hasattr(value, "elements") or len(value.elements) <= count:
+                break
+            if count < 0:
+                continue
+            result.append(value.elements[count])
+        return Expression(SymbolList, *result)
+
+
+class MatrixQ(Builtin):
+    """
+    <dl>
+    <dt>'MatrixQ[$m$]'
+        <dd>returns 'True' if $m$ is a list of equal-length lists.
+    <dt>'MatrixQ[$m$, $f$]'
+        <dd>only returns 'True' if '$f$[$x$]' returns 'True' for each
+        element $x$ of the matrix $m$.
+    </dl>
+
+    >> MatrixQ[{{1, 3}, {4.0, 3/2}}, NumberQ]
+     = True
+    """
+
+    rules = {
+        "MatrixQ[expr_]": "ArrayQ[expr, 2]",
+        "MatrixQ[expr_, test_]": "ArrayQ[expr, 2, test]",
+    }
+
+    summary_text = (
+        "Return 'True' if the given argument is a list of equal-length lists."
+    )
--- a/mathics/builtin/tensors.py
+++ b/mathics/builtin/tensors.py
@@ -445,26 +445,6 @@ class Outer(Builtin):
         return rec(lists[0], lists[1:], [])
 
 
-class MatrixQ(Builtin):
-    """
-    <dl>
-    <dt>'MatrixQ[$m$]'
-        <dd>returns 'True' if $m$ is a list of equal-length lists.
-    <dt>'MatrixQ[$m$, $f$]'
-        <dd>only returns 'True' if '$f$[$x$]' returns 'True' for each
-        element $x$ of the matrix $m$.
-    </dl>
-
-    >> MatrixQ[{{1, 3}, {4.0, 3/2}}, NumberQ]
-     = True
-    """
-
-    rules = {
-        "MatrixQ[expr_]": "ArrayQ[expr, 2]",
-        "MatrixQ[expr_, test_]": "ArrayQ[expr, 2, test]",
-    }
-
-
 class RotationTransform(Builtin):
     """
     <dl>
--- a/mathics/builtin/__init__.py
+++ b/mathics/builtin/__init__.py
@@ -160,6 +160,7 @@ for subdir in (
     "files_io",
     "intfns",
     "list",
+    "matrices",
     "moments",
     "numbers",
     "specialfns",
--- a/mathics/builtin/matrices/__init__.py
+++ b/mathics/builtin/matrices/__init__.py
@@ -1,0 +1,7 @@
+"""
+Matrices and Linear Algebra
+
+Construction and manipulation of Matrices.
+"""
+
+from mathics.version import __version__  # noqa used in loading to check consistency.
--- a/setup.py
+++ b/setup.py
@@ -164,6 +164,7 @@ setup(
         "mathics.builtin.fileformats",
         "mathics.builtin.files_io",
         "mathics.builtin.intfns",
+        "mathics.builtin.matrices",
         "mathics.builtin.moments",
         "mathics.builtin.numbers",
         "mathics.builtin.numpy_utils",


rocky (changes requested):
As this project grows larger, I believe we have to be more stringent and mindful of quality and in some cases this even means improving upon what we currently have. 

The code base is already large and in need of disorganization.

So some specifics here. I do not see a class field for `summary_text`.  I added that to the documentation a while back when someone else doing a similar activity wanted to add a new function. See https://mathics-development-guide.readthedocs.io/en/latest/extending/developing-code/extending/tutorial/2-help-markup.html

Also, the term "leaves" is no longer used. We should be using "elements" which is the WL term. A leaf of a tree or expression is different than an element of a tree or expression.  See https://mathics-development-guide.readthedocs.io/en/latest/extending/code-overview/glossary.html?highlight=glossary#term-element

https://reference.wolfram.com/language/ref/Diagonal.html lists two forms for Diagonal and this just does one of them. This is not a showstopper if it were important to get in Diagonal which I am not certain is the situation. 

As for the example, does this give the same result as https://reference.wolfram.com/language/ref/Diagonal.html on a non-square matix? 

Last and more important is where this is put. Since there are a large number of functions, I have been trying to organize functions along the classification in the Wolfram Language Docs. There this is listed as a function in the group [Parts of Matrices](https://reference.wolfram.com/language/guide/PartsOfMatrices.html)

This group which would a new module or file rather than inside "tensor" and those builtins inside "tensor" that are are not listed in [Tensors](https://reference.wolfram.com/language/guide/Tensors.html) such as MatrixQ and DiagonalMatrix would get summary text and would be moved to this new module

adamantinum:
> As this project grows larger, I believe we have to be more stringent and mindful of quality and in some cases this even means improving upon what we currently have.
> 
> The code base is already large and in need of disorganization.
> 
> So some specifics here. I do not see a class field for `summary_text`. I added that to the documentation a while back when someone else doing a similar activity wanted to add a new function. See https://mathics-development-guide.readthedocs.io/en/latest/extending/developing-code/extending/tutorial/2-help-markup.html
> 
> Also, the term "leaves" is no longer used. We should be using "elements" which is the WL term. A leaf of a tree or expression is different than an element of a tree or expression. See https://mathics-development-guide.readthedocs.io/en/latest/extending/code-overview/glossary.html?highlight=glossary#term-element
> 
> https://reference.wolfram.com/language/ref/Diagonal.html lists two forms for Diagonal and this just does one of them. This is not a showstopper if it were important to get in Diagonal which I am not certain is the situation.
> 
> As for the example, does this give the same result as https://reference.wolfram.com/language/ref/Diagonal.html on a non-square matix?
> 
> Last and more important is where this is put. Since there are a large number of functions, I have been trying to organize functions along the classification in the Wolfram Language Docs. There this is listed as a function in the group [Parts of Matrices](https://reference.wolfram.com/language/guide/PartsOfMatrices.html) not Tensors.
> 
> This group which would a new module or file rather than inside "tensor" and those builtins inside "tensor" that are are not listed in [Tensors](https://reference.wolfram.com/language/guide/Tensors.html) such as MatrixQ and DiagonalMatrix would get summary text and would be moved to this new module



> in need of disorganization

I completely agree on the necessity to reorganize code meanwhile it grows and I've tought for some days how to do it. 
I think you're right in following WL Docs, I suggest to create for each group a directory, rather than a file, and store each function of the group in its own file inside that. I find this tidier than having more function in a unique file. Maybe this can be implemented gradually in a separate branch.

Regarding Diagnal, I have checked with non-square matrices too, now I'm going to implement the missing parts.

mmatera:
In the line of @rocky 's comments, this routine belongs to the class of "Parts of Matrices" functions https://reference.wolfram.com/language/guide/PartsOfMatrices.html
Here we have these routines in the `mathics.builtin.list.eol` module. Maybe we can split this module accordingly.

Regarding @adamantinum comment of having a module for symbol, I think that would be too much: usually many symbols requires just few lines for its implementation. And if more core is necessary for an specific case, we can delegate the logic into another module on `mathics.algorithm`, as it was done for example in ``Part``.

rocky:
> I completely agree on the necessity to reorganize code meanwhile it grows and I've tought for some days how to do it. I think you're right in following WL Docs, I suggest to create for each group a directory, rather than a file, and store each function of the group in its own file inside that. I find this tidier than having more function in a unique file. Maybe this can be implemented gradually in a separate branch.


In addition to what mmatera reports, unfortunetaly, right now the documentation system is tied module system and we have a fixed number of levels. Which brings me to the point that "Parts of Matrices" is really part of [Matrices and Linear Algebra](https://reference.wolfram.com/language/guide/MatricesAndLinearAlgebra.html) which we don't currently have. That would then be a directory under builtins as you'll find for say "Integer Functions". In addition to creating the directory and its `__init__.py` (look at how others are done) there is some additional code needed to hook that into the system since this isn't completely auto-detected.

> 
> Regarding Diagnal, I have checked with non-square matrices too, now I'm going to implement the missing parts.

From my standpoint instead of extending Mathics like adding new primatives that are incomplete and will require people going over, a great activity for new people interested in the project is to help out cleaning it up. 

For example adding `summary_text` in those places where we don't have them on built-in function. And as mmatera suggests moving builtins to the proper place. In particular  "List Functions - Miscellaneous" has a number of builtin functions like this. I ran out of time in doing the work to move this to its proper place after having done a dozens of others.

It will also get you more familiar with the code.

Your attention to detail is appreciated.

rocky:
One other small thing - it helps to try things out when the pull request is from a _branch_ not master. See https://yangsu.github.io/pull-request-tutorial/

mmatera:
In my opinion, it is fine that if someone wants to contribute with some builtins just do it, instead of completing the existing ones. The condition I would put to accept a PR over Builtins are

- [ ] Fulfills the WL standard
- [ ] be properly formatted (having docstrings, summary, etc), 
- [ ] located in the right place
- [ ] having some meaningful tests.
- [ ] Passes all the tests.

mmatera:
Regarding the first point,

```
In[1]:= ?? DiagonalMatrix                                                       

Out[1]=
DiagonalMatrix[list] gives a matrix with the elements of list on the leading diagonal, and zero elsewhere. 

DiagonalMatrix[list, k] gives a matrix with the elements of list on the kth   diagonal.DiagonalMatrix[list, k, n] pads with zeros to create an n×n matrix.
        Attributes[DiagonalMatrix]={Protected}

```
while 

```
In[2]:= ?? Diagonal                                                             

Out[2]= 
Diagonal[m] gives the list of elements on the leading diagonal of the matrix m.
Diagonal[m, k] gives the elements on the k  diagonal of m.
        Attributes[Diagonal]={Protected}
```

So what is implemented here (partially) is ``Diagonal`` not ``DiagonalMatrix``.

TiagoCavalcante:
@rocky I'm for adding the pull request template, I'm going to add one to mathics-threejs-backend too.

rocky:
> @rocky I'm for adding the pull request template, I'm going to add one to mathics-threejs-backend too.

A number of those check items might not be applicable for backend. One that should be added in my opinion is that the pull request is not from a master (or main) branch.

adamantinum:
> > I completely agree on the necessity to reorganize code meanwhile it grows and I've tought for some days how to do it. I think you're right in following WL Docs, I suggest to create for each group a directory, rather than a file, and store each function of the group in its own file inside that. I find this tidier than having more function in a unique file. Maybe this can be implemented gradually in a separate branch.
> 
> In addition to what mmatera reports, unfortunetaly, right now the documentation system is tied module system and we have a fixed number of levels. Which brings me to the point that "Parts of Matrices" is really part of [Matrices and Linear Algebra](https://reference.wolfram.com/language/guide/MatricesAndLinearAlgebra.html) which we don't currently have. That would then be a directory under builtins as you'll find for say "Integer Functions". In addition to creating the directory and its `__init__.py` (look at how others are done) there is some additional code needed to hook that into the system since this isn't completely auto-detected.
> 

So, to fix the PR, should I create a directory called e.g. mathics/builtin/matrices with its own `__init__.py`, move there the functions of Matrix and Linear Algebra group, and add the new directory to `mathics/builtin/__init__.py`?

> > Regarding Diagnal, I have checked with non-square matrices too, now I'm going to implement the missing parts.
> 
> From my standpoint instead of extending Mathics like adding new primatives that are incomplete and will require people going over, a great activity for new people interested in the project is to help out cleaning it up.
> 
> For example adding `summary_text` in those places where we don't have them on built-in function. And as mmatera suggests moving builtins to the proper place. In particular "List Functions - Miscellaneous" has a number of builtin functions like this. I ran out of time in doing the work to move this to its proper place after having done a dozens of others.
> 
> It will also get you more familiar with the code.
> 
> Your attention to detail is appreciated.

Okay thanks, I'll work on that for my future PRs.

rocky:
> > > I completely agree on the necessity to reorganize code meanwhile it grows and I've tought for some days how to do it. I think you're right in following WL Docs, I suggest to create for each group a directory, rather than a file, and store each function of the group in its own file inside that. I find this tidier than having more function in a unique file. Maybe this can be implemented gradually in a separate branch.
> > 
> > 
> > In addition to what mmatera reports, unfortunetaly, right now the documentation system is tied module system and we have a fixed number of levels. Which brings me to the point that "Parts of Matrices" is really part of [Matrices and Linear Algebra](https://reference.wolfram.com/language/guide/MatricesAndLinearAlgebra.html) which we don't currently have. That would then be a directory under builtins as you'll find for say "Integer Functions". In addition to creating the directory and its `__init__.py` (look at how others are done) there is some additional code needed to hook that into the system since this isn't completely auto-detected.
> 
> So, to fix the PR, should I create a directory called e.g. mathics/builtin/matrices with its own `__init__.py`, move there the functions of Matrix and Linear Algebra group, 

Yes.

> and add the new directory to `mathics/builtin/__init__.py`?
> 

Yes, but also look in and change `setup.py`.
So

rocky:
One other thing to note. Several "Guides" may refer to the same built-in function. If the function has already been broken out into a sublevel or is covered by a "Guide" and that position is valid (that is, listed in a some guide on the Mathematica site) I wouldn't move it. 

Instead we just have to live with the fact that from a documentation perspective things aren't a strict hierarchy, but from the document it has to be like that until we revise it using something like sphinx and autodoc/autotest which might have a way to cross link better. But this is a big undertaking.

rocky:
> > One other thing to note. Several "Guides" may refer to the same built-in function. If the function has already been broken out into a sublevel or is covered by a "Guide" and that position is valid (that is, listed in a some guide on the Mathematica site) I wouldn't move it.
> > Instead we just have to live with the fact that from a documentation perspective things aren't a strict hierarchy, but from the document it has to be like that until we revise it using something like sphinx and autodoc/autotest which might have a way to cross link better. But this is a big undertaking.
> 
> Functions such as get_dimensions() defined in `builtin/tensors.py` but used in `builtin/matrices/` (e.g. Dimensions) are to keep in tensors and import in matrices or vice versa?

If you look at the Guides you will see `Dimensions` is listed both places. In Matrices and Linear Algebra it is in fact in the subsection "Parts of Matrices".

The simplest things is to keep things as they are, although a case could be made for putting it in Parts of Matrices since that is the lowest level at which it seems to be needed. 

However In my opinion there is a more fundamental organizational problem in that such helper kinds of things of a more general nature and that may be applicable to several builtins should be in totally separate modules from the boilerplate class code that is used to implement the builtin functions. 

But that's another battle for another time.

adamantinum:
I created the new group, moved there MatrixQ, and implemented Diagonal.
Moreover I noticed that ">>" in docstrings prevents `Information[]` displaying the text correctly, so I replaced the command line in the examples with that one of Mathics `In[1]:=` and `Out[1]:=`

TiagoCavalcante:
@adamantinum you can write this with the `>> ` at the start, `Information[]` is another issue.

TiagoCavalcante:
@adamantinum just to clarify why the `>> ` is important: with that in Mathics Django we can press a button to run the snippet.

mathics/builtin/tensors.py
```diff
@@ -89,10 +89,10 @@ class ArrayDepth(Builtin):
         'Length[Dimensions[$a$]]'.
     </dl>
 
-    >> ArrayDepth[{{a,b},{c,d}}]
-     = 2
-    >> ArrayDepth[x]
-     = 0
+    In[1]:= ArrayDepth[{{a,b},{c,d}}]
```
TiagoCavalcante, line 92: Still leaking this part.

adamantinum:
> @adamantinum you can write this with the `>> ` at the start, `Information[]` is another issue.

If I writea line with `>>` the text in that line and in the following is not displayed. If it's another issue I'll remove the commit.

adamantinum:
> @adamantinum just to clarify why the `>> ` is important: with that in Mathics Django we can press a button to run the snippet.

Ok I remove the commit

TiagoCavalcante:
> > @adamantinum you can write this with the `>> ` at the start, `Information[]` is another issue.
> 
> If I write a line with `>>`, the text in that line and in the following ones is not displayed calling `?NameOfMethod`. If it's another issue I'll remove the commit.

That is the expected behavior, `?` gives the short description, `??`` gives the full description.

adamantinum:
> > > @adamantinum you can write this with the `>> ` at the start, `Information[]` is another issue.
> > 
> > 
> > If I write a line with `>>`, the text in that line and in the following ones is not displayed calling `?NameOfMethod`. If it's another issue I'll remove the commit.
> 
> That is the expected behavior, `?` gives the short description, `??` gives the full description.

It doesn't work with ?? too.
However, I reverted the commmit.

mathics/builtin/matrices/partmatrix.py
```diff
@@ -0,0 +1,72 @@
...
+from mathics.builtin.base import Builtin
+from mathics.builtin.tensors import get_dimensions
+from mathics.core.expression import Expression
+from mathics.core.atoms import Integer
+
+
+class Diagonal(Builtin):
+    """
+    <dl>
+        <dt>'DiagonalMatrix[$m$]'
```
mmatera, line 19: <dt>Diagonal[$m$]</dt>

mathics/builtin/matrices/partmatrix.py
```diff
@@ -0,0 +1,72 @@
...
+        <dd>gives a list with the values in the diagonal of the matrix $m$.
+    </dl>
+
+    >> Diagonal[{{1, 2, 3}, {4, 5, 6}, {7, 8, 9}}]
+     = {1, 5, 9}
+
+    <dl>
```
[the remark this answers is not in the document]
adamantinum: Ok, I grouped the two `<dl>`s.
Regarding the wrong behavior of `??Method`, should I open a new issue or is there already one?

mathics/builtin/matrices/partmatrix.py
adamantinum: Oh... fixed

mathics/builtin/matrices/partmatrix.py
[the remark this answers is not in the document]
TiagoCavalcante: > Ok, I grouped the two `<dl>`s. Regarding the wrong behavior of `??Method`, should I open a new issue or is there already one?

@adamantinum I have checked it in Wolfram Cloud, the behavior there is the same as our current behavior, I'm not sure if that is desired though as I already have seen documentation in "chunks" between the code examples, so yes, you can open an issue.

mathics/builtin/matrices/partmatrix.py
```diff
@@ -0,0 +1,71 @@
...
+
+"""
+Parts of Matrices
+
+Set of methods for manipulating Matrices.
+"""
+
+
+from mathics.builtin.base import Builtin
+from mathics.builtin.tensors import get_dimensions
```
TiagoCavalcante, line 11: This import isn't used.

mathics/builtin/matrices/partmatrix.py
```diff
@@ -0,0 +1,71 @@
...
+from mathics.core.expression import Expression
+from mathics.core.atoms import Integer
```
TiagoCavalcante, line 13: Neither is this one.

mathics/builtin/matrices/partmatrix.py
```diff
@@ -0,0 +1,71 @@
...
+        <dt>'Diagonal[$m$]'
+        <dd>gives a list with the values in the diagonal of the matrix $m$.
+        <dt>'Diagonal[$m$, $k$]'
+        <dd>gives a list with the values in the $k$ diagonal of the matrix $m$.
+    </dl>
+
+    >> Diagonal[{{1, 2, 3}, {4, 5, 6}, {7, 8, 9}}]
+     = {1, 5, 9}
+
+
```
TiagoCavalcante, line 28: 3 blank lines here, it should be 0-1.

mathics/builtin/matrices/partmatrix.py
```diff
@@ -0,0 +1,71 @@
...
+        <dd>returns 'True' if $m$ is a list of equal-length lists.
+    <dt>'MatrixQ[$m$, $f$]'
+        <dd>only returns 'True' if '$f$[$x$]' returns 'True' for each
+        element $x$ of the matrix $m$.
+    </dl>
+
+    >> MatrixQ[{{1, 3}, {4.0, 3/2}}, NumberQ]
+     = True
+    """
+
```
TiagoCavalcante, line 67: It is leaking a `summary_text` here, I know it is not your code, but we are trying to improve the code, it is hard to add a summary text to 1000 builtins at once so we are doing it bit by bit.

mathics/builtin/matrices/partmatrix.py
```diff
@@ -0,0 +1,71 @@
...
+        result = []
+
+        for count, value in enumerate(expr.elements, start=diag.value):
+            if not hasattr(value, "elements") or len(value.elements) <= count:
+                break
+            if count < 0:
+                continue
+            result.append(value.elements[count])
+        return Expression("List", *result)
```
TiagoCavalcante, line 51: Use `Expression(SymbolList, *result)` here instead.

Import `SymbolList` from `mathics.core.symbols`.

mathics/builtin/matrices/partmatrix.py
```diff
@@ -0,0 +1,71 @@
...
+
+    summary_text = "Return a list with the diagonal elements of a given matrix"
+
+    def apply(self, expr, diag, evaluation):
+        "Diagonal[expr_List, diag_Integer]"
+
+        result = []
+
+        for count, value in enumerate(expr.elements, start=diag.value):
+            if not hasattr(value, "elements") or len(value.elements) <= count:
```
TiagoCavalcante, line 46: Use `has_form` instead:

```py
if not value.has_form("List", count):
```

mathics/builtin/matrices/partmatrix.py
```diff
@@ -0,0 +1,71 @@
...
+                break
+            if count < 0:
```
TiagoCavalcante, line 48: Why would the index be negative?

adamantinum: If you want to get a diagonal lower than the leading one you need to pass a negative `diag`.
It's the same behaviour of Mathematica; the if statement is needed because e.g. `Diagonal[matrix, -1] ` would return a list starting with the last element of the first row

mathics/builtin/matrices/partmatrix.py
adamantinum: With `has_form` it works only removing `not`, but then if the value of `diag` is greater than the available indices, mathics crashes.
So i should write
`if not value.has_form("List", count) or len(value.elements) <= count:`

TiagoCavalcante: You can keep this as is in this case.

adamantinum: Ok, I fixed the other things.

rocky (approved):
LGTM - thanks for hanging in there and making the changes. 

Some things I notice in case you want to add in a future one.

Please create the branch off of some name other than "master" in you local repository. (It should also help you as things change). 

Now that "Matrices and Linear Algebra" had been created "Constructing Matrices" can be added to that and IdentityMatrix and DiagonalMatrix would be moved there with summary_text (which in the future I guess will be called usage_text, but that's a separate process.

adamantinum:
> LGTM - thanks for hanging in there and making the changes.
> 
> Some things I notice in case you want to add in a future PR.
> 
> Please create the branch off of some name other than "master" in you local repository. (It should also help you as things change).
> 
> Now that "Matrices and Linear Algebra" had been created "Constructing Matrices" can be added to that and `IdentityMatrix` and `DiagonalMatrix` would be moved out of Tensors and in there with `summary_text` (which in the future I guess will be called usage_text, but that's a separate process.

Thank you for your advice.

Regarding summary_text, I noticed it's not displayed when `Method::usage is called`. Is it an already known issue?

rocky:
Not sure what you are talking about:

```
$ mathics

Mathics 4.1.0.dev0
on CPython 3.8.12 (default, Oct 21 2021, 14:26:19) 
using SymPy 1.10, mpmath 1.2.1, numpy 1.21.3

under certain conditions.

In[1]:= ?Diagonal
Return a list with the diagonal elements of a given matrix
Out[1]= Null
```

adamantinum:
> 

Well, `?Diagonal` works, but `Diagonal::usage` does not

```
In[1]:= ?Diagonal
Return a list with the diagonal elements of a given matrix
Out[1]:= Null

In[2]:= Diagonal::usage
Out[2]:= Diagonal::usage

```

rocky:
Ok - thanks for the more detailed information. That is something that needs to be addressed. Feel free to open a new issue or to work on this.

adamantinum:
Ok I'm going to open the issue, when I can I'll try to check.

mmatera:
> > LGTM - thanks for hanging in there and making the changes.
> > Some things I notice in case you want to add in a future PR.
> > Please create the branch off of some name other than "master" in you local repository. (It should also help you as things change).
> > Now that "Matrices and Linear Algebra" had been created "Constructing Matrices" can be added to that and `IdentityMatrix` and `DiagonalMatrix` would be moved out of Tensors and in there with `summary_text` (which in the future I guess will be called usage_text, but that's a separate process.
> 
> Thank you for your advice.
> 
> Regarding summary_text, I noticed it's not displayed when `Method::usage` is called. Is it an already known issue?

this is what I was talking about at the end of https://github.com/Mathics3/mathics-core/pull/223#issuecomment-1073211740

We should read the ``summary_text`` string in the ``builtin.contribute`` method and set that message value.

PR #218, opened 2022-03-18
Removing Expression "cache" from BaseElement.

mmatera:
Remove _cache property from Element and clear_cache method from ``BaseElement``, and move them to the class ``Expression``. These properties do not make sense for Elements that do not have sub-elements. Also, I added a couple of *ifs* 
to check that calls to ``clear_cache`` in builtins are performed just over expressions.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/218"></a>

### DIFF
--- a/mathics/algorithm/parts.py
+++ b/mathics/algorithm/parts.py
@@ -273,7 +273,8 @@ def walk_parts(list_of_list, indices, evaluation, assign_rhs=None):
         except MessageException as e:
             e.message(evaluation)
             return False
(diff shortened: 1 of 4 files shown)


rocky:
LGTM @mmatera Many thanks for doing this! 

And just in time for my weekend attempt to forge ahead in looking at Expression evaluation.

(Later, I'll post my thoughts/plan for weekend work.)

PR #219, opened 2022-03-18
doctest handling of "..." again

rocky:
$TestTracing was failing on the Django side. Here we
sync up code a little bit.

Ideally though, we would put more of this in the common routine.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/219"></a>

### DIFF
--- a/mathics/docpipeline.py
+++ b/mathics/docpipeline.py
@@ -57,7 +57,7 @@ def print_and_log(*args):
         logfile.write(string)
 
 
-def compare(result, wanted):
+def compare(result, wanted) -> bool:
     if result == wanted:
         return True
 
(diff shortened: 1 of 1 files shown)


TiagoCavalcante (approved):
LGTM

> Ideally though, we would put more of this in the common routine.

We can put it in another function and import from Django.

mmatera (approved):
LGTM

rocky:
> LGTM
> 
> > Ideally though, we would put more of this in the common routine.
> 
> We can put it in another function and import from Django.

@TiagoCavalcante If you'd handle I would appreciate it, Note though that django right now is set up to work with 4.0.0 and assuming this there would mean the dependencies on core would have to increase to the current version or greater of core.

Also if workflows starts using a github version, the protocol has to change from `git` to `http` since github recently disabled port access to git.

PR #220, opened 2022-03-19
Tolerate Unicode in parser testing

rocky:
`test_parser.py` has been failing for me for a while. Apparently ever since in mathicsscript we added these alternate standard Unicode symbols for some custom Wolfram ones.

Here we adapt tests where unicode symbols have been bleeding cause failures in parser testing because it expect no unicode characters. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/220"></a>

### DIFF
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -26,6 +26,8 @@ jobs:
         brew install llvm@11
         python -m pip install --upgrade pip
         LLVM_CONFIG=/usr/local/Cellar/llvm@11/11.1.0/bin/llvm-config pip install llvmlite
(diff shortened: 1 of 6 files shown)


TiagoCavalcante (approved):
LGTM

mmatera:
LGTM. In any case, I hope to arrive to a better solution when the rework of Boxes be finished.

rocky:
> LGTM. In any case, I hope to arrive to a better solution when the rework of Boxes be finished.

This is not a Boxing issue. This is the parser which works well before that. It is a scanner and unicode issue.

PR #227, opened 2022-03-20
Use python built-in functions to escape strings

hrueter:
This PR fixes the following behaviour
```
In[1]:= StringForm["Text1\\n Text2"]
Out[1]= Text1
         Text2
```
### Expected behavior
```
In[1]:= StringForm["Text1\\n Text2"]
Out[1]= Text1\n Text2
```
### Further fixes
Some character encoding related bugs have been fixed. In particular `Export` did not honor the `CharacterEncoding` option.

### DIFF
--- a/mathics/core/parser/convert.py
+++ b/mathics/core/parser/convert.py
@@ -29,12 +29,7 @@ class GenericConverter(object):
 
     @staticmethod
     def string_escape(s):
-        s = s.replace("\\\\", "\\").replace('\\"', '"')
-        s = s.replace("\\r\\n", "\r\n")
-        s = s.replace("\\r", "\r")
-        s = s.replace("\\n", "\n")
-        s = s.replace("\\t", "\t")
-        return s
+        return s.encode("raw_unicode_escape").decode("unicode_escape")
 
     def convert_Symbol(self, node: Symbol):
         if node.context is not None:
--- a/mathics/builtin/files_io/importexport.py
+++ b/mathics/builtin/files_io/importexport.py
@@ -1744,7 +1744,7 @@ class Export(Builtin):
         "$OptionSyntax": "System`Ignore",
     }
 
-    def apply(self, filename, expr, evaluation, **options):
+    def apply(self, filename, expr, evaluation, options={}):
         "Export[filename_, expr_, OptionsPattern[Export]]"
 
         # Check filename
--- a/test/test_importexport.py
+++ b/test/test_importexport.py
@@ -6,6 +6,7 @@ import pytest
 import sys
 from .helper import check_evaluation, session
 
+from mathics.builtin.atomic.strings import to_python_encoding
 
 # def test_import():
 #     eaccent = "\xe9"
@@ -23,7 +24,7 @@ def run_export(temp_dirname: str, short_name: str, file_data: str, character_enc
     file_path = osp.join(temp_dirname, short_name)
     expr = fr'Export["{file_path}", {file_data}'
     expr += (
-        ', CharacterEncoding -> f"{character_encoding}"' if character_encoding else ""
+        fr', CharacterEncoding -> "{character_encoding}"' if character_encoding else ""
     )
     expr += "]"
     result = session.evaluate(expr)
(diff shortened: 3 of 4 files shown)


mathics/core/parser/convert.py
```diff
@@ -29,12 +29,7 @@ def do_convert(self, node):
 
     @staticmethod
     def string_escape(s):
-        s = s.replace("\\\\", "\\").replace('\\"', '"')
-        s = s.replace("\\r\\n", "\r\n")
-        s = s.replace("\\r", "\r")
-        s = s.replace("\\n", "\n")
-        s = s.replace("\\t", "\t")
-        return s
+        return s.encode('raw_unicode_escape').decode('unicode_escape')
```
rocky, line 32: Much better - thanks!

rocky:
@hrueter Thanks for the PR. To painlessly fix the black error do this where mathics-core is checked out:

```
$ pip install pre-commit black
$ pre-commit  # this should run black and change the file
```

rocky:
@hrueter I am now seeing this PR causing warnings: when running pytest. For example: 

```
$ pytest test/test_linalg.py 
==================================================================================================== test session starts =====================================================================================================
platform linux -- Python 3.8.13, pytest-7.1.1, pluggy-1.0.0
rootdir: /src/external-vcs/github/Mathics3/mathics-core
collected 11 items                                                                                                                                                                                                           

test/test_linalg.py ...........                                                                                                                                                                                        [100%]

====================================================================================================== warnings summary ======================================================================================================
mathics/core/parser/convert.py:32
test/test_linalg.py::test_inverse[None-None-None-None]
  /src/external-vcs/github/Mathics3/mathics-core/mathics/core/parser/convert.py:32: DeprecationWarning: invalid escape sequence '\!'
    return s.encode("raw_unicode_escape").decode("unicode_escape")

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=============================================================================================== 11 passed, 2 warnings in 9.18s ===============================================================================================
```

This is because `\\'` is getting turned into `\'` and that wasn't happening before ith the more narrow replacement which I guess is only replacing _whitespace_ characters. 

So what was the motivation for this PR? Is it something other than aesthetics or was there actual breakage?  

Yes,

hrueter:
`\\` was always replaced by `\`. This was done through `s = s.replace("\\\\", "\\")`. Hence I am not quite sure how `\\'` could have turned into something else before.

I also noticed some deprecation warnings like `DeprecationWarning: invalid escape sequence '\!'`. We don't want to unescape it, but python thinks we might have made a mistake here. As far as I know, the solution would be to use raw strings instead.

In my case I am generating C-code using some WL scripts. In some files I want to include `printf` statements of strings including escape sequences like `\n`. In the input files we therefore using escaping the back slash, i.e. `\\n`, which gets unescaped to our desired `\n` when the C file is written.
Instead mathics would convert the `\\n` to a line break and there was no way around that behavior.

rocky:
> `\\` was always replaced by `\`. This was done through `s = s.replace("\\\\", "\\")`. Hence I am not quite sure how `\\'` could have turned into something else before.
> 

A lack of imagination on your part! This code works in strange and mysterious ways which we are all learning about.

> I also noticed some deprecation warnings like `DeprecationWarning: invalid escape sequence '\!'`. We don't want to unescape it, but python thinks we might have made a mistake here. As far as I know, the solution would be to use raw strings instead.
> 

Ok. But where? An investigation needs to be done understand what's going on here.

> In my case I am generating C-code using some WL scripts. In some files I want to include `printf` statements of strings including escape sequences like `\n`. In the input files we therefore using escaping the back slash, i.e. `\\n`, which gets unescaped to our desired `\n` when the C file is written. Instead mathics would convert the `\\n` to a line break and there was no way around that behavior.

Ok. Thanks for the information. If you could create the smallest test case that exhibits this breakage, then that would help isolate how we get the behaviors we are seeing. Thanks.

rocky:
> Ok. Thanks for the information. If you could create the smallest test case that exhibits this breakage before the change, then that would help isolate how we get the behaviors we are seeing. Thanks.

Never mind. (A lack of imagination or investigation on my part.) 

I see the minimal test case [here](https://github.com/hrueter/mathics-core/blob/4c72fd2bb57a0818f424b635606ac4754e341223/test/test_parser/test_convert.py#L90)

hrueter:
Here is a more detailed test case. This is the content of `write.math`
```
file = "test.c";
DeleteFile[file];
filepointer = OpenAppend[file];
WriteString[filepointer, "printf(\"test\\n\")"];
Close[filepointer]
```
and then I run
`mathics write.math`.

PR #228, opened 2022-03-20
Moved DiagonalMatrix and IdentityMatrix to group Matrices and Linear Algebra

adamantinum:
As in https://github.com/Mathics3/mathics-core/pull/215#pullrequestreview-915117546

### DIFF
--- a/mathics/builtin/matrices/constrmatrix.py
+++ b/mathics/builtin/matrices/constrmatrix.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+
+from mathics.builtin.base import Builtin
+from mathics.core.expression import Expression
+from mathics.core.atoms import Integer0
+
+
+class DiagonalMatrix(Builtin):
+    """
+    <dl>
+      <dt>'DiagonalMatrix[$list$]'
+      <dd>gives a matrix with the values in $list$ on its diagonal and zeroes elsewhere.
+    </dl>
+
+    >> DiagonalMatrix[{1, 2, 3}]
+     = {{1, 0, 0}, {0, 2, 0}, {0, 0, 3}}
+    >> MatrixForm[%]
+     = 1   0   0
+     .
+     . 0   2   0
+     .
+     . 0   0   3
+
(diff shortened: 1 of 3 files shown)


mathics/builtin/matrices/constrmatrix.py
```diff
@@ -0,0 +1,57 @@
...
+     .
+     . 0   2   0
+     .
+     . 0   0   3
+
+    #> DiagonalMatrix[a + b]
+     = DiagonalMatrix[a + b]
+    """
+
+    summary_text = "Return a diagonal matrix with the elements of a given list"
```
rocky, line 28: A small thing that I also noticed on the last PR. Summary text doesn't start out with a capital letter. You'll find mention and discussion of this in another draft PR. 

Also other than the two summary texts you've written I can't think of any other one that starts off "returns a".  Diagonal start out "gives a".  

One last small thing is that we want to eliminate the use of "#>" (a test that is run but not shown in documentation). These were added before we had proper pytests for such things that can be found in the test directory. 

You attention to detail is appreciated.

adamantinum: Ok. Regarding "#>", should I remove the whole line from tensor.py too?

rocky: Yes - if we could add these as pytests in the `test` directory that would be awesome. 

As mentioned before there is a lot of cleanup work that this project needs because it was greatly expanded before some basic problems were addressed.

rocky:
Looks good. I'll add the pytest for the removed "#>" so you see what I mean.

adamantinum:
> Looks good. I'll add the pytest for the removed "#>" so you see what I mean. 

Actually I have not deleted the "#>" yet because I was wating for put them in a pytest later...

rocky:
Well, I've done it now. But in the future if you'd do these things I would appreciate it.

adamantinum:
> Well, I've done it now in #229. But in the future if you'd do these things I would appreciate it. 

Yes, thank you

PR #229, opened 2022-03-20
Reorganize builtins

rocky:
Move a `#>` test into a unit test and Moved DiagonalMatrix and IdentityMatrix to group Matrices and Linear Algebra from @adamantian

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/229"></a>

### DIFF
--- a/test/test_constrmatrix.py
+++ b/test/test_constrmatrix.py
@@ -1,4 +1,4 @@
-from .helper import check_evaluation, session
+from .helper import session
 import pytest
 
 
@@ -8,7 +8,7 @@ import pytest
         (
             "DiagonalMatrix[a + b]",
             "DiagonalMatrix[a + b]",
-            "Leaves unchanged? Adapted from #> code that was in tensor.py",
+            "argument needs to be a list",
         ),
     ],
 )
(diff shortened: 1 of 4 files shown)


test/test_constrmatrix.py
```diff
@@ -0,0 +1,18 @@
...
+import pytest
+
+
+@pytest.mark.parametrize(
+    ("str_expr", "str_expected", "fail_msg"),
+    [
+        (
+            "DiagonalMatrix[a + b]",
+            "DiagonalMatrix[a + b]",
+            "Leaves unchanged? Adapted from #> code that was in tensor.py",
```
TiagoCavalcante, line 11: Maybe this could be `DiagonalMatrix is only defined for lists`.

TiagoCavalcante (approved):
LGTM - just see the comment.

test/test_constrmatrix.py
rocky: Ok. I looked at the doc and this is the way it is because clearly the expression `a + b` where neither a nor b has been defined is not an expression. It seems weird to me that someone would this property in this way.

TiagoCavalcante: Well, WL is a symbolic language, thus it seems logical that almost any symbol can be attached to a value, so I guess there isn't another way to test this.

rocky: A simpler test is use a single literal value like 1, or "a"

PR #230, opened 2022-03-20
Fix InputFileName

hrueter:
`InputFileName` was broken since 3e3802797677bdf075dfd92f3680c2c6a60055b3, because `INPUTFILE_VAR` is updated in another file.

### DIFF
--- a/mathics/builtin/files_io/files.py
+++ b/mathics/builtin/files_io/files.py
@@ -20,6 +20,7 @@ from itertools import chain
 
 from mathics_scanner import TranslateError
 from mathics.core.parser import MathicsFileLineFeeder, parse
+from mathics.core import read
 from mathics.core.read import (
     channel_to_stream,
     MathicsOpen,
@@ -61,7 +62,6 @@ INITIAL_DIR = os.getcwd()
 DIRECTORY_STACK = [INITIAL_DIR]
 
 INPUT_VAR = ""
-INPUTFILE_VAR = ""
 
 TMP_DIR = tempfile.gettempdir()
 SymbolPath = Symbol("$Path")
@@ -103,8 +103,7 @@ class InputFileName(Predefined):
     name = "$InputFileName"
 
     def evaluate(self, evaluation):
-        global INPUTFILE_VAR
-        return String(INPUTFILE_VAR)
+        return String(read.INPUTFILE_VAR)
 
 
 class EndOfFile(Builtin):
--- a/mathics/core/read.py
+++ b/mathics/core/read.py
@@ -12,6 +12,8 @@ from mathics.core.atoms import Integer, String
 from mathics.core.symbols import Symbol
 from mathics.core.streams import Stream, path_search, stream_manager
 
+INPUTFILE_VAR = ""
+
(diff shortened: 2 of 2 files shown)


mathics/builtin/files_io/files.py
```diff
@@ -103,6 +103,7 @@ class InputFileName(Predefined):
     name = "$InputFileName"
 
     def evaluate(self, evaluation):
+        return String(read.INPUTFILE_VAR)
         global INPUTFILE_VAR
         return String(INPUTFILE_VAR)
```
TiagoCavalcante, line 108: This is dead code, could you remove these 2 lines?

mathics/builtin/files_io/files.py
```diff
@@ -20,6 +20,7 @@
 
 from mathics_scanner import TranslateError
 from mathics.core.parser import MathicsFileLineFeeder, parse
+from mathics.core import read
```
TiagoCavalcante, line 23: Usually, we don't import modules, but just the things we need inside those modules, doing this makes circular imports less common.

```py
from mathics.core.read import INPUTFILE_VAR
```

TiagoCavalcante (changes requested):
LGTM - just 2 small things.

mathics/builtin/files_io/files.py
hrueter: I tried this, but it does not work for me on Python 3.8.10.
According to [this discussion](https://stackoverflow.com/questions/13034496/using-global-variables-between-files#comment104154371_45646477) I am not the only one observing this behavior.
Furthermore this is also the style used for the global variables in `settings.py`.

rocky: While whether or not `from x import` or `import` is used (and in my view this kind of thing is debatable), in Python this has no effect on circular imports. 

I tried a little test and in that it also shows why this doesn't change things: 

File a.py: 
```
from b import B
A = "A" + B
```

File b.py:

```
from a import A
b = "B" + A
```

File `__main__.py`: 

```
from a import A
from b import B

print(A + B)
```

The usual null `__init__.py` (but I am not sure that is needed here). 

Running `__main__.py`: 

```
python __main__.py 
  File "__main__.py", line 1, in <module>
    from a import A
  File "/tmp/myfoo/a.py", line 1, in <module>
    from b import B
  File "/tmp/myfoo/b.py", line 1, in <module>
    from a import A
ImportError: cannot import name 'A' from partially initialized module 'a' (most likely due to a circular import) (/tmp/myfoo/a.py)

```

TiagoCavalcante: @hrueter ok, thanks for explaining.

TiagoCavalcante (approved):

rocky:
@hrueter Sorry for introducing that bug - thanks for fixing. (There is way more code around reading streams that could stand improving.

mathics/builtin/files_io/files.py
mmatera: The reason to do not import full modules but just the needed elements is that it helps to debug these circular references, not because it avoids it.

PR #232, opened 2022-03-20
Fix Factor[a == b]

### LINKED ISSUE #231: Mathics crashes during Factor
### Observed behavior
Mathics crashes from this input
```
In[1]:= a == d b + d c // Factor
```
### Expected behavior as in Mathematica
```
In[1]:= a == d b + d c // Factor
Out[1]= a == (b + c) d
```
### Setup
Linux Mint 20.3
Mathics 4.1.0.dev0 on CPython 3.8.10 (default, Nov 26 2021, 20:14:08)  using SymPy 1.8, mpmath 1.2.1, numpy 1.17.4

TiagoCavalcante:
Fixes #231.

### DIFF
--- a/mathics/builtin/numbers/algebra.py
+++ b/mathics/builtin/numbers/algebra.py
@@ -592,14 +592,9 @@ class Factor(Builtin):
 
         try:
             result = sympy.together(expr_sympy)
-            numer, denom = result.as_numer_denom()
-            if denom == 1:
-                result = sympy.factor(expr_sympy)
-            else:
-                result = sympy.factor(numer) / sympy.factor(denom)
+            return from_sympy(sympy.factor(result))
         except sympy.PolynomialError:
             return expr
-        return from_sympy(result)
 
 
 class FactorTermsList(Builtin):


test/test_combinatorica.py
```diff
@@ -292,6 +292,11 @@ def test_permutations_groups_1_2():
             "m (1 + m) (4 - 2 m + 2 m ^ 2 + 3 m ^ 3 + m ^ 4 - m ^ 5 + m ^ 6) / 16",
             "Factor example in Polya polynomial 1.2.6, Page 26",
         ),
+        (
```
rocky, line 295: As ti suggests on line 24 of that file: 

```
# A number of examples from:
#  * Implementing Discrete Mathematics by Steven Skiena and
#  * Computation Discrete Mathematics by Sriram Pemmaraju and Steven Skiena.

# Page numbers below come from the first book
# Some tests have been altered to speed them up, or to make the intent
# more clear in a test.
```

This test should go in `test_algebra.py` not here. Down the line I think this will be even more explicit because that test will be `builtin/numbers/test_algebra.py` while this test which is about a package will stay here or maybe get moved to a directory called "test/package".

rocky: Also, it would be good to add this as an example with some text indicating that factor works with equations.

rocky: It occurs to me that if the example is the same thing as the pytest, there is an argument to be made that the same example doesn't need to be run twice. However, a case could be made the other way too. So I leave it for you to decide.

TiagoCavalcante: > Also, it would be good to add this as an example with some text indicating that factor works with equations.

@rocky I don't think so, a LOT of WL functions work with equations, and even more: there is nothing special with equations, if you remove those classes from the builtins you would have usual functions, and Factor also works with undefined functions.

rocky: While this may be true, the description for factor: "factors the polynomial expression expr" applies to an equation, so an example would be appreciated. 

If you compare what we have in the way of describing factor versus what is [here](https://reference.wolfram.com/language/ref/Factor.html) we are woefully lacking. An example where an equation is found is  https://wolfram.com/xid/0girwy-j72viv

That said, do as you wish.

rocky (approved):
LGTM - Thanks for doing this and going in this direction

PR #233, opened 2022-03-21
fix usage

### LINKED ISSUE #225: MessageName Symbol::usage not working
`Symbol::usage` returns nothing, while `?Symbol` works correctly.
```
Example:
In[1]:= ?Diagonal
Return a list with the diagonal elements of a given matrix
Out[1]:= Null

In[2]:= Diagonal::usage
Out[2]:= Diagonal::usage

```

mmatera:
This PR fixes #225 by adding the usage message from the ``summary_text`` property of a ``Builtin`` sub-class during the call to the ``contribute`` method.

### DIFF
--- a/mathics/builtin/base.py
+++ b/mathics/builtin/base.py
@@ -288,6 +288,8 @@ class Builtin(object):
         for form, formatrules in formatvalues.items():
             formatrules.sort()
 
+        if hasattr(self, "summary_text"):
+            self.messages["usage"] = self.summary_text
         messages = [
             Rule(
                 Expression("MessageName", Symbol(name), String(msg)),


TiagoCavalcante (approved):
LGTM - as simple as possible

rocky:
LGTM - thanks for doing.

rocky:
@mmatera I notice with this change that this also changes the behavior of `??` and removes the longer form.

``` mathics

Mathics 4.1.0.dev0
on CPython 3.8.12 (default, Oct 21 2021, 14:26:19) 
using SymPy 1.10, mpmath 1.2.1, numpy 1.21.3

under certain conditions.

In[1]:= ?Part
form n-tuples from a list
Out[1]= Null

In[2]:= ??Part
form n-tuples from a list
        
        Attributes[Part] = {NHoldRest, Protected, ReadProtected}
Out[2]= Null

In[3]:= 
```

versus what we had before: 

```
In[1]:= ??Part

          'Part[expr, i]'
            returns part i of expr.
        
        
        Extract an element from a list:
        
        Attributes[Part] = {NHoldRest, Protected, ReadProtected}
Out[1]:= Null
```

Do you know what's up? 

Down the line when we get into formatting, I believe we should have a way to handle the expanded doc formatting.

mmatera:
@rocky, I am aware of this. I started to write a discussion about this point, but I haven't finished. 

Briefly, when you ask for the short version of ``Information``, what is shown is just the ``symbol::usage`` string, and in the long-form, apart from this string, all the attributes and user-definitions of the symbol are also shown.
When I wrote the first implementation of ``Information``, I implemented it by building the  "symbol::usage" "on the fly" from the docstring. At some point, I think you changed this by looking at the property ``summary_text``, but just for the short form, while the long-form still looks at the docstring. 
The most compatible way (with the WL standard) would be to implement ``Information``  looking at the ``usage`` message (which can be replaced in runtime just by unprotecting and assigning a new string). And the ``usage`` message should be independent of the docstring (that is more suitable for full documentation). Later, I will try to extend this discussion.
 

For details, see https://reference.wolfram.com/language/ref/Information.html?q=Information

rocky:
Ok thanks for more information on this. The example given in your link is a little bit vague because it uses Sin which is pretty short and has one form. A more revealing description would be [GeoPosition](https://reference.wolfram.com/language/ref/GeoPosition.html).

Notice that there are several forms. Does the short information show _all_ of the form? The kind of object returned in a "usage" variable seems to be a string, while in Mathematica we return None when the `Information[]` function is called. Is that how Mathematica works? 

Note that in a guide mentioning GeoPosition there is a summary shown, not all of the forms are broken out and listed. See https://reference.wolfram.com/language/guide/Geodesy.html

The handling of `Information[]` has been a problem for as long as I have been working on this project, and until we get around to totally revising it, I don't think we are going to find a solution that suits all purposes. We have always been bending or deviating from Mathematica behavior in one way or another.

rocky:
I have been thinking about this a little more and with respect to what I in the Mathematica docs. Basically, it looks like `Information` should be its own object, and when the front-ends get that kind of object format depending on their capabilities, either as RsT, Sphinx, LaTeX, HTML, or (unicode or ASCII) text. 

But again this is work that I am not sure anyone is up for right now.

PR #235, opened 2022-03-21
Use assertEqual instead of assertEquals for Python 3.11 compatibility.

tirkarthi:
Deprecated aliases have been removed in python/cpython#28268


rocky:
@tirkarthi Thanks for the PR. 

To fix the autoblack problem, locally run in the mathics-code:

```
$ pip install pre-commit
$ pre-commit
```

It would be good to convert all of the unit tests (from the 2.7 days) to pytests, if you are up for it. If not we can just commit this and continue. Your thoughts?

tirkarthi:
Formatted with black. There is a migration tool from pytest I can give a try in another PR if I find time : https://github.com/pytest-dev/unittest2pytest

rocky:
Ok - thanks.

Issue #214, opened 2022-03-17
Crash during FullSimplify


hrueter:
### Observed behavior
Mathics crashes from the following input
```
In[1]:= a[x] + e f / (2 d) + c[x] // FullSimplify
```
### Expected behavior
```
In[1]:= a[x] + e f / (2 d) + c[x] // FullSimplify
Out[1]= e f / (2 d) + a[x] + c[x]
```
### Background
I tried to find a more minimal example, but reducing the input further eliminates the failure.

Curiously the following order of input works
```
In[1]:= a[x] + e f / (d) + c[x] // FullSimplify
Out[1]= e f / d + a[x] + c[x]

In[2]:= a[x] + e f / (2 d) + c[x] // FullSimplify
Out[2]= e f / (2 d) + a[x] + c[x]
```

### Setup
Linux Mint 20.3
Mathics 4.0.0 on CPython 3.8.10 (default, Nov 26 2021, 20:14:08)  using SymPy 1.8, mpmath 1.2.1, numpy 1.17.4

### Traceback
```
Traceback (most recent call last):
  File "/usr/local/bin/mathics", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.8/dist-packages/mathics/main.py", line 402, in main
    result = evaluation.evaluate(query, timeout=settings.TIMEOUT)
  File "/usr/local/lib/python3.8/dist-packages/mathics/core/evaluation.py", line 361, in evaluate
    result = run_with_timeout_and_stack(evaluate, timeout, self)
  File "/usr/local/lib/python3.8/dist-packages/mathics/core/evaluation.py", line 118, in run_with_timeout_and_stack
    return request()
  File "/usr/local/lib/python3.8/dist-packages/mathics/core/evaluation.py", line 332, in evaluate
    self.last_eval = query.evaluate(self)
  File "/usr/local/lib/python3.8/dist-packages/mathics/core/expression.py", line 1315, in evaluate
    expr, reevaluate = expr.evaluate_next(evaluation)
  File "/usr/local/lib/python3.8/dist-packages/mathics/core/expression.py", line 1449, in evaluate_next
    result = rule.apply(new, evaluation, fully=False)
  File "/usr/local/lib/python3.8/dist-packages/mathics/core/rules.py", line 67, in apply
    self.pattern.match(yield_match, expression, {}, evaluation, fully=fully)
[...]

PR #238, opened 2022-03-22
Handle unexpected SymPy exceptions in `Simplify[]`

mmatera:
This PR fixes issue #214, by catching the exception. In the future, we can prevent the exception for this particular case, but still would appear in other cases.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/238"></a>

### DIFF
--- a/mathics/builtin/numbers/algebra.py
+++ b/mathics/builtin/numbers/algebra.py
@@ -501,7 +501,13 @@ class Simplify(Builtin):
         if sympy_expr is None:
             return expr
         # Now, try to simplify using sympy
-        sympy_result = sympy.simplify(sympy_expr)
+        try:
+            sympy_result = sympy.simplify(sympy_expr)
+        except Exception:
+            # This handles the exception reported in 214.
+            # However, it seems that the crash is due to a
+            # problem in the implentation of the class SympyExpression
+            return expr
         # and bring it back
         result = from_sympy(sympy_result).evaluate(evaluation)
         return result
--- a/mathics/builtin/numbers/algebra.py
+++ b/mathics/builtin/numbers/algebra.py
@@ -504,9 +504,6 @@ class Simplify(Builtin):
         try:
             sympy_result = sympy.simplify(sympy_expr)
         except Exception:
-            # This handles the exception reported in 214.
-            # However, it seems that the crash is due to a
-            # problem in the implentation of the class SympyExpression
             return expr
         # and bring it back
         result = from_sympy(sympy_result).evaluate(evaluation)
--- a/test/test_algebra.py
+++ b/test/test_algebra.py
@@ -284,3 +284,14 @@ def test_simplify():
         ),
     ):
         check_evaluation(str_expr, str_expected)
+
+
+def test_fullsimplify():
+    for str_expr, str_expected, failure_message in (
+        (
+            " a[x] + e f / (2 d) + c[x] // FullSimplify",
+            "e f / (2 d) + a[x] + c[x]",
+            "issue #214",
+        ),
+    ):
+        check_evaluation(str_expr, str_expected, failure_message)


rocky (approved):
LGTM

TiagoCavalcante (approved):
LGTM - this is so weird, SymPyExpression seems to be so fragile but fixing Simplify fixed this for some versions of SymPy, and others not.

mathics/builtin/numbers/algebra.py
```diff
@@ -501,7 +501,13 @@ def apply(self, expr, evaluation):
...
             return expr
         # Now, try to simplify using sympy
-        sympy_result = sympy.simplify(sympy_expr)
+        try:
+            sympy_result = sympy.simplify(sympy_expr)
+        except Exception:
+            # This handles the exception reported in 214.
+            # However, it seems that the crash is due to a
+            # problem in the implentation of the class SympyExpression
+            return expr
```
TiagoCavalcante, line 510: This specific case I believe it is SymPy side.

rocky: I agree. We should get rid of this comment (and the misspelling in it) since the Python try says the same thing. And the fact that it was reported in an issue called 214 really is irrelevant. If this arcane information is needed, it is in fact stored in the git commit message.

rocky:
> LGTM - this is so weird, SymPyExpression seems to be so fragile but fixing Simplify fixed this for some versions of SymPy, and others not.

Probably something better down the line would be to write our own wrapper for SymPyExpression which does a try not just here but around _all_ SymPyExpressions.  The other thing I would like to see it do is have a way to trace the SymPy calls,
and for efficiency this could be done like the way we do it for other traces - we just replace the non-instrumented call with an traced version of it.

TiagoCavalcante:
> > LGTM - this is so weird, SymPyExpression seems to be so fragile but fixing Simplify fixed this for some versions of SymPy, and others not.
> 
> Probably something better down the line would be to write our own wrapper for SymPyExpression which does a try not just here but around _all_ SymPyExpressions. The other thing I would like to see it do is have a way to trace the SymPy calls, and for efficiency this could be done like the way we do it for other traces - we just replace the non-instrumented call with an traced version of it.

@rocky I don't think this is necessary, we kinda have a wrapper for SymPy expressions, but that is only used as a cache, the class I'm talking about is in `coversion.py`. That class [and the whole from_sympy method] need to be refactored.

rocky:
> > > LGTM - this is so weird, SymPyExpression seems to be so fragile but fixing Simplify fixed this for some versions of SymPy, and others not.
> > 
> > 
> > Probably something better down the line would be to write our own wrapper for SymPyExpression which does a try not just here but around _all_ SymPyExpressions. The other thing I would like to see it do is have a way to trace the SymPy calls, and for efficiency this could be done like the way we do it for other traces - we just replace the non-instrumented call with an traced version of it.
> 
> @rocky I don't think this is necessary, we kinda have a wrapper for SymPy expressions, but that is only used as a cache, the class I'm talking about is in `coversion.py`. That class [and the whole from_sympy method] need to be refactored.

Specifics on the location please. 

> > > LGTM - this is so weird, SymPyExpression seems to be so fragile but fixing Simplify fixed this for some versions of SymPy, and others not.
> > 
> > 
> > Probably something better down the line would be to write our own wrapper for SymPyExpression which does a try not just here but around _all_ SymPyExpressions. The other thing I would like to see it do is have a way to trace the SymPy calls, and for efficiency this could be done like the way we do it for other traces - we just replace the non-instrumented call with an traced version of it.
> 
> @rocky I don't think this is necessary, we kinda have a wrapper for SymPy expressions, but that is only used as a cache, the class I'm talking about is in `coversion.py`. That class [and the whole from_sympy method] need to be refactored.

Specifics on the location please. (In general if you and @mmatera would make a habit if spending a little extra time to be specific about what you mean, I would be grateful).  Here is an example of being detailed: 

when I run from Mathics3/mathics-core: 

```
$ PAGER=cat git grep -n " sympy[.]"
```

I am seeing about 200 calls to sympy functions. I'm okay with wrapping them all in a call that allows us to trace these and wrap them in `try` in case they fail, and if they do report an error which by the way is not done here but should be done.

No major refactor needed.

TiagoCavalcante:
@rocky the sensible function is this:

https://github.com/Mathics3/mathics-core/blob/0c84730e4d595ce8b9768db926f225a344edd2fe/mathics/core/convert.py#L140-L381

> I am seeing many locations where we have calls to sympy functions. I'm okay with wrapping all of the calls all in a call that allows us to trace and benchmark these on demand and wrap them in try in case they fail, and if they do report an error which by the way is not done here but should be done.
>
> No major refactor needed.

In some cases we are sure SymPy is going to work, e.g.:
```py
class F(Builtin):
  def apply(self, a, b, evaluation):
    "F[a_Integer, b_Integer]"
    return from_sympy(to_sympy(a) + to_sympy(b))
```

rocky:
> @rocky the sensible function is this:
> 
> https://github.com/Mathics3/mathics-core/blob/0c84730e4d595ce8b9768db926f225a344edd2fe/mathics/core/convert.py#L140-L381
> 
> > I am seeing many locations where we have calls to sympy functions. I'm okay with wrapping all of the calls all in a call that allows us to trace and benchmark these on demand and wrap them in try in case they fail, and if they do report an error which by the way is not done here but should be done.
> > No major refactor needed.
> 
> In some cases we are sure SymPy is going to work, e.g.:
> 
> ```python
> class F(Builtin):
>   def apply(self, a, b, evaluation):
>     "F[a_Integer, b_Integer]"
>     return from_sympy(to_sympy(a) + to_sympy(b))
> ```

I think this kind of petty optimization thinking has gotten us into trouble before. I realize you _think_ that there can't be a problem, until there is. And then we have of avoidable process we see here. 
 
Of course, the wrapper isn't here but in whatever to_sympy() and from_sympy() call.  I really doubt that the decorator calls and try/except would add even 1% overhead - prove me wrong. 

But this kind of thing it would allow for easier benchmarking and user-level tracing what's going on in sympy. 

Probably there is a way to do this by just capturing and replacing things by replacing the sympy module with another module that wraps all calls in it.  This kind of thing would more easily be done before getting into mathics would be difficult to do as a feature like we have for TraceBuiltins.

rocky:
It occurs to me that this would be better done as a decorator or as a `with`: 

```
with wrap_sympy(sympy.simplify, err_result) as symplify:
    sympy_result = simplify(sympy_expr)
```

TiagoCavalcante:
> I realize you think that there can't be a problem, until there is because of something someone overlooked somewhere and some really exceptional situation.

I'm saying that if the function I mentioned be fixed/re-written the errors would make sense, instead of a traceback to a random place the traceback would be to the correct line: the SymPy call.

> prove me wrong

You are not.

> It occurs to me that this would be better done as a decorator or as a context manager (`with`):
> 
> ```
> with wrap_sympy_fn(fn=sympy.simplify, result=err_result) as symplify:
>     sympy_result = simplify(sympy_expr)
> ```

Maybe this as we use multiple functions in some cases:

```py
with wrap_sympy(result=err_result) as sympy:
  # sympy is an object of a class that extends dict, that class overwrites __getitem__ so it wrap
  # the SymPy function on the fly.
  sympy_result = sympy.simplify(sympy_expr)
```

rocky:
I just realized how we could make the context manager handle dynamic tracing and benchmarking. Basically this would be a method off of the evaluation object which is generally everywhere. So: 

```
with evaluation.wrap_sympy(result=err_result) as sympy:
  # sympy is an object of a class that extends dict, that class
  # overwrites __getitem__ so it wrap the SymPy function on the fly.
  sympy_result = sympy.simplify(sympy_expr)
```

mathics/builtin/numbers/algebra.py
mmatera: Sorry for not responding before, but I had a busy day. I just left the message in case someone else had the chance to discover the origin of the exception.

mmatera: the comment has gone.

rocky: Ok, but I think in discussion we have come to feel more flexible solution would be to wrap this in a context manager method off of the evaluation. 

Using that here will mean that we won't have to touch this piece of code again. And this part of the code can be used for testing context manager features, even if this new context manager doesn't have lots of bells and whistles. 

Also, while it would be good to cover everything, it doesn't have to be done all at once.

PR #241, opened 2022-03-24
Adding beta function

mmatera:
Adding Euler's `Beta` function and ``LogGamma`` to the `mathics.builtin.special.gamma` module

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/241"></a>


mathics/builtin/specialfns/gamma.py
```diff
@@ -1,20 +1,256 @@
...
+    def apply_3(self, z, a, b, evaluation):
+        """Beta[z_, a_, b_]"""
+        # Here I needed to do that because the order of the arguments in WL
+        # is different from the order in mpmath. Most of the code is the same
+        # thatn in
+        if not all(isinstance(q, Number) for q in (a, b, z)):
+            return
+
+        args = (
+            Expression("Sequence", a, b, Integer0, z)
```
TiagoCavalcante, line 92: SymbolSequence

mathics/builtin/specialfns/gamma.py
```diff
@@ -1,20 +1,256 @@
...
+                return
+
+            result = call_mpmath(mpmath_function, tuple(float_args))
+            if isinstance(result, (mpmath.mpc, mpmath.mpf)):
+                if mpmath.isinf(result) and isinstance(result, mpmath.mpc):
+                    result = Symbol("ComplexInfinity")
+                elif mpmath.isinf(result) and result > 0:
+                    result = Expression("DirectedInfinity", Integer1)
+                elif mpmath.isinf(result) and result < 0:
+                    result = Expression("DirectedInfinity", Integer(-1))
```
TiagoCavalcante, line 113: SymbolDirectedInfinity

mathics/builtin/specialfns/gamma.py
```diff
@@ -1,20 +1,256 @@
...
+            float_args = [
+                arg.round().get_float_value(permit_complex=True) for arg in args
+            ]
+            if None in float_args:
+                return
+
+            result = call_mpmath(mpmath_function, tuple(float_args))
+            if isinstance(result, (mpmath.mpc, mpmath.mpf)):
+                if mpmath.isinf(result) and isinstance(result, mpmath.mpc):
+                    result = Symbol("ComplexInfinity")
```
TiagoCavalcante, line 109: SymbolComplexInfinity

mathics/builtin/specialfns/gamma.py
```diff
@@ -1,20 +1,256 @@
...
+                elif mpmath.isinf(result) and result > 0:
+                    result = Expression("DirectedInfinity", Integer1)
+                elif mpmath.isinf(result) and result < 0:
+                    result = Expression("DirectedInfinity", Integer(-1))
+                elif mpmath.isnan(result):
+                    result = Symbol("Indeterminate")
```
TiagoCavalcante, line 115: SymbolIndeterminate

TiagoCavalcante (approved):
LGTM - that seems a better place for Factorial and Factorial2

mathics/builtin/arithmetic.py
```diff
@@ -133,7 +134,11 @@ def apply(self, z, evaluation):
                 elif mpmath.isnan(result):
                     result = Symbol("Indeterminate")
                 else:
-                    result = from_mpmath(result)
+                    try:
```
rocky, line 137: Recently we had another "try except" the other day over something similar. 

The suggestion was to write a context handler that will handle this kind of stuff so that we not only do not duplicate boilerplate code like this, but we can customize it for things like tracing calls to SymPy (and here mpmath), get timings on that and so on. 

While that or this commit in of itself is not the end of the world, what has happened over time has that this kind of thing has gone on to the point where things have gotten pretty messy and while not gaining on general tools analogous to the benchmark stuff that could help everyone.

mmatera: @rocky, this is just what I meant when I say that each line we add to the builtins is affected by the refactor of the core. I just added something that I hoped was straightforward, but at the second like I hit this issue. 

Indeed, we could try here an approach using a context manager attached to the evaluation object, but that would imply touching the core. The best that I could do here is to add an inline comment talking about that idea.

rocky: Thanks for asking in advance.

Adding a context manager to wrap calls to sympy or mpmath It doesn't change how evaluation works and is I don't see how any of the contemplated changes to evaluation would have anything to do with this.  

In fact I suspect all of these context managers could be in the mathics.builtin.base module. Changes in the evaluator have largely been in mathics.core.

Boxing evaluation and code on the other hand is totally different. Boxing is a kind of evaluation and has to be intertwined with evaluation. I am sorry I haven't fully explained my vision for how things might work. 

(The short answer would be something like do it like a number of other bytecode evaluators and if you can mix in JIT afterwards as Jan suggested, great. But this is very handwavy and I although I don't have a clearer sense exactly of how things would go. But I think I have a much clearer sense than others here.)

mmatera: > Thanks for asking in advance.
> 
> Adding a context manager to wrap calls to sympy or mpmath It doesn't change how evaluation works and is I don't see how any of the contemplated changes to evaluation would have anything to do with this.
> 

If I understood right, the context manager should be attached to the evaluation object. Again, we can do that, but I still need to figure out how to implement it, and it does not seem to be as easy as adding a try-except block. The good thing, in this case, is that I do not need to patch many modules, but just ``mathics.core.arithmetic``, where the superclasses of numeric builtins are defined.

> In fact I suspect all of these context managers could be in the mathics.builtin.base module. Changes in the evaluator have largely been in mathics.core.
> 

Maybe it is possible, but your proposal was to attach it to the evaluation object. 

> Boxing evaluation and code on the other hand is totally different. Boxing is a kind of evaluation and has to be intertwined with evaluation. I am sorry I haven't fully explained my vision for how things might work.
> 

Yes, I have a similar feeling, but let's discuss it when we reach it.

rocky: > > Thanks for asking in advance.
> > Adding a context manager to wrap calls to sympy or mpmath It doesn't change how evaluation works and is I don't see how any of the contemplated changes to evaluation would have anything to do with this.
> 
> If I understood right, the context manager should be attached to the evaluation object. Again, we can do that, but I still need to figure out how to implement it, and it does not seem to be as easy as adding a try-except block. The good thing, in this case, is that I do not need to patch many modules, but just `mathics.core.arithmetic`, where the superclasses of numeric builtins are defined.
> 
> > In fact I suspect all of these context managers could be in the mathics.builtin.base module. Changes in the evaluator have largely been in mathics.core.
> 
> Maybe it is possible, but your proposal was to attach it to the evaluation object.

It would probably be simpler for me to just do it sometime.

PR #247, opened 2022-03-27
Start to organize unit tests

rocky:
Make the directory structure more closely resemble how it appears in the module space.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/247"></a>


test/test_convert.py
```diff
@@ -1,191 +0,0 @@
-#!/usr/bin/env python3
```
rocky, line 1: I am not sure why this show up as deleted since it was _moved_ to `test/parser/test_convert.py` above.

rocky:
I actually did more along these lines and have this around on my disk. However at some point things started to fail, and I hadn't been able to pinpoint why. 

So I'd like to stop this activity for now, see how this part goes, and then continue some other time.

test/test_convert.py
mmatera: stupid question: did you move it with `mv` or with `git mv`?

rocky: Should be fixed now.

mmatera (approved):
LGTM

PR #249, opened 2022-03-27
adding summary texts in builtin.numeric.algebra

mmatera:
This PR adds the ``summary_string`` field to ech ``Builtin`` in this module.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/249"></a>

### DIFF
--- a/mathics/builtin/numbers/algebra.py
+++ b/mathics/builtin/numbers/algebra.py
@@ -1230,7 +1230,7 @@ class ExpandAll(_Expand):
 
     """
 
-    summary_text = "expands products and powers, including negative integer powers"
+    summary_text = "expand products and powers, including negative integer powers"
 
     def apply_patt(self, expr, target, evaluation, options):
         "ExpandAll[expr_, target_, OptionsPattern[Expand]]"
--- a/mathics/builtin/numbers/algebra.py
+++ b/mathics/builtin/numbers/algebra.py
@@ -364,374 +364,13 @@ def find_exponents(expr, var):
     return sorted(result)
 
 
-class Cancel(Builtin):
-    """
-    <dl>
-      <dt>'Cancel[$expr$]'
-      <dd>cancels out common factors in numerators and denominators.
-    </dl>
-
-    >> Cancel[x / x ^ 2]
-     = 1 / x
-    'Cancel' threads over sums:
-    >> Cancel[x / x ^ 2 + y / y ^ 2]
-     = 1 / x + 1 / y
-
-    >> Cancel[f[x] / x + x * f[x] / x ^ 2]
-     = 2 f[x] / x
-    """
-
-    summary_text = "cancels common factors in rational expressions"
-    attributes = listable | protected
-
-    def apply(self, expr, evaluation):
-        "Cancel[expr_]"
-
-        return cancel(expr)
-
-
-class Simplify(Builtin):
-    """
-    <dl>
-    <dt>'Simplify[$expr$]'
-        <dd>simplifies $expr$.
-    <dt>'Simplify[$expr$, $assump$]'
-        <dd>simplifies $expr$ assuming $assump$ instead of $Assumptions$.
-    </dl>
-
-    >> Simplify[2*Sin[x]^2 + 2*Cos[x]^2]
-     = 2
-    >> Simplify[x]
-     = x
-    >> Simplify[f[x]]
-     = f[x]
-
-    Simplify over conditional expressions uses $\$Assumptions$, or $assump$
-    to evaluate the condition:
-    >> $Assumptions={a <= 0};
-    >> Simplify[ConditionalExpression[1, a > 0]]
-     = Undefined
-    The $assump$ option  override $\$Assumption$:
-    >> Simplify[ConditionalExpression[1, a > 0] ConditionalExpression[1, b > 0], { b > 0 }]
-     = ConditionalExpression[1, a > 0]
-    On the other hand, $Assumptions$ option does not override $\$Assumption$, but add to them:
-    >> Simplify[ConditionalExpression[1, a > 0] ConditionalExpression[1, b > 0], Assumptions -> { b > 0 }]
-     = Undefined
-    Passing both options overwrites $Assumptions with the union of $assump$ the option
-    >> Simplify[ConditionalExpression[1, a > 0] ConditionalExpression[1, b > 0], {a>0},Assumptions -> { b > 0 }]
-     = 1
-    >> $Assumptions={};
-    """
-
-    summary_text = "applies transformations to simplify the expression"
-    rules = {
-        "Simplify[list_List]": "Simplify /@ list",
-        "Simplify[rule_Rule]": "Simplify /@ rule",
-    }
-    options = {
-        "Assumptions": "$Assumptions",
-    }
-
-    def apply_with_assumptions(self, expr, assum, evaluation, options={}):
-        "%(name)s[expr_, assum_, OptionsPattern[]]"
-
-        # If the second argument is a rule, it means that
-        # it should be taken as an option.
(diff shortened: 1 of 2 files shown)


mathics/builtin/numbers/algebra.py
```diff
@@ -1106,7 +1120,16 @@ def apply(self, expr, evaluation):
         return variables
 
 
+# This Builtin belongs to mathics.builtin.list.eol
```
rocky, line 1123: Do you mind if we just move it there?

mathics/builtin/numbers/algebra.py
```diff
@@ -381,6 +381,7 @@ class Cancel(Builtin):
      = 2 f[x] / x
     """
 
+    summary_text = "cancels common factors in rational expressions"
```
rocky, line 384: This is weirdness of English which is the the only language which marks 3rd person present tense. 

When I compare with https://reference.wolfram.com/language/guide/RationalFunctions.html I see that the Imperative is used.  "cancel" versus "cancels". 

In my opinion both are correct but we should be consistent and uniform. Elsewhere for `Information` we use the imperative "get".  But we also use "gives" instead of "give". 

We should make everything in the same form. 

Your thoughts?

mathics/builtin/numbers/algebra.py
```diff
@@ -1884,6 +1915,7 @@ class Collect(_CoefficientHandler):
      = x h[4 Sin[x z]] + x ^ 3 h[1] + x y h[3] + x ^ 2 y h[3] + y ^ 2 h[4 Sin[x z]] + x y ^ 2 h[3] + y ^ 3 h[1]
     """
 
+    summary_text = "collect terms with a variable at the same power"
```
rocky, line 1918: Note that in contrast to many of the above, "tests" vs "test" "expands" vs "expand" this uses imperative as seems to be done on the WL site. "list" above is also the imperative tense.

mmatera: List is also a noun. Still, I liked more the present tense.

mathics/builtin/numbers/algebra.py
mmatera: Go ahead. I did not do here to make this PR shorter and clearer

mathics/builtin/numbers/algebra.py
rocky: You are correct. In that context "list of ..." it is a noun.

rocky:
I will admit that I probably shouldn't have done the method sorting as a separate PR since this makes it hard to see what's happened in the summary. 

@mmatera  - If this is too great a change, I can go back and unsort methods.

mathics/builtin/numbers/algebra.py
TiagoCavalcante: @rocky it is going to be displayed like this: `F[] gives` or `F[] give`, so I'm for the 3rd person.

rocky: @TiagoCavalcante There is a big misunderstanding here. So let me recap. 

I added "summary_text" originally as summary text. That is, these are used in description lists in summaries in Django for as you find in the section on "Tensors. It is [the way it is used on the Wolfram Reference site](https://reference.wolfram.com/language/guide/Tensors.html). 

Later, I did this hack to also use it as the "short description" as is found in `?variable`. Previously there was the full description with pseudo XML markup. So I thought this might be an improvement because it was less ugly. 

However in subsequent events and investigation I now realize this was a mistake. This mistake was uncovered when it was then extended _again_ to include the text given for the `::usage` variable.

In investigating the behavior of `Information` --  the underlying function behind  `?`, `??` and `::usage` --  we find that:

First,  in the latest versions of Mathematica there is no difference anymore between `?` and `??` they both are the definition field of an `InformationData` structure.  

Second, in contrast to "summary_text" which is perfectly fine as a string, this is not fine for the definition field of an `InformationData` structure which is some sort of Box-like structure. The advantage of the Box-like structure is that it can hold several different definitions whereas the string doesn't, unless you add more hacks which I strongly feel we shouldn't be doing. I would like to see fewer hacks. 

In sum:  __`summary_text` is summary text, not an InformationData definition field.__

Now there was mention of using some sort of fancy programming to generate all of this. If that consults summary_text , then that can change the verb as appropriate. But I don't know that it should since, that is about all forms not one the summary of all forms.

mmatera: My reasons to use present tense instead of imperative were two. 1. avoids to look that we just reaped them from WR and 2. ``summary_text`` would be a description of a symbol. So, in my very basic understanding of English grammar, descriptions are typically written in the simple present tense. For me, it also would make sense to use nominal phrases,  but I do not see how imperative phrases fit here. In any case, on this point, I would take  @rocky 's position,  as a native speaker, and because is who introduced the feature in the first place.

mmatera:
> I will admit that I probably shouldn't have done the method sorting as a separate PR since this makes it hard to see what's happened in the summary.
> 
> @mmatera - If this is too great a change, I can go back and unsort methods.

For me it is fine. I just wanted to make easier this PR to review for you and @TiagoCavalcante

mathics/builtin/numbers/algebra.py
mmatera: Regarding ``Information`` and ``usage`` strings, I just misunderstood the goal of the ``summary_text`` property. In that regard, I opened some days ago a discussion  (#236) to collect ideas about how to refactor the implementation of ``Information``

rocky: > My reasons to use present tense instead of imperative were two. 1. avoids to look that we just reaped them from WR

I too have thought about copying verbatim too much and have pondered over it. There is such a thing as fair use. Each summary is less than one sentence long and there are some places where our text is different even in this PR. I think it had to do with using the phrase "numerators and denominators" vs. "factors", but I don't have this code checked out to check it.

Not here, but more generally, any text that looks more like advertising hype rather than useful information I have been removing and replacing with what I hope is more useful user information. These generally occur in the top-level information rather than the summaries.

There is a difference between copying text and keeping the grammar forms the same. I don't know of any instance where there is a copyright violation based on the principle that the sentence phrases are in the 3rd person tense versus imperative tense.  Just as we try to organize the doc and modular structure around the structure of WL because that seems like a reasonable organization, here I was doing it with the grammar presentation. The less we have to think about stuff like this - just do it the same way - the better off we will be in the future. I realize it is much easier for me to follow the pattern "use imperative tense in summaries"  than it is for others. So in compensation, I am happy just to go over the silly stupid stuff like this. 


 > and 2. `summary_text` would be a description of a symbol. 

While the copying aspect I have thought a bit about, for the summaries themselves I have just been looking at the Django page with the list of summaries (again such as for Tensor) and I say: does this look reasonable? Is it about the same kind of feel you get when looking the corresponding guide?  Here the answer was yes, so good. When I see in a single list that verb tenses change, that to me is distracting. 

I haven't been trying to make an opinion about which would be the best verb tense to select if I were designing summary documentation.

Like many things Wolfram Language related, that verb form might not have been my choice but it is a _reasonable_ choice and probably was deliberate. So I see no reason to change it. 


> So, in my very basic understanding of English grammar, descriptions are typically written in the simple present tense. For me, it also would make sense to use nominal phrases, but I do not see how imperative phrases fit here. In any case, on this point, I would take @rocky 's position, as a native speaker, and because is who introduced the feature in the first place.

Thanks for your trust and confidence and I will try not to abuse it.

mmatera: @rocky, thanks for your patient explanation. So I am going forward with imperative tense summaries.

rocky:
Let's merge and continue if there is more stuff needed. (Well there is, I will be doing the stupid stuff of changing tense in other summaries when they aren't imperative.)

PR #250, opened 2022-03-28
WIP - First cut elements properties

rocky:
This is to give people some sense of what's up here. The useful properties aren't used yet.

Elsewhere I will provide a more full description of the observations, motivation, deficiencies and possible future direction.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/250"></a>

### DIFF
--- a/mathics/builtin/base.py
+++ b/mathics/builtin/base.py
@@ -432,18 +432,6 @@ class Builtin:
                     return s.get_name()[len(prefix) :], s
         return None, s
 
-    @property
-    def is_literal(self) -> bool:
-        """
-        True if the value can't change, i.e. a value is set and it does not
-        depend on definition bindings. That is why, in contrast to
-        `is_uncertain_final_definitions()` we don't need a `definitions`
-        parameter.
-        """
-        # FIXME: figure out what the right thing to do here is.
-        # For now we will be pessimistic.
-        return False
-
 
 class InstanceableBuiltin(Builtin):
     def __new__(cls, *args, **kwargs):
@@ -467,6 +455,18 @@ class InstanceableBuiltin(Builtin):
 
     def init(self, *args, **kwargs):
         pass
+
+    @property
+    def is_literal(self) -> bool:
+        """
+        True if the value can't change, i.e. a value is set and it does not
+        depend on definition bindings. That is why, in contrast to
+        `is_uncertain_final_definitions()` we don't need a `definitions`
+        parameter.
+        """
+        # FIXME: figure out what the right thing to do here is.
+        # For now we will be pessimistic.
+        return False
 
 
 class AtomBuiltin(Builtin):
--- a/mathics/builtin/base.py
+++ b/mathics/builtin/base.py
@@ -68,7 +68,7 @@ def has_option(options, name, evaluation):
 mathics_to_python = {}
 
 
-class Builtin(object):
+class Builtin:
     """
     This class is the base class for Builtin symbol definitions.
 
@@ -104,7 +104,7 @@ class Builtin(object):
     ```
     produces a ``Definitions`` object with just one definition, for the ``Symbol`` ``System`List``.
 
-    Notice that for creating a Bultinin, we must pass to the constructor the option ``expression=False``. Otherwise,
+    Notice that for creating a Builtin, we must pass to the constructor the option ``expression=False``. Otherwise,
     an Expression object is created, with the ``Symbol`` associated to the definition as the ``Head``.
     For example,
 
@@ -431,6 +431,18 @@ class Builtin(object):
                 if s.get_name().startswith(prefix):
                     return s.get_name()[len(prefix) :], s
         return None, s
+
+    @property
+    def is_literal(self) -> bool:
+        """
+        True if the value can't change, i.e. a value is set and it does not
+        depend on definition bindings. That is why, in contrast to
+        `is_uncertain_final_definitions()` we don't need a `definitions`
+        parameter.
+        """
+        # FIXME: figure out what the right thing to do here is.
+        # For now we will be pessimistic.
+        return False
 
 
 class InstanceableBuiltin(Builtin):
--- a/mathics/core/element.py
+++ b/mathics/core/element.py
@@ -444,6 +444,18 @@ class BaseElement(KeyComparable):
     def get_string_value(self):
         return None
 
+    @property
+    def is_literal(self) -> bool:
+        """
+        True if the value can't change, i.e. a value is set and it does not
+        depend on definition bindings. That is why, in contrast to
+        `is_uncertain_final_definitions()`, we don't need a `definitions`
+        parameter.
+
+        Each subclass should decide what is right here.
+        """
+        raise NotImplementedError
+
     def is_uncertain_final_definitions(self, definitions) -> bool:
         """
         Used in Expression.do_format() to determine if we should (re)evaluate
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -265,7 +265,6 @@ class Expression(BaseElement, NumericOperators):
         last_element = None
         for element in elements:
             converted_elt = from_python(element)
-            last_element = converted_elt
 
             # Test for the three properties mentioned above.
             if not converted_elt.is_literal:
@@ -278,6 +277,7 @@ class Expression(BaseElement, NumericOperators):
                 and last_element != converted_elt
             ):
                 self._is_sorted = False
+            last_element = converted_elt
 
             result.append(converted_elt)
 
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -265,19 +265,20 @@ class Expression(BaseElement, NumericOperators):
         last_element = None
         for element in elements:
             converted_elt = from_python(element)
+            last_element = converted_elt
+
+            # Test for the three properties mentioned above.
+            if not converted_elt.is_literal:
+                self._elements_fully_evaluated = False
+            if isinstance(converted_elt, Expression):
+                self._is_flat = False
             if (
                 self._is_sorted
                 and last_element is not None
                 and last_element != converted_elt
             ):
                 self._is_sorted = False
-            last_element = converted_elt
-            if isinstance(converted_elt, Expression):
-                self._is_flat = False
-                if not converted_elt.is_literal:
-                    self._elements_fully_evaluated = False
-            elif not converted_elt.is_literal:
-                self._elements_fully_evaluated = False
+
             result.append(converted_elt)
 
         return tuple(result)
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -6,7 +6,7 @@ import math
 import time
 
 import typing
-from typing import Any, Optional
+from typing import Any, Iterable, Optional
 from itertools import chain
 from bisect import bisect_left
 
@@ -176,12 +176,19 @@ class Expression(BaseElement, NumericOperators):
     # See if there's a way to get rid of this, or ensure that this isn't causing
     # a garbage collection problem.
 
-    def __new__(cls, head, *elts, **kwargs) -> "Expression":
+    def __new__(cls, head, *elements, **kwargs) -> "Expression":
         self = super().__new__(cls)
         if isinstance(head, str):
             head = Symbol(head)
         self._head = head
-        self._elements = tuple(from_python(element) for element in elts)
+
+        # Set some properties on elements that help us speed up evaluation.
+        # These are set in self._build_elements(elements)
+        #    self._elements_fully_evaluated, self._is_flat, self._is_sorted
+
+        self._elements = self._build_elements(elements)
+
+        # self._elements = tuple(from_python(element) for element in elts)
         self._sequences = None
         self._cache = None
         # comment @mmatera: this cache should be useful in BoxConstruct, but not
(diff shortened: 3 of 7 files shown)


mathics/builtin/base.py
```diff
@@ -432,6 +432,18 @@ def get_option_string(self, *params):
                     return s.get_name()[len(prefix) :], s
         return None, s
 
+    @property
+    def is_literal(self) -> bool:
```
mmatera, line 436: @rocky, I have a question: since just ``Expression`` and  ``Symbol`` are the only classes that should have a non-trivial ``evaluation`` method, wouldn't be enough to check class membership? (with ```isinstance(expr, (Symbol, Expression))```)

Furthermore, I imagined, for example, to have a superclass ``Evaluable`` from which ``Expression`` and ``[Evaluable]Symbol`` could have as a superclass (together with ``Element``).

rocky: > @rocky, I have a question: since just `Expression` and `Symbol` are the only classes that should have a non-trivial `evaluation` method, wouldn't be enough to check class membership? (with `isinstance(expr, (Symbol, Expression))`)
> 

This is dangerous kind of thinking. Note that Boxing also has a `is_literal` as well.  Making this a property things break when the class structure changes. Using `isinstance()` things may break in a way that is silent and hard to track down. 


> Furthermore, I imagined, for example, to have a superclass `Evaluable` from which `Expression` and `[Evaluable]Symbol` could have as a superclass (together with `Element`).

First the superficial comment: I never understood or like the word Instanceable. What in plain simple dumb longer-form language does Evaluable mean. It has an evaluate function? Or something more? What does Instanceable mean in longer dumb form? 

The more fundamental worry which is really the same concern as using `isinstance()` is that basically OO is evil because it has unintended consequences. See "Inheretance vs Delegation" in https://www.cs.unc.edu/~dewan/comp401/current/Class%20Notes/22_Delegation_Notes.pdf

We sometimes use what in Ruby is called a Mixin to add particular attributes and properties. For example I did this with NumericOperators.  But I suspect we should be careful and the current thought is not to rely too much on OO strict inheritance. Maybe this Evaluable thing is a Mixin, I don't know.  Rather than describe what the mechanism is, describe what it is you want to achieve.  

The docstring for `is_literal()` tries to describe why that exists and if that's not enough I've written elsewhere why it might be useful. If there are other things, please let me know and describe what's up?

rocky: I should also say that if `is_literal` is totally meaningless for the Builtin object, then the caller should be fixed. I wonder _why_ it is that this caller is calling "evaluate" on a builtin object that has no notion of whether it is a literal or not. 

Literalness or constantness _is_ a property of evaluation.

mathics/core/expression.py
```diff
@@ -232,6 +239,50 @@ def _flatten_sequence(self, sequence, evaluation) -> "Expression":
...
+        result = []
+        last_element = None
+        for element in elements:
+            converted_elt = from_python(element)
+            last_element = converted_elt
+
+            # Test for the three properties mentioned above.
+            if not converted_elt.is_literal:
+                self._elements_fully_evaluated = False
+            if isinstance(converted_elt, Expression):
```
rocky, line 273: Might consider something like `hasattr(converted_elt, "_elements")` ?

mathics/builtin/base.py
mmatera: > 
> First the superficial comment: I never understood or liked the word Instanceable. What in plain simple dumb longer-form language does Evaluable mean? It has an evaluate function? Or something more? What does Instanceable mean in longer dumb form?
> 

Instanceable has (at least for me) a clear meaning in Mathics: a (base) ``Builtin`` is a class that just stores information about how to build a ``Definition`` object. If you try to instanciate it, what you get is an ``Expression`` object.

 On the other hand, an ``InstanceableBuiltin`` is a kind of object that -apart from storing information about how to build a ``Definition``, can also be used as an ``Element`` in an expression. ``PatternObject`` is an example of this kind of classes.  It contains information that could be stored in an  ``Expression`` object, but the way in which the information is stored, and the methods are implemented are particular for each InstanceableBuiltin class. 



> The more fundamental worry which is really the same concern as using `isinstance()` is that basically OO is evil because it has unintended consequences because too much is implicit rather than explicit. See "Inheretance vs Delegation" in https://www.cs.unc.edu/~dewan/comp401/current/Class%20Notes/22_Delegation_Notes.pdf
>

OK, I am going to have a look
 
> We sometimes use what in Ruby is called a Mixin to add particular attributes and properties. For example I did this with NumericOperators. But I suspect we should be careful and the current thought is not to rely too much on OO strict inheritance. Maybe this Evaluable thing is a Mixin, I don't know. Rather than describe what the mechanism is, describe what it is you want to achieve.
> 

Yes, what I imagined that ``Evaluable`` could be something that supports an ``evaluate`` method, and then, an Expression (and a Symbol) would be at the time ``Element`` and ``Evaluable``. And if you define an ``InstancealeBuiltin``, you can decide if make it or not ``Evaluable``.  All the point is if you want to have an ``evaluate`` method in the base class ``Element`` or just to have them in objects on what you want to call it.




> The docstring for `is_literal()` tries to describe why that exists and if that's not enough I've written elsewhere why it might be useful. If there are other things, please let me know and describe what's up?


Yes, the docstring is clear about what this method is supposed to do. My question was about the reason for this implementation choice.

> I should also say that if is_literal is totally meaningless for the Builtin object, then the caller should be fixed. I wonder why it is that this caller is calling "evaluate" on a builtin object that has no notion of whether it is a literal or not.

a Builtin shouldn't have a ``is_literal`` because it is not an ``Element``. On the other hand, ``InstanceableBuiltin`` should have it, because they are at the time ``Builtin`` and ``Element``.

mathics/core/element.py
```diff
@@ -444,6 +444,18 @@ def get_sequence(self):
...
+    def is_literal(self) -> bool:
+        """
+        True if the value can't change, i.e. a value is set and it does not
+        depend on definition bindings. That is why, in contrast to
+        `is_uncertain_final_definitions()`, we don't need a `definitions`
+        parameter.
+
+        Each subclass should decide what is right here.
+        """
+        raise NotImplementedError
```
mmatera, line 457: OK, but at least ``Number``, ``String`` and  ``ByteArrayAtom`` should implement it and return ``True``.

mathics/core/expression.py
```diff
@@ -905,6 +956,20 @@ def get_sort_key(self, pattern_sort=False):
...
+    @property
+    def is_literal(self) -> bool:
+        """
+        True if the value can't change, i.e. a value is set and it does not
+        depend on definition bindings. That is why, in contrast to
+        `is_uncertain_final_definitions()` we don't need a `definitions`
+        parameter.
+        """
+        # Right now we are pessimisitic. We might consider changing this for
+        # Lists. Lists definitions can't be changed right?
```
mmatera, line 968: A list of numbers shouldn't change. For this, and other structures (like sparse arrays, associations, or subexpressions) I would like to consider using ``InstanceableBuiltin`` specific classes (@rocky, I know, you hate this name, but maybe we can find a better one...) or maybe classes that implement a part of the interface of ``Expression`` but that stores elements in a more suitable way for the specific propose.

mathics/core/symbols.py
```diff
@@ -454,6 +471,18 @@ def get_head_name(self):
...
 
+    @property
+    def is_literal(self) -> bool:
+        """
+        True if the value can't change, i.e. a value is set and it does not
+        depend on definition bindings. That is why, in contrast to
+        `is_uncertain_final_definitions()` we don't need a `definitions`
+        parameter.
+
+        Here, we have to be pessimistic and return True.
```
mmatera, line 482: here, we return ``False``

mathics/builtin/base.py
rocky: > a Builtin shouldn't have a `is_literal` because it is not an `Element`. On the other hand, `InstanceableBuiltin` should have it, because they are at the time `Builtin` and `Element`.

Ok. Something in boxing is treating a GraphicsBox as though it were an Element: 

```
  File "mathics/docpipeline.py", line 117, in test_case
    result = evaluation.evaluate(query)
  File "github/Mathics3/mathics-core/mathics/core/evaluation.py", line 343, in evaluate
    result = run_with_timeout_and_stack(evaluate, timeout, self)
  File "github/Mathics3/mathics-core/mathics/core/evaluation.py", line 98, in run_with_timeout_and_stack
    return request()
  File "github/Mathics3/mathics-core/mathics/core/evaluation.py", line 336, in evaluate
    return self.format_output(self.last_eval, format)
  File "github/Mathics3/mathics-core/mathics/core/evaluation.py", line 429, in format_output
    result = expr.format(self, "System`OutputForm")
  File "github/Mathics3/mathics-core/mathics/core/element.py", line 305, in format
    expr = self.do_format(evaluation, form)
  File "github/Mathics3/mathics-core/mathics/core/expression.py", line 550, in do_format
    expr = super().do_format(evaluation, form)
  File "github/Mathics3/mathics-core/mathics/core/element.py", line 262, in do_format
    expr = self.create_expression(
  File "github/Mathics3/mathics-core/mathics/core/expression.py", line 1772, in _create_expression
    return Expression(head, *elements)
  File "github/Mathics3/mathics-core/mathics/core/expression.py", line 189, in __new__
    self._elements = self._build_elements(elements)
  File "github/Mathics3/mathics-core/mathics/core/expression.py", line 271, in _build_elements
    if not converted_elt.is_literal:
AttributeError: 'GraphicsBox' object has no attribute 'is_literal'
```

mmatera: Add. It to instanceblebuiltin.

rocky: > Instanceable has (at least for me) a clear meaning in Mathics: a (base) `Builtin` is a class that just stores information about how to build a `Definition` object. If you try to instan[t]iate it, what you get is an `Expression` object.

Begin rant: 

I have no doubt that it has a meaning. In Mathics we call unescaping a string "escaping" it. And we call the nodes of a tree or an expression the "leaves".  And we name an evaluation method "apply". And we call incomplete HTML "XML".  (And I could go on with many other examples). 

Honestly I don't see the connection between Instanceable which isn't an English word and what you described in any way other than the most vague sense.

"Instanceable"  is someone trying to sound clever uses but doesn't realize they are being stupid. To the extent Instanceable has meaning it means that it is something that you are able create an instance (or object) of. Python  claims since 15 years ago everything is now an object, so that means very little: `str()` is an "instanceable" constructor because I can write `str(x)` and get back an instance of a string. 

I gather that this Instanceable thing is tied with Mathics expressions (not to be confused with the Mathics Expression class which refers to a compound expression).   Mathics expressions (derived from BaseElement) are very specific kinds of objects. 

So I kind of see how Instanceable fits in with the pattern of I'll just use a generic term like "data" and assume you know what I really am talking about. 

End rant. 

Some other time I will go over what you write and ask questions. I would like to be able to describe this in a more conventional and less vague way which seems very specific to Mathics.  That is where we probably need to start.

WL doesn't have a notion of Instanceable that I can find. Nor does it call "elements" "leaves", and so on.

mathics/core/expression.py
rocky: What I more dislike is talking about mechanisms ("consider using some particular class) rather than what we are trying to accomplish.

The specific question here. is about whether the function "List" function is allowed to be redefined it the meaning of in an S-expression such as `List[1, 2]` can vary.  I get that _after_ evaluation we may want a better or more efficient _representation_ of the result, but that was not the question here.

mathics/core/symbols.py
rocky: Thanks  - the docstring has been corrected.

mathics/builtin/base.py
rocky: > Add. It to instanceblebuiltin.

Ok - will do when I have time. Thanks!

rocky:
I looked at this InstanceableBuiltin class, and in doing that I get a different impression of what is going on there versus what was described about it here. Let me describe my view of that screwy class. And then I'll describe what to do about it (short story - ignore it), and where we are currently at.

First of all, as recall that 99% of all Python classes start out non-shared except in bytecode and code properties (name of file, location where the module for it is in a filesystem, etc). We typically say that when you create an object you are creating an _instance_ of that object. 

So now we come to this rare situation in 1% of all Python code, where you don't want the instance data to different, but instead be the same thing. This is called a "singleton" pattern. And one way to do this in Python is by hooking into  `__new__()` which has access to the object instance data portion.  Okay so at this point, the word "instance" in the name is misleading!

So now like any mystery story or "whodunnit" novel which is _not_ the way code should be written, we come to the last plot twist where although we want some parts shared, we want other parts to be not shared. So in the storage allocator code in `__new()__` there is code to change that. So maybe a more sophisticated and confusing name would be: UnUnSharedExceptWhereItIsShared.  Yes, this is silly and an exaggeration, but I use it to make a point. We are focusing on some particular behavior while completely missing what is important in giving a class a name which is what are we trying to do? 

Before coming to that, let me describe what I think would be the more conventional way to to this. You have a kind of unshared object, created using `__new()__`,  and from the usual `__init__()` object get access to this by _delegation_ or storing a pointer to this shared object from inside the normal unshared object instance. 

Ok - so now let's please forget about all of this mechanism with its the focus on sharing, and let us instead concentrate on the bigger picture. So here it is...

As we describe in the REPL pipeline diagram under [Session](https://mathics-development-guide.readthedocs.io/en/latest/extending/code-overview/session.html), there are basically two kinds of evaluation steps, one to produce an expression result from the user's input and another expression result to format the result so that it can be displayed.   

Displaying the resulting formatted expression may involve even more steps which may occur outside of Mathics Core, but presumably that result is in an expression form that easily used by whatever is going to render the result in the front end. For example the result is can be a single Atom like a String or a Number, and from that the front end can extract a value from a method call on the expression object.

The second kind of evaluation is usually more limited and simpler (conceptually if not necessarily in algorithm) because transformations go on only at fewer elements (nodes of the expression tree) which have Boxing routines associated with it. And the Boxing routines many times don't need to do a lot of computation (except where graphics is involved).  So the concept of whether formatting elements need to be sorted, ordered, flattened, listed, unevaluated, or threaded don't exist. 

_As a result, there was an (implicit?) observation that maybe a custom or specialized kind of expression would be useful here. It might use a different and more storage efficient representation of the parts of an element._

So we have someone coming along and thinking, hey, I now get this evaluation pipeline thing and I want specialization too for other kinds of expressions even though it is not related to formatting. I just learned about a literal value in programming languages and that seems like something that should be a specialized kind of expression like we have for formatted expressions/objects. 

Sure, great! Good observation. But let's deal with how to handle expression customization as its own problem and think about it new, independent of what feels to me the screwy way that it has been done for formatting or pattern objects. 

In this draft, I haven't gotten there yet. even though it may be fresh in mind. We are still working on the part _before the first real evaluation that takes place_. 

I am and have been well aware that alternate representations of an expression are desirable. That part will be coming up after we get through `rewrite_apply_eval` and after the pattern matching and rule replacement that follows that. 

This will be handled when we get to evaluation proper (misnamed the "apply" methods). I suspect lazy determination of alternate representations is the best way and place to handle this. 

However it will probably be true that after we do, that we'll have to go back into the other steps, and code and maybe classes that came before and alter these.

PR #253, opened 2022-03-28
Fix sympy_factor to allow equalities

hrueter:
This continues the fix PR #232 to the `Cancel` function, which made use of the same code.
Without it Mathics crashes from this input
`In[1]:= a == d b + d c // Factor`


TiagoCavalcante (approved):
LGTM - I haven't found any other case where `as_numer_denom` can be easily replaced.

PR #254, opened 2022-03-29
Handle expressions not representable by SymPy

hrueter:
This deals with some cases where it is safe to return the expression, when it cannot be handled by SymPy.

### DIFF
--- a/mathics/builtin/comparison.py
+++ b/mathics/builtin/comparison.py
@@ -510,6 +510,10 @@ def do_cmp(x1, x2) -> Optional[int]:
     s1 = x1.to_sympy()
     s2 = x2.to_sympy()
 
+    # return None if expression cannot be handled by SymPy
+    if s1 is None or s2 is None:
+        return None
+
     # Use internal comparisons only for Real which is uses
     # WL's interpretation of equal (which allows for slop
     # in the least significant digit of precision), and use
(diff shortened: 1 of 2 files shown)


TiagoCavalcante (approved):
LGTM

mmatera:
@hrueter,  could you some examples in which this change has an impact?

rocky:
> @hrueter, could you some examples in which this change has an impact?

Even better would be to add a test.

mmatera:
@hreuter, is this PR still valid?

hrueter:
> @hreuter, is this PR still valid?

You got my github handle wrong. ;-)

I will check whether this is still relevant. But first some new bugs have to be fixed, so that I can run my old test problem again.

hrueter:
I have added tests for the three functions that I am fixing here.
The test cases may seem a bit artificial, but they represent what I actually use in some WMA scripts.

hrueter:
The added documentation text is a little bit rough. If you have suggestions to describe what these tests actually prove, let me know.
WL does not really document what is supposed to happen in these cases, but this is what is observed in WMA.

rocky:
> The added documentation text is a little bit rough. If you have suggestions to describe what these tests actually prove, let me know. WL does not really document what is supposed to happen in these cases, but this is what is observed in WMA.

Thanks. I will try to look at later today.

rocky:
LGTM - thanks.

PR #258, opened 2022-04-02
fixing docstrings

mmatera:
more fixing in documentation and summary_texts

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/258"></a>


mathics/builtin/atomic/strings.py
```diff
@@ -382,9 +384,16 @@ def to_python_encoding(encoding):
...
+    <dl>
+    <dt>'$CharacterEncodings'
+    <dd>stores the list of available character encodings.
+    </dl>
+    """
+
     name = "$CharacterEncodings"
     value = "{%s}" % ",".join(map(lambda s: '"%s"' % s, _encodings.keys()))
-
+    summary_text = "avaliable character encodings"
```
rocky, line 396: avaliable -> available

mathics/builtin/atomic/strings.py
```diff
@@ -450,6 +462,7 @@ class Alphabet(Builtin):
 
     """
 
+    summary_text = "lowercase letters in an alphabet"
```
rocky, line 465: Would be nice to move this and the next one after messages.

mathics/builtin/inout.py
```diff
@@ -2164,6 +2218,8 @@ class SympyForm(Builtin):
     = exp(2) + 3*E
     """
 
+    summary_text = "translate expressions to sympy"
```
rocky, line 2221: BTW the person who worked on the project referred to the project as SymPy.

rocky (approved):
LGTM - but see comments.

mmatera:
> Why do we need value at all here? Is it anywhere except in that rule. If so, it would be nice to explain what it use is.

I do not know, but changing it would be a detour from the goal of this PR. Let's merge and iterate. In next PRs I am going to continue sorting the properties

PR #260, opened 2022-04-04
Expression tweaks

rocky:
Misc expression-related things found along the way
    
* Uppercase constant attribute flags (in evaluation module) and put that in alphabetic order
* Convert comments before classes as class doc strings
* Fix some small typos


mathics/core/expression.py
```diff
@@ -6,11 +6,25 @@
...
 
 import typing
-from typing import Any, Optional
+from typing import Any, Iterable, Optional
 from itertools import chain
 from bisect import bisect_left
 
 from mathics.core.atoms import from_python, Number, Integer
+
+# FIXME: adjust mathics.core.attributes to uppercase attribute names
```
rocky, line 15: @TiagoCavalcante would love to adjust mathics.core.attributes, but I ran out of time. 

What do you think? If you think this makes it clearer would you adjust the rest of the code? 

Basically when I was reading the code and saw flat, it took me a while to figure where this was set. Had it been uppercase that would have cleared things up.

TiagoCavalcante (approved):
LGTM

rocky:
Basically I made no progress in the expression-elements-properties branch because things are thornier. Things get converted into Expressions and updated mysteriously and which makes keeping elements properties of the expression harder than it should be. 

This branch is the part that is largely independent of that which I think moves the code in a more modular and maintainable fasihion.

The things here though are of the more generic kinds of refactoring that I think help. If this merges I will

mathics/core/expression.py
TiagoCavalcante: I'm not sure about this, the attributes are constants,

mmatera:
LGTM

mathics/core/expression.py
rocky: In C and other languages all uppercase indicates literal values (in C, C preprocessor literals). Even in Python there is some precedent for this. See https://docs.python.org/3/library/stat.html?module-stat#stat.ST_MODE

The other thing I had considered was to add `_attr` or something like that at the end of an attribute. A name like "flat"  to indicate an  constant attribute flag value is weird. And as I said I had to look around to find out what was up. Even _after_ I saw the import it took me another second to realize what was up. If this were marked a constant or named in fashion that suggests bitmap then things would have been fine.

PR #262, opened 2022-04-10
4 distinct of changes...

rocky:
* Add ImmutableValueMixin for objects that are literal-like and don't
  require evalation after creating. Use this property various place
* Go over Builtin docstring
* _Color -> _ColorObject
* remove named temporary files and 
* remove named temporary files and give them a recognizable preface

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/262"></a>

### DIFF
--- a/mathics/builtin/base.py
+++ b/mathics/builtin/base.py
@@ -92,11 +92,11 @@ mathics_to_python = {}
 
 class Builtin:
     """
-    A base class for a Built-in function symbols, like Integer, or variables like $SystemID,
-    and Built-in Objects like DateTimeObject
+    A base class for a Built-in function symbols, like List, or variables, like $SystemID,
+    and Built-in Objects, like DateTimeObject.
 
     Some of the class variables of the Builtin object are used to
-    create a definition object that built-in Symbol.  In particular,
+    create a definition object for that built-in symbol.  In particular,
     there are (transformation) rules, attributes, (error) messages,
     options, and other things.
 
@@ -105,6 +105,8 @@ class Builtin:
 
     Method names of a builtin-class that start with the word ``apply`` are evaluation methods that
     will get called when the docstring of that method matches the expression to be evaluated.
+
+    For example:
 
     ```
         def apply(x, evaluation):
@@ -473,10 +475,10 @@ class Builtin:
         depend on definition bindings. That is why, in contrast to
         `is_uncertain_final_definitions()` we don't need a `definitions`
         parameter.
+
+        Each subclass should decide what is right here.
         """
-        # FIXME: figure out what the right thing to do here is.
-        # For now we will be pessimistic.
-        return False
+        raise NotImplementedError
 
 
 class InstanceableBuiltin(Builtin):
--- a/mathics/builtin/base.py
+++ b/mathics/builtin/base.py
@@ -92,22 +92,35 @@ mathics_to_python = {}
 
 class Builtin:
     """
-    This class is the base class for Builtin symbol definitions.
+    A base class for a Built-in function symbols, like Integer, or variables like $SystemID,
+    and Built-in Objects like DateTimeObject
 
-    A Builtin class is a structure that holds the information needed to generate
-    a Definition object for a built-in Symbol, like transformation rules, attributes, options,
-    etc.
-    Method with names of the form ``apply*`` are considered replacement rules, for expressions
-    that match the pattern described in their docstrings. For example
+    Some of the class variables of the Builtin object are used to
+    create a definition object that built-in Symbol.  In particular,
+    there are (transformation) rules, attributes, (error) messages,
+    options, and other things.
+
+    Function application pattern matching
+    -------------------------------------
+
+    Method names of a builtin-class that start with the word ``apply`` are evaluation methods that
+    will get called when the docstring of that method matches the expression to be evaluated.
 
     ```
         def apply(x, evaluation):
-             '''F[x_Real]'''
+             "F[x_Real]"
              return Expression("G", x*2)
     ```
-    adds a ``BuilitinRule`` that implements ``F[x_]->G[x*2].
+
+    adds a ``BuiltinRule`` to the symbol's definition object that implements ``F[x_]->G[x*2]``.
+
+    As shown in the example above, leading argument names of the
+    function are the arguments mentioned in the names given up to the
+    first underscore ``_``.  So the single parameter in the above is
+    ``x``. The method must also have an evaluation parameter, and may
+    have an optional `options` parameter.
+
     If the ``apply*`` method returns ``None``, the replacement fails, and the expression keeps its original form.
-    Notice that the argument names must coincide with the name of the corresponding pattern in the docstring.
 
     For rules including ``OptionsPattern``
     ```
@@ -142,7 +155,6 @@ class Builtin:
     ```
     expr_list = Expression(SymbolList, Integer(1), Integer(2), Integer(3))
     ```
-
     """
 
     name: typing.Optional[str] = None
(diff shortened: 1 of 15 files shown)


mathics/builtin/base.py
```diff
@@ -70,22 +70,35 @@ def has_option(options, name, evaluation):
 
 class Builtin:
```
rocky, line 71: Why is this not called BuiltinBase like our other Base classes?

mmatera: maybe because we want to say that ``ArcTan`` is a ``Builtin`` symbol.

mathics/builtin/base.py
```diff
@@ -456,20 +480,13 @@ def __new__(cls, *args, **kwargs):
...
-        depend on definition bindings. That is why, in contrast to
-        `is_uncertain_final_definitions()` we don't need a `definitions`
-        parameter.
-        """
-        # FIXME: figure out what the right thing to do here is.
-        # For now we will be pessimistic.
-        return False
-
 
 class AtomBuiltin(Builtin):
```
mmatera, line 484: Maybe this should be a subclass of ``InstanceableBuiltin`` (or whatever be a better name for it), or vice-versa. Now, I think this is confusing...

mathics/core/symbols.py
```diff
@@ -194,20 +195,20 @@ def __cmp(self, other) -> int:
         return 0
 
 
-class Atom(BaseElement):
+class Atom(BaseElement, ImmutableValueMixin):
```
mmatera, line 198: ``Symbol`` is a subclass of ``Atom``, but must be evaluated. Maybe the feature ``InmutableValueMinin`` must be added to the subclasses (``Number``, ``String``, ``BitArray``, etc)

mathics/builtin/base.py
rocky: I don't understand. I might like to say ArcTan is a builtin function. Or that it is a trigonometric builtin function or many things like that. If there were common properties of built-in functions as opposed to built-in variables, then sure we could create a BuiltinFunction class which is a subclass of the Builtin (base class). 

But right now as far as I can tell, there are not properties of builtin functions that we seem to want to capture. Or trigonometric builtin functions either. And when they appear we can add them.

And `Builtin` is also used for builtin _objects_ like `DateTimeObject` and `ColorObject`. 

We work in Python where we build stuff up from. From an OO programmer's point of view, and a programmer using Mathics in particular, there are things like Base classes that you don't create instances from directly but must be subclassed. And `Builtin` is one of them. 

One of the things I personally find annoying is the looseness, vagueness (if not sometimes outright wrong) terminology backed by a lack of description of what is meant. And then that has probably lead to the object used in a way that it wasn't intended -- possibly, it is hard to tell. It has had very real tangible consequences that we even figure out how to improve the code because we can't even agree or figure what's going on. Instead it has been more like - here's a problem. Here is some code fix that addresses this. Having code that follows established conventions and principles suggests to us and others to determine when there is a problem how something should be addressed. 

Here, If there is code that instantiates a Builtin object and then starts using it (rather than subclassing it or using one of the subclasses) then that code is probably wrong.

mathics/builtin/base.py
rocky: `InstranceableBuiltin` or `SharedImplmenetationofBuiltinExceptWhenItIsn't` is some implemention of a Builtin. The "FIXME" says that it is a pessimistic guess because we have no idea. And as the FIXME suggests, this is bad. 

Probably instead inside Builtin where this has been moved, we should raise a RuntimeError because each real subclass should do the right thing. Notice that InstanceableBuiltin is _also_ a base class. If you create an object from that directly and start using it (rather than a subclass of that) then you are probably doing something wrong.

mathics/core/symbols.py
rocky: Yes, it _should_ be added to those, and I haven't done those yet, but I will add them.

mathics/builtin/base.py
mmatera: @rocky, OK. My unique concern is that most of the symbols are subclasses of ``Builtin``, so the change would affect almost all the builtin sub-modules. But yes, if you think it is relevant at this step, go ahead.

mathics/core/symbols.py
mmatera: My point is that (at the end) ``Atom`` shouldn't be ``ImmutableValueMixin``.

mathics/builtin/base.py
mmatera: I was not talking about the movement of the ``is_literal``, but about ``AtomBuiltin``which is something that we have had in the code for a while. As you were talking about re-examining this part of the class hierarchy, thought it was useful to say that this class still also confuses me.

mathics/core/symbols.py
rocky: You are correct. It has been removed now. Thanks!

mathics/builtin/base.py
rocky: @TiagoCavalcante - your thoughts?

mathics/builtin/base.py
rocky: @mmatera It confuses me too. I suspect down the line it will be clear how to consolidate classes. Right now I was just being cautious. 

The distinction between Builtin and Atom and why Builtin is not a subclass of Atom feels weird, except when you drill into the details we find that Atom is very heavyweight. And Builtin adds the evaluation convention stuff (how functions are matched, using the method and docstring properties) which is also heavyweight. So combining these probably doesn't make sense. But it also doesn't feel well organizied.

mathics/builtin/base.py
TiagoCavalcante: @rocky I'll say the truth, I don't see a problem with Builtin, it was very clear to me what it made after reading the documentation, and [for some unknown reason] it seems that renaming it to BuiltinBase would make its meaning less clear.

rocky: Ok. Then let's keep the name as it currently is.

rocky:
At this point I think all of the open concerns have been addressed.

mathics/core/symbols.py
```diff
@@ -196,18 +197,18 @@ def __cmp(self, other) -> int:
...
+    In other words, they are the some of units of an expression that we cannot
+    dig down deeper structurally.
 
     Of note is the fact that the Mathics ``Part[]`` function of an
     Atom object does not exist.
 
     Atom is not a directly-mentioned WL entity, although conceptually
     it very much seems to exist.
+
+    There are kind of leaf note is the Builtin, i.e. `ByteArray``, `CompiledCode`` or ``Image``.
```
mmatera, line 211: I do not get this sentence.

mathics/core/symbols.py
```diff
@@ -196,18 +197,18 @@ def __cmp(self, other) -> int:
...
 class Atom(BaseElement):
     """
-    Atoms are the leaves (in the common tree sense, not the Mathics
-    ``_elements`` sense) and Heads of an Expression or M-Expression.
+    Atoms are the (some) leaves and the Heads of an S-Expression or an M-Expression.
 
-    In other words, they are the units of an expression that we cannot
-    dig down deeper structurally.  Various object primitives i.e.
-    ``ByteArray``, `CompiledCode`` or ``Image`` are atoms.
+    In other words, they are the some of units of an expression that we cannot
```
mmatera, line 202: maybe "In other words, they are the expression's elements over which we cannot dig down deeper structurally."

mathics/core/symbols.py
rocky: Yes, this is a mangled sentence. Is this clearer: 

    The other kinds expression leaf is a ``Builtin``, e.g. `ByteArray``, `CompiledCode`` or ``Image``.

mathics/core/symbols.py
rocky: Here is what   I have now: 

    In other words, they are the expression's elements (leaves of the
    expression) which we cannot dig down deeper structurally.

    Of note is the fact that the Mathics ``Part[]`` function of an
    Atom object does not exist.

    Atom is not a directly-mentioned WL entity, although conceptually
    it very much seems to exist.

    The other kinds expression leaf is a Builtin, e.g. `ByteArray``, `CompiledCode`` or ``Image``.

If this is okay, I'll commit and push.

mathics/core/symbols.py
mmatera: OK, better :)

mathics/core/symbols.py
mmatera: OK. LGTM. Thanks!

PR #263, opened 2022-04-11
Revise bug template

rocky:


rocky:
At some point I will add "Reporting a Bug" to the developer docs.

PR #264, opened 2022-04-25
Expression elements properties

rocky:
Track properties of a compound expression: 

* whether the expression is flat
* whether we know the expression to be sorted (a false value includes "don't know if sorted")
* whether we know the expression to be fully evaluated (see caveat above)

In addition we start to mark Box elements as "literal" - that is, the are indivisible from the standpoint of evaluation and you don't "evaluate" these. Instead I guess we run a format-type evaluation method. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/264"></a>

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -78,6 +78,48 @@ symbols_arithmetic_operations = system_symbols(
 )
 
 
+def build_elements_with_properties(self, elements: Iterable) -> tuple:
+    """Build a tuple of Elements noting useful properties such as
+    whether the collection of elements is sorted, flat, or fully
+    evaluated.
+
+    Note: we add or set the following fields:
+      self._elements_fully_evaluated, self._is_flat, and self._is_sorted
+
+    """
+
+    # All of the properties start out optimistic (True) and are reset when that proves wrong.
+
+    # _elements_fully_evaluated is True if all elements have been fully evaluated.
+    # Strings, and Numbers are fully evaluated. Symbols like Null, True, and False may be up for debate.
+    self._elements_fully_evaluated = True
+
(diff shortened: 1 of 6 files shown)


TiagoCavalcante (approved):
LGTM

mmatera:
LGTM

mathics/core/expression.py
```diff
@@ -78,6 +78,48 @@
...
+    # _is_flat is True if all elements are atoms/leaves.
+    self._is_flat = True
+
+    # _is_sorted is True if elements do not need sorting
+    # elements with less than 2 items or all have the same value are sorted.
+    self._is_sorted = True
+
+    result = []
+    last_element = None
+    for element in elements:
```
mmatera, line 106: Just one detail:  in master, Mathics calls ``from_python`` over each element. I do not know if this was a good idea, but we can get a change in the behavior if we do not do that here.

mathics/core/expression.py
```diff
@@ -183,7 +225,13 @@ def __init__(self, head, *elements, **kwargs):
         if isinstance(head, str):
             head = Symbol(head)
         self._head = head
-        self._elements = tuple(from_python(element) for element in elements)
```
mmatera, line 186: This is what I was talking about before.

mathics/core/expression.py
```diff
@@ -78,6 +78,48 @@
 )
 
 
+def build_elements_with_properties(self, elements: Iterable) -> tuple:
```
mmatera, line 81: does this belongs to a class?

rocky:
Thanks everyone. Let's merge and continue....

PR #265, opened 2022-04-25
Revise some expression comments

rocky:
See  specific notes in the file for comments.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/265"></a>

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -636,14 +636,12 @@ class Expression(BaseElement, NumericOperators):
         self,
         evaluation: Evaluation,
     ) -> typing.Type["BaseElement"]:
-        """Apply transformation rules and expression evaluation to `evaluation` via
-        `rewrite_apply_eval_step()` until it tells us to stop or we hit some limit.
+        """
+        Apply transformation rules and expression evaluation to ``evaluation`` via
+        ``rewrite_apply_eval_step()`` until that method tells us to stop,
+        or unti we hit an $IterationLimit or TimeConstrained limit.
 
-        Note that this is a recusive process and
-        `rewrite_apply_eval_step()` may call us recursively.
-
-        Limits are either an evaluation iteration count or a timeout value.
-
+        Evaluation is a recusive:``rewrite_apply_eval_step()`` may call us.
         """
         if evaluation.timeout:
             return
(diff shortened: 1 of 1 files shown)


mathics/core/expression.py
```diff
@@ -636,14 +636,12 @@ def evaluate(
...
-        `rewrite_apply_eval_step()` until it tells us to stop or we hit some limit.
-
-        Note that this is a recusive process and
-        `rewrite_apply_eval_step()` may call us recursively.
-
-        Limits are either an evaluation iteration count or a timeout value.
+        """
+        Apply transformation rules and expression evaluation to ``evaluation`` via
+        ``rewrite_apply_eval_step()`` until that method tells us to stop,
+        or unti we hit an $IterationLimit or TimeConstrained limit.
```
rocky, line 642: I think this sasy exactly the same thing but is shorter and while also specific, e.g. mention of  `$IterationLimit` and `TimeConstrained`

mathics/core/expression.py
```diff
@@ -682,26 +679,25 @@ def evaluate(
...
-                # Trace evaluation...
+                # TraceEvaluation[] logging.
                 if evaluation.definitions.trace_evaluation:
                     evaluation.print_out(
                         "  " * evaluation.recursion_depth + "-> %s" % expr
                     )
                 iteration += 1
-                # Check if the iterationlimit was reached.
-                # we need to check on each step, in case that the expression
-                # changes its value. Maybe there is another way, for example,
```
rocky, line 703: I think that the way this is done is exactly right. I don't see anything suboptimal about this. 

Adding fields in the evaluation, even if it wasn't wrong (and I think it has a good chance of being wrong) would require extra work in maintaining.

mathics/core/expression.py
```diff
@@ -682,26 +679,25 @@ def evaluate(
                 if hasattr(expr, "options") and expr.options:
                     evaluation.options = expr.options
 
-                # This calls evaluate_next. This routine implements a single
-                # step in the evaluation, and determines if a fixed point
```
rocky, line 686: I tend to stay away from fancy words like "fixed point" when simple words like "will not change" will work. 

If the focus were on program correctness or proving mathematical properties of a program then a technical term would be useful. 

Most programmers don't think in terms of fixed points or idempotence, and to understand this code you really don't need to know about that.

mmatera:
LGTM

PR #269, opened 2022-05-01
fix 266

### LINKED ISSUE #266: Any result of Solve with 10 in it does not print 10
## Description

For example, `Solve[3*10^10==(0.1)^2/x^2,x]` gives `{{x -> -5.7735*^-7}, {x -> 5.7735*^-7}}` with no `10` in there.

## How to Reproduce

 `Solve[3*10^10==(0.1)^2/x^2,x]`

## Output Given

 `Solve[3*10^10==(0.1)^2/x^2,x]`

## Expected behavior

`{{x -> -5.7735*10^-7}, {x -> 5.7735*10^-7}}`

## Additional context

This might be because sympy returns `5.7735026918962578e-7`, and Mathics doesn't know how to deal with that.

mmatera:
This should fix #266. In this, there is a little, second-order issue: in WMA, 
```
In[1]:=-1.6*^-19
```
produces the output
```
Output[1]=-1.6 10^-19
```
and 
```
In[2]:=-1.6*^-19 // StandardForm
```
gives
```
Output[2]//StandardForm=-1.6`*^-19
```
while in the graphics interface,
imagen

On the other hand, with the current patch, Mathics gives, for OutputForm
```
Output[1]:=-1.6×10^-19
```
and for StandardForm
```
Output[2]//StandardForm=-1.6`*^-19
```

The difference seems to come from the way in which "×" is interpreted in the `boxes_to_text` method. We can take care of this after the refactor of ``MakeBoxes``


mathics/builtin/arithmetic.py
```diff
@@ -732,35 +732,35 @@ class Real_(Builtin):
 
     ## Formatting tests
     #> 1. * 10^6
```
rocky, line 734: Of course, in the long term we'd like to reduce `#>` kinds of tests. Eventually all of this homegrown kind of formatting and testing should go away.  The `#>` part, thanks to @mmatera, we can replace. 

But if this is too onerous, it can be done later.

rocky:
LGTM

mathics/builtin/arithmetic.py
mmatera: > Of course, in the long term we'd like to reduce `#>` kinds of tests. Eventually all of this homegrown kind of formatting and testing should go away. The `#>` part, thanks to @mmatera, we can replace.
> 
> But if this is too onerous, it can be done later.

@rocky, I can start to do that, but probably in a different PR.

PR #271, opened 2022-05-01
Testing whether an expression is fully evaluated

rocky:
We are getting errors in formatting, e.g. `[1, 2]`

Or infinite recursion errors when of the form:

```
->0 message(self=<mathics.core.evaluation.Evaluation object at 0x7fb2ed271670>, symbol='General', tag='notboxes...)
     called from file '/src/external-vcs/github/Mathics3/mathics-core/mathics/core/evaluation.py' at line 471
##1 format_output(self=<mathics.core.evaluation.Evaluation object at 0x7fb2ed271670>, expr=<Expression: System`String...)
     called from file '/src/external-vcs/github/Mathics3/mathics-core/mathic
##2 message(self=<mathics.core.evaluation.Evaluation object at 0x7fb2ed271670>, symbol='System`General', tag='n...)
     called from file '/src/external-vcs/github/Mathics3/mathics-core/mathics/core/evaluation.py' at line 505
##3 format_output(self=<mathics.core.evaluation.Evaluation object at 0x7fb2e
```

I have no idea why the behavior changes here. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/271"></a>


mmatera:
@rocky,  I think it is more that just a formatting issue: at some point, it seems to interrupt the evaluation before it is complete, like here:
```
b'  70 (34): TEST MakeBoxes[squared[args___], StandardForm] := SuperscriptBox[RowBox[{MakeBoxes[squared], "[", RowBox[Riffle[MakeBoxes[#]& /@ {args}, ","]], "]"}], 2]'
b'  71 (35): TEST squared[1, 2]'
b'  72 (36): TEST squared[1, 2] // TeXForm'
result =!=wanted
Test failed: Formatting Output in Manual / Language Tutorials
Manual
Result: 
Wanted: \text{squared}\left[1,2\right]^2
Additional output:
General::notboxes: SuperscriptBox[RowBox[{MakeBoxes[squared],"[",RowBox[Riffle[Map[Function[MakeBoxes[Slot[1]]], {1,2}], ","]],"]"}], 2] is not a valid box structure.
```
or here:
```
b" 204 ( 3): TEST extremes = Solve[f'[x] == 0, x]"
b" 205 ( 4): TEST f''[x] /. extremes // N"
result =!=wanted
Test failed: Curve sketching in Manual / Examples
Manual
Result: {8. (-3. + 2. 5. ^ 0.5) (3. - 2. 5. ^ 0.5) 5. ^ 0.5 / (10. - 3. 5. ^ 0.5) ^ 3. + 8. (-3. + 2. 5. ^ 0.5) / (10. - 3. 5. ^ 0.5) ^ 2. + 8. 5. ^ 0.5 / (10. - 3. 5. ^ 0.5) ^ 2., -8. (-3. - 2. 5. ^ 0.5) (3. + 2. 5. ^ 0.5) 5. ^ 0.5 / (10. + 3. 5. ^ 0.5) ^ 3. - 8. 5. ^ 0.5 / (10. + 3. 5. ^ 0.5) ^ 2. + 8. (-3. - 2. 5. ^ 0.5) / (10. + 3. 5. ^ 0.5) ^ 2.}
Wanted: {1.65086, -0.064079}
```

or here:
```
b" 206 ( 5): TEST inflections = Solve[f''[x] == 0, x] // N // Chop"
result =!=wanted
Test failed: Curve sketching in Manual / Examples
Manual
Result: {{x -> -1.0852 + 4.44089×10^-16 I}, {x -> -3.21463 - 6.66134×10^-16 I}, {x -> 4.29983 - 1.11022×10^-16 I}}
Wanted: {{x -> -1.0852}, {x -> -3.21463}, {x -> 4.29983}}
```

mmatera:
@rocky, I found one of the places where this get in troubles:
here: 
https://github.com/Mathics3/mathics-core/blob/3bdef1e77c70b057b1e3d19a9b07f315795cc4e7/mathics/core/evaluators.py#L76
and here:
https://github.com/Mathics3/mathics-core/blob/3bdef1e77c70b057b1e3d19a9b07f315795cc4e7/mathics/core/evaluators.py#L137

we should call `Expression._build_elements` in order to make `apply_N` to work.

I guess the other failures have a similar origin.

rocky:
> @rocky, I found one of the places where this get in troubles: here:
> 
> https://github.com/Mathics3/mathics-core/blob/3bdef1e77c70b057b1e3d19a9b07f315795cc4e7/mathics/core/evaluators.py#L76
> 
> 
> and here:
> https://github.com/Mathics3/mathics-core/blob/3bdef1e77c70b057b1e3d19a9b07f315795cc4e7/mathics/core/evaluators.py#L137
> 
> we should call `Expression._build_elements` in order to make `apply_N` to work.
> 
> I guess the other failures have a similar origin.

Thanks for investigating and finding these. 

I'd appreciate it if you have time if you'd go and fix these and continue. At some point I would be interested in benchmarking against the last release to see whether what we have done so far offers any measurable improvement. My gut feeling is that there specific bad cases that have been fixed by this work. But I don't have hard evidence, and this is needed also to guide us forward. (At the end of rewrite_apply_eval_step there are some suggested tests that I want to perform some time.)

In my opinion a function like `_build_elements` instead of a generic `tuple()` is useful even if `_build_elements()` in fact just formed a tuple and returned it. Because it indicates the function that is going on, and allows us in one place we control to adjust things like tracking properties. 

Of course the other reason which is why it was created there is so that we can save time in not having to

* convert from tuple to list and back to tuple when not needed,
* sort items,
* flatten them, and most importantly,
* evaluate them

mmatera:
OK, indeed, the problem is that in several places, we set 'Expression._element' without calling '_build_elements'. One possible patch would be
1) rename 'Expression._element' to 'Expression.__element'
2) define a setter for 'Expression._elements' that calls '_build_element'

With this change, all the tests passed locally...

TiagoCavalcante (approved):
LGTM

mmatera:
OK, so, the question now is if we want to change now all the occurrences of `Expression._elements` in the rest of the code to be `Expresssion.elements`,   remove the property methods `Expression._elements`, and rename back `Expresion.__element`. This is going to take some (mechanical) work,  but I think it would be clearer and tidier.

rocky:
> OK, so, the question now is if we want to change now all the occurrences of `Expression._elements` in the rest of the code to be `Expresssion.elements`, remove the property methods `Expression._elements`, and rename back `Expresion.__element`. This is going to take some (mechanical) work, but I think it would be clearer and tidier.

I am for cleaner and tidier. No matter how long it takes.

mmatera:
OK, so let's keep this until the next weekend. Then I can try to make all the replacements.

rocky:
With respect to using the `_is_flat` property to speed up a costly operation @mmatera noted the following doesn't work:

```diff
PAGER=cat git diff
diff --git a/mathics/core/expression.py b/mathics/core/expression.py
index a74b5e7c..41165076 100644
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -1174,7 +1174,7 @@ class Expression(BaseElement, NumericOperators):
         # Step 3: Now, process the attributes of head
         # If there are sequence, flatten them if the attributes allow it.
         # if not new._is_flat and not (SEQUENCE_HOLD | HOLD_ALL_COMPLETE) & attributes:
-        if not (SEQUENCE_HOLD | HOLD_ALL_COMPLETE) & attributes:
+        if not new._is_flat and not (SEQUENCE_HOLD | HOLD_ALL_COMPLETE) & attributes:
             # This step is applied to most of the expressions
             # and could be heavy for expressions with many elements (like long lists)
             # however, most of the times, expressions does not have `Sequence` expressions
```

But it should. 

Another idea for some time in the future is to replace`_is_fully_evaluated` with a list (or generator so we could have ranges) of element indices that need evaluation. That ways we don't have to scan long lists when there are only a few that need (re)evaluation. 

Of course, if all elements are fully evaluated, then the list is empty.

But this shouldn't be part of this PR. Just something for the future.

mmatera (approved):
LGTM

PR #273, opened 2022-05-01
Making expression-elements-properties be right...

mmatera:
* Ensure that Expression._build_elements is always call each time Expression._elements is modified
* DRY _build_elements

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/273"></a>

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -78,48 +78,6 @@ symbols_arithmetic_operations = system_symbols(
 )
 
 
-def build_elements_with_properties(self, elements: Iterable) -> tuple:
-    """Build a tuple of Elements noting useful properties such as
-    whether the collection of elements is sorted, flat, or fully
-    evaluated.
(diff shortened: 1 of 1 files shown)


mathics/core/expression.py
```diff
@@ -599,7 +557,19 @@ def do_format(self, evaluation, form):
 
     @property
     def elements(self):
-        return self._elements
+        return self.__elements
```
TiagoCavalcante, line 560: I didn't understant why change `_elements` to `__elements`.

mmatera: Because there are several places in which `Expression._elements` is set, without calling `_build_elements`. With this change, each time `Expression._elements` is set, now the setter is called.

TiagoCavalcante:
LGTM

rocky (approved):
LGTM

PR #277, opened 2022-05-02
removing __elements

mmatera:
OK, this was less painful than what I expected.  The question now is if we do not introduce too much overhead by replacing the direct access to a property value by a function call...

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/277"></a>


TiagoCavalcante (approved):
LGTM

PR #261, opened 2022-04-04
Summary text more2

mmatera:
Here, I completed `distances`, `files_io`,  and `intfns`  documentation.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/261"></a>

PR #278, opened 2022-05-03
_GraphicsElement -> _GraphicsElementBox / _GraphicsDirective

mmatera:
From the @rocky comment in #261, I realized that the current implementation of this class and their daughters was wrong and misleading. This PR fixes this and partially fix the way in which the documentation is generated. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/278"></a>

### DIFF
--- a/mathics/builtin/drawing/graphics_internals.py
+++ b/mathics/builtin/drawing/graphics_internals.py
@@ -19,7 +19,7 @@ from mathics.core.symbols import system_symbols_dict, Symbol
 
 class _GraphicsDirective(InstanceableBuiltin):
     def __new__(cls, *args, **kwargs):
-        # This ensures that all the graphics elements have a well formatted docstring
+        # This ensures that all the graphics directive have a well formatted docstring
         # and a summary_text
         instance = super().__new__(cls, *args, **kwargs)
         if not hasattr(instance, "summary_text"):
@@ -33,16 +33,11 @@ class _GraphicsDirective(InstanceableBuiltin):
                 + article
                 + split_name(cls.get_name(short=True)[:-3])
             )
-        #            clsname = cls.get_name()
-        #            if clsname[0] in ("A", "E", "I", "O", "U"):
-        #                instance.summary_text = f"boxes for an '{cls.get_name()}' element"
-        #            else:
-        #                instance.summary_text = f"boxes for a '{cls.get_name()}' element"
         if not instance.__doc__:
             instance.__doc__ = f"""
                 <dl>
                 <dt>'{cls.get_name()}[...]'
-                <dd>is a graphics directive that sets {cls.get_name().lower()[:-3]}
+                <dd>is a graphics directive that sets {cls.get_name().lower()[:3]}
                 </dl>
                 """
         return instance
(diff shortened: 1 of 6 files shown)


mathics/builtin/drawing/graphics_internals.py
```diff
@@ -16,8 +17,48 @@
...
+        if item is not None and not item.has_form(self.get_name(), None):
+            raise BoxConstructError
+        self.graphics = graphics
+
+    @staticmethod
+    def create_as_style(klass, graphics, item):
+        return klass(graphics, item)
+
+
 # Check if  _GraphicsElement shouldn't be a BoxConstruct instead of an InstanceableBuiltin
```
TiagoCavalcante, line 60: I think this comment is now necessary anymore.

TiagoCavalcante (approved):
LGTM, just look the comment

mathics/builtin/drawing/graphics_internals.py
```diff
@@ -16,8 +17,48 @@
 from mathics.core.symbols import system_symbols_dict, Symbol
 
 
+class _GraphicsDirective(InstanceableBuiltin):
```
rocky, line 20: This should be done in a test, not as code that gets run every time mathics is run.

mmatera: @rocky, what is what should be done in a test? 

The idea in this PR was not to solve the problem of generating summaries but to make clearer and more accurate the class tree, in the direction of the comment that Tiago ask me to remove now. 

*Background* 

When a ``Graphics`` expression is formatted by ``MakeBoxes``, each element in the graphics expression is translated in one of two kinds of elements: a graphic directive (like `Thickness` or `RGBColor`) or a drawable object (a `Point`, a `Line`, a `Disk`, etc). For example,
```
In[1]:= Graphics[{Tick, Red, Disk[]}]//MakeBoxes
```
leads to
```
GraphicsBox[{Tick, Red, DiskBox[{0, 0}]}]
```

Internally, now `Tick`,`Red` and `DiskBox[]` are stored not as Expression / Symbol objects, but as objects of classes derived from `InstanceableBuiltin`. 
In master, both kinds of classes are derived from `_GraphicsElement`, which is derived from ``InstanceableBuiltin``. However, in my opinion, this is wrong:  a directive and a boxed graphics element require different interfaces, and use different properties, even if we forget the summary_text synthesis.  


For this reason, I propose here to split `_GraphicsElement` into two different classes: one for basing graphic directives (which derive from `InstanceableBuiltin`) and the other for the boxed version of the  "drawable" objects (deriving from `BoxConstruction`).

With this, I can also define more accurate rules to produce automatic summaries. Here the summaries are produced "on the fly", but we could instead store them in a separate file, or add them into the classes.

rocky: Ok. Thanks for the explanation. I misread the code. But I still have a misgiving with the automatic summary generation, and what this kind of code might do in the future given that we know that all of this we want to redoc.

First, summaries about boxes are largely useless. Does  WMA give summaries or descriptions about _all_ of the autogenerated boxing routines it  defines? 

I would find it odd because such stuff is largely not useful to anyone. 

I did like and appreciate the thought and effort to clean up the class structure, even if only a little bit. 

However there has been a more fundamental problem there is a lot of stuff in `BaseElement` seems not to belong there since there is these other kinds of classes that are parts of S-Expressions or M-Expressions that do not inherit from these.  So that suggests that in the future stuff get removed from BaseElement, so that it really does represent the universal properties of an element of an S-Expression and everything that can be an element in an S-Expression use that. And this would include boxes. 

Lastly, and this kind of thing comes up a bit. Why do we start off with a PR and the implementation and _then_ find out what the motivation is afterwards? 

Having said this, I suppose this helps, and hopefully if we have to rip out code in the documentation stuff this won't be too tangled. So sure, go ahead and we'll revise later.

mmatera: > Ok. Thanks for the explanation. I misread the code. But I still have a misgiving with the automatic summary generation, and what this kind of code might do in the future given that we know that all of this we want to redo how function documentation works.
> 

OK. I keep just the part that generates the documentation for directives, but not for boxes.

> First, summaries about boxes are largely useless. Does WMA give summaries or descriptions about _all_ of the autogenerated boxing routines it defines?
> 
> I would find it odd because such stuff is largely not useful to anyone.
>
> I did like and appreciate the thought and effort to clean up the class structure, even if only a little bit.
> 

OK. At least, trying to implement this useless automatic summary helped me to understand better what this class was supposed to do.

> However there has been a more fundamental problem there is a lot of stuff in `BaseElement` seems not to belong there since there is these other kinds of classes that are parts of S-Expressions or M-Expressions that do not inherit from these. So that suggests that in the future stuff get removed from BaseElement, so that it really does represent the universal properties of an element of an S-Expression and everything that can be an element in an S-Expression use that. And this would include boxes.
> 

Yep, but these kinds of changes are at a different level. Let's keep that part until we start with the boxes refactor.

> Lastly, and this kind of thing comes up a bit. Why do we start off with a PR and the implementation and _then_ find out what the motivation is afterwards?
> 

I am sorry. I thought the change was self-explanatory. 

> Having said this, I suppose this helps, and hopefully if we have to rip out code in the documentation stuff this won't be too tangled. So sure, go ahead and we'll revise later.

Good. Let's merge this and iterate.

Issue #20, opened 2021-09-27
Expressions with undefined variables are slow


TiagoCavalcante:
Expressions with undefined variables are 16x slower than the equivalents defined variables.

```wl
In[1]:= x=1
Out[1]= 1

In[2]:= Print[Timing[x+1+1.4+2]]; Print[Timing[y+1+1.4+2]]; Print[Timing[1+1.4+2+x]]; Print[Timing[1+1.4+2+y]]
{0.000742186, 5.4}
{0.0123144, 4.4 + y}
{0.000772454, 5.4}
{0.0121675, 4.4 + y}
Out[2]= None
```

One way to fix it is to have a hash table (python dict) with the numeric variables (in the future we can expand it to have a hash table for each type of value).

If the variable isn't there and we are expecting a numeric value we stop searching for the variable in other hash tables (currently we just have the huge `definitions`).

To we have an order of magnitude of the performance improvement here is the time with a hardcoded `y`:
```wl
In[1]:= Timing[1+1.4+2+y]
Out[1]= {0.00349792, 4.4 + y}
```
And by hardcoded I mean:
```diff
- items = items.numerify(evaluation).get_sequence()
+ items = [Integer(1), Real(1.4), Integer(2), Symbol("y")]
```
The hardcoded sum is almost 3.5x (71%) faster.

rocky:
Wow! Yes, this is something we really need to fix sooner than later.

TiagoCavalcante:
I have been thinking about this and for a good improvement there should be added 2 more hash tables: the numbers one and a hash table for only delayed rules (`:=`).

rocky:
I'm in favor of trying something simple in code in  a branch as a feasibility study to get a more detailed picture of how this would work and have something more concrete to discuss. 

I think of this sort of like how an artist or cartoonist might do a rough sketch before working on the finished painting or cartoon.

mmatera:
@TiagoCavalcante , do you have a clue about why it takes so much time to add undefined symbols? I think that before adding more caching and hash-tables, we need to understand better why the evaluator has this behavior.

TiagoCavalcante:
@mmatera I think it is because it needs to check all the definitions, one way to verify this is evaluate that without adding builtins.

rocky:
> @mmatera I think it is because it needs to check all the definitions, one way to verify this is evaluate that without adding builtins.

[...]

PR #281, opened 2022-05-07
avoid Expression.element recalculation if possible

rocky:
@mmatera In #277 the property recalculation of elements properties and avoidance of calling `to_python()` was removing by virtue of called the setters property. Also, we weren't making use the flat property that was fixed as a result of #277. 

This preserves those properties and allows Expression creation to give (some of) these properties. To be added could be  add a way to indicate that `to_python()` on elements isn't needed. 

What could be done is to go over all calls to `Expression` to see if we can specify whether we can avoid calling `to_python` and if we determine element properties and set them up front.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/281"></a>

### DIFF
--- a/mathics/builtin/list/constructing.py
+++ b/mathics/builtin/list/constructing.py
@@ -87,7 +87,7 @@ class Array(Builtin):
                 return
             dims[index] = value
         if origins.has_form("List", None):
-            if len(origins.leaves) != len(dims):
+            if len(origins.elements) != len(dims):
                 evaluation.message("Array", "plen", dimsexpr, origins)
                 return
             origins = origins.get_mutable_elements()
@@ -151,7 +151,8 @@ class Normal(Builtin):
         if isinstance(expr, Atom):
             return
         return Expression(
-            expr.get_head(), *[Expression("Normal", leaf) for leaf in expr.leaves]
+            expr.get_head(),
+            *[Expression("Normal", element) for element in expr.elements],
         )
 
 
@@ -251,14 +252,14 @@ class Permutations(Builtin):
 
         rs = None
         if isinstance(n, Integer):
-            py_n = min(n.get_int_value(), len(li.leaves))
-        elif n.has_form("List", 1) and isinstance(n.leaves[0], Integer):
-            py_n = n.leaves[0].get_int_value()
+            py_n = min(n.get_int_value(), len(li.elements))
+        elif n.has_form("List", 1) and isinstance(n.elements[0], Integer):
+            py_n = n.elements[0].get_int_value()
             rs = (py_n,)
         elif (
-            n.has_form("DirectedInfinity", 1) and n.leaves[0].get_int_value() == 1
+            n.has_form("DirectedInfinity", 1) and n.elements[0].get_int_value() == 1
         ) or n.get_name() == "System`All":
-            py_n = len(li.leaves)
+            py_n = len(li.elements)
         else:
             py_n = None
 
@@ -274,7 +275,7 @@ class Permutations(Builtin):
         inner = structure("List", li, evaluation)
         outer = structure("List", inner, evaluation)
 
-        return outer([inner(p) for r in rs for p in permutations(li.leaves, r)])
+        return outer([inner(p) for r in rs for p in permutations(li.elements, r)])
 
 
 class Reap(Builtin):
@@ -479,7 +480,7 @@ class Tuples(Builtin):
         if n is None or n < 0:
             evaluation.message("Tuples", "intnn")
             return
-        items = expr.leaves
+        items = expr.elements
 
         def iterate(n_rest):
             evaluation.check_stopped()
@@ -491,7 +492,7 @@ class Tuples(Builtin):
                         yield [item] + rest
 
         return Expression(
-            SymbolList, *(Expression(expr.head, *leaves) for leaves in iterate(n))
+            SymbolList, *(Expression(expr.head, *elements) for elements in iterate(n))
         )
 
     def apply_lists(self, exprs, evaluation):
@@ -504,9 +505,9 @@ class Tuples(Builtin):
             if isinstance(expr, Atom):
                 evaluation.message("Tuples", "normal")
                 return
-            items.append(expr.leaves)
+            items.append(expr.elements)
 
         return Expression(
             SymbolList,
-            *(Expression(SymbolList, *leaves) for leaves in get_tuples(items)),
+            *(Expression(SymbolList, *elements) for elements in get_tuples(items)),
         )
--- a/mathics/builtin/list/constructing.py
+++ b/mathics/builtin/list/constructing.py
@@ -20,6 +20,7 @@ from mathics.core.convert import from_sympy
 
 from mathics.core.expression import (
     Expression,
+    identity_fn,
     structure,
 )
 from mathics.core.atoms import Integer
@@ -110,7 +111,7 @@ class Array(Builtin):
                     level.append(rec(rest_dims[1:], current + [index]))
                 return Expression(head, *level)
             else:
-                return Expression(f, *(Integer(index) for index in current))
+                return Expression(f, *current, element_conversion_fn=Integer)
 
         return rec(dims, [])
 
@@ -172,13 +173,12 @@ class Range(Builtin):
      = {0, 1 / 3, 2 / 3, 1, 4 / 3, 5 / 3, 2}
     """
 
-    summary_text = "generate a list of equispaced, consecutive numbers"
+    attributes = listable | protected
+
     rules = {
         "Range[imax_?RealNumberQ]": "Range[1, imax, 1]",
         "Range[imin_?RealNumberQ, imax_?RealNumberQ]": "Range[imin, imax, 1]",
     }
-
-    attributes = listable | protected
 
     summary_text = "form a list from a range of numbers or other objects"
 
@@ -194,7 +194,7 @@ class Range(Builtin):
             evaluation.check_stopped()
             result.append(from_sympy(index))
             index += di
-        return Expression(SymbolList, *result)
+        return Expression(SymbolList, *result, element_conversion_fn=identity_fn)
 
 
 class Permutations(Builtin):
@@ -239,10 +239,10 @@ class Permutations(Builtin):
     def apply(self, li, evaluation):
         "Permutations[li_List]"
         return Expression(
-            "List",
+            SymbolList,
             *[
                 Expression(SymbolList, *p)
-                for p in permutations(li.leaves, len(li.leaves))
+                for p in permutations(li.elements, len(li.elements))
             ],
         )
 
@@ -491,7 +491,7 @@ class Tuples(Builtin):
                         yield [item] + rest
 
         return Expression(
-            "List", *(Expression(expr.head, *leaves) for leaves in iterate(n))
+            SymbolList, *(Expression(expr.head, *leaves) for leaves in iterate(n))
         )
 
     def apply_lists(self, exprs, evaluation):
@@ -507,5 +507,6 @@ class Tuples(Builtin):
             items.append(expr.leaves)
 
         return Expression(
-            "List", *(Expression(SymbolList, *leaves) for leaves in get_tuples(items))
+            SymbolList,
+            *(Expression(SymbolList, *leaves) for leaves in get_tuples(items)),
         )
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -169,6 +169,29 @@ class ExpressionCache:
 
 
 class Expression(BaseElement, NumericOperators):
+    """
+    A Mathics M-Expression.
+
+    A Mathics M-Expression is a list where the head is a function designator.
+    (In the more common S-Expression the head is an a Symbol. In Mathics this can be
+    an expression that acts as a function.
+
+    positional Arguments:
+        - head -- The head of the M-Expression
+        - *elements - optional: the remainin elements
+
+    Keyword Arguments:
+
+        - element_properties -- optional: a dictionary describing properties
+                                of the colletion of elements.
+          Properties include:
+             _is_flat: bool   -- True inone of the elements is an Expression
+             _is_sorted: bool -- True if all of the elements are sorted. Of course this is true,
+                                 if there are less than 2 elements.
+             _fully_evaluated: bool
+                              -- True if none of the elements needs to be evaluated
+    """
+
     head: "Symbol"
     leaves: typing.List[Any]
     _sequences: Any
@@ -188,7 +211,14 @@ class Expression(BaseElement, NumericOperators):
         # These are set in self._build_elements(elements)
         #    self._elements_fully_evaluated, self._is_flat, self._is_sorted
 
-        self.elements = elements
+        element_properties = kwargs.pop("element_properties", None)
+        if element_properties is not None:
+            self._elements = elements
+            for field in ("_elements_fully_evaluated", "_is_flat", "_is_sorted"):
+                setattr(self, field, element_properties.get(field, False))
+        else:
+            # The self.elements setter method sets the elements collection properties.
+            self.elements = elements
 
         self._sequences = None
         self._cache = None
@@ -1150,25 +1180,30 @@ class Expression(BaseElement, NumericOperators):
                 eval_range(range(len(elements)))
                 # rest_range(range(0, 0))
 
+        # Step 2: Build a new expression. We take care not
+        # to evaluate elements, run to_python() on them in
+        # Expression construction, or convert Expresions elements from a tuple to a list
+        # and back if that can be avoided.
+
         if self._elements_fully_evaluated:
-            elements = self._elements
+            new = Expression(
+                head,
+                *self._elements,
+                element_properties={
+                    "_is_flat": self._is_flat,
+                    "_is_sorted": self._is_sorted,
+                    "_elements_fully_evaluated": self._elements_fully_evaluated,
+                }
+            )
+            elements = self.elements
         else:
             elements = self.get_mutable_elements()
             eval_elements()
-
-        # Step 2: Build a new expression. Notice that elements are given
-        # after creating the object, to avoid to call `from_python` on each element.
-
-        # FIXME: put this in a method
-        new = Expression(head)
-        new.elements = elements
-        # the setter in new.elements calls _build_elements, and it sets new._is_flat
-        # new._is_flat = self._is_flat
+            new = Expression(head, *elements)
 
         # Step 3: Now, process the attributes of head
         # If there are sequence, flatten them if the attributes allow it.
-        # if not new._is_flat and not (SEQUENCE_HOLD | HOLD_ALL_COMPLETE) & attributes:
-        if not (SEQUENCE_HOLD | HOLD_ALL_COMPLETE) & attributes:
+        if not new._is_flat and not (SEQUENCE_HOLD | HOLD_ALL_COMPLETE) & attributes:
             # This step is applied to most of the expressions
             # and could be heavy for expressions with many elements (like long lists)
             # however, most of the times, expressions does not have `Sequence` expressions
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -78,6 +78,10 @@ symbols_arithmetic_operations = system_symbols(
 )
 
 
+def identity_fn(arg: Any) -> Any:
+    return arg
+
+
 class BoxError(Exception):
     def __init__(self, box, form) -> None:
         super().__init__("Box %s cannot be formatted as %s" % (box, form))
@@ -1236,7 +1240,7 @@ class Expression(BaseElement, NumericOperators):
         for element in elements:
             element.unevaluated = False
 
-        # If HoldAllComplete is not an attribute,
+        # If HoldAllComplete Attribute (flag ``HOLD_ALL_COMPLETE``) is not set,
         # and the expression has elements of the form  `Unevaluated[element]`
         # change them to `element` and set a flag `unevaluated=True`
         # If the evaluation fails, use this flag to restore back the initial form
@@ -1263,8 +1267,9 @@ class Expression(BaseElement, NumericOperators):
                 new.elements = dirty_elements
                 elements = dirty_elements
 
-        # If the attribute FLAT is set, calls flatten with a callback
-        # that set elements as unevaluated too.
+        # If the Attribute ``Flat`` (flag ``FLAT``) is set, calls
+        # flatten with a callback that set elements as unevaluated
+        # too.
         def flatten_callback(new_elements, old):
             for element in new_elements:
                 element.unevaluated = old.unevaluated
@@ -1285,8 +1290,9 @@ class Expression(BaseElement, NumericOperators):
 
         # Step 5: Must we need to thread-rewrite the expression?
         #
-        # Threading is needed when head has the ``LISTABLE``
-        # Attribute.  ``Expression.thread`` rewrites the expression:
+        # Threading is needed when head has the ``Listable``
+        # Attribute (or flag ``LISTABLE``).
+        # ``Expression.thread`` rewrites the expression:
         #  ``F[{a,b,c,...}]`` as:
         #  ``{F[a], F[b], F[c], ...}``.
 
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -284,13 +284,18 @@ class Expression(BaseElement, NumericOperators):
         # _is_flat is True if all elements are atoms/leaves.
         self._is_flat = True
 
-        # _is_ordered is True if elements do not need ordering
-        # elements with less than 2 items or all have the same value are ordered.
-        # Ordering is a Attribute of some functions.
+        # _is_ordered is True if elements do not need ordering.
+        # Elements with less than 2 items or are ordered.
+        # Otherwise we'll try to use "==" to check for orderedness and if that fails
+        # we'll say the elements are not ordered.
+
+        # Note that *checking* a list is O(n) while sorting is O(n log n).
+        # Ordering is a Attribute is defined for some Mathics functions.
         self._is_ordered = True
 
         result = []
-        last_element = None
+
+        last_converted_elt = None
         for element in elements:
             converted_elt = conversion_fn(element)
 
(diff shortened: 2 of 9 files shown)


mathics/core/expression.py
```diff
@@ -1150,25 +1180,30 @@ def eval_range(indices):
                 eval_range(range(len(elements)))
                 # rest_range(range(0, 0))
 
+        # Step 2: Build a new expression. We take care not
+        # to evaluate elements, run to_python() on them in
+        # Expression construction, or convert Expresions elements from a tuple to a list
+        # and back if that can be avoided.
+
         if self._elements_fully_evaluated:
-            elements = self._elements
```
rocky, line 1154: This is where we started to go wrong.

mathics/core/expression.py
```diff
@@ -1150,25 +1180,30 @@ def eval_range(indices):
...
-        # FIXME: put this in a method
-        new = Expression(head)
-        new.elements = elements
-        # the setter in new.elements calls _build_elements, and it sets new._is_flat
-        # new._is_flat = self._is_flat
+            new = Expression(head, *elements)
 
         # Step 3: Now, process the attributes of head
         # If there are sequence, flatten them if the attributes allow it.
-        # if not new._is_flat and not (SEQUENCE_HOLD | HOLD_ALL_COMPLETE) & attributes:
```
rocky, line 1170: Last week this was failing. Probably as a result of the refactoring separating Box elements from general elements fixed this.

mmatera:
@rocky,
probably what is wrong is to call `from_python` here: https://github.com/Mathics3/mathics-core/blob/cf2fee14d0739e5edf2b5c05348f6095c161acc9/mathics/core/expression.py#L278

Apart from that, I think the solution in this PR is better than the former.

rocky:
> @rocky, probably what is wrong is to call `from_python` here:
> 
> https://github.com/Mathics3/mathics-core/blob/cf2fee14d0739e5edf2b5c05348f6095c161acc9/mathics/core/expression.py#L278
> 

Let's me first clear something up. This PR was about undoing the regression in performance caused by #277 and not about addressing other things which code might be improved. 

I am a little saddened by the fact that this [convoluted sequence of Expression building](https://github.com/Mathics3/mathics-core/blob/master/mathics/core/expression.py#L1159-L1164)  for setting variable `new` in `rewite_apply_eval_step`  _which was even documented_  (if not added) by you, was blithely removed as a result of of adding that  setter property, method, along with disabling some of the recently added properties that I had been talking about.

Instead of focusing on the problem at hand there is a this tangential `to_python()` thing you have been fixated on and bring it up any time there is something that is even closely related.

Yes, I know, as you've mentioned before that you didn't like the `to_python()` in the expression constructor because you feel it can take too much time. And an ineffective remedy offered in the past was to rearrange the order of `if` and `elif` statements. Or possibly this code which separated the the constructor call from element setting.

In this PR I start to show how I think this should be addressed: by allowing parameters on the constructor which specify something about the parameters. Such as whether they are already flat, or fully evaluated.

Whether elements need to be converted to Mathics objects is just yet another parameter, the conversion function,  that could be specified on element construction. 

In this PR, I started in that direction. However I realized when I wrote this that there is a bit more that could be done here. The question I had at the time was whether to go down that tangent now or later. I opted for later since time is limited and I wanted to get the breakage addressed.

So in this PR, you can explicitly specify properties on collections of elements so they don't have to be computed on the collection of elements. These properties currently are named `is_flat`, `_is_sorted` and `_elements_fully_evaluated`.  That last property `_elements_fully_evaluated` interacts with whether `to_python()` or _any_ other kind of conversion function. I had realized that, but was hoping to deal with cleaning this up later. 

As I said, correct thing would be to allow the conversion function to be specified as a keyword parameter If none is given, default conversion function to apply to elements when it is needed would be `to_python`.   I didn't mention this yesterday, because I had not checked the calls to Expression which would suggest whether that is correct or if not suggest what might be useful.

So let's look at some use cases. Take for example this block from [mathics.core.convert](https://github.com/Mathics3/mathics-core/blob/master/mathics/core/convert.py#L244-L251): 

```python
elif expr.is_Add:
        return Expression(SymbolPlus, *sorted([from_sympy(arg) for arg in expr.args]))
    elif expr.is_Mul:
        return Expression(SymbolTimes, *sorted([from_sympy(arg) for arg in expr.args]))
    elif expr.is_Pow:
        return Expression(SymbolPower, *[from_sympy(arg) for arg in expr.args])
    elif expr.is_Equality:
        return Expression(SymbolEqual, *[from_sympy(arg) for arg in expr.args])
```

There are two possibilities here. Either specify on the expression constructor that the elements properties are already converted, or specify a conversion function `from_sympy` is as the conversion function to use. In the first two cases above, we also probably also want to specify that the arguments are "sorted".  (I will come back to the use of the term "sorted" later). 

I tend to favor passing the conversion function `from_sympy()` since this removes an extra iteration over the elements. 

As for the word "sorted", a better word would be "ordered" since that is the word or Attribute name that WMA uses. I have the vague hope that if that were used that would make it less likely that @TiagoCavalcante when he pokes into the code is less tempted to remove it blithely because he doesn't understand its use as he has tried in the past.

In another example from earlier form that same function [mathics.core.convert](https://github.com/Mathics3/mathics-core/blob/master/mathics/core/convert.py#L223) we have:

```python
return Expression(SymbolTimes, Integer(-1), SymbolInfinity)
```

Here, we would indicate as a keyword argument to the Expression constructor that the parameters do not need conversion at all. That is, they are already in the form needed for a list of elements in a Mathics M-Expression. I believe there are numerous other examples of such kinds examples of this kind of thing in the code.

But right now, my own interest is assessing whether any of this has made a dent in this area of evaluation and to what extent. That will guide me as to how much focus I want to spend here. My inclination is not to go through the entire code base and customize the numerous Expression constructor calls like we previously have done to pass in a SymbolName as the head element in the past. For that I'd prefer is someone else undertake this, if this is the direction we  move in.

In advance of this, I'll try to update the PR so that we allow a keyword parameter for the kind of conversion function to be used, when it is needed.

Instead  of going through the codebase and adjusting Expression constructors, I'd rather go on and start to look at how rule replacement is done to investigate ways that can be improved. There has already been suggestion to hook into an efficient regular expression or some other pattern-matching library. 

Lastly in the area of `rewrite_apply_eval_step` improvement the last one I think that could be us use is to replace `_elements_fully_evaluated` with a list-like object that specifies which elements need to be evaluated. If the list is empty that could mean the same things as `_elements_fully_evaluated = True`. If the list is `None` that could mean evaluate everything. And for a small number of elements, listing the integer elements would that need evaluation would be fine. However for the very special case where there are large blocks of elements that need to be evaluated but it is not everything, a generator (`range()` is a kind of generator  would represent this efficiently. For example. If you know that the after the first two items, the next 100 need to be evaluated, then `[0, 1, range(2, 102)]` would indicate this. If on the other hand in say an association which is listed in key value pair format and all of the keys don't need to be evaluated, then `[range(1, len(n), 2)]` would handle this.

mmatera:
> ... 
> Instead of focusing on the problem at hand there is a this tangential `to_python()` thing you have been fixated on and bring it up any time there is something that is even closely related.


Sorry for the confusion. I understood that the problem was related to the call to 'from_python'. I didn't realize that in the merge I lost a line that took advantage of the new properties. Sorry for that too.


> 
> Yes, I know, as you've mentioned before that you didn't like the `to_python()` in the expression constructor because you feel it can take too much time. And an ineffective remedy offered in the past was to rearrange the order of `if` and `elif` statements. Or possibly this code which separated the the constructor call from element setting.
> 
> In this PR I start to show how I think this should be addressed: by allowing parameters on the constructor which specify something about the parameters. Such as whether they are already flat, or fully evaluated.
> 
> Whether elements need to be converted to Mathics objects is just yet another parameter, the conversion function, that could be specified on element construction.
> 

All the need of having a "from_python" conversion step inside the `__init__` method of `Expression` was about  making building expressions simper. Let's say, if you want to build an expression line '-F[x,{1,2,3}]', it is handy to write
```
-Expression("F", [1,2,3])
```
instead of
```
Expression(SymbolTimes, Integer(-1),  Expression(Symbol("F"), Expression(SymbolList, *(from_python(i) for in  [1,2,3])) ))
``` 
If we need to pass a parameter for the conversions, then I think that advantage is lost.


> In this PR, I started in that direction. However I realized when I wrote this that there is a bit more that could be done here. The question I had at the time was whether to go down that tangent now or later. I opted for later since time is limited and I wanted to get the breakage addressed.
> 
> So in this PR, you can explicitly specify properties on collections of elements so they don't have to be computed on the collection of elements. These properties currently are named `is_flat`, `_is_sorted` and `_elements_fully_evaluated`. That last property `_elements_fully_evaluated` interacts with whether `to_python()` or _any_ other kind of conversion function. I had realized that, but was hoping to deal with cleaning this up later.
> 

Just notice that in ``Expression("F", *[1,2,3])``   it makes sense to say that the elements (are going to be) "already evaluated" as soon as you convert them.


> As I said, the correct thing would be to allow the conversion function to be specified as a keyword parameter If none is given, > ...
> 
> ```python
>     elif expr.is_Add:
>         return Expression(SymbolPlus, *sorted([from_sympy(arg) for arg in expr.args]))
>     elif expr.is_Mul:
>         return Expression(SymbolTimes, *sorted([from_sympy(arg) for arg in expr.args]))
>     elif expr.is_Pow:
>         return Expression(SymbolPower, *[from_sympy(arg) for arg in expr.args])
>     elif expr.is_Equality:
>         return Expression(SymbolEqual, *[from_sympy(arg) for arg in expr.args])
> ```
> 

Here, I guess that better than telling '__init__' to use 'from_sympy', it would be to tell it that do not perform a new conversion.

> As for the word "sorted", a better word would be "ordered" since that is the word or Attribute name that WMA uses. I have the vague hope that if that were used that would make it less likely that @TiagoCavalcante when he pokes into the code is less tempted to remove it blithely because he doesn't understand its use as he has tried in the past.
> 

Agree

> In another example from earlier form that same function [mathics.core.convert](https://github.com/Mathics3/mathics-core/blob/master/mathics/core/convert.py#L223) we have:
> 
> ```python
> return Expression(SymbolTimes, Integer(-1), SymbolInfinity)
> ```
> 
> Here, we would indicate as a keyword argument to the Expression constructor that the parameters do not need conversion at all. That is, they are already in the form needed for a list of elements in a Mathics M-Expression. I believe there are numerous other examples of such kinds examples of this kind of thing in the code.
> 
> But right now, my own interest is assessing whether any of this has made a dent in this area of evaluation and to what extent. That will guide me as to how much focus I want to spend here. My inclination is not to go through the entire code base and customize the numerous Expression constructor calls like we previously have done to pass in a SymbolName as the head element in the past. For that I'd prefer is someone else undertake this, if this is the direction we move in.
> 

As far as we keep compatibility, we can make the change in the places that have more impact in the performance.

> In advance of this, I'll try to update the PR so that we allow a keyword parameter for the kind of conversion function to be used, when it is needed.
> 

I think the most important places to use it is in the evaluation process, and maybe in the apply_rules related functions.


> Instead of going through the codebase and adjusting Expression constructors, I'd rather go on and start to look at how rule replacement is done to investigate ways that can be improved. There has already been suggestion to hook into an efficient regular expression or some other pattern-matching library.
>

Indeed
 
> Lastly in the area of `rewrite_apply_eval_step` improvement the last one I think that could be us use is to replace `_elements_fully_evaluated` with a list-like object that specifies which elements need to be evaluated. If the list is empty that could mean the same things as `_elements_fully_evaluated = True`. If the list is `None` that could mean evaluate everything. And for a small number of elements, listing the integer elements would that need evaluation would be fine. However for the very special case where there are large blocks of elements that need to be evaluated but it is not everything, a generator (`range()` is a kind of generator would represent this efficiently. For example. If you know that the after the first two items, the next 100 need to be evaluated, then `[0, 1, range(2, 102)]` would indicate this. If on the other hand in say an association which is listed in key value pair format and all of the keys don't need to be evaluated, then `[range(1, len(n), 2)]` would handle this.

That could be a good next step. But the current changes could give us an idea of what is the impact in performance of following in this direction.  Most important, what I like about this is that it allows to write clearer code.

rocky:
> If we need to pass a parameter for the conversions, then I think that advantage is lost.

it isn't necessary, it is optional when it is felt that it is needed.

mathics/core/expression.py
```diff
@@ -1150,25 +1193,30 @@ def eval_range(indices):
...
+            new = Expression(
+                head,
+                *self._elements,
+                element_properties={
+                    "_is_flat": self._is_flat,
+                    "_is_ordered": self._is_ordered,
+                    "_elements_fully_evaluated": self._elements_fully_evaluated,
+                }
+            )
+            elements = self.elements
```
mmatera, line 1211: Even it is the same, shouldn´t be clearer to write `elements = new.elements`? or there is another reason?

mathics/core/expression.py
mmatera: we should avoid calling `evaluate` over Box elements. `Box` elements should be handled just as a `String`.

mathics/core/expression.py
```diff
@@ -1406,8 +1454,13 @@ def set_element(self, index: int, value):
     def shallow_copy(self) -> "Expression":
         # this is a minimal, shallow copy: head, elements are shared with
         # the original, only the Expression instance is new.
+
+        # FIXME: this should be encapulated in the constructor better.
         expr = Expression(self._head)
-        expr.elements = self._elements
+        expr._elements = self._elements
```
mmatera, line 1460: this could be put in the previous line as 
expr = Expression(self._head, self._elements, element_properties={"_please_do_not_convert_me"=True,})

mathics/core/expression.py
```diff
@@ -1895,7 +1948,7 @@ def __init__(self, head):
 
     def __call__(self, elements):
         expr = Expression(self._head)
-        expr.elements = tuple(elements)
+        expr.elements = elements
```
mmatera, line 1951: same as before...

mathics/core/expression.py
```diff
@@ -1936,7 +1989,7 @@ def slice(self, expr, py_slice):
             raise ValueError("Structure.slice only supports slice steps of 1")
 
         new = Expression(self._head)
-        new.elements = tuple(elements[lower:upper])
+        new.elements = elements[lower:upper]
```
mmatera, line 1992: same as before

mathics/core/expression.py
```diff
@@ -186,9 +210,25 @@ def __init__(self, head, *elements, **kwargs):
...
+        # Note: We don't allow specifying "_is_ordered"
+        # when "_elements_fully_evaluated" is False. And I suppose this
+        # might be right since we can't really know ordering if we things
+        # are not fully evaluated.
+        if element_properties is not None:
+            for field in ("_elements_fully_evaluated", "_is_flat", "_is_ordered"):
+                setattr(self, field, element_properties.get(field, False))
+            self._elements = (
+                elements
+                if self._elements_fully_evaluated
```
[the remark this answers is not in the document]
rocky: I am missing the point here.  The code does the right thing for these cases.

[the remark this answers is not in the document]
mmatera: Yes, but now, what would be the right parameters to pass  in element_properties?

mathics/core/expression.py
```diff
@@ -238,30 +280,31 @@ def _build_elements(self, elements: Iterable) -> tuple:
...
+        # _is_ordered is True if elements do not need ordering
+        # elements with less than 2 items or all have the same value are ordered.
+        # Ordering is a Attribute of some functions.
+        self._is_ordered = True
 
         result = []
         last_element = None
         for element in elements:
-            converted_elt = from_python(element)
+            converted_elt = conversion_fn(element)
```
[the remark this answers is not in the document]
rocky: I realize this  was a suggestion you made do after following what the direction I started. 

But that wasn't something I had in mind. In my view, either the caller makes properties, explicit when that is desired. 

As I have said in the past, am not a fan of isinstance(x, BaseElement) , especially now as the class structure has all sorts of quirks in it. We have seen that there are leaves of S-Expressions that aren't BaseElement.

mathics/core/expression.py
rocky: RIght now, I want to defer this for another some future iteration.

mathics/core/expression.py
rocky: Since this seems to mean more to you than me (I don't necessarily find one clearer than the other) ab5c21cc has this change.

mathics/core/expression.py
rocky: Yes, in the future this should be one Expression call. I think I mentioned that somewhere else. 

It is however not something I want to undertake right now.

mathics/core/expression.py
rocky: same as before.

mathics/core/expression.py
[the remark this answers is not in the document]
rocky: For `Expression("Plus", 50*[1])` doing it this way works and sets the properties correctly. 

As done in the `test_expression_constructor` you can also use
`Expression("Plus"  50*[1], element_properties={"_elements_fully_evaluated": True, "_is_flat": True, "is_ordered: True})` if you want be explicit about the elements properties so Expression does not try to evaluate, flatten or order the elements.

As for `Expression("Plus", Symbol("Global`a"), Integer(1))` best to just leave it as is. If you wanted to add element properties it would be what you said: `element_properties={"_elements_fully_evaluated: False", "_is_flat": True}`. 

You could specify that this isn't ordered, but that is also implied when elements are not fully evaluated. (How can something  be ordered when you don't have a value for everything).

This kind of thing isn't causing a problem. 

The problem place I am concerned about is when we have long lists. You find this for example in plotting.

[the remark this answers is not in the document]
mmatera: > For `Expression("Plus", 50*[1])` doing it this way works and sets the properties correctly.
> 
> As done in the `test_expression_constructor` you can also use `Expression("Plus" 50*[1], conversion_fn=Integer, element_properties={"_elements_fully_evaluated": True, "_is_flat": True, "is_ordered: True})` if you want be explicit about the elements properties so Expression does not try to evaluate, flatten or order the elements.
>

Following the line 225,
```
self._elements = (
                elements
                if self._elements_fully_evaluated 
                else (conversion_fn(element) for element in elements)
            )
```
if you set that the element is fully evaluated, then the conversion shouldn´t be done. 
So, I do not see where
```Expression("Plus", 50*[1], conversion_fn=Integer, element_properties={"_elements_fully_evaluated": True, "_is_flat": True, "is_ordered: True})```
applies conversion_fn over the elements. Probably it happens in another place.

 
> As for `` Expression("Plus", Symbol("Global`a"), Integer(1)) `` best to just leave it as is. If you wanted to add element properties it would be what you said: `element_properties={"_elements_fully_evaluated: False", "_is_flat": True}`.
> 
> You could specify that this isn't ordered, but that is also implied when elements are not fully evaluated. (How can something be ordered when you don't have a value for everything?)
> 

Agree

> This kind of thing isn't causing a problem.
> 
> The place I was concerned about is when we have long lists or lots of data. You find this for example in plotting.

OK, but if you want to build a Plot, usually you do not have a mixture of Python and Mathics native numbers: well you pass a NumPy array, well you pass a ``List`` ``Expression`` where all the elements are native numbers.
In any case, all this conversion stuff comes up just to make some typical code easier to read and write. 
It is unlike, (and probably, we should discourage)  to have code like

```
a=sympy.Symbol("a")
b = 20*[1.]
c = Icecream("mint")
expr = Expression("F", a , b, c, conversion_fn=funny_converter,  {"_elements_fully_evaluated: False", "_is_flat": True})
```

Also, I do not see the point of allowing
```
data = 50*[1]
expr = Expression(SymbolList, *data, conversion_fn=Integer)
```
which can be written in a clearer (and sometimes more efficient) way as
```
data = 50*[Integer1]
expr = Expression(SymbolList, *data, conversion_fn=None)
```
(or even)
```
data = 50*[Integer1]
expr = List(*data, conversion_fn=None)  # not sure there if this is supported now
```

File: mathics/core/expression.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (mmatera) [reply mathics/core/expression.py:291]: > As I have said in the past, am not a fan of isinstance(x, BaseElement) , especially now as the class structure has all sorts of quirks in it. We have seen that there are leaves of S-Expressions that aren't BaseElement.

Is there a better way to check if the element is a Mathics object? maybe 
```
isinstance(x, (Atom, Expression))
```  
in any case, I prefer this before doing something like  `getattr(x, "is_mathics_object", False)`

File: mathics/core/expression.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (mmatera) [reply mathics/core/expression.py:227]: Here is the example:
```
>> from mathics.session import MathicsSession
>> from mathics.core.expression import Expression, Integer
>> expr = Expression("Plus", 10*[1], element_conversion_fn=Integer, element_properties={"_elements_fully_evaluated": True,"_is_flat": True,"_is_ordered": True})
>> type(expr.elements[0])
<class 'list'>
```

File: mathics/core/expression.py [hunk 9]
```diff
@@ -560,8 +603,8 @@ def elements(self):
         return self._elements
 
     @elements.setter
-    def elements(self, values):
-        self._elements = self._build_elements(values)
+    def elements(self, values: Iterable, **kwargs: Optional[dict]):
```
mmatera, line 606: I guess that the setter never is called with this optional value.

mathics/core/expression.py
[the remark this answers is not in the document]
rocky: > > For `Expression("Plus", 50*[1])` doing it this way works and sets the properties correctly.
> > As done in the `test_expression_constructor` you can also use `Expression("Plus" 50*[1], conversion_fn=Integer, element_properties={"_elements_fully_evaluated": True, "_is_flat": True, "is_ordered: True})` if you want be explicit about the elements properties so Expression does not try to evaluate, flatten or order the elements.
> 
> Following the line 225,
> 
> ```
> self._elements = (
>                 elements
>                 if self._elements_fully_evaluated 
>                 else (conversion_fn(element) for element in elements)
>             )
> ```
> 
> if you set that the element is fully evaluated, then the conversion shouldn´t be done. So, I do not see where `Expression("Plus", 50*[1], conversion_fn=Integer, element_properties={"_elements_fully_evaluated": True, "_is_flat": True, "is_ordered: True})` applies conversion_fn over the elements. Probably it happens in another place.


My mistake here.  If you are passing a conversion function, you _can't_ set element properties as the code is currently written. The idea here is that if you have a conversion function `Expression._build_elements()` does the computation; it not has to iterate over the list anyway, and the property calculations are simple and not expensive.

So that should have been more simply: `Expression("Plus", 50*[1], conversion_fn=Integer)` . 

You have convinced me that this interface is a little too complicated. But right now this is not where I want to focus on. 

My take right now is to keep https://github.com/Mathics3/mathics-core/pull/282#partial-pull-merging off to the side for now and come back to it after we've made a full pass over evaluation. 

... 
>>  The place I was concerned about is when we have long lists or lots of data. You find this for example in plotting.
> 
> OK, but if you want to build a Plot, usually you do not have a mixture of Python and Mathics native numbers: well you pass a NumPy array, well you pass a `List` `Expression` where all the elements are native numbers. In any case, all this conversion stuff comes up just to make some typical code easier to read and write. It is unlike, (and probably, we should discourage) to have code like
> 
> ```
> a=sympy.Symbol("a")
> b = 20*[1.]
> c = Icecream("mint")
> expr = Expression("F", a , b, c, conversion_fn=funny_converter,  {"_elements_fully_evaluated: False", "_is_flat": True})
> ```

Right now I am thinking that there might be a special SympyExpression class for wrapping SymPy expressions that are used in Mathics. `from_sympy()`  would be used to extract a Mathics expression from this, but it likely that things would be kept in SymPy format as long a possible. This might have come up in https://github.com/Mathics3/mathics-core/issues/20#issuecomment-930319860 and the may get around slowness in https://github.com/Mathics3/mathics-core/issues/103#issuecomment-1009311644

I am not certain about this, but it is a possibility. 

> 
> Also, I do not see the point of allowing
> 
> ```
> data = 50*[1]
> expr = Expression(SymbolList, *data, conversion_fn=Integer)
> ```
> 
> which can be written in a clearer (and sometimes more efficient) way as
> 
> ```
> data = 50*[Integer1]
> expr = Expression(SymbolList, *data, conversion_fn=None)
> ```
> 
> (or even)
> 
> ```
> data = 50*[Integer1]
> expr = List(*data, conversion_fn=None)  # not sure there if this is supported now
> ```

I either don't understand what you are talking about and possibly don't agree, but I don't think it worth the effort debate any of this.

mathics/core/expression.py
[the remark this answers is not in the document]
rocky: Not something I want to think about or debug right now. Make a list somewhere and we can come back to these kinds of things when we have a better handle on performance. Thanks.

mathics/core/expression.py
rocky: Let me state a principle before coming back to whether to add the `kwargs` argument. 

In general, the kinds of things that need to get done takes precedence over mechanisms and syntactic sugar.  The whole idea of allowing elements to be set might be wrong, and might be avoided with a better interface and possibly better modularity of the code. 

The reason `kwargs` was added was to allow being able to set elements and specify properties without going through the Expression  constructor.   The place that used that could have done so probably the same and skipped a call by calling `self._build_elements()` directly and then probably the leading `_` in `_build_elements()` should be removed. 

So  `**kwargs` could be removed which would make the elements setter probably more normal and less surprising. 

As with the other comments here, all of this is futzing around at the expense of devoting time to more important places.

It's not that I don't think this shouldn't be addressed, I can make note of this and after a pass is made over evaluation if such things are marked or there are alternative interfaces, then we should come back to this and address these kinds of things.

rocky: Removed in a1280cc9cec1c0335630ca1e5d5e7465ab8a7095

mathics/core/expression.py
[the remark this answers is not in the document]
rocky: I wrote: 
> You have convinced me that this interface is a little too complicated. 

However in looking at use cases such as those found in Range of constructing.py, I think this may be the right thing. 

It may turn out that we don't need to set element properties at all (or rarely) when we have a pre-formed Expression.

I am not certain about that. But this is something that can be dealt with later.

mathics/builtin/list/constructing.py
```diff
@@ -110,7 +111,7 @@ def rec(rest_dims, current):
                     level.append(rec(rest_dims[1:], current + [index]))
                 return Expression(head, *level)
             else:
-                return Expression(f, *(Integer(index) for index in current))
+                return Expression(f, *current, element_conversion_fn=Integer)
```
mmatera, line 114: This is the case in which I think it makes sense to tell the constructor that the elements are '_fully_evaluated'  (after the conversion)

mmatera (approved):
OK, I think now at least the change in the interface goes in the right direction. Let's  merge this and iterate.

mathics/builtin/list/constructing.py
rocky: With the most recent changes and in tracing execution it feels to me that the changes in https://github.com/Mathics3/mathics-core/blob/9c91a7641eaae6536103201dd73f724c55d0c09e/mathics/core/expression.py#L303-L309 pretty much handle detection whether evaluation is needed in the most egregious situations. 

If you can find an example where that is not so, I would be interested in seeing an example.

For me right now, the question is more about how often we will pass a fully formed list and know its properties. Right now I think things are pretty isolated and harmless that it is not worth worrying about too much (unless of course you have an specific example in mind that is likely go occur.)

mmatera: The point is that in this line you create an expression with all the elements "fully evaluated". So, you could pass this information to the constructor, at the time you want to convert the (pythonspeed.com n) ints to (mathics) Integer. In any case, as I told you before, I think this is ready to merge.

PR #282, opened 2022-05-08
Avoiding element conversion in the setter for Expression.element

mmatera:
@rocky, as I told you, I agree with the idea of passing options to the __init__ method of Expression. What I was talking about is the conversion step. Elements should be converted from python just if they are not instances of ``BaseElement``. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/282"></a>

### DIFF
--- a/test/test_format.py
+++ b/test/test_format.py
@@ -375,7 +375,7 @@ def test_makeboxes_outputform_text(
         ("Subsuperscript[a, p, q]", "a_p^q", None),
         (
             "Integrate[F[x],{x,a,g[b]}]",
-            "\\int_a^{g\\left[b\\right]} F\\left[x\\right] \uf74cx",
+            "\\int_a^{g\\left[b\\right]} F\\left[x\\right] \\, dx",
             None,
         ),
         ("a^(b/c)", "a^{\\frac{b}{c}}", None),
--- a/test/test_format.py
+++ b/test/test_format.py
@@ -375,7 +375,7 @@ def test_makeboxes_outputform_text(
         ("Subsuperscript[a, p, q]", "a_p^q", None),
         (
             "Integrate[F[x],{x,a,g[b]}]",
-            "\\int_a^{g\\left[b\\right]} F\\left[x\\right] \\, dx",
+            "\\int_a^{g\\left[b\\right]} F\\left[x\\right] \uf74cx",
             None,
         ),
         ("a^(b/c)", "a^{\\frac{b}{c}}", None),
--- a/mathics/algorithm/parts.py
+++ b/mathics/algorithm/parts.py
@@ -69,11 +69,11 @@ def set_part(varlist, indices, newval):
                 raise PartDepthError
             try:
                 if pos > 0:
-                    part = cur._elements[pos - 1]
+                    part = cur.elements[pos - 1]
                 elif pos == 0:
                     part = cur.get_head()
                 else:
-                    part = cur._elements[pos]
+                    part = cur.elements[pos]
             except IndexError:
                 raise PartRangeError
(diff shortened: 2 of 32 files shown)


test/test_format.py
```diff
@@ -375,7 +375,7 @@ def test_makeboxes_outputform_text(
         ("Subsuperscript[a, p, q]", "a_p^q", None),
         (
             "Integrate[F[x],{x,a,g[b]}]",
-            "\\int_a^{g\\left[b\\right]} F\\left[x\\right] \uf74cx",
+            "\\int_a^{g\\left[b\\right]} F\\left[x\\right] \\, dx",
```
mmatera, line 378: I found this by chance: In LaTeX, the integral measure should be put with a 'd' instead the unicode character.
For some reason, the changes in this PR fixed it.

rocky: Ok. Put in another PR just for  that or just change it on mater.

rocky:
> @rocky, as I told you, I agree with the idea of passing options to the **init** method of Expression. What I was talking about is the conversion step. Elements should be converted from python just if they are not instances of `BaseElement`.
> 

Whether to do the conversion step should be a controlled by a keyword parameter since the caller knows that.

rocky:
Sigh - we are working at cross purposes and working on the same code. Please see the update 7f4aa786460b1d0db3ec299ea9417d4529a2b11f

mmatera:
> Sigh - we are working at cross purposes and working on the same code. Please see the update [7f4aa78](https://github.com/Mathics3/mathics-core/commit/7f4aa786460b1d0db3ec299ea9417d4529a2b11f)

Yes, I was merging it. I did it because I though it was easier to check the pros and cons of both slightly approaches. For example, I would expect that is better to tell the `__init__` method that is not necessary to do a conversion is better that passing a function. Also, it would be better to check over which elements is necessary to do the conversion instead of converting all by default. What do you think about?


> 
> Feel free comment on the differences in approach.
> 
> I decided not to add the conversion function as in the element property parameter because it isn't really an property about the elements collection. I think of it as about how the creation gets done.

I have the same feeling.

> 
> From [#281 (comment)](https://github.com/Mathics3/mathics-core/pull/281#issuecomment-1120415110)
> 
> > In advance of this, I'll try to update the PR so that we allow a keyword parameter for the kind of conversion function to be used, when it is needed.

The problem with the conversion function is that, if you know you need something different than `from_pỳthon`, probably you could apply it in the caller function. My guess is that is better to tell to '__init__' when such a call is not needed.

rocky:
> The problem with the conversion function is that, if you know you need something different than `from_pỳthon`, probably you could apply it in the caller function. 

No I strongly believe these kinds of setup code should _not_ be done in the caller. The constructor should be constructing expressions rather than parts of construction split many places in the code. 

I see code like: 

```
        expr = Expression(self._head)
        expr._elements = self._elements
```

several times. There kind of thing should be make explicit and put in the Expression constructor. 

It is true that most of the time you either the expression has been converted or not. However there are cases when what we want is "Integer", or "from_sympy",  in addition to "from_python".

mmatera:
> No I strongly believe these kinds of constructor setup code should _not_ be done in the caller. The constructor should be constructing expressions rather than parts of construction split many places in the code.
> 
> I see code like:
> 
> ```
>         expr = Expression(self._head)
>         expr._elements = self._elements
> ```

For this case, I agree. But here we do not need in principle any conversion.  What I do not have in mind is a case in which we gain something by telling to the constructor which conversion function should use. The constructor should construct the expression, but not necessarily to infer how to convert the ingredients.

mmatera:
Superseded by  #281

PR #283, opened 2022-05-08
Contexts and message assignment

mmatera:
This PR fixes the behavior of assignments to Message in protected symbols. Also, pytests in ``pytest_context.py`` where improved to check the compatibility with the WMA behavior.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/283"></a>

### DIFF
--- a/mathics/core/definitions.py
+++ b/mathics/core/definitions.py
@@ -100,8 +100,8 @@ class Definitions(object):
         self._packages = []
         self.current_context = "Global`"
         self.context_path = (
-            "Global`",
             "System`",
+            "Global`",
         )
         self.trace_evaluation = False
         self.timing_trace_evaluation = False
--- a/test/test_format.py
+++ b/test/test_format.py
@@ -375,7 +375,7 @@ def test_makeboxes_outputform_text(
         ("Subsuperscript[a, p, q]", "a_p^q", None),
         (
             "Integrate[F[x],{x,a,g[b]}]",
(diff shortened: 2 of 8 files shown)


test/test_format.py
```diff
@@ -375,7 +375,7 @@ def test_makeboxes_outputform_text(
         ("Subsuperscript[a, p, q]", "a_p^q", None),
         (
             "Integrate[F[x],{x,a,g[b]}]",
-            "\\int_a^{g\\left[b\\right]} F\\left[x\\right] \uf74cx",
```
rocky, line 378: Ideally this kind of thing would be in a separate PR if not separate commit. It's not strictly necessary but something that would be nice to keep in mind. Thanks.

mathics/core/definitions.py
```diff
@@ -100,8 +100,8 @@ def __init__(
         self._packages = []
         self.current_context = "Global`"
         self.context_path = (
-            "Global`",
             "System`",
+            "Global`",
```
rocky, line 104: Do I understand correctly, that are context path had been wrong previously? 

Other than this message problem,  how can we tell? 

Thanks in advance for the education here.

mmatera: Yep: in WMA, "System`" is before that "Global`" in ContextPath, and in the order of name resolution.

rocky: Ok - thanks for the clarification. Given this, it probably should me mentioned in the CHANGES.rst as this is kind of a big deal.  (And the name of the PR doesn't give any clue that this is going on either.)

PR #284, opened 2022-05-08
fix format pytest for integrate in tex

mmatera:
As I mentioned before, a recent modification of the code, fixed the behavior of `boxes_to_tex`.  
 Here I fixed the pytest accordingly.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/284"></a>


mmatera:
That is funny: this works on my system, but not here...

rocky:
> That is funny: this works on my system, but not here...

It is a sign of several that Mathics isolating the environment, e.g. whether unicode is available or not.  

My suggestion is to mark this skipped for now. At some point we will come back to how operator formatting in Mathics should be handled.

PR #289, opened 2022-05-17
Reduce unnecessary calls to from_python ...

rocky:
in Expression._build_elements().

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/289"></a>

### DIFF
--- a/mathics/builtin/lists.py
+++ b/mathics/builtin/lists.py
@@ -957,18 +957,24 @@ class _IterationFunction(Builtin):
     def apply_max(self, expr, imax, evaluation):
         "%(name)s[expr_, {imax_}]"
 
-        index = 0
-        imax = imax.evaluate(evaluation)
-        imax = imax.numerify(evaluation)
-        if isinstance(imax, Number):
-            imax = imax.round()
-        imax = imax.get_float_value()
-        if imax is None:
-            if self.throw_iterb:
-                evaluation.message(self.get_name(), "iterb")
-            return
+        # Even though `imax` should be an integeral value, its type does not
+        # have to be an Integer.
+        if isinstance(imax, Integer):
+            py_max = imax.get_int_value()
+        else:
+            imax = imax.evaluate(evaluation)
+            imax = imax.numerify(evaluation)
+            if isinstance(imax, Number):
+                imax = imax.round()
+            py_max = imax.get_float_value()
+            if py_max is None:
+                if self.throw_iterb:
+                    evaluation.message(self.get_name(), "iterb")
+                return
+
         result = []
-        while index < imax:
+        index = 0
+        while index < py_max:
             evaluation.check_stopped()
             try:
                 result.append(expr.evaluate(evaluation))
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -201,11 +201,6 @@ class Expression(BaseElement, NumericOperators):
     leaves: typing.List[Any]
     _sequences: Any
 
-    # __new__ seems to be used because BaseElement does some
-    # questionable stuff using new.
-    # See if there's a way to get rid of this, or ensure that this isn't causing
-    # a garbage collection problem.
-
     def __init__(self, head, *elements, **kwargs):
         super().__init__(self)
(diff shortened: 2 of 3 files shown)


mathics/core/expression.py
```diff
@@ -297,7 +299,12 @@ def _build_elements(
 
         last_converted_elt = None
         for element in elements:
-            converted_elt = conversion_fn(element)
+            # FIXME: we should have isinstance(element(BaseElement, Builtin))
```
mmatera, line 302: Actually, what should not be converted are InstanceableBuiltins. And I agree: these elements in the future should be a subclass of BaseElement.

mmatera (approved):
LGTM

mathics/core/expression.py
rocky: Ok. Will change that later. 

I also wonder if this shouldn't be done instead in from_python.... benchmarking will tell us though.

rocky: Perhaps Instanceable Builtins would be better named BuiltinElement to replace something about the mechanism (is some sort of thing that we get an instance out of) with something about its intended use (acts as an element in an Expression).

mmatera: Yep, I agree.

mmatera: > Ok. I will change that later.
> 
> I also wonder if this shouldn't be done instead in from_python.... benchmarking will tell us though.

Actually, it is already done here:
https://github.com/Mathics3/mathics-core/blob/1106ff85ac40e5d359232a4e168118d516f73691/mathics/core/atoms.py#L1076

The problem is that making function calls in Python is onerous. This is why, for example, I would prefer to avoid the call at all instead of calling something like `identity_fn`.
Also, having an optional parameter adds some overhead:
```
li = [i for i in range(10)]        

def identity_fn(x):
    return x

def  f0(*u):
    return None


def  f1(*u, g=identity_fn):
    return None

def  f2(*u, **kwargs):
    g = kwargs.get("g", identity_fn)
    return None

%timeit f0(li)  
%timeit f1(li)
%timeit f2(li)
```
gives
```
50.4 ns ± 1.04 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
58.7 ns ± 4.85 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
79 ns ± 2.55 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```
This is why I think that we do not gain too much in most cases by doing the conversion inside `_build_elements`. Mostly because in the most common calls (which come from the calls inside `rewrite_apply_eval_step`) are that the arguments are already Mathics Elements.

mmatera: @rocky, the point here was not to decide which of these functions are better defined but to have an idea of the overall factor introduced by adding optional parameters in a very frequently called function.   If you call 1000 times a 50ns function, the 50ns becomes 50 in microseconds. 
On the other hand, if you are not worried about the cost of a call, you shouldn't worry either about the number of calls to `from_python`: most of the time, this function does nothing, because, on its first line, it checks if the conversion is needed.

mathics/builtin/lists.py
```diff
@@ -957,18 +957,22 @@ def apply_range(self, expr, i, imax, evaluation):
...
-        imax = imax.evaluate(evaluation)
-        imax = imax.numerify(evaluation)
-        if isinstance(imax, Number):
-            imax = imax.round()
-        imax = imax.get_float_value()
-        if imax is None:
-            if self.throw_iterb:
-                evaluation.message(self.get_name(), "iterb")
-            return
+        if isinstance(imax, Integer):
```
mmatera, line 960: Nop: In WMA, 
```
In[2]:= Do[Print[i],{2.3}]                                                      
i
i
```
So, imax could be perfectly a Real number.

mmatera: In any case, I agree with the proposed change.

rocky: Ok - Thanks for the information. 'll make note of this in a comment somewhere.

PR #290, opened 2022-05-17
remove __init__ method from BaseElement

mmatera:
BaseElement contains a ``__init__`` method that just initializes some attributes needed in ``Expression``.  I propose to remove it, both to have a cleaner interface, and to avoid those other elements like `Symbol` or `Integer` call it when are created.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/290"></a>

### DIFF
--- a/mathics/core/element.py
+++ b/mathics/core/element.py
@@ -109,14 +109,15 @@ class BaseElement(KeyComparable):
     # FIXME: kwargs seems to be is needed because Real.__new_() takes a parameter
     # and magically that gets turned into kwargs here.
     # Figure out how to address this.
-    def __init__(self, *args, **kwargs):
-        self.options = None
-        self.pattern_sequence = False
-        # This property would be useful for a BoxExpression
-        # (see comment in mathocs.core.expression.) However,
-        # WL has a way to handle the connection between
-        # an expression and a Box expression ``InterpretationBox``.
-        # self.unformatted = self  # This may be a garbage-collection nightmare.
+
+    #    def __init__(self, *args, **kwargs):
+    #        self.options = None
+    #        self.pattern_sequence = False
+    # This property would be useful for a BoxExpression
+    # (see comment in mathocs.core.expression.) However,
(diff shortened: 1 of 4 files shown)


mathics/core/element.py
```diff
@@ -106,18 +106,6 @@ class BaseElement(KeyComparable):
...
-        self.options = None
-        self.pattern_sequence = False
-        # This property would be useful for a BoxExpression
-        # (see comment in mathocs.core.expression.) However,
-        # WL has a way to handle the connection between
-        # an expression and a Box expression ``InterpretationBox``.
-        # self.unformatted = self  # This may be a garbage-collection nightmare.
-
     # comment @mmatera: The next method have a name that starts with ``apply``.
     # This obstaculizes to define ``InstanceableBuiltin``
```
rocky, line 110: If this commit goes in this comment is no longer valid and it would be nice to remove it then.

rocky:
@mmatera This is good and I see a reason it shouldn't be commited.

mathics/core/element.py
mmatera: Do you mean the one about ``apply_rule``? It is about the next method.

rocky: Ah - you are correct. I thought it applied to the one above not following. 

Reading it though, it has some spelling and grammar mistakes. But that can be dealt with some other time.

mmatera:
BTW, I would like to implement the change suggested in the comment. @rocky, would you agree in making that change?

rocky:
> BTW, I would like to implement the change suggested in the comment. @rocky, would you agree in making that change?

Yes. On the Expression evaluation benchmarking side, it looks like there needs to be more code refactoring for us to see more pervasive improvement. 

So better structuring of the code is sorely needed.  It annoys me that we can't make performance progress because of intertwined code because it wasn't better documented previously, and therefore not better understood and therefore modularity and hacks were added that now need to be understood and/or undone.

PR #291, opened 2022-05-17
improving Share[] 

mmatera:
In WMA, the command  ``Share[]`` tries to recover memory by replacing repeated expressions by references to a single instance of the expression. For example, if we have two definitions
```
In[1]:= table=Table[i^2,{i,1000}];                                              
In[2]:= a=F[table,2];                                                           
In[3]:= b=F[table,3];                                                           
In[4]:= MemoryAvailable[]                                                       
Out[4]= 9147990016
In[5]:= Share[]                                                                 
Out[5]= 17104
``` 
in a way that 17104 bytes are recovered by replacing (internally) the three lists in the definitions of `table`,`a` and `b`  by a single list expression, that is referenced in the three places.

In Mathics, ``Share[]`` is doing nothing, and returns always `0`. This PR tries to improve this by calling explicitly the garbage collector.  

reduces the amount of used memory 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/291"></a>

### DIFF
--- a/mathics/builtin/system.py
+++ b/mathics/builtin/system.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 
 """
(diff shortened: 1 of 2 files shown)


rocky:
Makes sense to me. I am for merging when you think ready.

mathics/builtin/system.py
```diff
@@ -596,16 +596,20 @@ class Share(Builtin):
 
     def apply_0(self, evaluation) -> Integer:
         """Share[]"""
+        totalmem = psutil.virtual_memory().available
```
rocky, line 599: psutil is currently an _optional_ requirement.

rocky:
Looks like the Share doc should be updated. It currently says nothing is done.

mathics/builtin/system.py
mmatera: OK, so I can put this version in an "if" statement, and update the documentation accordingly.

mmatera:
@rocky, when you have the chance, please have a look at the docstring, and if it is OK, merge this.

PR #292, opened 2022-05-18
BaseElement.apply_rules-> BaseElement.do_apply_rules

mmatera:
Following https://github.com/Mathics3/mathics-core/pull/290#issuecomment-1129348529, here I implemented the changes to make  `BoxConstruct`  inherits from BaseElement.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/292"></a>

PR #293, opened 2022-05-18
Expression and Symbol are EvalMixin

mmatera:
This follows #292 and #186. In this PR, ``BaseExpression``   is simplified by removing certain methods that just make sense in its daughter classes. In particular, `get_option_values` is defined just for `Symbol` and `Expression`. Also, these classes are the only ones for which the methods `evaluate` and `evaluate_next` makes sense. Now, these methods are not included in ``BaseExpression`` but in a new class ``EvalMixin``. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/293"></a>


mathics/core/element.py
```diff
@@ -546,3 +465,34 @@ def to_sympy(self, **kwargs):
 
     def to_mpmath(self):
         raise NotImplementedError
+
+
+class Evaluable:
```
rocky, line 470: Evaluable isn't an English word, but I can't think of what the right word is.

mmatera: ``Computable`` maybe?

rocky: How about `EvalMixin`.  We use [`Mixin`](https://en.wikipedia.org/wiki/Mixin) in `ImmutableValueMixin` and at one time I used it for NumericOperators, but I think that name with "Mixin" was objected to.

rocky: BTW in case you had not caught on before, I didn't like the name Instanceable either, and part of the reason there is that Instanceable is also not an English word that is posing as an English word. In addition it doesn't convey anything, but that's a separate thing as well. 

Continuing a convention that is objectionable just gives me cause to remind me of this fact. Please don't give me more reminders of what I dislike about this code base. Thanks.

mmatera: ``EvalMixin`` could work.  In any case, as you probably have noticed, I am not particularly brilliant in English. Sorry for that. In any case, what I would like to have is a class that collects that part of the interface, and that allows us to recognize if it makes sense to attempt the evaluation.

rocky: Read the Wiki link on Mixin.  As a Mixin is how we seem to be using this class.

That reminds me that when in looking things over before in the change from apply_rewrite_rule to do_rewrite_rule I had thought about whether we meant "apply" or "eval" here, and we indeed do mean "apply" here. 

As I've said before those _other_  apply_xxx functions would more properly be called "eval" methods. So it is ironic that this renaming had to be done to support some other misfeature I guess because it was less work to change this than all of the misnamed eval methods.

rocky:
@mmatera from my standpoint not only is this good, but it would be good to get it into master since https://github.com/Mathics3/mathics-core/tree/expression-interface-refactor is based off of this. 

Your thoughts?

mmatera:
> @mmatera from my standpoint not only is this good, but it would be good to get it into master since https://github.com/Mathics3/mathics-core/tree/expression-interface-refactor is based off of this.
> 
> Your thoughts?

@rocky,  if you agree, then let's merge this and iterate.

mmatera:
Just a last change: I remove the duplicated inheritance in `BoxExpression` (now, `BuiltinElement` already has `BaseElement` as a base class)

rocky:
> Just a last change: I remove the duplicated inheritance in `BoxExpression` (now, `BuiltinElement` already has `BaseElement` as a base class)

Great! Please merge in after CI gives green. Thanks.

PR #294, opened 2022-05-18
By by is true

mmatera:
Following in the direction of looking a simpler and clearer interface for core.BaseElements, in this PR the use of `is_true()` was completely removed. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/294"></a>


rocky:
LGTM - it does make it work kmore like a Lisp or Lisp-like interpreter does this.

PR #295, opened 2022-05-23
More pervasive use of Symbols in Expressions and other things of that kind

rocky:
There will be a lot more of things like this which are needed in the expression-interface-refactor

This splits out some of the trivial stuff that can go in now.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/295"></a>


TiagoCavalcante (approved):
LGTM

PR #296, opened 2022-05-24
create_infix calls Expression() with Integer ..

rocky:
not an int argument. This will be required in the future for
Expression() - otherwise use to_expression().

One more place leaves -> elements.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/296"></a>


mmatera:
LGTM

PR #297, opened 2022-05-25
Range speedup and One more Expression() conversion

rocky:
Have to hold off on the Range conversion problem until I have time to figure out what's wrong

<strike>
Check this out.

Using Python list comprehension with range:

```
In[1]:= Timing[Range[1000]][[1]]
Out[1]= 0.00679923
```

Previous method:

```
In[1]:= Timing[Range[1000]][[1]]
Out[1]= 0.0269802

```
</strike>

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/297"></a>

### DIFF
--- a/mathics/builtin/numbers/algebra.py
+++ b/mathics/builtin/numbers/algebra.py
@@ -346,14 +346,14 @@ def find_all_vars(expr):
     return variables
 
 
-def find_exponents(expr, var):
+def get_exponents_sorted(expr, var) -> list:
     """
-    Find all exponents of var in expr
+    Return a sorted list of exponents of var in expr
     """
     f = expr.to_sympy()
     x = var.to_sympy()
     if f is None or x is None:
-        return {0}
+        return [Integer0]
 
     result = set()
     for t in f.expand(power_exp=False).as_ordered_terms():
@@ -971,7 +971,7 @@ class CoefficientList(Builtin):
             # single & multiple variables cases
             if not form.has_form("List", None):
                 return Expression(
-                    "List",
+                    SymbolList,
                     *[
                         _coefficient(
                             self.__class__.__name__, expr, form, Integer(n), evaluation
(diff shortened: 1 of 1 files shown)


mathics/builtin/list/constructing.py
```diff
@@ -185,6 +185,14 @@ class Range(Builtin):
     def apply(self, imin, imax, di, evaluation):
         "Range[imin_?RealNumberQ, imax_?RealNumberQ, di_?RealNumberQ]"
 
+        if (
+            isinstance(imin, Integer)
+            and isinstance(imax, Integer)
+            and isinstance(di, Integer)
+        ):
+            result = [Integer(i) for i in range(imin.value, imax.value, di.value)]
```
rocky, line 193: I had considerd using `x.get_int_value()` and balked. That method name has alway felt clumsy and here, I don't see what the benefit is. If there are _x_ could be one of _several_ numeric types, then sure, having a common method name to do whatever it takes to get an int makes sense.

mathics/builtin/list/constructing.py
```diff
@@ -185,6 +185,14 @@ class Range(Builtin):
...
+            return Expression(SymbolList, *result)
```
rocky, line 194: In the expression refact branch one also has the option if indicating that *result is flat, fully evaluated, and ordered iff di > 0.

mathics/builtin/numbers/algebra.py
```diff
@@ -1351,11 +1351,11 @@ def apply(self, expr, form, h, evaluation):
             return Expression(SymbolDirectedInfinity, Integer(-1))
 
         if not form.has_form("List", None):
-            return Expression(h, *[i for i in find_exponents(expr, form)])
+            return Expression(h, *[i for i in get_exponents_sorted(expr, form)])
```
rocky, line 1354: As mentioned above, in the expression refactor branch we can make use of knowing that coefficients will be ordered.

TiagoCavalcante (approved):
LGTM

PR #298, opened 2022-05-26
Faster Range[] most of the time...

rocky:
Using Python range(). Previous attempt was simply off by one by one.
We get a 5x speed improvement with Python range().

[This time, for sure!](https://youtu.be/yIBdVkJ9L-k?t=66)

### DIFF
--- a/mathics/builtin/list/constructing.py
+++ b/mathics/builtin/list/constructing.py
@@ -185,6 +185,14 @@ class Range(Builtin):
     def apply(self, imin, imax, di, evaluation):
         "Range[imin_?RealNumberQ, imax_?RealNumberQ, di_?RealNumberQ]"
 
+        if (
+            isinstance(imin, Integer)
+            and isinstance(imax, Integer)
+            and isinstance(di, Integer)
+        ):
+            result = [Integer(i) for i in range(imin.value, imax.value + 1, di.value)]
+            return Expression(SymbolList, *result)
+
         imin = imin.to_sympy()
         imax = imax.to_sympy()
         di = di.to_sympy()


mathics/builtin/numbers/algebra.py
```diff
@@ -1351,9 +1351,13 @@ def apply(self, expr, form, h, evaluation):
             return Expression(SymbolDirectedInfinity, Integer(-1))
 
         if not form.has_form("List", None):
```
rocky, line 1353: We could probably write a specialized version of `has_form("List", None) which I believe comes up a bit. Here, all you need to do is check that Head is `SymbolList`.  

We should benchmark to see if it is worth it though.

mmatera: Maybe even better would be to define List as a BuiltinExpression, which has at least two advantages:
* it would allow to check if is a List by using the standard way `isinstance(x, List)`
* Provides a way to have a simpler evaluation method: ``List`` is a Locked symbol, so all its attributes are always the same.

rocky: I wholeheartedly agree!  Something like this has been in the back of my mind for a while.

The other aspect here that I would like to see in this object is an optional representation as a Python or NumPy list (when that is possible) so that we reduce conversion for those operations that are done in SymPy, numpy, mpmath, or Python. 

But that can come later.

As I have asked before, in Mathematica can you redefine List in any way? If not, the parser could form List literals directly.

mmatera: > I wholeheartedly agree! Something like this has been in the back of my mind for a while.
> 
> The other aspect here that I would like to see in this object is an optional representation as a Python or NumPy list (when that is possible) so that we reduce conversion for those operations that are done in SymPy, numpy, mpmath, or Python.
> 

Yep: I guess it could be helpful in some places like in linear algebra or graphics packages. 

> But that can come later.
> 
> As I have asked before, in Mathematica can you redefine List in any way? If not, the parser could form List literals directly.

In WL, ``List`` is ``Locked`` and ``Protected``, so you can only modify its behavior by defining upvalues
(something like ``List[MySymbol, a,b]^:=g[a,b]``) (because the rule is attached to ``MySymbol`` and not to ``List``.

TiagoCavalcante (approved):
LGTM - such a speedup!

mathics/builtin/numbers/algebra.py
rocky: > In WL, `List` is `Locked` and `Protected`, so you can only modify its behavior by defining upvalues (something like `List[MySymbol, a,b]^:=g[a,b]`) (because the rule is attached to `MySymbol` and not to `List`.


Ok - I am not sure I understand this fully. But this means that whenever the parser sees "List" it is a Mathics List and therefore it could go directly into this ListExpression object, right? 

BTW when ListExpression exists, then Range would return that rather than an Expression.

ListExpression with the property elements_fully_evaluated means we have a list literal and that might be an important property to note in of itself.

rocky:
> LGTM - such a speedup!

I am getting the impression that a lot of the slowness is because we aren't using Python objects and functions when we can. There is a lot of time spent in converting into and out of Mathics structures.  

And here, although WL offers extends Python's version of range by allowing increments of rational numbers, in practice it is rarely used. I guess for simplicity (a toy Mathics interpreter) there was just this one-size-fits-all kind of way of doing things. But that means that for 99% of the situations where Range is used with an integer, we were 5 times slower. 

If one wanted to improve things in the rare cases that remain such as `Range[5.]` or with rational increments, more could be done. But for now I just want get the common case and come back to this when we need to squeeze out the last bit.  

As I look at things overall I see a lot of room for improvement all over the place, that is far greater than what we would get by handing the other lesser forms of Range.

mathics/builtin/numbers/algebra.py
mmatera: > 
> Ok - I am not sure I understand this fully. But this means that whenever the parser sees "List" it is a Mathics List and therefore it could go directly into this ListExpression object, right?

Yep

> 
> BTW when ListExpression exists, then Range would return that rather than an Expression.
> 

Exactly

> ListExpression with the property elements_fully_evaluated means we have a list literal and that might be an important property to note in of itself.

yes, because then, in construction like `F[{....}]`, the `evaluation` method of the expression does not need to try to evaluate the list.

rocky: > yes, because then, in construction like F[{....}], the evaluation method of the expression does not need to try to evaluate the list.

Actually, that is currently how expression-interface-refactor works [right now]( https://github.com/Mathics3/mathics-core/blob/expression-interface-refactor/mathics/core/expression.py#L1262-L1270).

PR #300, opened 2022-05-26
Reduce Expression() constructor conversions in datentime...

rocky:
Although this helps a little here, the main payoff will be when _all_ of the Expressions get sorted into those that may require*conversion -- to_expression() -- and those that *don't* --  Expression(). 

That other branch will then kick in and we should see a bigger improvement overall.

However to the extent possible, I am doing this little by little rather than one huge PR. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/300"></a>

### DIFF
--- a/mathics/builtin/datentime.py
+++ b/mathics/builtin/datentime.py
@@ -20,11 +20,13 @@ from mathics.core.atoms import Integer, Real, String, from_python
 from mathics.core.attributes import hold_all, no_attributes, protected, read_protected
 from mathics.core.evaluation import TimeoutInterrupt, run_with_timeout_and_stack
 from mathics.core.element import ImmutableValueMixin
-from mathics.core.expression import Expression
-from mathics.core.symbols import Symbol
+from mathics.core.expression import Expression, to_expression
+from mathics.core.symbols import Symbol, SymbolList
 from mathics.core.systemsymbols import (
     SymbolAborted,
     SymbolInfinity,
+    SymbolNull,
+    SymbolRowBox,
 )
 
 from mathics.settings import TIME_12HOUR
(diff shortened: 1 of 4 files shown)


mathics/builtin/datentime.py
```diff
@@ -96,6 +98,12 @@ def total_seconds(td):
 else:
     total_seconds = timedelta.total_seconds
 
+SymbolAbsoluteTime = Symbol("AbsoluteTime")
+SymbolDateObject = Symbol("DateObject")
+SymbolDateString = Symbol("DateString")
+SymbolGregorian = Symbol("Gregorian")
+SymbolPause = Symbol("Pause")
+
```
rocky, line 106: I think this is a better way to handle Symbols that largely appear here and won't in other places (as opposed to adding them to systemsymbols). 

For those things though that are common systemsymbols is fine.

mathics/builtin/datentime.py
```diff
@@ -1067,10 +1084,10 @@ class Now(Predefined):
     summary_text = "current date and time"
 
     def evaluate(self, evaluation):
-        return Expression("DateObject").evaluate(evaluation)
+        return Expression(SymbolDateObject.evaluate(evaluation))
 
 
-if sys.platform != "win32" and ("Pyston" not in sys.version):
+if sys.platform != "win32" and not hasattr(sys, "pyston_version_info"):
```
rocky, line 1090: The preferred way for checking for Pyston is like this. See https://github.com/pyston/pyston#checking-for-pyston-at-runtime

mathics/builtin/datentime.py
```diff
@@ -665,14 +673,15 @@ def apply_makeboxes(self, datetime, gran, cal, tz, fmt, evaluation):
...
+            fmt = Expression(SymbolList, String("DateTimeShort"))
+        fmtds = Expression(SymbolDateString, datetime, fmt).evaluate(evaluation)
         if fmtds is None:
             return
         # tz = Expression("ToString", tz).evaluate(evaluation)
-        tz = int(tz.to_python())
-        tz = String(str(tz))
-        return Expression("RowBox", Expression("List", "[", fmtds, "  GTM", tz, "]"))
+        tz_string = String(str(int(tz.to_python())))
+        return to_expression(
```
TiagoCavalcante, line 682: Why `to_expression` here?

rocky: Because there are strings in the elements. I could wrap the string arguments in String, but here I was lazy. 

I will probably revise in the expression interface refactor branch where I can also list the element properties.

PR #301, opened 2022-05-27
Make a pass over color_directives + more symbols..

rocky:
The additional systemsymbols are mostly used in builtin/arithmetic.py which I'll use in a future PR. I though many of these like Apply are general enough to be here.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/301"></a>

### DIFF
--- a/mathics/builtin/colors/color_directives.py
+++ b/mathics/builtin/colors/color_directives.py
@@ -15,7 +15,7 @@ from mathics.builtin.base import (
 )
 from mathics.builtin.drawing.graphics_internals import _GraphicsDirective, get_class
 from mathics.core.element import ImmutableValueMixin
-from mathics.core.expression import Expression
+from mathics.core.expression import Expression, to_expression
 from mathics.core.atoms import (
     Integer,
     Real,
@@ -25,8 +25,11 @@ from mathics.core.atoms import (
 )
 
 from mathics.core.symbols import Symbol, SymbolList
+from mathics.core.systemsymbols import SymbolApply
 
 from mathics.core.number import machine_epsilon
+
+SymbolOpacity = Symbol("Opacity")
 
 
 def _cie2000_distance(lab1, lab2):
@@ -142,7 +145,7 @@ class Opacity(_GraphicsDirective):
 
     def init(self, item=None, *args, **kwargs):
         if isinstance(item, (int, float)):
-            item = Expression("Opacity", MachineReal(item))
+            item = Expression(SymbolOpacity, MachineReal(item))
             super(Opacity, self).init(None, item)
         self.opacity = item.leaves[0].to_python()
 
@@ -226,7 +229,7 @@ class _ColorObject(_GraphicsDirective, ImmutableValueMixin):
         return self.to_rgba()
 
     def to_expr(self):
-        return Expression(self.get_name(), *self.components)
+        return to_expression(self.get_name(), *self.components)
 
     def to_rgba(self):
         return self.to_color_space("RGB")
@@ -401,15 +404,15 @@ class ColorDistance(Builtin):
 
             def compute(a, b):
                 return Expression(
-                    "Apply",
+                    SymbolApply,
                     distance_function,
                     Expression(
-                        "List",
+                        SymbolList,
                         Expression(
-                            "List", *[Real(val) for val in a.to_color_space("LAB")]
+                            SymbolList, *[Real(val) for val in a.to_color_space("LAB")]
                         ),
                         Expression(
-                            "List", *[Real(val) for val in b.to_color_space("LAB")]
+                            SymbolList, *[Real(val) for val in b.to_color_space("LAB")]
                         ),
                     ),
                 )
@@ -436,7 +439,7 @@ class ColorDistance(Builtin):
                         return
                     else:
                         return Expression(
-                            "List",
+                            SymbolList,
                             *[distance(a, b) for a, b in zip(c1.leaves, c2.leaves)],
                         )
                 else:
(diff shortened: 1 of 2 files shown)


mathics/builtin/colors/color_directives.py
```diff
@@ -401,15 +404,15 @@ def apply(self, c1, c2, evaluation, options):
 
             def compute(a, b):
                 return Expression(
-                    "Apply",
+                    SymbolApply,
                     distance_function,
                     Expression(
-                        "List",
+                        SymbolList,
```
rocky, line 410: @mmatera As I go over the conversion for the Expression interface I am seeing so many Literal constants so many places that, I suspect having that part in alone will be a big win. 

And I suspect we can get that in faster than going over all of the Expressions and sorting the whites from the colored (to use my favorite laundry analogy). 

I also think that neither precludes the other and both  Expression interface change and handling Lists will be an improvement on just one. 

The main thing I worry about is making sure Lists goes in knowing that underneath there are those 3 properties: fully evaluated, ordered, and flat.

mathics/builtin/colors/color_directives.py
```diff
@@ -401,15 +404,15 @@ def apply(self, c1, c2, evaluation, options):
...
                         Expression(
-                            "List", *[Real(val) for val in a.to_color_space("LAB")]
+                            SymbolList, *[Real(val) for val in a.to_color_space("LAB")]
```
rocky, line 412: The idom `*[conversionfn(val) for val in collection]` should encapsulated in some sort of List constructor. Probably not _the_ object constructor which, like Expression should assume everything is converted, but more like the way it is in `to_expression()`.

mathics/builtin/colors/color_directives.py
mmatera: > @mmatera As I go over the conversion for the Expression interface I am seeing so many Literal constants so many places that, I suspect having that part in alone will be a big win.
> 


Do you mean, to use always the plane constructor (the one without the type conversion)? I agree with that.  But mainly, because we can avoid the constructor doing a lot of extra work in cases where the conversions are not needed, what happens in the rewrite_and_apply method, and (probably) the ``do_apply_rule``.


> And I suspect we can get that in faster than going over all of the Expressions and sorting the whites from the colored (to use my favorite laundry analogy).
> 

Yep, this would clarify more what is the landscape we have in front.

> I also think that neither precludes the other and both Expression interface change and handling Lists will be an improvement on just one.
> 
> The main thing I worry about is making sure Lists goes in knowing that underneath there are those 3 properties: fully evaluated, ordered, and flat.

I am not sure to understand this last point: I `List` where a different class, you shouldn´t worry about sorting:  `List` never have the attribute `Orderless`.  They also do not have the attribute `Flat`, but in that case, maybe you gain something if you know that a list is already flat when it is an element of another expression.

mathics/builtin/colors/color_directives.py
mmatera: something like `to_mathics_list`?

rocky: Yes, something like `to_mathics_list()`

mathics/builtin/colors/color_directives.py
rocky: > > @mmatera As I go over the conversion for the Expression interface I am seeing so many Literal constants so many places that, I suspect having that part in alone will be a big win.
> 
> Do you mean, to use always the plane constructor (the one without the type conversion)? I agree with that. But mainly, because we can avoid the constructor doing a lot of extra work in cases where the conversions are not needed, what happens in the rewrite_and_apply method, and (probably) the `do_apply_rule`.

There is a part where I agee, and another part where you may be mistaken. I will elaborate and explain below.

> 
> > And I suspect we can get that in faster than going over all of the Expressions and sorting the whites from the colored (to use my favorite laundry analogy).
> 
> Yep, this would clarify more what is the landscape we have in front.
> 
> > I also think that neither precludes the other and both Expression interface change and handling Lists will be an improvement on just one.
> > The main thing I worry about is making sure Lists goes in knowing that underneath there are those 3 properties: fully evaluated, ordered, and flat.
> 
> I am not sure to understand this last point: I `List` where a different class, you shouldn´t worry about sorting: `List` never have the attribute `Orderless`. They also do not have the attribute `Flat`, but in that case, maybe you gain something if you know that a list is already flat when it is an element of another expression.


Some operations like "Plus" are "Flat", "Orderless", and "Listable". While the properties of the elements of a List are not of use to the List function, it _is_ of use to the a "Plus" function, which it gets from querying the List elements structure.

I will make a stab at a ListExpression class over the weekend so as to make some of these ideas more concrete, and to insure that I convey the characteristics that I think are useful and important.

However in this stab, there will be some deliberate short-cuts and places for correction and improvement. Some of this is because I would like to have something working that we can iterate on, and already I see that as happens so often, since the code base is large and uses List everywhere it will probably be a chore to convert everything initially.

So I imagine some parts will be done and other parts will need to be done. But it will be clear how to do the remaining part. Similar to the situation with Symbol where we have many places where we are still using Symbol("X") instead of SymbolX.

Before doing this though, there will be some work done to adjust the expression-interface-refactor branch so that it can work with our currently unsorted (sloppy) use of Expression.  Right now that branch can't be in master because it rejects stuff that isn't sorted. And it will take a long time for Expressions to get sorted. 

The main change will be to get the interfaces adjusted and to start useing the ElementsProperties mutable record.

While the expression-interface-refactor branch can't be used in general, it is however very useful for getting a sense of the performance improvement we will see, because for small limited kinds of things, small benchmark test in particular, we this code does work. 

The other places that will need correction is that I intend to start off subclassing Expression which is too large. What I imagine will be needed is a BaseExpression that is stripped down that can serve as the the Base class for both (compound) Expression and ListExpression. In the future Associations and possibly Boxing kinds of things might also fit here. You probably know about that more than I.

rocky:
Committing. Although not significant in of itself, it will be good not to leave this stranded in what is to come.

PR #184, opened 2022-03-06
Format tests

mmatera:
This PR incorporates an extended set of tests for expression formatting. This was intended to check that #181 behaves equally or better than master. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/184"></a>

PR #302, opened 2022-05-28
Improving format compatibility

mmatera:
This follows #184. In this PR there are several small changes that improve the compatibility between the formatted text in Mathics and WMA. Also, I made some progress in decoupling the BaseElement interface from the formatting interface.
This would reduce the entanglement we have in the formatting process.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/302"></a>


test/test_format.py
```diff
@@ -8,523 +8,736 @@
...
-    #    assert isinstance(format_result,  BoxConstruct)
-    if msg:
-        assert (
-            format_result.boxes_to_text(evaluation=session.evaluation) == str_expected
-        ), msg
-    else:
-        str_format = format_result.boxes_to_text(evaluation=session.evaluation)
-        assert str_format == str_expected
 
+# Set this to False in case mathml tests must be considered xfail. With True, ensures the
```
rocky, line 22: expected fail or expected skip.  Fail means you deliberately want something to fail. Skip means you don't have the means or resources to do something.

test/test_format.py
```diff
@@ -8,523 +8,736 @@
...
+    "Integrate[F[x], {x, a, g[b]}]": {
+        "msg": "Non trivial SubsuperscriptBox",
+        "text": {
+            "System`StandardForm": "Subsuperscript[∫, a, g[b]]\u2062F[x]\u2062\uf74cx",
+            "System`TraditionalForm": "Subsuperscript[∫, a, g(b)]\u2062F(x)\u2062\uf74cx",
+            "System`InputForm": "Integrate[F[x], {x, a, g[b]}]",
+            "System`OutputForm": "Integrate[F[x], {x, a, g[b]}]",
+        },
+        "mathml": {
+            "System`StandardForm": '<mrow><msubsup><mo>∫</mo> <mi>a</mi> <mrow><mi>g</mi> <mo>[</mo> <mi>b</mi> <mo>]</mo></mrow></msubsup> <mo form="prefix" lspace="0" rspace="0.2em">\u2062</mo> <mrow><mi>F</mi> <mo>[</mo> <mi>x</mi> <mo>]</mo></mrow> <mo form="prefix" lspace="0" rspace="0.2em">\u2062</mo> <mrow><mtext>\uf74c</mtext> <mi>x</mi></mrow></mrow>',
```
rocky, line 324: I get saddened about stuff that has lots of gibberish in it like setting rspace to 0.2em, but I don't have a good solution for this right now.

rocky:
I really like the intent and I would like to see all of this go in. 

It is a bit over my head to understand so I guess I'll trust what you have done. The comments I left are comment in case you want to act on it.

 I think there are various spelling typos too that I saw, but right now my feeling is to deal with this later. 

One thing I'd ask is that you separate the the changes into smaller parts so we don't have 14 files of changes with +849-789. Thanks for doing this and listening.

test/test_format.py
mmatera: @rocky,  in this case, *xfail* means that the core is still failing to fulfill the expected behavior (the one in WMA). The good news is that with the last changes here, we pass all the tests. Of course, we can add more.
Still. we could leave some "fragile" tests (those involving complex mathml/svg/xml code)   marked with xfail. 

 In any case, I have to rewrite this statement in the comment.

test/test_format.py
mmatera: I could mark these tests as "fragile" and put them in the set of tests that we allow fails.  The good part to have these tests is that we notice when an "unrelated" change, has an impact on how the formatting is processed. In the current state of things, this happens a lot...

test/test_format.py
rocky: > @rocky, in this case, _xfail_ means that the core is still failing to fulfill the expected behavior (the one in WMA). The good news is that with the last changes here, we pass all the tests. Of course, we can add more. Still. we could leave some "fragile" tests (those involving complex mathml/svg/xml code) marked with xfail.
> 
> In any case, I have to rewrite this statement in the comment.

My understanding is that things that are not implemented yet are skipped with the reason "not implemented yet", not expected to fail. An "expected to fail" kind of test should never all succeed. Divide by 0 is the kind of expected failure.

mmatera:
Superseeded by #311

PR #304, opened 2022-05-28
Miscelleanous to_expression/Expression separation and some leaf->element correction

rocky:
These are random droplets amongst a sea of needed changes,. They happen to be some of the changes I made in the expression-interface-factor branch that I need to dumb down that I need before ListExpression can go on.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/304"></a>


rocky:
Merging in to simply my local branching structure which is quite complex.

PR #305, opened 2022-05-28
Revise Expression to use ElementsProperties

rocky:
ElementsProperties is a mutable record.

Note that addition work is needed to categorize existing Expression() calls into `to_expression()` or
`Expression()`.

After that is done, we should see a big improvement in speed due to a
reduction of unneeded calls to convert elements.

Also, there should be improvements as we start setting the elements
properties in the calls to be more precise.

Until then, the performance here will be slightly worse with this
commit, since we have the worst of both worlds: we have to do conversion
on each Expression() call as well as conversion that may already be done in setup to calling
Expression which will be required (not optional) in the future.

A specialization of Expression, ListExpression() will be using this as the base code to work off of.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/305"></a>

### DIFF
--- a/mathics/builtin/numbers/numbertheory.py
+++ b/mathics/builtin/numbers/numbertheory.py
@@ -9,7 +9,7 @@ import sympy
 
 from mathics.core.evaluators import apply_N
 from mathics.builtin.base import Builtin, SympyFunction
-from mathics.core.expression import Expression
+from mathics.core.expression import Expression, to_expression
 from mathics.core.symbols import Symbol
 from mathics.core.atoms import (
     Integer,
@@ -99,8 +99,8 @@ class Divisors(Builtin):
         "Divisors[n_Integer]"
         if n == Integer0:
             return None
-        return Expression(
-            SymbolList, *sympy.divisors(n.to_sympy()), element_conversion_fn=from_sympy
+        return to_expression(
+            SymbolList, *sympy.divisors(n.to_sympy()), elements_conversion_fn=from_sympy
         )
 
 
--- a/mathics/builtin/numbers/numbertheory.py
+++ b/mathics/builtin/numbers/numbertheory.py
@@ -100,7 +100,7 @@ class Divisors(Builtin):
         if n == Integer0:
             return None
         return to_expression(
-            SymbolList, *sympy.divisors(n.to_sympy()), elements_conversion_fn=from_sympy
+            SymbolList, *sympy.divisors(n.value), elements_conversion_fn=Integer
         )
 
 
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -9,7 +9,6 @@ import typing
 from typing import Any, Callable, Iterable, Optional, Tuple, Union
 from itertools import chain
 from bisect import bisect_left
-from recordclass import RecordClass
 
 from mathics.core.atoms import from_python, Number, Integer
 
@@ -27,7 +26,7 @@ from mathics.core.attributes import (
     sequence_hold as SEQUENCE_HOLD,
 )
 from mathics.core.convert import sympy_symbol_prefix, SympyExpression
-from mathics.core.element import ensure_context
+from mathics.core.element import ensure_context, ElementsProperties
 from mathics.core.evaluation import Evaluation
 from mathics.core.interrupt import ReturnInterrupt
 from mathics.core.number import dps
@@ -87,38 +86,6 @@ class BoxError(Exception):
         super().__init__("Box %s cannot be formatted as %s" % (box, form))
         self.box = box
         self.form = form
-
-
-class ElementsProperties(RecordClass):
-    """Properties of Expression elements that are useful in evaluation.
-
-    In general, if you have some set of properties that you know should
-    be set a particular way, but don't know about the others, it is safe
-    to set the unknown properties to False.
-
-    When *all* of the properties are unknown, use a `None` value in
-    the Expression.properties field instead of creating an
-    ElementsProperties object with everything set False.
-    """
-
-    # True if none of the elements needs to be evaluated.
-    elements_fully_evaluated: bool = False
-
-    # is_flat: True if none of the elements is an Expression
-    # Some Mathics functions allow flattening of elements. Therefore
-    # it can be useful to know if the elements are already flat
-    is_flat: bool = False
-
-    # is_ordered: True if all of the elements are ordered. Of course this is true,
-    # if there are less than 2 elements. Ordered is an Attribute of a
-    # Mathics function.
-    #
-    # In rewrite_eval_apply() if a function is not marked as Ordered this attribute
-    # has no effect which means it doesn't matter how it is set. So
-    # when it doubt, it is always safe to set is_ordered to False since at worst
-    # it will cause an ordering operation on elements sometimes. On the other hand, setting
-    # this True elements are not sorted can cause evaluation differences.
-    is_ordered: bool = False
 
 
 # ExpressionCache keeps track of the following attributes for one Expression instance:
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -6,9 +6,10 @@ import math
 import time
 
 import typing
-from typing import Any, Callable, Iterable, Optional, Union
+from typing import Any, Callable, Iterable, Optional, Tuple, Union
 from itertools import chain
 from bisect import bisect_left
+from recordclass import RecordClass
 
 from mathics.core.atoms import from_python, Number, Integer
 
@@ -81,15 +82,43 @@ symbols_arithmetic_operations = system_symbols(
 )
 
 
-def identity_fn(arg: Any) -> Any:
-    return arg
-
-
 class BoxError(Exception):
     def __init__(self, box, form) -> None:
         super().__init__("Box %s cannot be formatted as %s" % (box, form))
         self.box = box
         self.form = form
+
+
+class ElementsProperties(RecordClass):
+    """Properties of Expression elements that are useful in evaluation.
+
+    In general, if you have some set of properties that you know should
+    be set a particular way, but don't know about the others, it is safe
+    to set the unknown properties to False.
+
+    When *all* of the properties are unknown, use a `None` value in
+    the Expression.properties field instead of creating an
+    ElementsProperties object with everything set False.
+    """
+
+    # True if none of the elements needs to be evaluated.
+    elements_fully_evaluated: bool = False
+
+    # is_flat: True if none of the elements is an Expression
+    # Some Mathics functions allow flattening of elements. Therefore
+    # it can be useful to know if the elements are already flat
+    is_flat: bool = False
+
+    # is_ordered: True if all of the elements are ordered. Of course this is true,
+    # if there are less than 2 elements. Ordered is an Attribute of a
+    # Mathics function.
+    #
+    # In rewrite_eval_apply() if a function is not marked as Ordered this attribute
+    # has no effect which means it doesn't matter how it is set. So
+    # when it doubt, it is always safe to set is_ordered to False since at worst
+    # it will cause an ordering operation on elements sometimes. On the other hand, setting
+    # this True elements are not sorted can cause evaluation differences.
+    is_ordered: bool = False
 
 
 # ExpressionCache keeps track of the following attributes for one Expression instance:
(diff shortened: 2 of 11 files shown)


rocky:
Finally! This is the intermediate Expression interface refact that I can work off of and that presumably passes all tests.

This can stay as a branch until others look over. There will be additional PRs that merge into this one. In particular the ListExpression work will want to merge into this base. 

So in contrast to the other PRs where  I have been committing partial small pieces into master, this one folks can take more time to look at and think about since this is really the thing we want to focus on and the follow-on ListExpression.

mathics/builtin/numbers/numbertheory.py
```diff
@@ -99,8 +99,8 @@ def apply(self, n, evaluation):
         "Divisors[n_Integer]"
         if n == Integer0:
             return None
-        return Expression(
-            SymbolList, *sympy.divisors(n.to_sympy()), element_conversion_fn=from_sympy
+        return to_expression(
```
TiagoCavalcante, line 102: I don't get the `to_expression`

mathics/builtin/numbers/numbertheory.py
```diff
@@ -99,8 +99,8 @@ def apply(self, n, evaluation):
...
+            SymbolList, *sympy.divisors(n.to_sympy()), elements_conversion_fn=from_sympy
```
TiagoCavalcante, line 103: `elements_conversion_fn=Integer`

mathics/core/expression.py
```diff
@@ -620,7 +597,9 @@ def elements(self):
 
     @elements.setter
     def elements(self, values: Iterable):
-        self._elements = self._build_elements(values)
+        self._elements = tuple(values)
+        # Set to build self.elements_properties on next evaluation()
+        self.elements_properties = None
```
TiagoCavalcante, line 569: This doesn't seem like the right place.

TiagoCavalcante (approved):
LGTM

mathics/builtin/numbers/numbertheory.py
rocky: If this is just about the fact that this could be Expression with an integer conversion, I'll have to check and you may be right!

However if this is about `to_expression()` in general, begin rant....

What we had been seeing is that way too much time is being spent in the Expression() construction calling from_python() on the elements because we allowed flexible forms like:

```
Expression("Plus", 1, 2)
Expression(SymbolPlus, 1, 2)
Expression(SymbolPlus  1, Integer(2))
```

and so on. While it might not be all that important to use:

```
Expression(SymbolBesselJ, Integer(n), Integer(z))
```

instead of:

```
Expression("BesselJ", n, z)
```

inside the class BesselJ apply (or really eval) method. The fact that Expression has to do the conversion, impacts _other_ calls where it _does_ matter that we don't want to slow down an Expression constructor.

I think if this like traffic jams we sometimes get in NYC. Let's say you are a casual driver who just wants to travel from the suburbs north to a place further out in the suburbs but the route happens to cross one of the few major East-West bridges that daily commuters use to drive into the NYC. Because there is so much traffic on the bridge it now sprawls into the north entrance to the bridge and the road you are traveling on just happens to have an exit that connects to the bridge. So now you are impacted even though you aren't really contributing to the traffic jam.

(I don't know if this analogy helps, but I thought it neat to mention.)


So this is where `to_expression()` comes it. If you want to pass anything other than:

```
Expression(SymbolPlus, Integer1, Integer2)
```

such as the forms shown above, then write this as `to_expression("Plus, 1, 2)` or whatever variation you want.

In contrast to `Expression()` we also allow you to specify a conversion function, so that above could also be written `to_expression("Plus", 1, 2, elements_conversion_fn=Integer)` and that will avoid using the generic `from_python()`. For a list of two items this has very little impact. However for a function an expression like `Range[1000]` we have a bit by telling `to_expression` we are working with integers and needing `from_python()` to figure this out.

(Currently the code doesn't do this, but a small change now will allow this in case you want to do.)

Lastly, because it been annoying and frustrating let me describe a little history.

It was rightly noticed that there were too many `from_python()` calls. So the remedy was to change the order of `if` conditions or suggest that we could inline stuff (if not here then definitely other places) because "function calls are too slow", or that certain forms of function calls are too slow. Or that we should Cythonize everything which from a developer standpoint is a huge time drain waiting for Cython do its thing for some code that will change in a minute or two.

Instead, what I think was and is needed is to understand how the code works at a higher level such as I have tried to describe and understand how we can organize the code so that we don't need conversions all over the place. This and the rewrite/apply pattern matching are making us at least one order of magnitude too slow.

It may be that after the code is reorganized so that things are sorted, `from_python()` will be *slower* because our pattern of use is the opposite: where before most of the time we were asking to convert objects that were already converted, in the future when calls are sorted we will _rarely_ ask to convert something already converted because we call `Expression()` when we don't want conversion and `to_expression()` when we do.

And I am finding something similar in say `N_apply()` where there is this expression customization that has been expanded and probably needs to be reworked. And we have tons of `has_form("List", None)`  or Expression("List", ... ) that we wouldn't need to convert if we had a cleaner base to start out with which understands how to deal with Lists more efficiently

In general we expanded from am messy base and expanding from a clean base makes less work.


Ok. Sorry to rant.

mathics/builtin/numbers/numbertheory.py
rocky: We may be able to use Expression here.... checking...

rocky: You are correct we can use Integer here. And we can use `n.value` since we know n is an Integer. 

It also appears that the elements will always be sorted because that's how sympy.divisors works. 
But to_expression will figure that out. It has to make a pass over the arguments anyway so checking for sortedness comes pretty much for free.

As noted elsewhere though all of this will get replaced by something like 

```
ListExpression(*sympy_divisors(n.value), Integer)
```

and by doing this we can save both forms of the Mathics list as well as a Python/SymPy/mpmath/numpy - convenient form too.

mathics/core/expression.py
rocky: I am getting the sense that adding an element setter property was a good idea in the first place. 
Right now though I don't want to go through the code to remove all of the callers.

Because elements is a setter, more than on argument generally is not passed. We also don't want to aggressively compute element properties here because we don't know if the expression that this is in will be evaluated and there can be a lot of elements.  

So the safest thing to do is to set this to `None` which you will find mentioned in docstrings as indicating that on the next  rewrite_apply_eval  step properties need to be computed.

mmatera:
@rocky, in general, I agree with the idea of removing the conversions inside the constructor. I couldn't finish today reviewing the changes, and there are some things that I am not sure to understand. In particular, what is the impact of using the ``RecordClass`` library to store the properties? Have you run some benchmarks?

rocky:
> @rocky, ... In particular, what is the impact of using the `RecordClass` library to store the properties? Have you run some benchmarks?

https://bitbucket.org/intellimath/recordclass/src/master/ describes benchmarks the author has done.

test/core/test_expression_constructor.py
```diff
@@ -1,38 +1,40 @@
-from mathics.core.expression import Expression
+from mathics.core.expression import Expression, to_expression, ElementsProperties
```
mmatera, line 1: @rocky, I started to wonder if `build_expression` or `create_expression` wouldn't be a better name for  `to_expression`: I used `to_expression` to denote a method that takes an object and converts it into an expression. On the other hand, the function `to_expression` is taking several elements and properties, and with them, making conversions and finally assembling an Expression. 
I ask about this now, because at some point we need to do a large number of changes over all the modules.

mathics/core/rules.py
```diff
@@ -62,6 +62,8 @@ def yield_match(vars, rest):
             else:
                 result = new_expression
 
+            if isinstance(result, Expression) and result.elements_properties is None:
```
mmatera, line 65: In this method, we apply a rule to an expression to produce a new expression. I guess that in certain cases, the new expression does not need to be reevaluated, and hence, the element_properties that we compute here are not used. Is there a reason to make this here?

test/core/test_expression_constructor.py
rocky: `to_xxx` as a function takes  whatever you give and turns it into `xxx`. As top-level functions already have 

* `to_python`, 
* `to_sympy` 
* `to_real_or_none`
* `to_integer_or_none`
* `to_complex`
* `to_boxes`
* `to_sympy_matrix`
* `to_mpmath_matrix`

among others. 

The `or_none` in these I kind of think of as superfluous and the `or_none` should be removed, since `None` is the value that gets returned when a value doesn't fit and we don't want to throw an error. So this fits into the pattern way more nicely that `build` or `create` do.

As an instance method, as you noted there is there is a `to_expression()` instance method. Notice there are also `to_python()`, and `to_sympy()` _instance_ methods and again this fits into this pattern of having both instance methods and top_level methods that return the thing after the `to_` as an object. 

More generally in OO programs, a `to_xxx` pattern is what is generally what is used as a helper function to create an instance of some sort of special object.

So in sum, yes, there is a pattern and this is just continuing the pattern. I am sorry that this pattern was not familiar or noticed either in this code or in general. And that was a part of into why I wanted to start  a `ListExpression()` object even though right now it is woefully inadequate in terms of being an upward-compatible specialization of `Expression()` and needs a lot of work still such as to create a base class that both of these inherit from.

Know your strengths and also know your limitations. And although among others I would not consider myself having a flair for API design, among the people we have doing it here, it makes me seem more talented. I suppose a part of it is just that I've seen other interpreters and lots of other code Python and others.

mathics/core/rules.py
rocky: I first was noticed this empirically - if you don't have it the code will currently crash. And there are discussions in this PR which show why.  We have the setter and there is code that uses that. One of the places where the setter method is used is right here in this method. That could be replaced, but I am leaving that for some other time or some other person. 

The setter-property method sets `self.element_properties` to `None`.  See the discussion above in this PR.

There is a good deal of this code here, that can be improved but it will require other parts of the code to be more sensitive to elements properties. 

That is why we are using a _mutable_ records rather than a named tuple which isn't mutable. 

If it wasn't clear, I repeat that what I have seen in benchmarking is that there are some big bit wins here, especially if you look at an overall sequence of operations such as you'd get by running the entirety of the combinatorica tests. 

But there are also some regressions and a number of those regressions  when you look at specific individual operations. And these can eliminated and some cases turned into even bigger wins after the rest of the code is adjusted to understand that there are elements properties and these other new features using alternative representations (in the case of List) and objects with specialized evaluation functions. 

Right now though we need to introduce some additional and non-optimal checks or some pessimistic behavior to bridge the gaps because we also want _working_ code. 

I  do however have a branch which removes the pessimism and doesn't bridge gaps. I use this for benchmarking performance. Needless to say it is broken in general. When there is something I want to benchmark I incrementally fix the brokenness around that so that I can benchmark some specific feature. And those changes to fix the brokenness then go into master. That is why earlier in the weekend there were what seemed like (and to some extent _were_) random adjustments.

But until  all the gaps are fixed, we have to live with live with some regressions in specific cases, especially when there is a plan in mind for how those regressions will be reversed. 


>  hence, the element_properties that we compute 

Note that I write what in English is awkward "element_properties" rather than the more natural "element_properties" because it is a property of the _all_ elements of the Expression object. And yes, technically an Expression is an Element too, but I want to emphasize the fact these are properties of a _collection_. If folks want to casually refer to this as Element properties where Element refers to the Expression object rather than the elements attribute of the Expression that's okay. For myself, I want to keep emphasizing that this is about elements or the elements field of the Expression element, not a single element in elements.

test/core/test_expression_constructor.py
mmatera: > `to_xxx` as a function takes whatever you give and turns it into `xxx`. As top-level functions already have
> 
>     * `to_python`,
> 
>     * `to_sympy`
> 
>     * `to_real_or_none`
> 
>     * `to_integer_or_none`
> 
>     * `to_complex`
> 
>     * `to_boxes`
> 
>     * `to_sympy_matrix`
> 
>     * `to_mpmath_matrix`
> 
> 
> among others.

OK, but in all the cases in the list, the method takes an object of a certain class and returns an expression. The new function, on the other hand, takes several elements and builds the expression. 

> Know your strengths and also know your limitations. And although among others I would not consider myself having a flair for API design, among the people we have doing it here, it makes me seem more talented. I suppose a part of it is just that I've seen other interpreters and lots of other code Python and others.


I guess that I didn't express myself well. Sorry for that. The aim was not to question your experience and knowledge. I asked you about this because I made myself one idea about what this pattern ("to_*" ) means, and the change you proposed seemed to me was in contradiction with that idea. So, I thought that I was missing something.  As I told you many times, I am OK with any name you choose, but I would like to understand the choice to avoid screwing up in future rounds.

rocky: > 
> OK, but in all the cases in the list, the method takes an object of a certain class and returns an expression. The new function, on the other hand, takes several elements and builds the expression.
> 

I think this is a red herring. First, most of the conversion functions like `to_python` or `to_real`  input and output something that isn't considered a compound object. For the few that do, the places that want to use these happen to have the collections already stored in a single variable, so there isn't a need or convenience to break this out. You won't find many class constructors in Mathics that benefit by taking an arbitrary number of arguments and then need to bundle those as a collection. So the to_xxx helper method is just mirroring that aspect of the class.

The other place where it comes up so far and for the same reason is `to_mathics_list()` which is a construction helper for `ListExpression()`

An analogy would be like saying we can't create a ternary "op" functions because all of the other ops take one or two arguments.  

So in sum, the `*elements` is an aspect or convenience dictated by how you want to call the function. The name of a function generally isn't directed by the form you want to pass arguments to the function.

mmatera (approved):
Thanks for all the clarifications. LGTM

rocky:
@mmatera given #313  which combines this and #311 (which ideally would have more changes).  I would like to see this merged it and then #313 rebased off of this so that it just has the trailing changes and then that merged in.

In general, more smaller PRs are better than ones that combines two. 

Your thoughts?

PR #308, opened 2022-05-29
List expression lite

rocky:
First cut at ListExpression
    
Many things don't work though. See comment-out code for avenues to explore.

@mmatera @TiagoCavalcante Feel free to hack away at this. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/308"></a>

PR #311, opened 2022-05-29
MathML and TeXForm output changes

mmatera:
This completes #302 with some changes needed for making the proposal in https://github.com/Mathics3/mathics-core/pull/308/files/7a7457cc549480435daea118cc3536ee57b5efde#r884201751
works.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/311"></a>


mathics/core/symbols.py
```diff
@@ -424,12 +424,9 @@ def __str__(self) -> str:
         return self.name
 
     def atom_to_boxes(self, f, evaluation) -> "String":
```
rocky, line 426: The return type "String" -> "_BoxedString"

mathics/builtin/box/inout.py
```diff
@@ -465,40 +469,37 @@ def apply_list(self, boxes, evaluation):
...
     def to_expression(self):
         """
         returns an evaluable expression.
         """
-        items = (item.to_expression() for item in self.items)
+        items = tuple(
+            item.to_expression() if isinstance(item, BoxExpression) else item
+            for item in self.items
+        )
         result = Expression("RowBox", Expression("List", *items))
```
rocky, line 503: RowBox doesn't work here?

mathics/core/symbols.py
mmatera: Done

mathics/builtin/box/inout.py
mmatera: This is the `to_expression` method, so what we want here is an `Expression` object. This is needed because during the evaluation, there are   expressions like
```
expr = Expression(Symbol("F"), RowBox(String("bla")))
```
which could appear when the user defines a custom `MakeBoxes` rule. 

Then, maybe `F` has specific rules when acting over a `RowBox`, so, when you call 
```
expr.evaluate(evaluation)
```
you need the interface of an expression. The methods are then implemented by calling `to_expression` and then delegating the method to the `Expression` you get there object.

mathics/builtin/box/inout.py
```diff
@@ -134,10 +133,11 @@ def render(format, string, in_text=False):
             return format % encode_tex(string, in_text)
 
         if text.startswith('"') and text.endswith('"'):
-            if show_string_characters:
-                return render(r'\text{"%s"}', text[1:-1], in_text=True)
-            else:
-                return render(r"\text{%s}", text[1:-1], in_text=True)
+            # if show_string_characters:
```
TiagoCavalcante, line 140: Why remove this option? (to say the truth I like it)

TiagoCavalcante (changes requested):

mathics/builtin/box/inout.py
mmatera: @TiagoCavalcante, this is because in WMA, in TeXForm you do not have double quotes even if you ask `FullForm` or `InputForm`:

```
In[1]:= "Hola"//InputForm                                                       

Out[1]//InputForm= "Hola"

In[2]:= "Hola"//InputForm//TeXForm                                              

Out[2]//TeXForm= \text{Hola}

In[3]:= "Hola"//FullForm//TeXForm                                               

Out[3]//TeXForm= \text{Hola}
```

To reproduce this behavior in Mathics, we need to discard the value of `show_string_characters`.

mmatera: @TiagoCavalcante,
I think I managed to keep the behavior and the option. Please, have a look when you have time.

TiagoCavalcante (approved):
LGTM

mathics/builtin/box/inout.py
rocky: This makes perfect sense. Thanks for the explanation. 

Given this that we need an Expression, then please adjust the `Expression()` calls to the appropriate form. For example use the pre-defined SymbolRowBox that appears in `mathcis.core.systemsymbols`, or turn these into top-level function calls `to_epxression()` calls. For the list expression, if we can a ListExpression here and that works, that would be nice, otherwise as a fallback either use `to_expression()` or `SymbolList()` as before. Thanks.

rocky: Ok I see this change in https://github.com/Mathics3/mathics-core/pull/313  

So this PR is obsolete if we just merge the other one in, right?

mmatera: @rocky,  Now, this has some small changes regarding the other. Once I finish with the changes that you asked for, I can merge this to master, or wait that you to merge #308 and then merge this.

rocky: I've just merged #308. So if you could rebase off of master and then finish that would be great. Thanks!

mmatera: OK.

CHANGES.rst
```diff
@@ -12,6 +12,7 @@ Enhancements
 * The order of the context name resolution (and ``$ContextPath``) switched putting ``"System`"`` before ``"Global`"`.
 * In assignment to messages associated with symbols, the attribute ``Protected`` is not having into account, folliwing the standard in WMA. With this and the above change, Combinatorical 2.0 works as written.
 * ``Share[]`` performs an explicit call to the Python garbage collection and returns the amount of memory free.
+* Improving the compatibility of ``TeXForm`` and ``MathMLForm`` outputs with WMA.
```
rocky, line 15: If you are up to it, maybe a little bit about how the compatibility was changed. I looked hard to find _where_ the `mtext` tag was change to the `mn` tag.  I guess what is going on is that we are doing better about detecting the type and using the OO behavior. So that `mn` tag` code was already there, but it is now just getting used properly. 

I see that quotes in TeXForm get changed as well.  Personally, I think it would be helpful as a user to know that in TeXForm some quotes may disappear and that in MathForm some mtext tags may become mn tags.  But I realize this is a matter of taste.

rocky:
LGTM. Left one remark about CHANGES.rst. This is a huge PR so it is hard to understand. As far as I can tell, one part of it is about improving compatibility of formatted output. I wonder if that had been split into its own PR whether those smaller PRs would be easier to deal with.  That said, there is good stuff here whether one or two PRs.

CHANGES.rst
mmatera: What is puzzling to me is to see how and where to document these changes. There is a rationale for why a number in StandardForm / OutputForm must be tagged with ``<mn>`` in MathMLForm, while in InputForm are tagged as 
``<mtext>``: InputForm is supposed to represent how an expression must be introduced, while StandardForm/OutputForm
tries to represent what it actually is: a number.

In the case of quotes in `TeXForm[InputForm[..]]`,  the problem is that I am not sure that the WMA output is right. However, if we want to be able to load WMA packages in Mathics, we need to keep the maximum compatibility. In any case, now it is more clear than before what we need to change in order to have the other behavior.

mmatera:
> LGTM. Left one remark about CHANGES.rst. This is a huge PR so it is hard to understand. As far as I can tell, one part of it is about improving compatibility of formatted output. I wonder if that had been split into its own PR whether those smaller PRs would be easier to deal with. That said, there is good stuff here whether one or many PRs.


Actually, it was a "small" PR. The problem is that since all the code was so strongly entangled, each time I try to split this, the more changes I need to introduce in order to make the parts work. So, after adding a comment in CHANGES.rst, I will merge this as it is, and I will try to write an explanation about how the formatting is actually working, and how I think we can improve this.

rocky (approved):
LGTM - CHANGES.rst are much more informative. Thanks.

PR #313, opened 2022-05-30
List expression lite plus format trailing changes

mmatera:
In this PR #311 is merged over #308


mathics/builtin/box/inout.py
```diff
@@ -465,41 +485,41 @@ def apply_list(self, boxes, evaluation):
...
-            else:
-                elements.append(item)
-        elements = Expression(SymbolList, *elements)
-        self._elements = tuple((elements,))
+        self.items = tuple((check_item(item) for item in items))
+        self._elements = None
 
     def to_expression(self):
         """
         returns an evaluable expression.
```
rocky, line 516: I meant to mention this from the #311. Please let us make the same change were we use the simpler more natural `->Expression` rather than gobbledigook. 

However, where a comment would be most appreciated is in line 522 with something along the lines of why we need to use `Expression()` and not `RowBox()` so that someone (probably me) in the future doesn't get confused again. 

Also, if this is the merge of ListExpression "lite", then ListExpression(*items) ot `to_mathics_list()` neither work?

mathics/builtin/box/inout.py
```diff
@@ -172,8 +180,20 @@ def render(format, string, in_text=False):
             else:
                 return render("%s", text)
 
+    def get_head(self) -> Symbol:
+        return SymbolString
+
+    def get_head_name(self) -> "str":
```
rocky, line 186: Removing the quotes here would be nice since they aren't needed.

rocky:
A number of places in test code that number now appear with `<mn>` tags instead of `<mtext>`. 

However I don't see exactly where that is in the code. And it would be nice to mention this in CHANGES as well.

mmatera:
@rocky, Since #308 was merged, this branch now does not have sense. In incorporated changes according to the comments in #311.

Regarding the change ``<mn>``<->``<mtext>``, it was just to reproduce the behavior of WMA as faithful as possible. I added a line in CHANGES mentioning the change in the behavior, and added a comment in `test/format/test_format.py` regarding this.

PR #314, opened 2022-05-30
Add TimingContextManager()

rocky:
this can be used as a context manager to time specific parts of the code.

### DIFF
--- a/mathics/core/util.py
+++ b/mathics/core/util.py
@@ -42,6 +42,29 @@ def timeit(method):
     return timed
 
 
+class TimeitContextManager:
+    """Add this as a context manager to time parts of the code.
+
+    For example:
+        with TimeitContextManger("testing my loop"):
+           for x in collection:
+               ...
+    """
+
+    def __init__(self, name: str):
+        self.name = name
+
+    def __enter__(self):
+        # print(f"{date.today()}	{method_name} starts")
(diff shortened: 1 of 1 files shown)


mathics/core/util.py
```diff
@@ -42,6 +42,29 @@ def timed(*args, **kw):
...
+    def __init__(self, name: str):
+        self.name = name
+
+    def __enter__(self):
+        # print(f"{date.today()}	{method_name} starts")
+        self.ts = time.time()
+
+    def __exit__(self, exc_type, exc_value, exc_tb):
+        te = time.time()
+        elapsed = int((te - self.ts) * 1000)
```
TiagoCavalcante, line 63: Maybe `float` instead of `int`?

mathics/core/util.py
```diff
@@ -42,6 +42,29 @@ def timed(*args, **kw):
...
+        if elapsed > MIN_ELAPSE_REPORT:
+            print("%r  %2.2f ms" % (self.name, elapsed))
```
TiagoCavalcante, line 65: Here it is treated as a float.

mathics/core/util.py
rocky: Ok - that makes sense and is simpler. Done in d874cf08

PR #315, opened 2022-05-31
Expression("XXX"..) -> Expression(SymbolXX..) ...

rocky:
more places. This also means that in some cases there are new
Symbol's defined.

Use to_expression() where we can't use SymboXXX.

I gouldn't get SymbolSetAttributes SymbolClearAttributes working,
so that is left for later.


mathics/algorithm/integrators.py
```diff
@@ -23,24 +26,24 @@
...
     if interval.has_form("System`List", 3, None):
         intervals = []
-        intvar = interval.leaves[0]
+        intvar = interval.elements[0]
         if not isinstance(intvar, Symbol):
             evaluation.message("ilim", interval)
             return None
-        boundaries = [a for a in interval.leaves[1:]]
+        boundaries = [a for a in interval.elements[1:]]
         if any([b.get_head_name() == "System`Complex" for b in boundaries]):
```
TiagoCavalcante, line 45: get_head() here?

TiagoCavalcante (approved):
LGTM

mathics/algorithm/integrators.py
rocky: Actually it would be the equivalent of `Rest[]`. We don't have any special function for that, (see for example the eval method for Rest), although it would make sense if we did. And why this needs to be wrapped in a comprehension so I will try removing that.

PR #318, opened 2022-05-31
Next batch of Expression("XXX") removal

rocky:


rocky:
For the curious, these are basically what is needed to get through the combinatorica benchmark test without having to convert Expression elements. 

The main thing remaining is not to have to convert the head element from a `str` in Expression()

PR #319, opened 2022-06-01
The next batch of symbol names

rocky:
These were driven by strings in the "head" parameter off of Expression running the combinatorica benchmarks. 

Alas, this is not all of them yet. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/319"></a>


mathics/builtin/compilation.py
```diff
@@ -166,7 +171,7 @@ def _pythonized_mathics_expr(*x):
         if cfunc is None:
             evaluation.message("Compile", "comperr", expr)
             args = Expression("List", *names)
```
TiagoCavalcante, line 173: SymbolList?

TiagoCavalcante (approved):
LGTM

mmatera (approved):
LGTM

mathics/builtin/compilation.py
rocky: I will try ListExpression when I get a chance. 

In general, although the PR strays outside my immediate interest - get to a state where we are in terms of the combinatorica benchmark - I do see stuff like this a bit.  There have been specific situations, not here in particular though, where I leave something for later when it looks like things deserve more care than a simple mechanical change.

PR #321, opened 2022-06-01
removing flatten_x_sequence from Atoms and subclasses

mmatera:
This PR removes two superfluous methods from the `Atom` interface and a superfluous call to them.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/321"></a>

PR #322, opened 2022-06-01
Remove flatten from atoms

mmatera:
This PR is a little bit more controversial than #321. Here, we also remove `flatten_with_respect_to_head` from the Atom interface.
From the point of view of computational cost, this adds a little overhead (a function call to `isinstance` against a call to a *do nothing* method). On the other hand, we gain clarity on the interface.

Thoughts?

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/322"></a>


rocky:
> This PR is a little bit more controversial than #321. Here, we also remove `flatten_with_respect_to_head` from the Atom interface. From the point of view of computational cost, this adds a little overhead (a function call to `isinstance` against a call to a _do nothing_ method). On the other hand, we gain clarity on the interface.
> 
> Thoughts?
> 

There is too much worry about the speed of `isinstance` versus a method call that does nothing. Let Python interpreters worry about that. 

Here though I would prefer `hasattr()` over `isinstance()` for two reasons. First is that what is expression-like is subject to change. In a non-OO language interface-oriented language like Rust, Go, and Javascript, you care about whether the method exists because there is no hierarchy. And the confusion that can arise from OO inheritance is one reason OO is on the decline.

mmatera:
> 
> There is too much worry about the speed of `isinstance` versus a method call that does nothing. Let Python interpreters worry about that.

My problem is that when we compare times with WMA, the time consumed by the function call in Mathics is comparable with the time of the whole evaluation in WMA. But OK, let's the interpreter to handle this.


> 
> Here though I would prefer `hasattr()` over `isinstance()` for two reasons. First is that what is expression-like is subject to change. In a non-OO language interface-oriented language like Rust, Go, and Javascript, you care about whether the method exists because there is no hierarchy. And the confusion that can arise from OO inheritance is one reason OO is on the decline.

OK. Done.

rocky:
> > There is too much worry about the speed of `isinstance` versus a method call that does nothing. Let Python interpreters worry about that.
> 
> My problem is that when we compare times with WMA, the time consumed by the function call in Mathics is comparable with the time of the whole evaluation in WMA. But OK, let's the interpreter to handle this.
> 

And my problem with this approach is that we have gone down that path and the code has gotten more complicated, while we haven't gotten the speed gain we would get if we solved the problem at a higher level which is where the problem really lies.

When you write "consumed by function call" you mean the time consumed by _all_ of the function calls. And more or most often specifically function calls that convert data.

A solution to this we've seen is to reduce the number of function calls by being smarter about creating expressions into groups that don't need conversions. And reducing conversions in evaluation. Down the line we can reduce this further. This already seems have a far greater impact on the speed, and leave us with code that is more maintainable, scalar, and modular.


> > Here though I would prefer `hasattr()` over `isinstance()` for two reasons. First is that what is expression-like is subject to change. In a non-OO language interface-oriented language like Rust, Go, and Javascript, you care about whether the method exists because there is no hierarchy. And the confusion that can arise from OO inheritance is one reason OO is on the decline.
> 
> OK. Done.

Thanks! 

If we can get this into a shape where it passes CI and doesn't have conflict, we should merge it in so there aren't further conflicts.

PR #324, opened 2022-06-01
More symbol names 5

rocky:
A bit of upheaval since we moved the basic operators Plus, TImes, Divide, Power, Abs from systemsymbols to symbols where it is used in NumericOperators.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/324"></a>


mathics/core/symbols.py
```diff
@@ -650,41 +663,97 @@ def is_uncertain_final_definitions(self, definitions) -> bool:
...
-SymbolTraditionalForm = Symbol("TraditionalForm")
-SymbolTeXForm = Symbol("TeXForm")
-SymbolMathMLForm = Symbol("MathMLForm")
+    you can instead have:
+        abs(Integer(-8))
+        Integer(1) + Integer(2)
+    """
+
+    def __abs__(self) -> BaseElement:
+        return self.create_expression(SymbolAbs, self)
```
TiagoCavalcante, line 716: I didn't find (at least not in GitHub) where `create_expression` is defined.

TiagoCavalcante (approved):
LGTM

mathics/core/symbols.py
mmatera: I had to introduce this weird way to add a method to BaseElement because it was the only way to break the cyclic import. I hope that before the next release we are able to get rid of it.

mmatera: Yes, sure. What I tried to mean is that a more or less short term goal would be to finish disentangling the modules that used to be a single very large file. I guess we are not too far to reach that point, but it is fine if we have a 4.1 release when the evaluation refactory be done

PR #325, opened 2022-06-02
First cut at list expression evaluation specialization

rocky:
We see a noticeable win by specializing `rewrite_apply_eval_step()` .

Here are timings for:

```
Timing[Do[{a,a,a,a,a,a,a,a,a,a,a,a,a,a,a},{10}]][[1]]
```

```
4.0.0
Out[1]= 0.00109942

4.1.0 With list specialization (this branch with additional customization)
Out[3]= 0.00070406

4.1.0 without list specialization (current master)
Out[1]= 0.00212048
```

However, the above was done by forcing `to_mathics_list` on the list `{a,a...}` which is outside of this PR. 

When further work is done to more pervasively use `ListExpression()` when possible, we should see a benefit like this automatically.  

Like the ongoing work in the "more-Symbol-names-x" branches, these will be separate PRs.

Also, moving the refactored `evaluate_elements()` of ListExpression back into Expression or ultimately BaseExpression will be done in a future PR. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/325"></a>


mathics/core/list.py
```diff
@@ -58,6 +83,31 @@ def is_literal(self) -> bool:
...
+        In general, this step is time consuming, complicated, and involved.
+        However for a ListExpression, things are much much simpler and faster because
+        we don't need the rewrite/apply phases of evaluation.
+        """
+
+        if self.elements_properties is None:
+            self._build_elements_properties()
+
+        if not self.elements_properties.elements_fully_evaluated:
+            self.evaluate_elements(evaluation)
```
mmatera, line 108: @rocky, one possible problem with this line is that represents a change of behaviour regarding the same method in the superclass: in ``Expression`` this function does not modifies ``self``, but creates a new element and returns it.

A second problem is that this neglects upvalues. For example, consider the following code:
```
In[1]:= a={F[1],2};
In[2]:= {F[x_],t___}^:=Q[t];
In[3]:=a
```
the output must be `Q[2]`. However, with this implementation, the output would be `{F[x],2}`

Finally, the other problem is how ``Unevaluate`` elements are handled. This feature was broken for a while, but we should try to fix it eventually.

rocky: > @rocky, one possible problem with this line is that represents a change of behaviour regarding the same method in the superclass: in `Expression` this function does not modifies `self`, but creates a new element and returns it.
> 

Ok. Good observation and one that can be easily fixed. What is a specific scenario where this important? 

> A second problem is that this neglects upvalues. For example, consider the following code:
> 
> ```
> In[1]:= a={F[1],2};
> In[2]:= {F[x_],t___}^:=Q[t];
> In[3]:=a
> ```
> 
> the output must be `Q[2]`. However, with this implementation, the output would be `{F[x],2}`
> 

Ok. I will add a test case for this and mark it "skipped" (not expected failure), and when we address this we can remove the skipped pytest tag.

> Finally, the other problem is how `Unevaluate` elements are handled. This feature was broken for a while, but we should try to fix it eventually. Here is completely lost.

Where specifically is this handled in `Expression.rewrite_apply_step` ?

mmatera:
@rocky, look at the comments in #331

mathics/builtin/lists.py
```diff
@@ -562,10 +562,14 @@ class List(Builtin):
     summary_text = "specify a list explicitly"
     attributes = locked | protected
 
-    def apply(self, items, evaluation):
-        """List[items___]"""
-        items = items.get_sequence()
```
rocky: @mmatera I would appreciate it if we would stop taking a variable of one type (some sort of Mathics expression and converting it into a variable of a different type (a list or tuple). 

In Basic or maybe some sort of interactive programming this kind of thing is done a bit. But in "production" code is is frowned upon, and in fact in typed programming languages it is disallowed.

rocky:
Note: with the change from #331 the performance of 

```
Timing[Do[{a,a,a,a,a,a,a,a,a,a,a,a,a,a,a},{10}]][[1]]
```

is more than doubled in this branch: 0.00770123. 

However again this is corrected by catching conversion to a `ListExpression` earlier in the code which again is outside of this PR, but you can get a hint of how to hack the code so that covers this particular case (but breaks things in general).

With that we slow down a little to 0.00117201 which is pretty close to the 4.0.0 behavior

mathics/core/list.py
mmatera: > 
> > Finally, the other problem is how `Unevaluate` elements are handled. This feature was broken for a while, but we should try to fix it eventually. Here is completely lost.
> 
> Where specifically is this handled in `Expression.rewrite_apply_step` ?

In `Expression.rewrite_apply_eval_step`, here: https://github.com/Mathics3/mathics-core/blob/5b2fb73854d76d6536182faa7e83d28d4b9c6a51/mathics/core/expression.py#L1223

In any case, this is broken by now. When I have the chance I am going to try to fix it.

rocky: What are test-case examples that show the brokenness both for a general Expression and a ListExpression. 

But at any rate, I am happy to hold off looking at this if you're going to look at. Basically we want to get the Expression part closed out so we can release (hopefully by July which is when the next round of NumFocus adjunct applications starts). And then after the release you'll have free ground in working on Boxing and Formatting.  

Going over the rewrite and apply phases, I guess will come after that.

rocky:
> When I have the chance I am going to try to fix it.

@mmatera @TiagoCavalcante Thinking about this more, unless we can find problems that appear here that are not in master, let's merge this in now. And @mmatera can iterate on it in another PR.

TiagoCavalcante (approved):
LGTM

mmatera (approved):
LGTM

PR #331, opened 2022-06-03
List expression evaluation 2

mmatera:
The changes that I proposed to #325.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/331"></a>

### DIFF
--- a/test/builtin/test_assignment.py
+++ b/test/builtin/test_assignment.py
@@ -238,3 +238,7 @@ def test_predecrement():
     check_evaluation(
         "--5", "4", failure_message="Set::setraw: Cannot assign to raw object 5."
     )
+
+
+def test_assign_list():
+    check_evaluation("G[x_Real]=x^2; a={G[x]}; {x=1.; a, x=.; a}", "{{1.}, {G[x]}}")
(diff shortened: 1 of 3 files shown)


test/builtin/test_assignment.py
```diff
@@ -238,3 +238,7 @@ def test_predecrement():
     check_evaluation(
         "--5", "4", failure_message="Set::setraw: Cannot assign to raw object 5."
     )
+
+
+def test_assign_list():
+    check_evaluation("G[x_Real]=x^2; a={G[x]}; {x=1.; a, x=.; a}", "{{1.}, {G[x]}}")
```
mmatera, line 244: The idea of this test was to check that the stored list in ``a`` is not altered during the evaluation. Curiously, we pass this test with the current implementation of ``ListExpression.rewrite_apply_eval_step``. This happens because during the regular evaluation, the ``ListExpression`` stored in ``a`` is converted back into an ``Expression`` in the ``apply_rule`` call.

rocky: Thanks!

PR #338, opened 2022-06-05
Symbol, Expression changes

rocky:
These changes get us up to date on being about to run mathics/test/core
tests without conversion in Expression()

Some ListExpression opportunties were omitted for now. Some files
still have _other_ Expressions that need to be converted.

More later...

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/338"></a>


TiagoCavalcante:
LGTM

rocky:
I have about 370 more Expression("xxx", ...) occurrences to go over as well as about the same number of Expression(SymbolList occurancess. 

So merging and iterating.

rocky:
And even after that, a second pass probably could be done to improve individual calls everywhere by setting properties in calls or moving calls from the conversion class to the non-conversion class. Or changing how we do list detection by removing "has_form" and use instead "isinstance(x, ListExpression)", etc. 

Down the line we can and should add Symbol creation automatically when defining a Builtin class so we can as for say the head of a built-in expression. 

But right now I just want to get through this, see some measurable improvements and make a release where we can claim partial victory over something we had intended to get done.

PR #340, opened 2022-06-06
Do max and integer hacks

rocky:
About the removal of x.get_int_value() when x is an Integer. 

The name `get_int_value()` in general has always struck me as a holdover from early C++  or Java days. 

In Python and especially interpreters, it is common for a wrapped native object like a String, or an Int, to have a ".value" field. 

Down the line we could make this a property instead of a field, but I don't know why that not doing that seems less annoying than `.get_int_value()`.

Your thoughts? 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/340"></a>


mmatera:
@rocky, actually, Integer already has a property ``value`` that holds its value (which is always a Python int).  ``get_int_value`` could make sense if we store some kind of "arbitrary size integer" but not with the current implementation. If you remove this method, I am not going to miss it.

mathics/builtin/atomic/strings.py
```diff
@@ -739,8 +739,8 @@ def apply(self, s, n, expression, evaluation):
 
     def apply_truncated(self, s, n, m, expression, evaluation):
         "StringRepeat[s_String, n_Integer, m_Integer]"
-        py_n = n.get_int_value() if isinstance(n, Integer) else 0
-        py_m = m.get_int_value() if isinstance(m, Integer) else 0
+        py_n = n.value if isinstance(n, Integer) else 0
```
mmatera, line 742: m and n must be integers, according to the rule.

mmatera (approved):
LGTM

mathics/builtin/atomic/strings.py
rocky: Good point.

rocky:
Something I thought interesting when benchmarking `Do[]]`.   I basically "slowed" the code down by adding a function and wrapping that in a `try` block. This is only noticeable when the "do" is done once or a small number of times. 

However this was done so as not to duplicate code in a fast/common path were Python's _range_ is used with the branch where we do things the slow and general way.  When there are a lot of iterations the win in using `range` far exceeds the slowdown in subroutine call and the initial `try` setup.  

The cases we _care_ about though are where `Do` has a lot of iterations, not where there are few because there, the differences is negligible.

PR #344, opened 2022-06-06
adding test for Unevaluated over symbols

mmatera:
@rocky, this PRs adds the test case that shows the issue with ``Unevaluated``. I marked it as xfail. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/344"></a>

### DIFF
--- a/test/test_evaluation.py
+++ b/test/test_evaluation.py
@@ -98,7 +98,9 @@ def test_evaluation(str_expr: str, str_expected: str, message=""):
         ),
     ],
 )
-@pytest.mark.skip
+@pytest.mark.skip(
+    reason="the right behaviour was broken since we start to use Symbol as singleton, to speedup comparisons."
+)
 def test_unevaluate(str_setup, str_expr, str_expected, message):
     if str_setup:
         evaluate(str_setup)
--- a/test/test_evaluation.py
+++ b/test/test_evaluation.py
@@ -94,7 +94,7 @@ def test_evaluation(str_expr: str, str_expected: str, message=""):
             "a =.; F[a, x_Real, a] := List[x]^2;a=4.;",
             "F[Unevaluated[a], a, Unevaluated[a]]",
             "{16.}",
-            "Here, the second a is kept unevaluated because the bug.",
+            "Here, the second ``a`` is kept unevaluated because of a bug.",
         ),
     ],
 )
--- a/test/test_evaluation.py
+++ b/test/test_evaluation.py
@@ -62,6 +62,53 @@ def test_evaluation(str_expr: str, str_expected: str, message=""):
 
 
 @pytest.mark.parametrize(
+    "str_setup,str_expr,str_expected,message",
+    [
+        (
+            "F[x___Real]:=List[x]^2; a=.4;",
+            "F[Unevaluated[a], a, Unevaluated[a]]",
+            "F[Unevaluated[a], 0.4, Unevaluated[a]]",
+            None,
+        ),
+        (
+            "F[x___Real]:=List[x]^2; a=.4;",
+            "F[Unevaluated[b], b, Unevaluated[b]]",
+            "F[Unevaluated[b], b, Unevaluated[b]]",
+            "the second argument shouldn't be ``Unevaluated[b]``",
+        ),
+        (
+            "G[x___Symbol]:=List[x]^2; a=.4;",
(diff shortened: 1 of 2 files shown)


rocky:
> @rocky, this PRs adds the test case that shows the issue with `Unevaluated`. I marked it as xfail.
> 

I though we went over the difference between an expected failure which is not appropriate for marking bugs, versus skipped tests which _are_ what are used to remind you that you have a bug. 

I think we have this in other tests too. It is wrong and should be changed. Thanks.

> [](https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/344)

test/test_evaluation.py
```diff
@@ -61,6 +61,53 @@ def test_evaluation(str_expr: str, str_expected: str, message=""):
...
+            "G[x___Symbol]:=List[x]^2; a=.4;",
+            "G[Unevaluated[b], b, Unevaluated[b]]",
+            "F[Unevaluated[b], b, Unevaluated[b]]",
+            "the second argument shouldn't be ``Unevaluated[b]``",
+        ),
+        (
+            "a =.; F[a, x_Real, a] := List[x]^2;a=4.;",
+            "F[Unevaluated[a], a, Unevaluated[a]]",
+            "{16.}",
+            "Here, se second a is kept unevaluated because the bug.",
```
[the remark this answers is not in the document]
mmatera: The test checks the "correct behavior" (the one in WMA). It is marked as "skipped" because right now, Mathics does not pass the test.

rocky:
> > I think we have this in other tests too. It is wrong and should be changed. Thanks.
> 
> OK, I am on it.

Many thanks for your tolerance and understanding.

test/test_evaluation.py
```diff
@@ -98,7 +98,9 @@ def test_evaluation(str_expr: str, str_expected: str, message=""):
         ),
     ],
 )
-@pytest.mark.skip
+@pytest.mark.skip(
+    reason="the right behaviour was broken since we start to use Symbol as singleton, to speedup comparisons."
+)
```
rocky, line 103: Thanks!

rocky (approved):
LGTM

test/test_evaluation.py
rocky: > the right behaviour was broken since we start to use Symbol as singleton, to speedup comparison

@mmatera I've been thinking about this. I suspect like a number of other things in this codebase, there was a fundamental flaw in thinking about Symbols and how bindings on Symbols work. The same sort of mess we see with M-Expressions and Elements (called oddly leaves), and how boxing and formatting should operate, along with the weirdly named InstanceableBuiltin. 

In other words, I don't think this is a simple thing that anyone will be able to fix by changing just the code in rewrite_apply_eval_step.

mmatera: I have the following idea: what about if, when we find an Element of the form ``Unevaluated[s]`` with ``s`` an Atom, we replace it with ``Sequence[s]``. Now, we have an Expression, and then, we can tag it without any problem.  Just before applying rules, all the sequences must be flattened anyway, so this adds an overhead just if there are elements tagged as ``Unevaluated``. If the rules fail to be applied, we go back to the expression with ``Sequence[s]`` on it, and then, depending on the value of the property ``unevaluated`` we can replace it back with ``Unevaluated[s]``.

rocky: > I have the following idea:

Please, no. I have the following idea: learn more about how conventional interpreters work and make this code fit more into that framework. And then make the proposal if this isn't more simply handled in that framework. 

We may eventually do something like this. But I'd like do this starting from a framework that doesn't feel so homegrown and ignorant of common practice. 

Remember the days when string search and replace on MathML text for special strings like "->"  was proposed, because of a lack of understanding expression-driving formatting worked? I hope we can agree that the shotgun approach of search and replace on output strings isn't as good changing ToString for the Rule operator to output the unicode symbol when it is appropriate. 

Right now, this feel in that kind of category.

rocky: If you want to write the fix outside of the Python code (i.e. "core")  such as in a file that gets autoloaded, e.g. a `.m` then have at it! 

Otherwise please wait.

mmatera: Actually, that could be a good solution. I am going to check if it is possible.

PR #347, opened 2022-06-06
Symbols and Expressions involving _Integer

rocky:
As always much more can be done. But this is what I can fit in a session

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/347"></a>


mathics/builtin/drawing/image.py
```diff
@@ -517,7 +519,7 @@ def apply_resize_width_height(self, image, width, height, evaluation, options):
         h = self._get_image_size_spec(old_h, height)
         if h is None or w is None:
             return evaluation.message(
-                "ImageResize", "imgrssz", Expression(SymbolList, width, height)
+                "ImageResize", "imgrssz", to_mathics_list(width, height)
```
TiagoCavalcante, line 522: Maybe `ListExpression(Integer(width), Integer(height))`?

mathics/builtin/intfns/divlike.py
```diff
@@ -339,7 +340,7 @@ class Quotient(Builtin):
     }
     summary_text = "integer quotient"
 
-    def apply(self, m, n, evaluation):
+    def apply(self, m: Integer, n: Integer, evaluation):
         "Quotient[m_Integer, n_Integer]"
         py_m = m.get_int_value()
```
TiagoCavalcante, line 345: `.value`?

TiagoCavalcante (approved):
LGTM

mathics/builtin/drawing/image.py
rocky: Look at the return signature on `_get_image_spec()`

rocky: There is some question as to whether we should return optional float. Or possibly whether we should use Real() along with ListExpression().  

However here I want to punt and come back to this at some later stage. And it's probably not worth putting any effort into for this particular place because this is done only when there is an error, in which case it is okay for that to be slow.

mathics/builtin/arithfns/basic.py
```diff
@@ -240,8 +240,7 @@ class Minus(PrefixOperator):
 
     def apply_int(self, x, evaluation):
         "Minus[x_Integer]"
-
-        return Integer(-x.to_sympy())
```
rocky, line 244: Here is an example of the looseness of indicating what type we have may have contributed to the sloppiness.

rocky:
270 or so more occurrences to get to. Merging and will continue tomorrow.

PR #354, opened 2022-06-08
TimeConstrained

mmatera:
Now it seems that also works in pyston.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/354"></a>

### DIFF
--- a/mathics/builtin/datentime.py
+++ b/mathics/builtin/datentime.py
@@ -1118,17 +1118,17 @@ if sys.platform != "win32":
         # that TimeConstraint has bugs.
 
         # Consider testing via unit tests.
-        >> TimeConstrained[Pause[5];x,.1]
+        >> TimeConstrained[Pause[.5];x,.1]
          = $Aborted
 
-        >> TimeConstrained[Pause[5];Integrate[Sin[x],x], .1, Integrate[Cos[x],x]]
+        >> TimeConstrained[Pause[.5];Integrate[Sin[x],x], .1, Integrate[Cos[x],x]]
          = Sin[x]
 
         >> s=TimeConstrained[Integrate[Sin[x] ^ 3, x], a]
          : Number of seconds a is not a positive machine-sized number or Infinity.
          = TimeConstrained[Integrate[Sin[x] ^ 3, x], a]
 
-        >> a=5.; s
+        >> a=.5; s
          = Cos[x] (-3 + Cos[x] ^ 2) / 3
         """
 
--- a/mathics/builtin/datentime.py
+++ b/mathics/builtin/datentime.py
@@ -25,6 +25,7 @@ from mathics.core.list import ListExpression
 from mathics.core.symbols import Symbol, SymbolList
 from mathics.core.systemsymbols import (
     SymbolAborted,
+    SymbolFailed,
     SymbolInfinity,
     SymbolNull,
     SymbolRowBox,
@@ -1040,14 +1041,21 @@ class Pause(Builtin):
 
     def apply(self, n, evaluation):
         "Pause[n_]"
-        sleeptime = n.to_python()
-        if not isinstance(sleeptime, (int, float)) or sleeptime < 0:
+        sleeptime = float(n.to_python())
+        if not isinstance(sleeptime, float) or sleeptime < 0:
             evaluation.message(
                 "Pause", "numnm", Expression(SymbolPause, from_python(n))
             )
             return
-
-        time.sleep(sleeptime)
+        # This checks each 10ms if the evaluation
+        # was stopped.
+        while sleeptime > 0.01:
+            sleeptime = sleeptime - 0.01
+            time.sleep(0.01)
+            if evaluation.timeout:
+                return SymbolNull
+        if sleeptime > 0:
+            time.sleep(sleeptime)
         return SymbolNull
 
 
@@ -1088,7 +1096,7 @@ class Now(Predefined):
         return Expression(SymbolDateObject.evaluate(evaluation))
 
 
-if sys.platform != "win32" and not hasattr(sys, "pyston_version_info"):
+if sys.platform != "win32":
 
     class TimeConstrained(Builtin):
         r"""
@@ -1105,24 +1113,24 @@ if sys.platform != "win32" and not hasattr(sys, "pyston_version_info"):
         the evaluation continues after the timeout. However, at the end of the evaluation, the function will return '$Aborted' and the results will not affect
         the state of the \Mathics kernel.
 
-        """
 
         # FIXME: these tests sometimes cause SEGVs which probably means
         # that TimeConstraint has bugs.
 
         # Consider testing via unit tests.
-        # >> TimeConstrained[Integrate[Sin[x]^1000000,x],1]
-        # = $Aborted
+        >> TimeConstrained[Pause[5];x,.1]
+         = $Aborted
 
-        # >> TimeConstrained[Integrate[Sin[x]^1000000,x], 1, Integrate[Cos[x],x]]
-        # = Sin[x]
+        >> TimeConstrained[Pause[5];Integrate[Sin[x],x], .1, Integrate[Cos[x],x]]
+         = Sin[x]
 
-        # >> s=TimeConstrained[Integrate[Sin[x] ^ 3, x], a]
-        #  : Number of seconds a is not a positive machine-sized number or Infinity.
-        #  = TimeConstrained[Integrate[Sin[x] ^ 3, x], a]
+        >> s=TimeConstrained[Integrate[Sin[x] ^ 3, x], a]
+         : Number of seconds a is not a positive machine-sized number or Infinity.
+         = TimeConstrained[Integrate[Sin[x] ^ 3, x], a]
 
-        # >> a=1; s
-        # =  Cos[x] (-5 + Cos[2 x]) / 6
+        >> a=5.; s
+         = Cos[x] (-3 + Cos[x] ^ 2) / 3
+        """
 
         attributes = hold_all | protected
         messages = {
--- a/mathics/core/evaluation.py
+++ b/mathics/core/evaluation.py
@@ -54,7 +54,7 @@ def _thread_target(request, queue) -> None:
         queue.put((False, exc_info))
 
 
-def kill_thread(thread):
+def kill_thread(thread) -> bool:
     """
     Tries to kill a thread.
     If successful, returns True; otherwise, False.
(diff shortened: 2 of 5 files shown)


mathics/core/evaluation.py
```diff
@@ -117,10 +149,11 @@ def run_with_timeout_and_stack(request, timeout, evaluation):
     thread.join(timeout)
     if thread.is_alive():
         evaluation.timeout = True
-        while thread.is_alive():
-            pass
-        evaluation.timeout = False
+        time.sleep(0.01)
+        if not kill_thread(thread):
+            print("thread couldn't be stopped.")
```
rocky, line 154: I wonder if there is something more appropriate than `print`?

mathics/builtin/datentime.py
```diff
@@ -1105,24 +1113,24 @@ class TimeConstrained(Builtin):
...
 
-        """
 
         # FIXME: these tests sometimes cause SEGVs which probably means
         # that TimeConstraint has bugs.
 
         # Consider testing via unit tests.
-        # >> TimeConstrained[Integrate[Sin[x]^1000000,x],1]
-        # = $Aborted
+        >> TimeConstrained[Pause[5];x,.1]
```
rocky, line 1121: The problem with code like this is that now tests are 5 seconds times two longer when most of the time it doesn't take that long. 

And this shows a problem with doctests. If you want to write a good test, then it really doesn't make a good example. 

In Python for example you could write an operation and time it, and _then_ pick a pause number that is _a little bit_ large than the last operation run. 

That is a longer way of saying "Consider testing via unit tests". 

 BTW do we know _why_ there were  SEGV's?

mathics/core/evaluation.py
```diff
@@ -54,6 +54,38 @@ def _thread_target(request, queue) -> None:
         queue.put((False, exc_info))
 
 
+def kill_thread(thread):
```
rocky, line 57: Small thing if you are up for it -> bool would be nice. Maybe even a type annotation for thread if that is easily accessible.

test/builtin/test_datentime.py
```diff
@@ -8,17 +9,28 @@
 
 
 @pytest.mark.skipif(
-    sys.platform in ("win32",) or hasattr(sys, "pyston_version_info"),
+    sys.platform in ("win32",),  # or hasattr(sys, "pyston_version_info"),
+    reason="pyston and win32 does not do well killing threads",
+)
+def test_timeremaining0():
+    str_expr = "TimeConstrained[Integrate[Sin[x]^1000000, x], 0.9]"
```
rocky, line 16: Do we really need something complicated and something that takes a lot of time like this? 

Maybe something like `TimeConstrained[1 + 2, 0.00001]` would work and would be a lot faster. 

(I tried in current master and it doesn't work. In that I learn we don't have a `$TimeUnit`.

test/builtin/test_datentime.py
```diff
@@ -8,17 +9,28 @@
...
+    result = evaluate(str_expr)
+
+    assert result is None or result == Symbol("$Aborted")
```
rocky, line 19: Elsewhere is a suggestion for how to do this in a way were you can be more certain of what the result should be.

rocky:
> Now it seems that also works in pyston.
> 

That reminds me, it would be nice to have another gitflow that _demonstrates_ that this works in pyston. And maybe one for Python 3. 10 as well.

test/builtin/test_datentime.py
rocky: I tried `TimeConstrained[Integrate[Sin[x]^10000, x], 0.5]` and that give Aborted too. This is simpler and faster. Can we get this the exponent smaller and the delay time smaller still?

mathics/core/evaluation.py
```diff
@@ -117,10 +149,11 @@ def run_with_timeout_and_stack(request, timeout, evaluation):
     thread.join(timeout)
     if thread.is_alive():
         evaluation.timeout = True
-        while thread.is_alive():
-            pass
-        evaluation.timeout = False
+        time.sleep(0.01)
```
rocky, line 152: When I see a `time.sleep(x)` I always feel that something is fishy. This kind of think should only be used as a _last_ resort. 

Will think about this some more.

rocky: I wonder if 0.01 is related to `$TimeUnit` which we don't have defined.

mmatera:
OK, I thought I had marked this as "draft" buti think I didn't. What I was trying to see if with the new implementation, this code works on windows too. I tried this in my computer with pyston and it seems to work.

mathics/core/evaluation.py
mmatera: Is related with the overhead I see in my computer. This is something to refine.

mmatera: Regarding *sleep*, is what Pause does in WL. Is there a better way to do it?

test/builtin/test_datentime.py
mmatera: The goal of this test is to check if the thread is killed, or not. Ideally, you reach the wall time before you saturates the resources and rise a segmentation fault.

mathics/builtin/datentime.py
mmatera: Yep: the problem is in sympy. To find it I needed such a onerous test.

mathics/core/evaluation.py
mmatera: Sure, but this was supposed to be a draft. Tonmorow I am going to replace this with a message.

mathics/core/evaluation.py
rocky: I suspect if you were to do this  with the Python's [multiprocessing](https://docs.python.org/3/library/multiprocessing.html) module also mentioned in https://www.geeksforgeeks.org/python-different-ways-to-kill-a-thread/  it would work in Windows, PyPy, and pyston without the sleeps and with more reliable (non-GIL) threads and therefore also faster. 

The only downside is that it requires a CPU with more than one core, but that is pretty much all CPU's except for little ARM matchbox processors. But those won't run this anyway.

mathics/core/evaluation.py
rocky: > > Now it seems that also works in pyston.
> 
> That reminds me, it would be nice to have another gitflow that _demonstrates_ that this works in pyston. And maybe one for Python 3. 10 as well.

@mmatera are you up for adding this?

test/builtin/test_datentime.py
rocky: And if something goes wrong, you leave the developer with a long-running mess.

rocky:
> OK, I thought I had marked this as "draft" but think I didn't. What I was trying to see if with the new implementation, this code works on windows too. I tried this in my computer with pyston and it seems to work.

What I see in a lot of the "drafts" are lots of commit of the form "merge", "print", "word", "word2", "revert" ... 

When working like this, I think you'd be better doing this in your own personal fork. When you are close, squash all of the noise down. 

It is useful to get some sense of what's up so we lose the head's up on what random thing is getting worked on, but maybe that could be done some other way.

mathics/core/evaluation.py
mmatera: Yes, but I am not sure right now how to do it.

test/builtin/test_datentime.py
mmatera: The idea is to merge this into the point in which we are sure that such a long-running mess does not happen. But to see that the evaluation was really stopped, I need something that complains alone. 

Another possible test would be
```a=0;TimeConstrained[Do[Pause[.1];a=a+1,{10000}], .1];a<10``` 
but the problem with it is that almost all the evaluation happens inside Mathics, where we are checking the status of the `timeout` flag. 

What I am trying to test with this expression are evaluations that happen in sympy or numpy, where we do not have a way to fix a wall time.

mathics/core/evaluation.py
mmatera: I have to check it. What decides me to use threads instead the process is that we are working over the same ``evaluation.definitions`` object. So, it makes sense to use a what that allows sharing memory.  

On the other hand, at some point the multithread module should support this "feature" in the other platforms, isn't it?

mathics/core/evaluation.py
rocky: Ok - I will do. I have an opinion on this too. 

The main thing is to do this as a separate workflow so we don't increase the overall time to completion. And since ubuntu is the fastest and doesn't seem to have quotas on the number of runs (I guess partly because it is a free OS so the CI provider can spin up an unlimited number of these), this suggests we do this like we do in Cython in Ubuntu only.  

In my opinion we get value in testing say Python 3.6, whether there is value in testing Python 3.6 across *all* platforms is questionable. 

And down the line maybe we should switch to pyston across all platforms, especially Windows where testing takes the longest.

mathics/core/evaluation.py
rocky: The docs say it runs on Windows and GNU/Linux. Given that I can't imagine why it wouldn't also be supported on on MacOS. 

And I think you will find it is faster. Please give it a try.

mathics/builtin/datentime.py
rocky: Ok. Thanks for the investigation. One way to let others know this is to add it in a comment. And if it is onerous, then it is fine to add this in a pytest kind (marked that this is known to SEGV).

test/builtin/test_datentime.py
rocky: > I tried `TimeConstrained[Integrate[Sin[x]^10000, x], 0.5]` and that gives Aborted too. This is simpler and faster. Can we get this the exponent smaller and the delay time smaller still?

I see that instead of 0.5, a value of 0 has to be the best.

mathics/builtin/datentime.py
mmatera: The choice of 5secs instead of 0.5 allows to me to see if the process was killed or not in real-time. Now, I change it to .5.

rocky: Or even 0. The other point though while it is great that you are doing all of these experiments as tests, does everyone have to know about the experiments that work and the ones that don't work (which generally happen more often).

I mean there has to be a way to be less noisy but on the other hand let folks have some sense of what's up.

test/builtin/test_datentime.py
rocky: Another possibility that is easily done in pytest is to have long-running tests performed via an OS Environment variable. Then we can arrange a CI workflow (one in parallel) that doesn't run the entire suite but just the choice long-running tests.

mathics/builtin/datentime.py
mmatera: Now ``TimeConstrained[Integrate[Sin[x]^1000000,x],1]`` was moved into a pytest

mathics/core/evaluation.py
rocky: > What decides me to use threads instead the process is that we are working over the same evaluation.definitions object

This might be true if there weren't the Global Interpreter Lock (GIL) that slows things down here.

PR #358, opened 2022-06-10
List expression 3

rocky:
More of the same.

There are some unconverted Expressions and ListExpressions remaining. These will require special attention I don't have time for right now. So that will be another PR. 

Until then I want to get the easier changes out of the way. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/358"></a>

### DIFF
--- a/mathics/builtin/compilation.py
+++ b/mathics/builtin/compilation.py
@@ -8,13 +8,19 @@ When LLVM and Python libraries are available, compilation produces LLVM code.
 
 
 import ctypes
+from types import FunctionType
 
-
-from mathics.core.evaluators import apply_N
 from mathics.builtin.base import Builtin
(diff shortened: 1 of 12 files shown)


mmatera (approved):
LGTM

rocky:
FYI In an aborted attempt, I realized how to reduce some of the circular dependencies between atom (which contains from_python) and Expression which has conversion routines in the form of to_expression (and right now Expression() which calls from_python(). 

Basically the to_expression from expression.py should be pulled out the one in Expression will just disappear soon)  and the same thing for to_mathics_list from ListExpression, and the same thing for from_expression of atom. These can be in a module for from/to conversions. Once conversions are  _out_ of the lower level atom and expression modules, we con't have circular imports. 

I tried that in the PR, but got stuck.

PR #359, opened 2022-06-10
Final Expression -> ListExpression

rocky:
Roadmap: 

* After this there are about 200 occurrences of `Expression("string")` to remove. 
* Even after this, we need to probably sort out some reduced number of  `Expression` from `to_expression`.
* Then the conversion in `Expression()` can be removed  and we win in performance!
* Sometime here or after the next step, I invite others to do their own benchmarks
* Then we can remove `has_form` on lists. And we can probably simplify other "has_form"s as well
* Then we can release, but I will note there are still large opportunities still in Expression and ListExpression. A big one I would like to see is reducing a deconversion from Mathics to Python data for list literals which will be helpful in plotting. This was just the first "rough" pass. 
* Then @mmatera can start on the Boxing and Formatting stuff that he's long wanted to do. That too will eventually result in speedups.

Originally I was going to do _all_ of `Expression()` before converting `ListExpression()` so that we could get a benchmark of the one versus the other.  I have though _some_ benchmarks for Expression on the combinatorica tests. 

`ListExpression()` never allowed conversion like `Expression()`.  So by doing things intermingled, I've reduced the work needed to sort out Expression vs to_expression as well as the remaining Expression() go over since some Expressions were ListExpressions.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/359"></a>


TiagoCavalcante:
LGTM

PR #360, opened 2022-06-10
Speed up tests by reducing numbers...

rocky:
Here there are a number of tests where having large numbers really is
showning much benefit. For example whether we wan the 500 digit versus
the 1000000 digit is kind of pointless and just takes time.

Or how about a maximum number of iterations. In the end we get this same
result: too many iterations.

For plotting and image processing though, I suppose there is benefit in
taking the time since in the end you see a prettier plot. In the case of
images, well, we use reasonable size images.

(The time for plotting eventually I expect to see drop.)

The same things should be done for the pytests. However the combinatorica tests like plotting tests need to stay as the are. Those tests are exactly those from a published book. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/360"></a>

### DIFF
--- a/mathics/builtin/atomic/numbers.py
+++ b/mathics/builtin/atomic/numbers.py
@@ -507,13 +507,13 @@ class RealDigits(Builtin):
     >> RealDigits[19 / 7]
      = {{2, {7, 1, 4, 2, 8, 5}}, 1}
 
-    The 10000th digit of  is an 8:
-    >> RealDigits[Pi, 10, 1, -10000]
-    = {{8}, -9999}
+    The 500th digit of Pi is 2:
+    >> RealDigits[Pi, 10, 1, -500]
+    = {{2}, -499}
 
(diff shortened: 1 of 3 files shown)


rocky:
I should also say, that I looked over the `--time-each` option of docpipeline.py and found it lacking. Too often I am seeing numbers without insight. 

Parsing and comparing output strings has never been a problem and probably never will be a problem. So listing those times is to me noise.  What I _wanted_ was a way to see timings of tests that were _over_ a certain threshold, not all of them. This suggests it would be good to have a parameter to that option to indicate larger than what should we start reporting time. This is what is done in `mathics.core.util` in `MIN_ELAPSE_REPORT`. BTW that timeit stuff if not other stuff should be moved out of `core` and up a level. Timing is not a core thing. 

Finally I see a misspelled variable name `enlapsed_time` which should be `elapsed_time`

PR #362, opened 2022-06-10
using ``installed`` property to skip tests

mmatera:
This PR is a previous step for adding a Pyston test workflow (see #368). The proposed changes allow skipping the tests associated with symbols whose evaluation requires not available external libraries. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/362"></a>

To check that indeed this works with a minimal environment, I added here a workflow to check that in standard Ubuntu Python.

### DIFF
--- a/.github/workflows/ubuntu-minimal.yml
+++ b/.github/workflows/ubuntu-minimal.yml
@@ -1,0 +1,34 @@
+name: Mathics (ubuntu-minimal)
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    env:
+      NO_CYTHON: 1
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        python-version: [3.9]
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        sudo apt-get update -qq && sudo apt-get install -qq liblapack-dev llvm-dev
+        python -m pip install --upgrade pip
+        # Can remove after next Mathics-Scanner release
+        # python -m pip install -e git+https://github.com/Mathics3/mathics-scanner#egg=Mathics-Scanner[full]
+    - name: Install Mathics with full dependencies
+      run: |
+        make develop
+    - name: Test Mathics
+      run: |
+        make -j3 check
--- a/mathics/doc/common_doc.py
+++ b/mathics/doc/common_doc.py
@@ -37,7 +37,7 @@ from typing import Callable
 
 from mathics import builtin
 from mathics import settings
-from mathics.builtin import get_module_doc
+from mathics.builtin import get_module_doc, check_requires_list
 from mathics.core.evaluation import Message, Print
 from mathics.doc.utils import slugify
 
@@ -665,6 +665,8 @@ class Documentation(object):
                                     # iterated below. Probably some other code is faulty and
                                     # when fixed the below loop and collection into doctest_list[]
                                     # will be removed.
+                                    if not docsubsection.installed:
+                                        continue
                                     doctest_list = []
                                     index = 1
                                     for doctests in docsubsection.items:
@@ -906,13 +908,8 @@ class MathicsMainDocumentation(Documentation):
         object to the chapter, a DocChapter object.
         "section_object" is either a Python module or a Class object instance.
         """
-        installed = True
-        for package in getattr(section_object, "requires", []):
-            try:
-                importlib.import_module(package)
-            except ImportError:
-                installed = False
-                break
+        installed = check_requires_list(getattr(section_object, "requires", []))
+
         # FIXME add an additional mechanism in the module
         # to allow a docstring and indicate it is not to go in the
         # user manual
@@ -949,17 +946,12 @@ class MathicsMainDocumentation(Documentation):
         operator=None,
         in_guide=False,
     ):
-        installed = True
-        for package in getattr(instance, "requires", []):
-            try:
-                importlib.import_module(package)
-            except ImportError:
-                installed = False
-                break
+        installed = check_requires_list(getattr(instance, "requires", []))
 
         # FIXME add an additional mechanism in the module
         # to allow a docstring and indicate it is not to go in the
         # user manual
+
         if not instance.__doc__:
             return
         subsection = DocSubsection(
@@ -1058,6 +1050,8 @@ class PyMathicsDocumentation(Documentation):
                 and var.__module__[: len(self.pymathicsmodule.__name__)]
                 == self.pymathicsmodule.__name__
             ):  # nopep8
+                if not check_requires_list(var):
+                    continue
                 instance = var(expression=False)
                 if isinstance(instance, Builtin):
                     self.symbols[instance.get_name()] = instance
@@ -1105,13 +1099,7 @@ class PyMathicsDocumentation(Documentation):
         chapter = DocChapter(builtin_part, title, XMLDoc(text))
         for name in self.symbols:
             instance = self.symbols[name]
-            installed = True
-            for package in getattr(instance, "requires", []):
-                try:
-                    importlib.import_module(package)
-                except ImportError:
-                    installed = False
-                    break
+            installed = check_requires_list(getattr(instance, "requires", []))
             section = DocSection(
                 chapter,
                 strip_system_prefix(name),
@@ -1329,8 +1317,12 @@ class DocGuideSection(DocSection):
         # A guide section's subsection are Sections without the Guide.
         # it is *their* subsections where we generally find tests.
         for section in self.subsections:
+            if not section.installed:
+                continue
             for subsection in section.subsections:
                 # FIXME we are omitting the section title here...
+                if not subsection.installed:
+                    continue
                 for doctests in subsection.items:
                     yield doctests.get_tests()
 
--- a/mathics/algorithm/optimizers.py
+++ b/mathics/algorithm/optimizers.py
@@ -350,9 +350,11 @@ def find_root_newton(f, x0, x, opts, evaluation) -> (Number, bool):
     return x0, True
 
 
+native_optimizer_messages = {}
+
 native_local_optimizer_methods = {
     "Automatic": find_minimum_newton1d,
-    "newton": find_minimum_newton1d,
+    "Newton": find_minimum_newton1d,
 }
 
 native_findroot_methods = {
@@ -360,6 +362,7 @@ native_findroot_methods = {
     "Newton": find_root_newton,
     "Secant": find_root_secant,
 }
+native_findroot_messages = {}
 
 
 def is_zero(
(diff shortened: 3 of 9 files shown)


mathics/doc/common_doc.py
```diff
@@ -138,6 +138,28 @@
 test_result_map = {}
 
 
+# TODO: Use this when checking requires for modules.
+requires_lib_cache = {}
+
+
+def check_requires_list(requires: list) -> bool:
+    global requires_cache
```
rocky, line 146: This line does nothing.

mathics/doc/common_doc.py
```diff
@@ -1058,6 +1072,9 @@ def __init__(self, module=None):
                 and var.__module__[: len(self.pymathicsmodule.__name__)]
                 == self.pymathicsmodule.__name__
             ):  # nopep8
+                if not check_requires_list(var):
+                    print("  skiping ", var.__name__)
```
rocky, line 1076: skiping -> skipping. But I'd be more explicit to why it is skipped, an optional package is not installed.

mathics/doc/common_doc.py
mmatera: I put this for clarity, but OK.

mathics/doc/common_doc.py
mmatera: This print actually never happens, because we are not generating documentation for pymathics modules. I just removed it.

mathics/doc/common_doc.py
rocky: Huh? There is no variable `requires_cache`.  The typical way of adding clarity is in a docstring. For example:

```
"""
Check if module names  in ``requires`` can be imported and return True if they can or False if not.

This state value is also recorded in dictionary `requires_lib_cache` keyed by package name and is used to determine whether to skip trying to get information from the package.
"""
```

rocky:
Overall this looks good and is much needed. I'd like to try it out though and haven't had a chance.

mathics/doc/common_doc.py
mmatera: You are right! I changed the name. The variable was `requires_lib_cache`, which was defined just above.

mathics/doc/common_doc.py
```diff
@@ -138,6 +138,31 @@
...
+# TODO: Use this when checking requires for modules.
+requires_lib_cache = {}
+
+
+def check_requires_list(requires: list) -> bool:
+    """
+    Check if module names in ``requires`` can be imported and return True if they can or False if not.
+
+    This state value is also recorded in dictionary `requires_lib_cache` keyed by module name and is used to determine whether to skip trying to get information from the module."""
+    global requires_lib_cache
```
mmatera, line 150: This is what I had in mind. Having or not this line does not seems to change anything, just to state that `require_lib_cache` is a variable at the level of the module.

mmatera:
> Overall this looks good and is much needed. I'd like to try it out though and haven't had a chance.

Actually, there are several things to improve in this module, but I refrain to do it in order to make this PR as simple as possible to get something that works with the basic Pyston test workflow.

mathics/doc/common_doc.py
rocky: Please don't add a global. It is frowned upon generally.  Some would argue that this should be passed as a parameter, and that is the typical way you convey this - in the interface. 


I know this kind of thing, a creative use of  `global` seems innovated. I used to (and still sometimes do) add `pass` statements where in most programming languages an end closer would appear. However it puts off most programmers so I generally don't do it except in code where I am pretty much the only one who will ever use this. 

There is a place for creativity, but here and this way sadly isn't the place for that.

rocky:
> > Overall this looks good and is much needed. I'd like to try it out though and haven't had a chance.
> 
> Actually, there are several things to improve in this module, but I refrain to do it in order to make this PR as simple as possible to get something that works with the basic Pyston test workflow.

Sure, I just want to double check and try it so there isn't something kind of silly like there was in `--time-each` option of docpipeline.py

rocky:
> Actually, there are several things to improve in this module, 

Please record those items as  issues. Thanks.

I just tried without `scikit_image` using pyston-2.3.3. I get: 

```
python mathics/docpipeline.py -x
Testing Mathics 4.1.0.dev0
...
Test failed: BinaryImageQ in Reference of Built-in Symbols / Graphics, Drawing, and Images
Reference of Built-in Symbols
Result: False
Wanted: True
Additional output:
General::pyimport: Binarize[] is not available. Python module "skimage" is not installed.
Test failed: BinaryImageQ in Reference of Built-in Symbols / Graphics, Drawing, and Images
Reference of Built-in Symbols
Result: False
Wanted: True
Additional output:
General::pyimport: Binarize[] is not available. Python module "skimage" is not installed.

```

mmatera:
> > Actually, there are several things to improve in this module,
> 
> Please record those items as issues. Thanks.
> 
> I just tried without `scikit_image` using pyston-2.3.3. I get:
> 
> ```
> python mathics/docpipeline.py -x
> Testing Mathics 4.1.0.dev0
> ...
> Test failed: BinaryImageQ in Reference of Built-in Symbols / Graphics, Drawing, and Images
> Reference of Built-in Symbols
> Result: False
> Wanted: True
> Additional output:
> General::pyimport: Binarize[] is not available. Python module "skimage" is not installed.
> Test failed: BinaryImageQ in Reference of Built-in Symbols / Graphics, Drawing, and Images
> Reference of Built-in Symbols
> Result: False
> Wanted: True
> Additional output:
> General::pyimport: Binarize[] is not available. Python module "skimage" is not installed.
> ```

this is in this branch?

rocky:
> > > Actually, there are several things to improve in this module,
> > 
> > 
> > Please record those items as issues. Thanks.
> > I just tried without `scikit_image` using pyston-2.3.3. I get:
> > ```
> > python mathics/docpipeline.py -x
> > Testing Mathics 4.1.0.dev0
> > ...
> > Test failed: BinaryImageQ in Reference of Built-in Symbols / Graphics, Drawing, and Images
> > Reference of Built-in Symbols
> > Result: False
> > Wanted: True
> > Additional output:
> > General::pyimport: Binarize[] is not available. Python module "skimage" is not installed.
> > Test failed: BinaryImageQ in Reference of Built-in Symbols / Graphics, Drawing, and Images
> > Reference of Built-in Symbols
> > Result: False
> > Wanted: True
> > Additional output:
> > General::pyimport: Binarize[] is not available. Python module "skimage" is not installed.
> > ```
> 
> this is in this branch?

I believe so: 

```
$ git status
On branch improve_install_check_in_docs
Your branch is up to date with 'origin/improve_install_check_in_docs'.
git log | head
commit 4fde248a5dd2e88eeb3a50707f9789c646751061
Author: mmatera <matera@fisica.unlp.edu.ar>
Date:   Fri Jun 10 19:18:33 2022 -0300

    adding docstring. fixing variable name

commit 53a52465e1f6ee1254a5e916daf90aa8f5821906
Author: mmatera <matera@fisica.unlp.edu.ar>
Date:   Fri Jun 10 19:05:06 2022 -0300
```

mathics/doc/common_doc.py
mmatera: Since ```mathics.builtin.base``` also benefits from using ```check_requires_list``` I moved the code there.

mmatera:
> > > Actually, there are several things to improve in this module,
> > 
> > 
> > Please record those items as issues. Thanks.
> > I just tried without `scikit_image` using pyston-2.3.3. I get:
> > ```
> > python mathics/docpipeline.py -x
> > Testing Mathics 4.1.0.dev0
> > ...
> > Test failed: BinaryImageQ in Reference of Built-in Symbols / Graphics, Drawing, and Images
> > Reference of Built-in Symbols
> > Result: False
> > Wanted: True
> > Additional output:
> > General::pyimport: Binarize[] is not available. Python module "skimage" is not installed.
> > Test failed: BinaryImageQ in Reference of Built-in Symbols / Graphics, Drawing, and Images
> > Reference of Built-in Symbols
> > Result: False
> > Wanted: True
> > Additional output:
> > General::pyimport: Binarize[] is not available. Python module "skimage" is not installed.
> > ```
> 
> this is in this branch?

Now I added a workflow testing against a basic installation of Python 3.9 in ubuntu. #368 should show that this also works in Pyston.

rocky:
> . https://github.com/Mathics3/mathics-core/pull/368 should show that this also works in Pyston.

That is a draft and this isn't.  That may show that there is a particular way to install Pyston that works. However I've been running Pyston with scikit-image installed for a while even without this. 

What i am wondering about though is whether this addresses the problem I encountered. Not whether there is some scenario where we can make pyston work. 

It would help if you can describe a specific setup and test case where things were failing without this patch and then with this they don't fail.

mmatera:
> > . #368 should show that this also works in Pyston.
> 

> That is a draft and this isn't. That may show that there is a particular way to install Pyston that works. However I've been running Pyston with scikit-image installed for a while even without this PR.
> 

No: the idea is not to show that Pyston can run with scikit-image, but that can run without it. 

> What i am wondering about though is whether this addresses the problem I encountered. Not whether there is some scenario where we can make pyston work.
> 
> It would help if you can describe a specific setup and test case where things were failing without this patch and then with this they don't fail.


With master, if scikit image is not installed, there are tests that fail (as you show above). However, we say that scikit is an optional package, so the CI should show in green if the environment does not have scikit available. The same happens with scipy. This PR fixes that and shows that it works by adding a workflow that has a minimal installation.

mmatera:
The problem with having a workflow with the full set of libraries in Pyston is that most of the time is invested in recompiling scipy, scikit and other heavy libraries. When that libraries are available, then the tests that assume that the libraries are there run.  
having a preinstalle 
I am looking for possibilities for downloading a preinstalled environment with all the optional libraries, but still I have not give with a solution. On the other hand, if we are looking to do performance tests in the evaluation process, I think that also makes sense to check it over a minimal environment.

mmatera:
@rocky,  in #370 I tried a workflow for Pyston with the full set of optional libraries. It took ~60 min, against ~6 min with the reduced set (#368). 48 minutes were used to compile the libraries.

> > . #368 should show that this also works in Pyston.
> 
> That is a draft and this isn't. That may show that there is a particular way to install Pyston that works. However I've been running Pyston with scikit-image installed for a while even without this PR.
> 
> What i am wondering about though is whether this addresses the problem I encountered. Not whether there is some scenario where we can make pyston work.
> 
> It would help if you can describe a specific setup and test case where things were failing without this patch and then with this they don't fail.

@rocky,  in #370 I tried a workflow for Pyston with the full set of optional libraries. It took ~60 min, against ~6 min with the reduced set (#368). 48 minutes were used to compile the libraries. So, if we are trying to check that the mathics core runs properly in Pyston, I think we should try it against the basic installation. There is where this PR is needed: if you just add the workflow to master, several tests are not passed (like the ones in the `mathics.builtin.drawing.image` module).

mathics/doc/common_doc.py
mmatera: Regarding this, probably the cache for the previous results is not needed: Python have also a cache inside `importlib`, so the global variable does not provide too much advantage.

.github/workflows/ubuntu-minimal.yml
```diff
@@ -0,0 +1,34 @@
+name: Mathics (ubuntu-minimal)
```
mmatera, line 1: This workflow was added to check that indeed with the changes the tests pass with a minimal installation.

mmatera:
Superseeded.

PR #363, opened 2022-06-10
fixing NIntegrate to make the internal integrators available.

mmatera:
This PR fixes a few bugs I found in ``NIntegrate`` trying to make it works when scipy is not available. Also, I marked here ``NIntegrate``, and ``_BaseFinder`` daughter classes as builtin that requires scipy, to avoid running their doctests when scipy is not available (because the results are going to be slightly different).

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/363"></a>


rocky (approved):
LGTM -  @mmatera we should get this when you are satisfied or finished.

mmatera:
Let's merge and iterate

PR #364, opened 2022-06-12
leaves->elements at the Pattern boundary

rocky:
We need to change leaves to elements in Pattern, but that will be some
other round in the future.

This inches forward though.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/364"></a>

### DIFF
--- a/mathics/builtin/base.py
+++ b/mathics/builtin/base.py
@@ -869,13 +869,13 @@ class BoxExpression(BuiltinElement):
                 default[option] = parse_builtin_rule(value)
         return default
 
-    def boxes_to_text(self, leaves, **options) -> str:
+    def boxes_to_text(self, elements, **options) -> str:
         raise BoxConstructError
 
-    def boxes_to_mathml(self, leaves, **options) -> str:
+    def boxes_to_mathml(self, elements, **options) -> str:
         raise BoxConstructError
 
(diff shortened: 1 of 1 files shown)


mathics/builtin/base.py
```diff
@@ -1012,28 +1012,28 @@ def from_expression(expr):
...
                     else:
                         return CountableInteger(py_n, upper_limit=True)
                 elif CountableInteger._support_infinity:
                     if (
                         n.get_head_name() == "System`DirectedInfinity"
-                        and len(n.leaves) == 1
+                        and len(n.elements) == 1
                     ):
-                        if n.leaves[0].get_int_value() > 0:
+                        if n.elements[0].get_int_value() > 0:
```
TiagoCavalcante, line 1036: `.value` here.

TiagoCavalcante (approved):
LGTM

mathics/builtin/base.py
rocky: The broader context: 

```
                if isinstance(n, Integer):
                    ....
                elif CountableInteger._support_infinity:
                    if (...):
                        if n.elements[0].get_int_value() > 0:
```

So no, not here just yet.  I still think `get_int_value()` in general is a name from the bad-old-days, but addressing this is a separate concern and set of PRs.

PR #369, opened 2022-06-13
More expression conversions 1

rocky:
In this there even more new ListExpression appearances (and I had thought I got them all previously). 

After this there, are about 180 more occurrences to get through.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/369"></a>


TiagoCavalcante (approved):
LGTM

mmatera:
@rocky, there is one more in `mathics.builtin.numbers.algebra`.
Also, there are a few other places where `Expression(SymbolList,...)` are created:
* when the parser finds something like `{...}`
* in `Element.do_format`
* in `Expression.atom_list_constructor`

and maybe other places

rocky:
> @rocky, there is one more in `mathics.builtin.numbers.algebra`. Also, there are a few other places where `Expression(SymbolList,...)` are created:
> 
> * when the parser finds something like `{...}`
> * in `Element.do_format`
> * in `Expression.atom_list_constructor`
> 
> and maybe other places

There are plenty more places, yes. I have just been reducing them. Anyone free to jump in and fix these. One thing that tends to be lacking is a coordinated effort some particular activity, like things

Instead lurch off in different unexpected directions and this slows down progress in agreed-upon areas.

mmatera:
> > @rocky, there is one more in `mathics.builtin.numbers.algebra`. Also, there are a few other places where `Expression(SymbolList,...)` are created:
> > 
> > * when the parser finds something like `{...}`
> > * in `Element.do_format`
> > * in `Expression.atom_list_constructor`
> > 
> > and maybe other places
> 
> There are plenty more places, yes. I have just been reducing them. Anyone free to jump in and fix these. One thing that tends to be lacking is a coordinated effort some particular activity, like things
> 
> Instead lurch off in different unexpected directions and this slows down progress in agreed-upon areas.

@rocky, I mentioned this because I understood you wanted that in the following step to use ``isinstance(expr, ListExpression)`` as a replacement of ``expr.hasform("List", None)``. 

Regarding the explicit occurrences, I think you have removed all of them.

rocky:
Instead lurch off in different unexpected directions and this slows down progress in agreed-upon areas.
> 
> @rocky, I mentioned this because I understood you wanted that in the following step to use `isinstance(expr, ListExpression)` as a replacement of `expr.hasform("List", None)`.
> 

Ok it is not that this isn't useful so much as uncoordinated. 

In https://github.com/Mathics3/mathics-core/pull/359 I list the plan I am following. And when I give a plan or a task I try to stay pretty much to it. 

Here again is part of the list 

* After this there are about 200 occurrences of Expression("string") to remove.
* Even after this, we need to probably sort out some reduced number of Expression from to_expression.
* Then the conversion in Expression() can be removed and we win in performance!
* Sometime here or after the next step, I invite others to do their own benchmarks
* Then we can remove has_form on lists. And we can probably simplify other "has_form"s as well

The things mentioned a bit down the lst. I realize I should elaborate though on how I've doing and thinking about removing the conversion to Expression().

I imagine there will be alternate versions of this and ListExpression that do explicit checks of their parameters and that  can be switched to either as an option on startup or possibly even dynamically. Assuming the tests cover all of the places you mention, we will catch those places in that.

PR #381, opened 2022-06-19
Changes to allow PyPy to build and mostly run ...

rocky:
we can get through pytests and the Gries & Schneider tests.

There is probably a PyPy bug or weirdness around `Binarize`:

```
In[1]:= img = Import["ExampleData/lena.tif"]
Out[1]= -Image-
In[2]:= Binarize[img]
terminate called after throwing an instance of 'pybind11::error_already_set'
  what():  IndentationError: ('unexpected indent', ('<string>', 2, 1, '        class pybind11_static_property(property):\n', 2))
```

But let's leave that for later.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/381"></a>

### DIFF
--- a/mathics/builtin/numbers/calculus.py
+++ b/mathics/builtin/numbers/calculus.py
@@ -2033,7 +2033,7 @@ class NIntegrate(Builtin):
     >> NIntegrate[Exp[x],{x,-Infinity, 0},Tolerance->1*^-6, Method->"Internal"]
      = 1.
     >> NIntegrate[Exp[-x^2/2.],{x,-Infinity, Infinity},Tolerance->1*^-6, Method->"Internal"]
-     = 2.50664
+     = 2.5066...
 
     """
 
--- a/mathics/builtin/system.py
+++ b/mathics/builtin/system.py
@@ -536,7 +536,7 @@ class MemoryInUse(Builtin):
     """
     <dl>
       <dt>'MemoryInUse[]'
-      <dd>Returns the amount of memory used by the definitions object. If we can't determine this we return -1.
+      <dd>Returns the amount of memory used by all of the definitions objects if we can determine that; -1 otherwise.
     </dl>
 
     >> MemoryInUse[]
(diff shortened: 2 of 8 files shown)


mathics/builtin/numbers/calculus.py
```diff
@@ -2033,7 +2033,7 @@ class NIntegrate(Builtin):
     >> NIntegrate[Exp[x],{x,-Infinity, 0},Tolerance->1*^-6, Method->"Internal"]
      = 1.
     >> NIntegrate[Exp[-x^2/2.],{x,-Infinity, Infinity},Tolerance->1*^-6, Method->"Internal"]
-     = 2.50664
+     = 2.5066...
```
rocky, line 2036: This isn't related to PyPy per se, but running mathics _withouth_ scipy. 
Here we get I think 2.50663

mmatera: The NIntegrate implementation using "Internal" as a method should not depend on scipy. Maybe ``Exp`` does?

rocky: Maybe, I didn't investigate.

mathics/builtin/system.py
```diff
@@ -535,7 +536,7 @@ class MemoryInUse(Builtin):
     """
     <dl>
       <dt>'MemoryInUse[]'
-      <dd>Returns the amount of memory used by the definitions object.
+      <dd>Returns the amount of memory used by the definitions object. If we can't determine this we return -1.
```
TiagoCavalcante, line 539: This sounds more fluid to me:
"Returns the amount of memory used by the definitions object if we can determine it, -1 otherwise."

TiagoCavalcante (approved):
LGTM

mathics/builtin/system.py
rocky: I agree - thanks!  Change made in in 04fcb35a

PR #386, opened 2022-06-19
more expression conversions

mmatera:
more canonicalization, more removing implicit conversions.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/386"></a>

### DIFF
--- a/mathics/builtin/atomic/symbols.py
+++ b/mathics/builtin/atomic/symbols.py
@@ -513,7 +513,9 @@ class Information(PrefixOperator):
                 lines.append(
                     Expression(
                         SymbolHoldForm,
-                        Expression(up and "UpSet" or "Set", lhs(rule.pattern.expr), r),
+                        Expression(
+                            up and SymbolUpSet or SymbolSet, lhs(rule.pattern.expr), r
+                        ),
                     )
                 )
 
--- a/mathics/builtin/atomic/symbols.py
+++ b/mathics/builtin/atomic/symbols.py
@@ -277,7 +277,7 @@ class Definition(Builtin):
                     Expression(
                         SymbolHoldForm,
                         Expression(
-                            up and SymbolUpSet or SymbolSet, lhs(rule.pattern.expr), r
+                            SymbolUpSet if up else SymbolSet, lhs(rule.pattern.expr), r
                         ),
                     )
                 )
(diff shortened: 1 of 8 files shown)


mathics/builtin/atomic/symbols.py
```diff
@@ -513,7 +513,9 @@ def print_rule(rule, up=False, lhs=lambda k: k, rhs=lambda r: r):
                 lines.append(
                     Expression(
                         SymbolHoldForm,
-                        Expression(up and "UpSet" or "Set", lhs(rule.pattern.expr), r),
+                        Expression(
+                            up and SymbolUpSet or SymbolSet, lhs(rule.pattern.expr), r
```
TiagoCavalcante, line 517: This is quite confusing, wouldn't `UpSet if up else Set` be better?

TiagoCavalcante (approved):
LGTM

mathics/builtin/atomic/symbols.py
mmatera: I just left it as I found it...

rocky: Yeah, I think this code is a holdover from the python 2.6- days before the ternary if-else operator was available. 

I find it amazing how much work we still have to do to get this to current Python standards.  If you grep for 'class .*(object)` you'll find 57 occurrences of this.

rocky:
> LGTM

Me too.

mathics/builtin/drawing/plot.py
```diff
@@ -807,7 +807,11 @@ def to_number(x):
                 multiple_colors = True
 
             if not multiple_colors and not self.never_monochrome:
-                colors = [Expression(SymbolRGBColor, *mpl_colors[0])]
+                colors = [
+                    Expression(
+                        SymbolRGBColor, *(MachineReal(cc) for cc in mpl_colors[0])
```
rocky, line 812: Actually, for this kind of thing, to_expression is better and let me explain why....

WIth Expression, we need to loop over the list to get element properties. So we do two loops. One here and one in Expression. 

Also with to_expression we record the fact that the elements here all come from MachineReal, and so in the future if we want to do the inverse conversion we may have head start here. 

Finally, I do thinks removing the generator loop simplifies things.

mathics/builtin/atomic/symbols.py
mmatera: I wonder if there are tools that apply this kind of "code upgrade" to the code automatically...

PR #388, opened 2022-06-20
Fix fn call mismatches in process_assign_other

rocky:
In particular:
* process_assign_recursion_limit
* process_assign_iteration_limit
* process_assign_module_number

only take 3 args, not 6

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/388"></a>

### DIFF
--- a/mathics/builtin/assignments/assignment.py
+++ b/mathics/builtin/assignments/assignment.py
@@ -5,17 +5,13 @@ Forms of Assignment
 
 
 from mathics.builtin.base import Builtin, BinaryOperator
-from mathics.core.symbols import Symbol
-
+from mathics.builtin.assignments.internals import _SetOperator
+from mathics.core.attributes import hold_all, hold_first, protected, sequence_hold
+from mathics.core.definitions import PyMathicsLoadException
+from mathics.core.symbols import SymbolNull
 from mathics.core.systemsymbols import (
     SymbolFailed,
 )
-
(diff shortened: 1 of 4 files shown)


rocky:
@mmatera looks like you added this back in September of 2021. Looks to me like a pretty straightforward case of parameter mismatches. Not sure way we aren't catching these in tests though.

mmatera:
> @mmatera looks like you added this back in September of 2021. Looks to me like a pretty straightforward case of parameter mismatches. Not sure way we aren't catching these in tests though.

How did you catch this? Maybe we can add a test for this.

rocky:
> > @mmatera looks like you added this back in September of 2021. Looks to me like a pretty straightforward case of parameter mismatches. Not sure way we aren't catching these in tests though.
> 
> How did you catch this? Maybe we can add a test for this.

Glad you asked. I was looking to see if we can run KnotTheory now. It sets `$MaxRecursion = 512;`. 
See  https://github.com/Mathics3/mathics-core/tree/add-KnotTheoryPackage for the branch. 
It needs the string conversion code the way it was before that last change. 

More generally I am pleased to see that this and Rubi now don't take forever to load. However neither work, and I think part of it is a package scoping problem, because variables that get defined in the package seem to disappear.

TiagoCavalcante (approved):
LGTM

PR #392, opened 2022-06-20
Reduce Symbol("Null") remove dup definition...

rocky:
A couple replaces in assignment.py are covered in another PR

Notable is that parser and its test keep the Symbol("Null").
Not sure though if this is necessary. If so, leave for later.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/392"></a>


TiagoCavalcante (approved):
LGTM

PR #395, opened 2022-06-20
Add copy fns to ListExpression and ...

rocky:
Change `__str__` to make a clearer distinction between `ListExpression` and `Expression`.

Also one more Expression() conversion. It appears there are a couple more to go.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/395"></a>

### DIFF
--- a/mathics/builtin/list/rearrange.py
+++ b/mathics/builtin/list/rearrange.py
@@ -27,6 +27,8 @@ from mathics.core.list import ListExpression
 from mathics.core.symbols import Atom, Symbol, SymbolTrue
 from mathics.core.systemsymbols import SymbolMap
 
+SymbolReverse = Symbol("Reverse")
+
 
 def _test_pair(test, a, b, evaluation, name):
     test_expr = Expression(test, a, b)
@@ -668,7 +670,9 @@ class Reverse(Builtin):
         if py_levels and py_levels[0] < 1:  # if py_level is not None, it's sorted
             py_levels = None
         if py_levels is None:
-            evaluation.message("Reverse", "ilsmp", Expression("Reverse", expr, levels))
+            evaluation.message(
+                "Reverse", "ilsmp", Expression(SymbolReverse, expr, levels)
+            )
         else:
             return Reverse._reverse(expr, 1, py_levels, evaluation)
 
--- a/mathics/core/list.py
+++ b/mathics/core/list.py
@@ -50,9 +50,9 @@ class ListExpression(Expression):
         # here...
         self._format_cache = None
 
-    # # Add this when it is safe to do.
-    # def __repr__(self) -> str:
-    #     return "<ListExpression: %s>" % self
+    # Add this when it is safe to do.
+    def __repr__(self) -> str:
+        return "<ListExpression: %s>" % self
 
     # @timeit
     def evaluate_elements(self, evaluation):
@@ -115,6 +115,17 @@ class ListExpression(Expression):
             self.evaluate_elements(evaluation)
 
         return self, False
+
+    def shallow_copy(self) -> "ListExpression":
+        """
+        For an Expression this does something with its cache.
+        Here this does not need that complication.
+        """
+        return ListExpression(
+            *self._elements, elements_properties=self.elements_properties
+        )
+
+    copy = shallow_copy
 
 
 def to_mathics_list(
--- a/mathics/core/list.py
+++ b/mathics/core/list.py
@@ -125,7 +125,14 @@ class ListExpression(Expression):
             *self._elements, elements_properties=self.elements_properties
         )
 
-    copy = shallow_copy
+    def copy(self, reevaluate=False) -> "Expression":
+        expr = ListExpression(self._head.copy(reevaluate))
+        expr._elements = tuple(element.copy(reevaluate) for element in self._elements)
+        expr.options = self.options
+        expr.original = self
+        expr._sequences = self._sequences
+        expr._format_cache = self._format_cache
+        return expr
 
 
 def to_mathics_list(


mathics/builtin/list/rearrange.py
```diff
@@ -668,7 +670,9 @@ def apply(self, expr, levels, evaluation):
         if py_levels and py_levels[0] < 1:  # if py_level is not None, it's sorted
             py_levels = None
         if py_levels is None:
-            evaluation.message("Reverse", "ilsmp", Expression("Reverse", expr, levels))
+            evaluation.message(
```
mmatera, line 673: Naughty line, it escaped to my sight and checks...

mathics/core/list.py
```diff
@@ -116,6 +116,17 @@ def rewrite_apply_eval_step(self, evaluation) -> Tuple[Expression, bool]:
 
         return self, False
 
+    def shallow_copy(self) -> "ListExpression":
+        """
+        For an Expression this does something with its cache.
+        Here this does not need that complication.
+        """
+        return ListExpression(
```
mmatera, line 124: We could check it with the example `First[Timing[Fold[#1+#2&, Range[750]]]]`  and see the difference...

rocky: Here is what I get for the current master versus this code (just rebased off of master): 

```
07:34:41 [master] $ mathics

Mathics 4.1.0.dev0
on CPython 3.8.12 (heads/v2.3.4.1_release:4a6b4d3504, Jun  3 2022, 15:46:12)
using SymPy 1.10.1, mpmath 1.2.1, numpy 1.22.4, cython 0.29.30
...
In[1]:= First[Timing[Fold[#1+#2&, Range[750]]]]
Out[1]= 1.8438

07:35:05 [listexpression-copy-fns] $ mathics

Mathics 4.1.0.dev0
on CPython 3.8.12 (heads/v2.3.4.1_release:4a6b4d3504, Jun  3 2022, 15:46:12)
using SymPy 1.10.1, mpmath 1.2.1, numpy 1.22.4, cython 0.29.30

under certain conditions.

In[1]:= First[Timing[Fold[#1+#2&, Range[750]]]]
Out[1]= 0.463508
```

Do you get something like this too?

rocky: Actually, I skimped on the `ListExpression.copy` operation before. In 26dfd6d4 it is more faithful.

The timing I now get is 0.673966. And when I go to back to 4.0.0 and run this I get 0.70535.

But again I am curious if you get something similar.
```suggestion
        return ListExpression(
```

TiagoCavalcante: Really nice improvement!

TiagoCavalcante (approved):
LGTM

mathics/core/list.py
mmatera: This was tested using pyston as the Python interpreter:


This branch:

```

In[1]:= First[Timing[Fold[#1+#2&, Range[750]]]]
Out[1]= 0.341739
```



2.0
```
In[1]:= First[Timing[Fold[#1+#2&, Range[750]]]]
Out[1]= 0.202057
```


3.1
```
In[1]:= First[Timing[Fold[#1+#2&, Range[750]]]]
Out[1]= 0.242977
```


master (4.1.0dev):
```
In[1]:= First[Timing[Fold[#1+#2&, Range[750]]]]
Out[1]= 1.17943
```

rocky: The timings are missing 4.0.0 which is what I have been using to compare against. When we release, that is what I had in mind to compare against in reports.

That said, it might be interesting and useful to compare against older releases. But part of the problem we run into is that especially for 1.0 there was a lot that just wasn't supported so we are comparing two different programming languages. 

And finally with respect to say matching the 3.1 behavior. The caching thing was broken somewhere along the way from 4.0.0 (which is _another_ reason why that makes a good comparison). 

At some point we'll understand what's up here and add the corresponding caching in, or more likely understand what's up with the slowness and address this some other way. 

From what I can see there is still a lot that could be done _just on the list expression part_ !

See https://github.com/Mathics3/mathics-core/discussions/394

mmatera: Here are the timings against CPython 3.9.7:


this branch:
```
In[1]:= First[Timing[Fold[#1+#2&, Range[750]]]]
Out[1]= 0.855883
```


master (4.1.0dev):
```
In[1]:= First[Timing[Fold[#1+#2&, Range[750]]]]
Out[1]= 2.7332

```



4.0.1:

```
In[1]:= First[Timing[Fold[#1+#2&, Range[750]]]]
Out[1]= 0.435961
``` 


Notice that in the following old versions, Cython is used:


3.1.0
```
In[1]:= First[Timing[Fold[#1+#2&, Range[750]]]]
Out[1]= 0.43468
```

2.0.0
```
In[1]:= First[Timing[Fold[#1+#2&, Range[750]]]]
Out[1]= 0.455832
```

rocky: Ok. Again my take is that between 4.0.0 and here we broke Expression caching although I think it is still around in some form. As I've said before probably in removing changing `__new__` to  `__init__`  but that's just wild guess. 

What I believe to be the case is that overall things have sped up since 4.0.0.  

I am not sure though whether we want to work on the caching per se, or whether the time lost here could be made up by improving ListExpression integration into the system. It is also possible that the time could be made up by improving how rules are implemented which forms the basis of the implementation of Fold.

rocky:
```
In[1]:= TraceBuiltins[Fold[#1+#2 & , Range[3]], Timing-> True]
TraceBuiltins::optx: Timing -> True is not a supported option for TraceBuiltins[].
count     ms Builtin name
  150      0 List
   64      9 Set
   51     48 CompoundExpression
   34      0 Function
   32      4 Unprotect
   32      0 ClearAttributes
   30      2 Protect
   30      0 SetAttributes
   20      0 Test
   19      0 RegisterImport
   16     91 Get
   16      0 Path
   15      0 Append
   15     18 Block
   15      0 Length
   15      0 SameQ
   15     14 If
   15      0 Last
   15      0 Most
   15      2 SetDelayed
   15      0 RegisterExport
   14      1 Times
    7      1 _MPMathFunction
    4      0 N
    4      0 _EqualityOperator
    4      0 Power
    3      0 I
    2      0 Map
```

`Get` and `CompoundExpression` seem to stick out.

rocky: > The last time I test it, removing the code associated to the cache produced a notorious fall in the performance

Well, this time it hasn't :-) 

If you want to investigate where Expression caching broke and report how to reinstate that, please do. 

I am not against caching. Just homegrown, somewhat complex undocumented and poorly understood caching. Which might also have caused huge memory leaks. 

I do think there is a place for expression caching, but _smart_ and simple caching preferably using an existing Python module like [lru](The last time I test it, removing the code associated to the cache produced a notorious fall in the performance).

rocky: > how good was the caching mechanism when it really worked, against the new approach.

A clarification. It is possible that caching was broken before ListExpression was created. I think caching got broken _inadvertently_ in a refactor of the code. 

In my view, preserving fragile optimizatons built on top of a unmodular code base isn't going to get us to where we need to get to. In some cases like here you do need to take a step or two back to get a better foundation before we can make significant progress. That caching was added more as a low-level workaround rather than understanding the big picture and trying to streamline that. 

" Well, you can put lipstick on a hog and call it Monique, but it's still a pig."

rocky: >  made the comparisons against 2.0 and 3.0 because 

This is a frequent pattern. I do X, and instead of repeating which is the simplest helpful thing to compare against you do Y which is close. Then I  wonder why did you do Y instead of X? _Then_ then we get the reason. 

If you vary from something it would be good to understand why beforehand rather than the constant why did you do Y instead of X questions and answers. 

Now back to benchmarking and some _analysis_

```
                  3.9.4    pyston-2.3.3 (no Cython)
81736f9ac3a4378   1.8      0.7
4.0.0-benchmark   1.1      0.5

                  3.9.4C   pyston-2.3.3
81736f9ac3a4378   1.5      0.7
4.0.0-benchmark   1.0      0.5


Performance diff  for pyston vs 3.9 (no Cython)
81736 vs 4.0.0    1.6      1.4

3.9.4 vs Pyston   81736f9  4.0.0-benchmark
                  2.6      2.2
3.9.C vs Pyston   81736f9  4.0.0-benchmark
                  2.1      2
```

In the above I compared CPython (no Cython) performance with Pyston 2.3.3 (no Cython). Later I tried 3.9.4 with Cython. 

In the first set of comparisons we see that in both cases, for the Fold example, 4.0.0 is faster than the current master (8173...) as @mmatera  reports. 


However the gap between the 4.0.0 and and master is smaller when either Pyston (0.7 / 0.5 vs 1.8 / 1.1) or Python with Cython  (1.5 / 1.0 vs 1.8 / 1.1) is used.  However note that in both cases Pyston is faster than Cython in running Fold on either master or 4.0.0-benchmark and the discrepancy between the code (cached, versus not) is smaller in the Pyston case.

_This says to me that the interpreter is doing some of the same kinds of caching that the homegrown Python caching is doing._    And pyston is doing a better job at this.  

_I would prefer to let Pyston (or Cython) do its caching rather than have it done in our code._ These interpreters may even improve over time there without wasting our precious effort. What they won't and cannot do is take our code and make it scalable, modular, and understandable.  The kind of efficiencies we get by storing information about a list, or performing customized evaluation based on an Expression type, neither of these interpreters is going to do either and those a probably bigger wins anyway. 

Yes, there is some benefit in adding some sort of lower-level cache here, just not where we should be spending our time right now. 

I suggested in https://github.com/Mathics3/mathics-core/discussions/387#discussioncomment-2985411 that we see if there is any benefit in using Cython when pyston is used. Haven't heard back on that. Most of these comments were drafted while waiting on cython to compile the code for the above tests. It takes something like 2 minutes to Cythonize the small set of files we have marked. And the lifetime of that compilation my only last a few minutes if you happen to be working on one of those files. And it can also cause inconsistencies in running when when a file is out of sync.

PR #398, opened 2022-06-22
Reverse Cython default.

rocky:
Note that for Pyston this slow things down.
For PyPy we can't use it either.

There is benefit in doing this over CPython (probably up to 3.9), but
since it takes minutes to run Cython, for development this often isn't a
big win since it can happen that files you are working on change and so you'll need to
run again, with the 2 minutes again. Or files will stale, sometimes
inadvertently.

So running Cython is worth the effort before building a
wheel (i.e. before release) or using in a docker container.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/398"></a>

### DIFF
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -9,7 +9,6 @@ on:
 jobs:
   build:
     env:
-      NO_CYTHON: 1
       LDFLAGS: "-L/usr/local/opt/llvm@11/lib"
       CPPFLAGS: "-I/usr/local/opt/llvm@11/include"
     runs-on: macos-latest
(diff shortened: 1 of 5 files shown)


mmatera (approved):
LGTM

mmatera:
> Looks like some new problem with installing LLVM 11 via `brew` on MacOSX. This means all OSX builds will fail.
> 
> As far as I can tell, something has changed on the Workflows CI side.

It seems that this does not have anything to do with this PR. I think it is safe to merge it.

mmatera:
> > Looks like some new problem with installing LLVM 11 via `brew` on MacOSX. This means all OSX builds will fail.
> > As far as I can tell, something has changed on the Workflows CI side.
> 
> It seems that this does not have anything to do with this PR. I think it is safe to merge it.

It seems that the issue is already solved. I just needed to rerun the tests by hand.

PR #399, opened 2022-06-22
Add more conversion modules

rocky:
Move out of atom the conversion modules to mathics.core.convert

Conceptually and organizationally this helps me out. I think it also reduces the likelihood of circular imports. Smaller more self-contained modules. In general conversion modules pull in a lot. The basic class types like Expression and the kinds of Atoms should not be pulling in a lot. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/399"></a>


mmatera (approved):
LGTM

PR #401, opened 2022-06-25
Removing conversions from Expression.__init__

mmatera:
and more replacements from `leaves` to `elements`.

The nontrivial changes are essentially those proposed in the first patch in #400.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/401"></a>

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -187,7 +187,7 @@ class Expression(BaseElement, NumericOperators, EvalMixin):
         - *elements - optional: the remaining elements
 
     Keyword Arguments:
-        - element_properties -- properties of the collection of elements
+        - elements_properties -- properties of the collection of elements
     """
 
     head: "Symbol"
@@ -199,9 +199,9 @@ class Expression(BaseElement, NumericOperators, EvalMixin):
     ):
         self.options = None
         self.pattern_sequence = False
-        if isinstance(head, str):
-            # We should fix or convert to to_expression all nonSymbol uses.
-            head = Symbol(head)
+        assert isinstance(head, BaseElement)
+        assert isinstance(elements, tuple)
+        assert all(isinstance(e, BaseElement) for e in elements)
 
         self._head = head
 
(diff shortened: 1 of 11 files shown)


mathics/core/pattern.py
```diff
@@ -355,34 +361,34 @@ def yield_head(head_vars, _):
             # and not OneIdentity &
             # (expression.get_attributes(evaluation.definitions) |
             # expression.get_head().get_attributes(evaluation.definitions)):
-            new_expression = Expression(self.head, expression)
-            for leaf in self.leaves:
-                leaf.match_count = leaf.get_match_count()
-                leaf.candidates = [expression]
-                # leaf.get_match_candidates(
-                #    new_expression.leaves, new_expression, attributes,
+            new_expression = Expression(self.head.expr, expression)
```
mmatera, line 364: Here there is one of the three non trivial changes: `self.head` is a `Pattern` object, but `Expression` needs  a `BaseElement`.

mathics/core/expression.py
```diff
@@ -214,12 +214,27 @@ def __init__(
...
-        # self.elements_properties = elements_properties
+        # self._elements, self.elements_properties = convert_expression_elements(
+        #    elements, lambda x:x
+        # )
+        # assert isinstance(self._elements, tuple)
+        self._elements = elements
+
+        self._build_elements_properties()
+        #
+        # comment mmatera: I found that in certain cases, the elements_properties
```
mmatera, line 225: Here is the reason why test_series.py failed.

TiagoCavalcante (approved):
LGTM

mathics/core/pattern.py
```diff
@@ -131,7 +131,7 @@ def get_elements(self):
         return self.expr.get_elements()
 
     # Compatibily with old code. Deprecated, but remove after a little bit
-    get_leaves = get_elements
+    get_elements = get_elements
```
rocky, line 134: This does nothing.

mathics/core/expression.py
```diff
@@ -199,9 +199,9 @@ def __init__(
     ):
         self.options = None
         self.pattern_sequence = False
-        if isinstance(head, str):
-            # We should fix or convert to to_expression all nonSymbol uses.
-            head = Symbol(head)
+        assert isinstance(head, BaseElement)
```
rocky, line 202: Now that this is the case, we should change the signature in line 197 or so to:

`self, head: BaseElement, ...`

mathics/core/expression.py
```diff
@@ -199,9 +199,9 @@ def __init__(
...
+        assert isinstance(elements, tuple)
```
rocky, line 203: I have been wavering on this this. When you pass `*x` the result is always a tuple. But I guess it doesn't hurt to check.

mmatera: I put it just to check that in all the platforms and versions, we get the right type. We can comment out these checks now.

mathics/core/pattern.py
mmatera: There was too many Ctrl+%... I fixed it now.

mathics/core/expression.py
mmatera: Yep. It seems we gain 25% of the time in the doctests avoiding these asserts...

rocky:
LGTM and Thanks!

PR #402, opened 2022-06-25
Benchmarking+futher improvements fully working

mmatera:
@rocky,  here I added the changes from #401 that make that this branch passes all the tests.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/402"></a>

### DIFF
--- a/mathics/algorithm/optimizers.py
+++ b/mathics/algorithm/optimizers.py
@@ -14,11 +14,10 @@ from mathics.core.atoms import (
     Integer2,
     Integer3,
     Integer10,
-    MachineReal,
     Number,
     Real,
-    from_python,
 )
+from mathics.core.convert.python import from_python
 from mathics.core.evaluation import Evaluation
 from mathics.core.evaluators import apply_N
 from mathics.core.expression import Expression
--- a/mathics/builtin/arithfns/basic.py
+++ b/mathics/builtin/arithfns/basic.py
@@ -31,9 +31,9 @@ from mathics.core.atoms import (
     Rational,
     Real,
     String,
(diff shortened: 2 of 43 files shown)


rocky:
The way I have been handling this is just to rebase off of master. So after #401 goes in, let's do that..

rocky:
When I rebased and ran `pytest -s test/package/test_combinatorica_benched.py` I see a categorical gain in speed. 

https://github.com/Mathics3/mathics-core/commit/ed8c5d5207d61abed8fd0dffa60e0089d3d4cd12 has the details.

I expected a slowdown would happen when fixing bugs though.

mmatera:
> When I rebased and ran `pytest -s test/package/test_combinatorica_benched.py` I see a categorical gain in speed.
> 
> [ed8c5d5](https://github.com/Mathics3/mathics-core/commit/ed8c5d5207d61abed8fd0dffa60e0089d3d4cd12) has the details.
> 
> I expected a slowdown would happen when fixing bugs though.

Actually, we still could get something better if we could use the `expressions_properties` parameter in the constructor. However, it seems that in some places we fail to predict what the internal `_build_elements_properties` would produce.

rocky:
> Actually, we still could get something better if we could use the `expressions_properties` parameter in the constructor. However, it seems that in some places we fail to predict what the internal `_build_elements_properties` would produce.

https://github.com/Mathics3/mathics-core/discussions/400#discussioncomment-3013481 has one place that needs fixing.

mmatera:
Yes, I have an eye on it. But it seems that three are several places with similar problems.

PR #403, opened 2022-06-26
fix build element properties in Expression.__init__ and Expression.rewrite_apply_eval

mmatera:
This PR fixes the issue that screwed the sorting property and made test_series.py fail.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/403"></a>

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -209,8 +209,8 @@ class Expression(BaseElement, NumericOperators, EvalMixin):
         self._head = head
         self._elements = elements
         self.elements_properties = elements_properties
-        if elements_properties is None:
-            self._build_elements_properties()
+        #        if elements_properties is None:
+        #            self._build_elements_properties()
 
         self._sequences = None
         self._cache = None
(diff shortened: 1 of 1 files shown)


mathics/core/expression.py
```diff
@@ -208,20 +208,9 @@ def __init__(
...
-        #                                                           self, (self.elements_properties,
-        #                                                                  elements_properties)
-        #                                                           )
-        # After things are clean we should not run _build_elements_properties. This is done
-        # when needed in self.rewrite_apply_eval()
-        # self.elements_properties = elements_properties
-        self._build_elements_properties()
+        self.elements_properties = elements_properties
+        if elements_properties is None:
+            self._build_elements_properties()
```
rocky, line 213: This is done inside rewrite_apply_eval where it is needed.

mmatera: If I remove lines 212 and 213, all the pytest crashes, probably because we check for elements_properties in another place.
I am investigating it.

mmatera: Now 212 and 213 are not needed anymore.

PR #406, opened 2022-07-10
Misc type fixes and test tweaks

rocky:
most of these were caught by mypy. There are many many more.

test_assumptions.py: move to a more specific location and and remove
improper xfail

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/406"></a>


TiagoCavalcante (approved):
LGTM

mathics/core/element.py
```diff
@@ -348,6 +348,13 @@ def equal2(self, rhs: Any) -> Optional[bool]:
             return self == rhs
         return None
 
+    def evaluate(self, evaluation) -> "BaseElement":
```
mmatera, line 351: I thought we had agree in that this method does not belong to BaseElement, but just to those that are subclasses of `EvalMixin`.

mmatera (approved):
Apart from the comment about restoring the method `evaluate` in BaseElement, LGTM

mathics/core/element.py
rocky: Hmmm. Locally, I have changed this to move this code into `EvalMixin` which I'll commit after local testing.

I get why we have `EvalMixin` but now I am wondering from first principles (rather than code mechanics) what kinds of BaseElements are _not_ legit to run evaluate?

mmatera: > 
> I get why we have `EvalMixin` but now I am wondering from first principles (rather than code mechanics) or from semantics what kinds of BaseElements are _not_ legitimate to run evaluate?

The main reason to have `EvalMixin` was to split and simplify the interface of BaseElement, and its atomic subclasses. 
The typical cases of Elements that shouldn't be evaluated are:

* Numbers
* Strings
* Boxes

In `Expression.evaluation` we can skip the evaluation of those elements just by checking  `hasattr(element, "evaluate")`  or `isinstance(element, EvalMixin)`.

rocky: I guess I wasn't clear. I said I get why `EvalMixin` was added. I don't get from a high-level or semantic reason _why_ you can't evaluate Numbers, Strings, or Boxes. All elements whether they are leaves (in the normal sense of leaf) in an Expression tree logically can be evaluated. The evaluation of a String or a Number is its value - the evaluation parameter is ignored, but that's okay.  For a Box, it is something similar if not the same. 

One can make tests for a class (currently named EvalMixin) independent of whether it has an `evaluation()` method.

The reason Element _should_ have an evaluation method is for conceptual simplicity and uniformity.

mmatera: OK, you are right: any element that is not an `EvalMixin` can be evaluated, and the evaluation returns the same value. The question was if it makes sense to have this method. If the goal is to have the smallest interface on the base elements, probably it doesn't.  For example, if someone is implementing a new Builtin,  do we want to allow that in an apply method something like this?
```
    def apply(self, *elements, evaluation):
    """<Rule>"""
    evaluated_elements = (e.evaluate(evaluation) for e in elements)
    ....
```

or we would like to force that a test is done:

```
    def apply(self, *elements, evaluation):
    """<Rule>"""
    evaluated_elements = (e.evaluate(evaluation) if hasattr(e, "evaluate") else e for e in elements)
    ....
```

rocky: >   or we would like to force that a test is done:

You don't test, you add an `evaluate()` method that returns `self`.

rocky: Yes, I think it belongs to BaseElement. And sameQ probably belongs there as well. 

As I look at the evaluate code I see there is stuff like this in expression `eval_range()`:

```
                        if isinstance(element, EvalMixin):
                            new_value = element.evaluate(evaluation)
                            # We need id() because != by itself is too permissive
                            if id(element) != id(new_value):
                                recompute_properties = True
                                elements[index] = new_value

```

Here it feels what we need is something that indicates that the class _isn't_ Idempotent (rather than it _can_ be evaluated) and this could just be a class attribute. 

However all of this would take a bit more work. So as a transition let us keep EvalMixin but try to move away from it or reduce its need for when we can remove it and we have the same effect as that above code.

rocky: > The main reason to have `EvalMixin` was to split and simplify the interface of BaseElement,

Yes, make as simple as possible, _but no simpler_.  evaluation is a fundamental property of an element.  Stuff like `get_float_value()`,  `get_atoms()`  and a number of other things like that aren't. _Those_ are the parts that need to be removed. However we need more work to clean this up.  We inherited a mess here.

mmatera:
@rocky, in that case, merge this when you think it is ready.

PR #410, opened 2022-07-11
Fix pythonized expressions generated by Compile

mmatera:
`f=Compile[{x}, expr]` tries to produce a compiled (llvm-lite) version of the expression, in a way that `f[.1]` should return
the value of `expr/.x->1`. Internally, it creates a Python callable object, which is used in the implementation of other symbols like `NIntegrate` or `Plot`. When the llvm compiler fails to produce that callable, a *pythonized* (callable) version of the expression is generated. In both cases, the arguments of the callable must be native Python int/float/complex values. 
However, in the current implementation (master), the *pythonized* version does not convert those values into Mathics objects, and then when the callable is called, the evaluation fails.
This PR fixes this issue by adding the corresponding conversion.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -93,7 +93,6 @@ Bugs
 ++++
 
 * ``First``, ``Rest`` and  ``Last`` now handle invalid arguments.
-* ``N`` now handles arbitrary precision numbers when the number of digits is not specified.
 *  ``Set*``: fixed issue #128.
 *  ``SameQ``: comparison with MachinePrecision only needs to be exact within the last bit Issue #148.
 * Fix a bug in ``Simplify`` that produced expressions of the form ``ConditionalExpression[_,{True}]``.
@@ -103,6 +102,8 @@ Bugs
 * Streams used in MathicsOpen are now freed and their file descriptors now released. Issue #326.
 * Some temporary files that were created are now removed from the filesystem. Issue #309.
 * There were a number of small changes/fixes involving ``NIntegrate`` and its Method options. ``Nintegrate`` tests have been expanded.
+* Fix a bug in handling arguments of pythonized expressions, that are produced by ``Compile`` when the llvmlite compiler fails.
+
 
 4.0.1
 -----


rocky:
I see only a change to `CHANGES.rst`. Is that your intent?

mmatera:
> I see only a change to `CHANGES.rst`. Is that your intent?

For some reason, the other changes were not committed when I created the PR. Now they are, together with the corresponding tests.

rocky:
LGTM.   

BTW using llvm instead of a real compiler. We will never get WMA performance this way.

 However I guess like the docstuff we'll have to live with this until the interpreter come more like a normal interpreter and we have time to get around to it.

TiagoCavalcante (approved):
LGTM - I agree that LLVM won't make us reach WMA, but all bugs should be fixed

mmatera:
Actually, this PR was less related to llvm compilation than to be able to convert expressions in Python callables. This is necessary for using numeric libraries like scipy to perform numerical optimization and integration.

mmatera:
What is probably wrong here is that in order to convert another expression into a Python callable, we need to evaluate an expression. Probably this should be encapsulated in a function, (like `apply_N`) and use that function in `Compile.apply`.

mmatera:
> > In[1]:= NIntegrate[BesselJ[0, x], {x, 0, 2}]
> > NIntegrate::mtdfail: The specified method failed to return a number. Falling back into the internal evaluator.
> > Out[1]= NIntegrate[BesselJ[0, x], {x, 0, 2}]
> 
> It would be great to add this as a _pytest_ to be run when scipy is available.
> 
> > Probably this should be encapsulated in a function, (like apply_N) and use that function in Compile.apply.
> 
> Yes, I think this is correct. 

OK. I am going to start a branch with this test and make the conversion from Expressions to callable objects more modular. The question is whether that should or shouldn't be part of the next release.

> Of course, I have been thinking about what it would take to get Mathics performance to where it should be. Virtually all of the more performant interpreters are not tree interpreters like Mathics.[1](#user-content-fn-**-a7ac73b895bfda455d1dfe902c285789)
>

OK, not sure how to make that work, but if you have an idea, go ahead! As we start disentangling the core from the builtins, it becomes more feasible to test different parsing and evaluation schemes, without breaking too much the evaluation logic or the rest of the code.   

 
> That means that eventually we will have to create a set of _instructions_ which to start out you should think of as not-OO methods like `N_apply()` where self is passed as a parameter.
> 

The OO way to define builtins is a nice Pythonic way to define properties and replacement rules for built-in symbols. However,  for "evaluation" functions like `N_apply`, `evaluate`, or `format`, which are highly recursive,  very used functions, I think the best would be to have them as non-member functions.  


> In my view this is _also_ better modularity because the whole class structure for a mathics primitive is ["too clever by a half"](https://www.merriam-webster.com/dictionary/too%20clever%20by%20half).
> 
> Over time the other thing that has been slowly annoying me and grating on me is calling "apply" what everyone else calls "eval". In the past when I have said something is "boneheaded wrong" what i mean is that it is wrong and then proliferated everywhere such as the term "leaf" when "element" or generic node in a graph is intended.
> 

My problem with this is that you probably are right, and the name `apply_` is not the best choice. However, to change it would imply a huge work of changing names, without a clear warranty that the new name is really the better name.  As I understand this, "apply_" methods define (builtin)replacement rules, and do not represent an evaluation by itself. An evaluation requires to check properties of the different symbols involved in an expression, and applying rules associated to these symbols in a certain order.  
So, N.apply_N (as defined now) defines a rule that consists in calling a certain class of evaluation over one argument (the evaluation is now called `mathics.core.evaluators.apply_nvalues`, but maybe should be renamed as `mathics.core.evaluators.eval_n`). Once we converge on how we interpret this, we could plan how to rename the methods.

> It would be awesome if we didn't proliferate the "apply" wrongness. If we ever are to get to the stage where we have an instruction interpreter, it will be more apparent the terminology wrongness. Therefore it might be nice to rename `N_apply` to `N_eval` and any other `X_apply` things to `X_eval`.
> 

With this, I totally agree. 

> I know it seems like a silly thing, but language and precision _do_ matter. It hinders anyone who _is_ familiar with interpreters from wanting to get involved because things are so confusing. Personally, I find I have to mentally say X really means Y.
> 

In this part, I rely on your expertise. From the inner side, the only  interpreter that I had studied its implementation  in deep is Mathics.  

> If Mathics is your first interpreter you study it isn't so irksome. However if you then try to go to any other interpreter, you'll be just as confused because the terminology used in other interpreters will not match.
> ## Footnotes
> 
>     1. David Korn asserts that his tree interpreter ksh is pretty fast, but I think that is only relative to other POSIX shells, not interpreters like Python. [leftwards_arrow_with_hook](#user-content-fnref-**-a7ac73b895bfda455d1dfe902c285789)

rocky:
> OK, not sure how to make that work, but if you have an idea, go ahead! As we start disentangling the core from the builtins, it becomes more feasible to test different parsing and evaluation schemes,

To be precise, parsing currently is fine. It is just evaluation that is a problem. 

> > That means that eventually we will have to create a set of instructions which to start out you should think of as not-OO methods like N_apply() where self is passed as a parameter.

I got this a little wrong n that the method is called `apply_N()`, and it _doesn't_ need a `self` parameter which is also good and that I hope will always be the case. If there is something from "self" that is needed that should be extracted and passed as an explicit parameter. 

> However, to change it would imply a huge work of changing names, without a clear warranty that the new name is really the better name. As I understand this, "apply_" methods define (builtin)replacement rules, and do not represent an evaluation by itself. 

I wasn't suggesting or intending to rename this now. At some point we'll get to the real (as opposed to misnamed) apply step and that's where we should start to thing about revising. My current thought is at that point, we can have a period where we account method names of both forms "apply_" and "eval_".  And this is kind of like what we went through with Expression, and leaves->elements, and some of the documentation stuff. (And some of this is still ongoing). 

> OK. I am going to start a branch with this test and make the conversion from Expressions to callable objects more modular. The question is whether that should or shouldn't be part of the next release.

In my mind _definitely_ the next release. We tend to at the last minute try to push too much disruptive stuff in and that delays things.  However with respect to 

> > eval_N and any other apply_X things to eval_X 

> With this, I totally agree.

This is a pretty simple change. Also `N_apply` was added _this_ release and it is a simple search and replace to fix it.

PR #412, opened 2022-07-12
Make a pass over Stream fns...

rocky:
OutputStream leaving a temporary file for a doctest that didn't even
involve OutputStream. Adjust the doctest so that it now does refer to
OutputStream (in the output that is expected to be shown)

Some other small doc tweaks done too.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/412"></a>


TiagoCavalcante (approved):
LGTM

PR #413, opened 2022-07-12
N_apply -> N_eval when it is not a class method

rocky:


mmatera (approved):
LGTM

TiagoCavalcante (approved):
LGTM

rocky:
> In this like of thougnt, since `apply_*` methods define `BuiltinRule` s, analog to the (replacement) `Rule` s like (`patt->repl`), maybe a more appropriated name would be `repl_*`.

I don't understand or agree. We talk about methods and functions being evaluated all the time. The name of the method is what it does. `"x ".strip()` strips `x`. 

eval_N evaluates N on its arguments. It doesn't replace anything or apply anything. All of that has been done previously.

rocky:
This focus on how we got there as opposed to what the function does is misguided. As we are seeing, in the future there will be other ways to get there other than from apply and rewrite.

mmatera:
Maybe this could help:
https://reference.wolfram.com/language/tutorial/EvaluationOfExpressions.html

rocky:
> Maybe this could help: https://reference.wolfram.com/language/tutorial/EvaluationOfExpressions.html

This is very useful and it is definitely something we need to keep in mind as we implement things. As with most things in their docs they never use the a term in a way that conflicts with conventional usage, although terms may slightly different or less used from historical precedent. 

Note that "evaluation" and "evaluate" are used a lot while "apply" not so much. However when "apply" is used, it is used properly generally to mean that you take a function and and some list data (the arguments) and the act of combining them is the process of "apply"ing.  Note that the word "replace" is rarely used. 

Coming back to our situation, while it is important to follow this at a high level, this document is not about writing an interpreter, and does not suggest how an interpreter implement the higher-level ideas. 

And it is here in this area, when there isn't a corresponding thing mentioned used in WMA that I suggest we use conventional names and ideas rather than someone's misunderstood version. 

There is a well established precedent for names and processes in functional languages and in term-rewriting languages.  Lets both stick to the WMA terms and when that doesn't apply use the common functional and term-rewriting terms.

mmatera:
Another thing that could help :  'evaluate' always produces a new Expression. Apply methods can return (python) None if the rule fails to be applied.

rocky:
> Another thing that could help : 'evaluate' always produces a new Expression. Apply methods can return (python) None if the rule fails to be applied.

You mean "Eval methods"? An eval method returns whatever interface it wants. "eval"rather than "evaluate" was intended to suggest the more common term used in functional programming. I don't see that someone is going to confuse `evaluate()` with `eval_X()`.  The are similar in idea and that is suggested by the "eval" part that they both share. 

But their return types are as you noted are different.  Python has a `randint()` and a `randrange()` and they both work with random numbers, but no one would suggest that they should both return the same type because they both start with `rand`.

PR #414, opened 2022-07-13
Modularize Compile

mmatera:
This PR proposes to split the definition of `Compile.apply` from the function that converts an expression into a Python callable.
Also, a typo that prevented that scipy.optimizers be loaded was fixed.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/414"></a>

### DIFF
--- a/mathics/builtin/compile/__init__.py
+++ b/mathics/builtin/compile/__init__.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 
-
 try:
     import llvmlite
 
@@ -15,5 +14,3 @@ if has_llvmlite:
     from .compile import _compile
     from .base import CompileArg, CompileError
     from .types import *
-
-from .topython import *
--- a/mathics/builtin/compile/__init__.py
+++ b/mathics/builtin/compile/__init__.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 
+
 try:
     import llvmlite
 
@@ -14,3 +15,5 @@ if has_llvmlite:
     from .compile import _compile
     from .base import CompileArg, CompileError
     from .types import *
+
+from .topython import *
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -1347,12 +1347,13 @@ class Expression(BaseElement, NumericOperators, EvalMixin):
 
         n_evaluation = kwargs.get("n_evaluation")
         head = self._head
+        if head is SymbolFunction:
+            from mathics.builtin.compile import expression_to_callable_and_args
+
+            vars, expr_fn = self.elements
+            return expression_to_callable_and_args(expr_fn, vars, n_evaluation)
+
         if n_evaluation is not None:
-            if head is SymbolFunction:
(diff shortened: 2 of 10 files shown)


mathics/builtin/compile/__init__.py
```diff
@@ -14,3 +27,96 @@
...
+# _Complex not implemented
+permitted_types = {
+    Expression(SymbolBlank, SymbolInteger): int_type,
+    Expression(SymbolBlank, SymbolReal): real_type,
+    SymbolTrue: bool_type,
+    SymbolFalse: bool_type,
+}
+
+
+class CompileDuplicateArgName(Exception):
```
rocky, line 41: Please let's not put clases in __init__.py

mathics/core/expression.py
```diff
@@ -998,6 +998,7 @@ def rewrite_apply_eval_step(self, evaluation) -> typing.Tuple["Expression", bool
         # @timeit
         def eval_elements():
             nonlocal recompute_properties
+
```
rocky, line 1001: I think we are having blank wars here. 

I think we needs to set 

```
trim_trailing_whitespace: “true
``` 

in `.editorconfig`.

mathics/builtin/compile/__init__.py
mmatera: OK. I left this here because I was not sure where is the best place to put it: some options I had in mind are

* mathics.builtin.compile.topython
* mathics.core.convert.python
* mathics.core.evaluators
* mathics.core.expression
* Another option?

rocky: Smaller modules would be appreciated. So given though choices above, a new module in `mathics.builtin.compile` is preferable.

mathics/core/expression.py
mmatera: In any case, if I remove the blank line, flake8 complains about it.

rocky:
Would be nice use underscores to separate words so something like `topython` doesn't get read as "top ython". 

But thinking about this more. It looks like all other conversion modules are called "convert", so this  module name seems to fit better in the pattern already established.

test/builtin/test_compile.py
```diff
@@ -44,20 +47,38 @@
...
         assert len(expr.elements) == 3
         code = expr.elements[2]
         assert isinstance(code, CompiledCode)
-        print(code.cfunc)
         y = code.cfunc(x)
+        print(str_expr, code.cfunc)
+        assert abs(y - res) < 1.0e-6
+
+
+def test_pythonize_code_1():
```
rocky, line 69: A very small thing, but what this test seems to be doing is testing builtin functions with symbols in them. The logic by which this is about pythonizing (which isn't really a word) is at best confusing and jarring jargon. 

A suggestion for an alternate name might be `test_builtin_fns_with_symbols`

mmatera: Sure.

rocky:
@mmatera I just had a chance to go over this in detail - Excellent work!  If this is ready to go into master I think it would be a great addition!

mmatera:
Actually, I didn't see that slowdown when I tried this PR. This week I am on hollydays, away from my computer. Next week I can investigate this.

PR #418, opened 2022-07-17
Add vector section and add sort order for chapters/sections

rocky:
This does really two independent things related to going over the docs.

The first commit adds an "Operation on Vectors" section following the organization in https://reference.wolfram.com/language/guide/OperationsOnVectors.html 

(Down the line, I would like to reduce what is in the "miscellaneous list functions" and in "linear algebra".)

The second commit adds a sort-key on how this should appear in the Django and LaTeX docs. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/418"></a>


rocky:
I should note that there is a corresponding change to mathics-django which will fail until something like this gets to master,.

TiagoCavalcante (approved):
LGTM

rocky:
After this branch got merge there seem to be more tests and tests are taking considerably longer. Will be investigating what's up when I get a chance.

PR #420, opened 2022-07-18
Add vector section but revert test (and chapter) sorting for now.

rocky:
This code keeps the vector split from #418 but remove the sorting part.

Something in the sorting part of #418 causes testing slow down  a lot, like several minutes in `Plot[]` routines like `Plot[KelvinKer[x]]` and `Plot[KlevenKei[x]`

I don't believe it is the code in the sorting tests that is a problem, but instead more a latent performance problem in our code base that sorting happens to trigger.  Until this is isolated and fixed, I am reverting this for now.  There will be a separate PR in the future to reinstate this after the problem is more thoroughly isolated. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/420"></a>

### DIFF
--- a/mathics/builtin/__init__.py
+++ b/mathics/builtin/__init__.py
@@ -174,12 +174,13 @@ disable_file_module_names = (
 
 for subdir in (
     "arithfns",
-    "atomic",
     "assignments",
+    "atomic",
     "box",
     "colors",
     "distance",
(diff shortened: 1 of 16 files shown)


TiagoCavalcante (approved):
LGTM

PR #423, opened 2022-07-19
Update UnsameQ to accept any number of arguments

Kalsofar:
Before:
```
In[1]:= 1 =!= 2 =!= 3 =!= 4
Out[1]:= 1 =!= 2 =!= 3 =!= 4
```

After:
```
In[1]:= 1 =!= 2 =!= 3 =!= 4
Out[1]= True
```


rocky:
Nicely done!  To fix the "black" problem run: 

```
$ pip install pre-commit
$ pre-commit install
```

(I need to update developer docs to mention this.)

Some other things that might be updated are the `CHANGES.rst` file and `test/test_compare.py` should be moved into `test/builtin/test_comparison.py` after some relative paths are changed. However I can handle this if you can't or don't want to.

Kalsofar:
Let me fix the black issue and also move `test/test_compare.py` as you suggested.

rocky:
Thanks for doing this!

PR #424, opened 2022-07-20
Add BernoulliB

rocky:
and generalize `_MPMathFunction` `nargs` to allow a set of acceptable arguments

@TiagoCavalcante every time I encounter the attribute bitset it feels to me that those should be capitalized to make it more clear that this is a constant and a bitmask. 

If it okay to make the change throughout, I'll do that in a separate PR.  

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/424"></a>

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,3 +1,6 @@
+# THis is an EditorConfig file
+# https://EditorConfig.org
+
 root = true
 
 [*]
@@ -10,8 +13,15 @@ indent_size = 4
 [*.yml]
 indent_style = space
 indent_size = 2
+end_of_line = lf
+insert_final_newline = true
 
 [*.py]
 indent_style = space
 indent_size = 4
-trim_trailing_whitespace: true
+end_of_line = lf
+insert_final_newline = true
+
+# Tab indentation (no size specified)
+[Makefile]
+indent_style = tab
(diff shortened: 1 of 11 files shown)


TiagoCavalcante:
@rocky I'm not sure, constants should be uppercase, attributes are constants, but they aren't used as if they were, as their values don't matter.
Attributes are used more as a "pipeable magic", making them uppercase makes it more clear that they are constants, but what they are isn't relevant as long as they "just work".
I would prefer style over ... uppercase here, but you have way more experience in Python than me, so what you think it's the better, I agree.

rocky:
> @rocky I'm not sure, constants should be uppercase, attributes are constants, but they aren't used as if they were, as their values don't matter. Attributes are used more as a "pipeable magic", making them uppercase makes it more clear that they are constants, but what they are isn't relevant as long as they "just work". I would prefer style over ... uppercase here, but you have way more experience in Python than me, so what you think it's the better, I agree.

Sor precedent for uppercase in Python see the section around https://docs.python.org/3/library/os.html?#os.O_RDONLY where above it says: 

> The following constants are options for the flags parameter to the [open()](https://docs.python.org/3/library/os.html?#os.open) function. They can be combined using the bitwise OR operator |.

These values in fact come from C convention and the `O_` probably was selected to mean "option". 

Another example is https://docs.python.org/3/library/stat.htmlstat.S_ISUID  where it says before: 

> The following flags can also be used in the mode argument of [os.chmod()](https://docs.python.org/3/library/os.html#os.chmod):The variables below define the flags used in the [ST_MODE]

Again this is a direct copy of how it works in C where the `S_` prefix probably stands for "stat" the function that uses this.

So this reminds me that to follow this properly we should add `A_`  The intent here was that it makes it clear that all of these constants or bitmasks are related. 

As for implementing, I think I'll start off by just continuing along this line.  Perhaps adding `A_`. By doing this, it can be done pretty much incrementally, and then at the end a change is made to attributes and we remove the `xx as XX`.

PR #428, opened 2022-07-21
Create ByteArray section and populate it.

rocky:
Also, make more explicit where "check_requires" comes from.
It was imported the wrong way, and we don't do this kind of things
here. Implicit imports make sense only at the top level.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/428"></a>


rocky:
Merging in and continuing, but we'll keep the branch open for review. Function attributes may need adjusting on the new functions.

PR #449, opened 2022-07-24
extract from LoadModule as an eval method and ...

rocky:
apply_nvalues -> eval_nvalues

Some attribute renaming and isort imports done as well.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/449"></a>


rocky:
Should also mention that `eval_load_module()` is used in another branch I haven't pushed where we can run doctests and extract documentation from pymathics modules.

mmatera:
LGTM

PR #451, opened 2022-07-24
using tuple as sort key

mmatera:
This is a first attempt to reimplement the sort keys as tuples.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/451"></a>

### DIFF
--- a/mathics/builtin/drawing/image.py
+++ b/mathics/builtin/drawing/image.py
@@ -2079,8 +2079,8 @@ class Image(Atom):
     def default_format(self, evaluation, form):
         return "-Image-"
 
-    # FIXME: return type should be a specific kind of Tuple, not a list.
-    def get_sort_key(self, pattern_sort=False) -> list:
+    # FIXME: return type should be a specific kind of Tuple, not a tuple.
+    def get_sort_key(self, pattern_sort=False) -> tuple:
         if pattern_sort:
             # If pattern_sort=True, returns the sort key that matches to an Atom.
             return super(Image, self).get_sort_key(True)
@@ -2089,7 +2089,7 @@ class Image(Atom):
             # but with a `2` instead of `1` in the 5th position,
             # and adding two extra fields: the length in the 5th position,
             # and a hash in the 6th place.
-            return [1, 3, SymbolImage, tuple(), 2, len(self.pixels), hash(self)]
+            return tuple((1, 3, SymbolImage, tuple(), 2, len(self.pixels), hash(self)))
 
     def sameQ(self, other) -> bool:
         """Mathics SameQ"""
--- a/mathics/builtin/drawing/image.py
+++ b/mathics/builtin/drawing/image.py
@@ -2079,7 +2079,6 @@ class Image(Atom):
     def default_format(self, evaluation, form):
         return "-Image-"
 
-    # FIXME: return type should be a specific kind of Tuple, not a tuple.
     def get_sort_key(self, pattern_sort=False) -> tuple:
         if pattern_sort:
             # If pattern_sort=True, returns the sort key that matches to an Atom.
@@ -2089,7 +2088,7 @@ class Image(Atom):
             # but with a `2` instead of `1` in the 5th position,
             # and adding two extra fields: the length in the 5th position,
             # and a hash in the 6th place.
-            return tuple((1, 3, SymbolImage, tuple(), 2, len(self.pixels), hash(self)))
+            return (1, 3, SymbolImage, tuple(), 2, len(self.pixels), hash(self))
 
     def sameQ(self, other) -> bool:
         """Mathics SameQ"""
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -729,45 +729,42 @@ class Expression(BaseElement, NumericOperators, EvalMixin):
                 else:
                     pattern += 20
             if pattern > 0:
-                return tuple(
-                    (
-                        2,
-                        pattern,
-                        1,
-                        1,
-                        0,
-                        head.get_sort_key(True),
-                        tuple(element.get_sort_key(True) for element in self._elements),
-                        1,
-                    )
+                return (
+                    2,
+                    pattern,
+                    1,
+                    1,
+                    0,
+                    head.get_sort_key(True),
+                    tuple(element.get_sort_key(True) for element in self._elements),
+                    1,
                 )
-
             if head is SymbolPatternTest:
                 if len(self._elements) != 2:
-                    return tuple((3, 0, 0, 0, 0, head, self._elements, 1))
+                    return (3, 0, 0, 0, 0, head, self._elements, 1)
                 sub = list(self._elements[0].get_sort_key(True))
                 sub[2] = 0
                 return tuple(sub)
             elif head is SymbolCondition:
                 if len(self._elements) != 2:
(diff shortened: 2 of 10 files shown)


mmatera:
@rocky,

Notice that I implemented this as `tuple` instead of a specific kind of `Tuple`, because different types of elements involve different kinds and numbers of elements.

mathics/core/expression.py
```diff
@@ -729,43 +729,45 @@ def get_sort_key(self, pattern_sort=False) -> list:
...
+                        1,
+                    )
+                )
 
             if head is SymbolPatternTest:
                 if len(self._elements) != 2:
-                    return [3, 0, 0, 0, 0, head, self._elements, 1]
-                sub = self._elements[0].get_sort_key(True)
+                    return tuple((3, 0, 0, 0, 0, head, self._elements, 1))
+                sub = list(self._elements[0].get_sort_key(True))
```
mmatera, line 748: This is probably the reason for using `list` instead of `tuple`: Here the return value is obtained from calling the method on one of the elements, and then modifying it. Probably there is also a better way to implement the same...

rocky:
> @rocky,
> 
> Notice that I implemented this as `tuple` instead of a specific kind of `Tuple`, because different types of elements involve different kinds and numbers of elements.

If I have it right, the key property behind selection of the datatype used was that the comparison operators are defined and have a multi-level property. In other words, it was desired that there be something like 5 or 6 categories that form a hierarchy to be used in comparison. 

For example (and I don't know if I have the details I have correct) but the gives the idea: 

If both objects are expressions, then go to some deeper level of comparison. However if one object is an atom or numeric expression, then it (the atom or numeric expression) comes first. If both objects are a numeric expression, then go to some deeper level of comparison, but if one is an atom and another that comes first. if both objects are atoms, then ... 

Or in other words if we wanted to compare an atom to an expression we might have: 

```
atom, numeric expression, expression, object-specific-item 
(0   1                    1           5)                        <   (1, 1, 0, 2) 

```

Just because of the first position in the tuple. Here 0 means the element is of that type and 1 means it is not of the type,

Okay, with this out of the way, we can now get to the aspect of length of tuple and whether it can be fixed in length or not. 

Assuming we always use positive integers (and I don't see a reason why that can't be the case), we can always pad a tuple with zeros and it has no effect. 

But why bother to do this? 

Well, because what we have here is a [named tuple](https://docs.python.org/3/library/typing.html) which as the term suggests is a tuple you can give specific meanings to positions, like x-coordinate, y-coordinate, z-coordinate, time-coordinate.

For us position 0 is the "is an atom" position. position 1 is the "a an numeric expression" position and so on.

mmatera:
> ```
> atom, numeric expression, expression, object-specific-item 
> (0   1                    1           5)                        <   (1, 1, 0, 2) 
> ```
> 
> Just because of the first position in the tuple. Here 0 means the element is of that type and 1 means it is not of the type,
> 

Yep

> Okay, with this out of the way, we can now get to the aspect of length of tuple and whether it can be fixed in length or not.
> 
> Assuming we always use positive integers (and I don't see a reason why that can't be the case), we can always pad a tuple with zeros and it has no effect.

In the case of expressions, there are elements of the tuple that are `BaseElement`, or lists of `BaseElement`, so they are not all integer numbers. 

> 
> But why bother to do this?
> 
> Well, because what we have here is a [named tuple](https://docs.python.org/3/library/typing.html) which as the term suggests is a tuple you can give specific meanings to positions, like x-coordinate, y-coordinate, z-coordinate, time-coordinate.
> 
> For us position 0 is the "is an atom" position. position 1 is the "is a numeric expression" position and so on.

I tried to add some descriptions about what is the meaning of each field. However, depending on the kind of object, the number and kind of fields are different, so I think you are right in that this is a NamedTuple, but I am not sure that it can be implemented using just integers: For Expressions, the tuples represent a kind of tree structure, in a way that if two expressions have the same head,  a (recursive) comparison over the elements is performed. 

In any case, it should be a less homebrew way to implement this kind of sorting over tree structures...

rocky:
> I tried to add some descriptions about what is the meaning of each field. However, depending on the kind of object, the number and kind of fields are different, so I think you are right in that this is a NamedTuple, but I am not sure that it can be implemented using just integers: For Expressions, the tuples represent a kind of tree structure, in a way that if two expressions have the same head, a (recursive) comparison over the elements is performed.
> 
> In any case, it should be a less homebrew way to implement this kind of sorting over tree structures...

Just leave it like it is on this PR and we'll come back and document and generalize later. I think we'll need more OO classes.

mathics/builtin/drawing/image.py
```diff
@@ -2079,8 +2079,8 @@ def do_copy(self):
     def default_format(self, evaluation, form):
         return "-Image-"
 
-    # FIXME: return type should be a specific kind of Tuple, not a list.
-    def get_sort_key(self, pattern_sort=False) -> list:
+    # FIXME: return type should be a specific kind of Tuple, not a tuple.
```
rocky, line 2082: This comment can be removed. It was just a placeholder for the line below.

mathics/builtin/drawing/image.py
```diff
@@ -2089,7 +2089,7 @@ def get_sort_key(self, pattern_sort=False) -> list:
             # but with a `2` instead of `1` in the 5th position,
             # and adding two extra fields: the length in the 5th position,
             # and a hash in the 6th place.
-            return [1, 3, SymbolImage, tuple(), 2, len(self.pixels), hash(self)]
+            return tuple((1, 3, SymbolImage, tuple(), 2, len(self.pixels), hash(self)))
```
rocky, line 2092: Here and many other places the `tuple()` is superfluous. Given that the return type now is listed as `tuple`, in my opinion that is sufficient. `dict({1: 2})` in my view isn't any clearer than `{1: 2}`.

mathics/core/expression.py
rocky: This kind of behavior where you want the 0th element sort key and modify the portion of based on `head` should be wrapped inside a function.

mathics/builtin/drawing/image.py
mmatera: OK, I removed the `tuple(...)`. I left the `dict(...)` because it was there before. In a subsequent round, I can remove them.

mathics/core/expression.py
mmatera: @rocky, I am not sure how to implement what you suggested. In any case, in a following round, I would like to reimplement all this chuck of code, by specializing the class for Patterns.

mmatera:
@rocky, I think this is ready to merge. However, this would produce an incompatibility with all the other packages. Maybe we should wait until after the release to merge it.

rocky:
> @rocky, I think this is ready to merge. However, this would produce an incompatibility with all the other packages. Maybe we should wait until after the release to merge it.

I am in the process of updating natlang and graph and it doesn't seem like this kind of change to both is a big deal. 

 The reason the next release goes to 5.0.0 is because the API not compatible from 4.0.0, so from that standpoint it good to get all of the API breaking changes out of the way so we have the possibility of doing 5.0.1 or 5.1.0 releases because the API isn't incompatible like this one. 

That said, I am always for not delaying releases when there is work that will delay things. Here though I think it is just a matter of committing what has already been done.

PR #452, opened 2022-07-24
Yet another pass over describing "get_sort_key()"

rocky:
It so happens `get_sort_key()` also comes up in the pymathics graph module I have been working to get fully converted. 

I spent a little more time then to try to understand this, and describe it.


mmatera (approved):
LGTM

rocky:
There will of course be more passes and this isn't perfect. But it is a start.

PR #453, opened 2022-07-25
Add pymathics to docs

rocky:
This doesn't fully work but it is a step in the right direction.

The context for pymathics modules should have the Pymathics module name, e.g. `Pymathics``Graph` or `Pymathics``Natlang`` rather than simply `Pymathics`.

But this messes up both testing and interactive use.

@mmatera if you can move this forward that would be good. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/453"></a>

### DIFF
--- a/mathics/core/definitions.py
+++ b/mathics/core/definitions.py
@@ -1,11 +1,12 @@
 # -*- coding: utf-8 -*-
 
-import pickle
 
-import os
 import base64
-import re
 import bisect
+import inspect
+import os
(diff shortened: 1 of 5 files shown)


mmatera (approved):
LGTM

rocky:
While this code has many problems remaining, the number of changes is already quite large. So we'll merge and iterate. 
I hope we can make more progress on this.

PR #459, opened 2022-07-25
Remove crash when name not found.

rocky:
Expand a little and simplify some.

Note that this would make an idea pymathics package

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/459"></a>

### DIFF
--- a/mathics/builtin/atomic/strings.py
+++ b/mathics/builtin/atomic/strings.py
@@ -59,32 +59,54 @@ _regex_shortest = {
 }
 
 
+# A better thing to do would be to write a pymathics module that
+# covers all of the variations. Here we just give some minimal basics
+
+# Data taken from:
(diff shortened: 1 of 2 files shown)


mathics/builtin/atomic/strings.py
```diff
@@ -479,8 +504,8 @@ class Alphabet(Builtin):
 
     def apply(self, alpha, evaluation):
         """Alphabet[alpha_String]"""
-        alphakey = alpha.get_string_value()
-        alphakey = alphabet_alias[alphakey]
+        alphakey = alpha.value
+        alphakey = alphabet_alias.get(alphakey, alphakey)
```
mmatera, line 508: Do the same in line 574.

rocky: Was done in 258346e

Issue #332, opened 2022-06-03
Invalid syntax  result from Integrate command.


nasser1:
Using 4.0 on Linux, I get the following

>mathics

```
Mathics 4.0.0
on CPython 3.10.4 (main, Apr  2 2022, 09:04:19) [GCC 11.2.0]
using SymPy 1.10.1, mpmath 1.2.1, numpy 1.22.4

Copyright (C) 2011-2021 The Mathics Team.
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under certain conditions.
See the documentation for the full license.

Quit by evaluating Quit[] or by pressing CONTROL-D.

In[1]:= Integrate[(1 - x^3)^(2/3)/x,x]
Out[1]= --1 ^ (2 / 3) x ^ 2 Gamma[-2 / 3] hyper[{-2 / 3, -2 / 3}, {1 / 3}, 1 / x ^ 3] / (3 Gamma[1 / 3])

```

Notice the `--1` output. This is an error. Mathematica says

`PreDecrement::rvalue: 1 is not a variable with a value, so its value cannot be changed.`

Btw, as a side issue, what is `hyper[` function above? There is no such function in Mathematica.

But the main problem is generating `--1`

rocky:
> Btw, as a side issue, what is hyper[ function above? 

It probably is SymPy's [hyper()](https://docs.sympy.org/latest/modules/functions/special.html?highlight=hyper#sympy.functions.special.hyper.hyper) and probably appears because there is no Mathematica equivalent. So this is a translation from SymPy results into the Wolfram Language. In that regard to your other bug https://github.com/mathics/Mathics/issues/1591 which is also related to conversion from SymPy into the Wolfram Language.

Personally, I don't think I will be looking at this for a while. However if you are interested in digging deeper you might check to see what would happen if you wrote this in SymPy and then compare the results.

DUOLabs333:
This is because Mathics removes the parentheses from the result. Here is what sympy returns: `-(-1)**(2/3)*x**2*gamma(-2/3)*hyper((-2/3, -2/3), (1/3,), x**(-3))/(3*gamma(1/3))`.

nasser1:
> removes the parentheses from the result

Thanks for finding this.  Removing parentheses from result is not good thing to do. Hopefully someone familiar with Mathics code could fix this.

DUOLabs333:
[...]

PR #460, opened 2022-07-26
Add parentheses around negative numbers in formatting output in infix notation - issue 332

mmatera:
This PR fixes formatting  issue in ``Power`` that was behind the one reported in #332.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/460"></a>

### DIFF
--- a/mathics/builtin/inout.py
+++ b/mathics/builtin/inout.py
@@ -209,12 +209,11 @@ def parenthesize(precedence, element, element_boxes, when_equal):
         element = element.elements[0]
 
     if element.has_form(("Infix", "Prefix", "Postfix"), 3, None):
-        element_prec = element.elements[2].get_int_value()
+        element_prec = element.elements[2].value
     elif element.has_form("PrecedenceForm", 2):
-        element_prec = element.elements[1].get_int_value()
+        element_prec = element.elements[1].value
     # For negative values, ensure that the element_precedence is at least the precedence. (Fixes #332)
     elif isinstance(element, (Integer, Real)) and element.value < 0:
-        print("element is a signed number")
         element_prec = precedence
     else:
         element_prec = builtins_precedence.get(element.get_head_name())
--- a/mathics/builtin/inout.py
+++ b/mathics/builtin/inout.py
@@ -206,10 +206,14 @@ def parenthesize(precedence, element, element_boxes, when_equal):
 
     while element.has_form("HoldForm", 1):
         element = element.elements[0]
+
     if element.has_form(("Infix", "Prefix", "Postfix"), 3, None):
         element_prec = element.elements[2].get_int_value()
     elif element.has_form("PrecedenceForm", 2):
         element_prec = element.elements[1].get_int_value()
+    # For negative values, ensure that the element_precedence is at least the precedence. (Fixes #332)
+    elif isinstance(element, Integer) and element.value < 0:
+        element_prec = precedence
     else:
         element_prec = builtins_precedence.get(element.get_head_name())
     if precedence is not None and element_prec is not None:
--- a/mathics/builtin/inout.py
+++ b/mathics/builtin/inout.py
@@ -27,12 +27,13 @@ from mathics.builtin.lists import list_boxes
 from mathics.builtin.options import options_to_rules
 
 from mathics.core.atoms import (
-    String,
-    StringFromPython,
     Integer,
+    Number,
     Real,
     MachineReal,
     PrecisionReal,
+    String,
+    StringFromPython,
 )
 from mathics.core.element import EvalMixin
 from mathics.core.expression import Expression, BoxError
@@ -212,7 +213,8 @@ def parenthesize(precedence, element, element_boxes, when_equal):
     elif element.has_form("PrecedenceForm", 2):
         element_prec = element.elements[1].get_int_value()
     # For negative values, ensure that the element_precedence is at least the precedence. (Fixes #332)
-    elif isinstance(element, Integer) and element.value < 0:
+    elif isinstance(element, (Integer, Real)) and element.value < 0:
+        print("element is a signed number")
         element_prec = precedence
     else:
         element_prec = builtins_precedence.get(element.get_head_name())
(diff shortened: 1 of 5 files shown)


mathics/builtin/inout.py
```diff
@@ -206,10 +206,14 @@ def parenthesize(precedence, element, element_boxes, when_equal):
...
     while element.has_form("HoldForm", 1):
         element = element.elements[0]
+
     if element.has_form(("Infix", "Prefix", "Postfix"), 3, None):
         element_prec = element.elements[2].get_int_value()
     elif element.has_form("PrecedenceForm", 2):
         element_prec = element.elements[1].get_int_value()
+    # For negative values, ensure that the element_precedence is at least the precedence. (Fixes #332)
+    elif isinstance(element, Integer) and element.value < 0:
+        element_prec = precedence
```
TiagoCavalcante, line 216: I'm not sure about this, this really seems a bug in parser that is passed forward and fixed somewhere later. @rocky @mmatera ?

mmatera: @TiagoCavalcante, is not a problem in the parser, but in the way expressions involving Power are formatted:
`Power[a,b]` is formatted as `a^b`, while `Power[a c,b]` is formatted as `(a c)^b`. However, negative numbers must be handled different: `Power[-1, b]` should be formatted as `(-1)^b`, but in master, it is formatted as a with any other positive number  `-1^b`. This is what this PR is fixing.

TiagoCavalcante: Now I understand, thanks for the explanation!

TiagoCavalcante (approved):
LGTM

mathics/builtin/inout.py
rocky: But is this simply limited to Integers? Wouldn't there be the same problem with any sort of _numeric_ literal, like -5.3 ? 

The other thing I wonder about is under exactly which conditions it is needed. It is needed for Infix of which StandardForm and other Forms follow. But plain Infix is above in line 209 or 210 and this doesn't hit that. 

(The get_int_value()'s could be simplified.) 

I suspect we are also going to have to make another pass over this code because we "canonicalize" or order Polynomials the opposite order, but fixing that is going to cause lots of changes in the expected output behavior so that will have to be deferred.

mathics/builtin/inout.py
```diff
@@ -212,7 +213,8 @@ def parenthesize(precedence, element, element_boxes, when_equal):
     elif element.has_form("PrecedenceForm", 2):
         element_prec = element.elements[1].get_int_value()
     # For negative values, ensure that the element_precedence is at least the precedence. (Fixes #332)
-    elif isinstance(element, Integer) and element.value < 0:
+    elif isinstance(element, (Integer, Real)) and element.value < 0:
+        print("element is a signed number")
```
rocky, line 217: Stray print.

mathics/builtin/inout.py
```diff
@@ -212,7 +213,8 @@ def parenthesize(precedence, element, element_boxes, when_equal):
     elif element.has_form("PrecedenceForm", 2):
         element_prec = element.elements[1].get_int_value()
     # For negative values, ensure that the element_precedence is at least the precedence. (Fixes #332)
-    elif isinstance(element, Integer) and element.value < 0:
+    elif isinstance(element, (Integer, Real)) and element.value < 0:
```
rocky, line 216: And Complex literal?

mmatera: ``Complex`` and ``Rational`` seems to be handled properly without any change:
```
In[1]:= (2+3 I)^q
Out[1]= (2 + 3 I) ^ q

In[2]:= (2/3)^q
Out[2]= (2 / 3) ^ q
```

mathics/builtin/inout.py
mmatera: > But is this simply limited to Integers? Wouldn't there be the same problem with any sort of _numeric_ literal, like -5.3 ?
> 

Thanks for noting this!

> The other thing I wonder about is under exactly which conditions it is needed. It is needed for Infix of which StandardForm and other Forms follow. But plain Infix is above in line 209 or 210 and this doesn't hit that.
> 

I have to check it...

> (The get_int_value()'s could be simplified.)
> 

I changed this in this part of the code, but it appears in many places of the module. I am going to hold off changing the other cases until we face the formatting refactor.

> I suspect we are also going to have to make another pass over this code because we "canonicalize" or order Polynomials the opposite order, but fixing that is going to cause lots of changes in the expected output behavior so that will have to be deferred.

Yep. This is another scale.

rocky:
LGTM - was more involved than initially thought. Thanks for seeing this through thoug.

mmatera:
> LGTM - was more involved than initially thought. Thanks for seeing this through thoug.

OK. I am going to fix the `Infix` case in another round, to avoid making this too long.

rocky:
> > LGTM - was more involved than initially thought. Thanks for seeing this through thoug.
> 
> OK. I am going to fix the `Infix` case in another round, to avoid making this too long.

I had checked this before and unless I got something wrong, I didn't see a problem with Infix after the correction:

Old: 

```
In[1]:= 1*(-1)^n // Infox
Out[1]:= --1 ^ n
```

New: 

```
In[1]:= -1*(-1)^n // Infix
Out[1]= -(-1) ^ n
```

mmatera:
> > > LGTM - was more involved than initially thought. Thanks for seeing this through though.
> > 
> > 
> > OK. I am going to fix the `Infix` case in another round, to avoid making this too long.
> 
> I had checked this before and unless I got something wrong (which is possible based on the specifics of the code) I didn't see a problem with Infix after the correction:
> 
> Old:
> 
> ```
> In[1]:= -1*(-1)^n // Infix
> Out[1]:= --1 ^ n
> ```
> 
> New:
> 
> ```
> In[1]:= -1*(-1)^n // Infix
> Out[1]= -(-1) ^ n
> ```

I got a problem for custom infix expressions:

```
In[1]:= Infix[F[-1,-2,2]]
Out[1]= -1~F~ -2 ~F~2
```
that should be

```
Out[1]= (-1)~F~(-2)~F~2
```

But this could be delayed for the next round...

PR #462, opened 2022-07-26
Collect subexpressions.

### LINKED ISSUE #285: Problem with Collect[ , , ]
The Collect[ , , ]  command doesn't work well.

code:

Solve[Q[x,y]==1/6(x+y+1/x+1/y+y/x+x/y)(Q[x,y]-Q[x,0]-Q[0,y]+Q[0,0])+1/2(x+y)Q[0,0]+(1/3 x+1/6(y+1/x+y/x+1))(Q[x,0]-Q[0,0]) + (1/3 y+1/6(x+1/y+x/y+1))(Q[0,y]-Q[0,0]),Q[x,y]];
%[[1,1,2]]; 
Coefficient[%,Q[0,y]]//Factor//TeXForm
Collect[%%,Q[_,_],Factor]//TeXForm
(*I just solve an equation and first check the coefficient of Q[0,y], and then write it down*)

 and it gives:

-\frac{y \left(-1+x\right) \left(1+y\right)}{x+x^2+y-6 x y+x^2 y+y^2+x y^2}

\frac{Q\left[x,0\right] \left(-1-x-y\right) \left(-x-y+2 x y\right)}{x+x^2+y-6 x y+x^2 y+y^2+x y^2}

(*The coefficient of Q[0,y] exists, but Q[0,y] does not appear in the Collect[, ,] expression. Actually it doesn't work well for Collect[%%,Q[x,0],Factor] as well. gives some misterious results*)

## Your Environment
Mathics 4.0.0

mmatera:
This PR fixes #285.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/462"></a>


mathics/builtin/numbers/algebra.py
```diff
@@ -621,6 +621,11 @@ def powers_list(pf):
                         SymbolPlus, *[c[i] for c in contrib]
                     ).evaluate(evaluation)
                 return powers
+            else:
+                for i, pat in enumerate(var_pats):
+                    if match(pf, pat, evaluation):
+                        powers[i] = Integer(1)
```
TiagoCavalcante, line 627: Integer1

TiagoCavalcante (approved):
LGTM

Kalsofar:
I *think* it also needs to handle this case:
```
In[1]:= Collect[ q[0, x] q[0, y] + q[0, x], q[0, x] ]
Out[1]= q[0, x] + q[0, x] q[0, y]
```

Expected output:
```
In[1]:= Collect[ q[0, x] q[0, y] + q[0, x], q[0, x] ]
Out[1]= q[0, x] (1 + q[0, y])
```

mathics/builtin/numbers/algebra.py
rocky: Personally I'd appreciate a comment here  explaining that we are using matching terms raised to the power 1 to make sure when the term is multiplied it isn't forgotten, and maybe a little example. 

A simple example reduced from  @Kalsofar's can be obtained by the zero parameter, e.g. 

```
Collect[ q[x] q[y] + q[[x],  q[x] ]
```

rocky:
> I _think_ it also needs to handle this case:
> 
> ```
> In[1]:= Collect[ q[0, x] q[0, y] + q[0, x], q[0, x] ]
> Out[1]= q[0, x] + q[0, x] q[0, y]
> ```
> 
> Expected output:
> 
> ```
> In[1]:= Collect[ q[0, x] q[0, y] + q[0, x], q[0, x] ]
> Out[1]= q[0, x] (1 + q[0, y])
> ```

I just tried and it doesn't "Collect" or simplify. We produce: 

```
q[0, x] + q[0, x] q[0, y]
``` 

However, without this change we produce an incorrect result: 

```
q[0, x] + q[0, y]
```

mmatera:
> > I _think_ it also needs to handle this case:
> > ```
> > In[1]:= Collect[ q[0, x] q[0, y] + q[0, x], q[0, x] ]
> > Out[1]= q[0, x] + q[0, x] q[0, y]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Expected output:
> > ```
> > In[1]:= Collect[ q[0, x] q[0, y] + q[0, x], q[0, x] ]
> > Out[1]= q[0, x] (1 + q[0, y])
> > ```
> 
> I just tried and it doesn't "Collect" or simplify. We produce:
> 
> ```
> q[0, x] + q[0, x] q[0, y]
> ```
> 
> However, without this change we produce an incorrect result:
> 
> ```
> q[0, x] + q[0, y]
> ```

This was another bug in the code. I think I fixed it now.

mathics/builtin/numbers/algebra.py
```diff
@@ -587,21 +601,24 @@ def coeff_power_internal(self, expr, var_exprs, filt, evaluation, form="expr"):
             var_pats = [Pattern.create(var) for var in var_exprs]
 
         # ###### Auxiliary functions #########
-        def key_powers(lst):
+        def key_powers(lst: list) -> int:
```
rocky, line 604: I think we can return a float, right? So it would be `Union[int, float]`, I think.

mathics/builtin/numbers/algebra.py
```diff
@@ -587,21 +601,24 @@ def coeff_power_internal(self, expr, var_exprs, filt, evaluation, form="expr"):
...
             key = Expression(SymbolPlus, *lst)
             key = key.evaluate(evaluation)
             if key.is_numeric(evaluation):
                 return key.to_python()
```
rocky, line 608: Or maybe a comment or an assertion here if this is supposed to always be an integer.

mmatera: Indeed, it could be a float...

rocky:
LGTM

PR #463, opened 2022-07-26
Add `from_bool()`, explicit System` in "systemsymbols" module

rocky:
and some small doc changes.

I have given up on the hope of getting pymathics`graphics working fully before a 5.0.0 release probably this weekend. 

networkx has changed a bit since two years ago when I last looked at this (probably for the better). And the thought of coming out with a new release that depends on an outdated networkx isn't appealing either. 

However, to the extent I can,  changes to mathics-core that assist in simplifying that I'd like to get in. These changes were motivated by that. 

Also, although I'd like to get pymathics namespacing fixed, I have given up on that too. 

 Pymathics modules should not be living in the same top-level namespace. We should have Pymathics'Natlang'XXX and Pymathics'Graph'XXX rather than having the both going to Pymathics'XXX and possibly clash.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/463"></a>


mmatera (approved):
LGTM

PR #470, opened 2022-07-27
Elliptic Integral related Doc fixes

rocky:
See also previous commit 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/470"></a>


rocky:
Side note: I note: I saw that I opened this issue from about a year ago https://github.com/mathics/Mathics/issues/1539

I thought it would be nice to address something I saw back then.

rocky:
In these, often I matched on the number of parameters when there were 1 to 3 argument forms. I don't know if Mathematica gives an error or just returns the same thing when the number of parameters doesn't match. 

As with the Attributes settings, this is something that we should check.

TiagoCavalcante (approved):
LGTM

PR #473, opened 2022-07-28
Allow text link in URL

rocky:
This PR really has to separable things. The first part is to allow text links in url. The lack of this I found annoying and it is a direct result of a homegrown implementation that is hard to understand as well as being incomplete.

Because there are substitutions going on various places, I couldn't use either double quote which gets substituted in HTML for &quote) or single quote which is interpreted as code. Vertical bar is used in regexps and escaping that makes this ugly re even more ugly. So I used colon. 

The second aspect of this PR has to do with the builtins. The thought is on each builtin start of with the corresponding Wikipedia name if one exists with this name linked to the corresponding wiki page, and then give Sympy (when appropriate) and WMA links. 

Again we are hamstrung by limitations of the homegrown system in that this all has to appear on one line in the docstring.  And then there is the problem that list items aren't handled properly, which is why this is in parenthesis.

I spent probably too much time trying find something that would render reasonably well in Django and LaTeX. 
Screenshots of examples are below. The PDF rendering in chrome doesn't seem to show the links as links. In Django these appear as blue.

AiryAi-Tex
AiryAi-Django

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/473"></a>

I would love to figure out how to get rid of the two column format in LaTeX which is inappropriate for built-in function descriptions.

### DIFF
--- a/mathics/builtin/specialfns/elliptic.py
+++ b/mathics/builtin/specialfns/elliptic.py
@@ -24,8 +24,8 @@ class EllipticE(SympyFunction):
       <dt>'EllipticE[$m$]'
       <dd>computes the complete elliptic integral $E$($m$).
 
-      <dt>'EllipticE[ϕ|$m$]'
-      <dd>computes the complete elliptic integral of the second kind $E$($m$|$ϕ$).
+      <dt>'EllipticE[phi|$m$]'
+      <dd>computes the complete elliptic integral of the second kind $E$($m$|phi).
     </dl>
 
     Elliptic curves give Pi / 2 when evaluated at zero:
@@ -132,7 +132,7 @@ class EllipticK(SympyFunction):
 
     def apply_default(self, args, evaluation):
         "%(name)s[args___]"
-        evaluation.message("EllipticK", "argx", Integer(len(m.elements)))
+        evaluation.message("EllipticK", "argx", Integer(len(args.elements)))
 
     def apply(self, m, evaluation):
         "%(name)s[m_]"
--- a/mathics/builtin/specialfns/elliptic.py
+++ b/mathics/builtin/specialfns/elliptic.py
@@ -1,9 +1,8 @@
 """
-Elliptic Integrals
+<url>:Elliptic Integrals: https://en.wikipedia.org/wiki/Elliptic_integral</url>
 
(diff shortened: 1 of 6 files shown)


mathics/builtin/specialfns/elliptic.py
```diff
@@ -24,8 +24,8 @@ class EllipticE(SympyFunction):
       <dt>'EllipticE[$m$]'
       <dd>computes the complete elliptic integral $E$($m$).
 
-      <dt>'EllipticE[ϕ|$m$]'
-      <dd>computes the complete elliptic integral of the second kind $E$($m$|$ϕ$).
+      <dt>'EllipticE[phi|$m$]'
+      <dd>computes the complete elliptic integral of the second kind $E$($m$|phi).
```
TiagoCavalcante, line 28: `$phi$`?

TiagoCavalcante (approved):
LGTM - we could use `<a href="link">...</a>`?

mmatera:
I like this. However, maybe we should start thinking about migrating Builtin docstrings to a more standard format...

rocky:
> LGTM - we could use `<a href="link">...</a>`?

No because in our screwy system, on HTML the quote is replaced to `&quote` or something like that. And the complicated re gloms "em" with "url" for some inexplicable reason.  I feel like I am dancing around a turd mine. 

If you or others want to figure out how to get this done, by all means go ahead! 

And fix the other screwy parts like the extra spacing that a "li "tag causes and figure out how to get it to ignore newlines for reflowed text.  And figure how to get it do deal with mathematical text and symbols

At some point we will need to make a plan to ditch this and migrate to something more standard.

rocky:
> I like this. However, maybe we should start thinking about migrating Builtin docstrings to a more standard format...

I couldn't agree more!  But I think fixing up performance (which probably means making this look more like a traditional interpreter which will also fix some of the scoping bugs) and making this more scalable and modular has to come first. 

Alas this is going to take a while.

mathics/builtin/specialfns/elliptic.py
rocky: Ah - ok. But in my mind what is really needed is this junk to be able to understand Math mode. 

The unicode symbols were removed when I notice that we can't handle this in building the LaTeX doc. Sigh.

mmatera:
> 
> I would love to figure out how to get rid of the two column format in LaTeX which is inappropriate for built-in function descriptions.

adding `\singlecolumn` in the parameters of `\documentclass` ?

PR #482, opened 2022-08-01
rebase decoupling_format

mmatera:
As promised, just after the last 5.0.0 release, here we go again with the proposal for decoupling the formatting methods.

## Some background.

The goal of this PR is to simplify the interface of `BaseElement` and derivative classes, by removing the code associated to build the different formatted expressions. A side benefit would be also to reduce a little bit the number of circular imports.

The doubt that I still have is if `String` and `_BoxedString` should be the same class, contained in a specific module (let's say, ``mathics.core.string``). This could help to reduce more the number of circular references and conversions, but also would make the interface for String more complicated. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/482"></a>


mathics/builtin/box/graphics.py
```diff
@@ -1066,7 +1070,9 @@ def init(
             if len(item.elements) not in (1, 2, 3):
                 raise BoxConstructError
             content = item.elements[0]
-            self.content = content.format(graphics.evaluation, "TraditionalForm")
+            self.content = format_element(
+                content, graphics.evaluation, "TraditionalForm"
```
rocky, line 1074: One thing I noticed in general looking things over is that "format_element" or ".format", really should be taking a Symbol parameter as we did for Expression.  If I recall correctly, when I looked at the code, the first thing it does is call Symbol on the `str` parameter. 

This can come later though. But I thought I would point it out.

mathics/core/evaluation.py
```diff
@@ -450,13 +451,15 @@ def format_output(self, expr, format=None):
...
         elif format == "xml":
-            result = Expression(SymbolStandardForm, expr).format(
-                self, "System`MathMLForm"
+            result = format_element(
+                Expression("StandardForm", self, expr), "System`MathMLForm"
             )
         elif format == "tex":
-            result = Expression(SymbolStandardForm, expr).format(self, "System`TeXForm")
+            result = format_element(
+                Expression("StandardForm", self, expr), "System`TeXForm"
```
rocky, line 461: This no longer works.

mathics/builtin/box/inout.py
```diff
@@ -23,13 +25,13 @@
 SymbolMakeBoxes = Symbol("System`MakeBoxes")
```
rocky, line 25: Now is defined in systemsymbols

mathics/core/formatter.py
```diff
@@ -1,7 +1,43 @@
...
+
+SymbolComplex = Symbol("Complex")
+SymbolDivide = Symbol("Divide")
+SymbolI = Symbol("I")
+SymbolMinus = Symbol("Minus")
+SymbolPlus = Symbol("Plus")
+SymbolRational = Symbol("Rational")
+SymbolTimes = Symbol("Times")
+SymbolString = Symbol("String")
+
```
rocky, line 39: Most of these are now defined elsewhere.

mathics/core/formatter.py
```diff
@@ -119,3 +155,391 @@ def add_conversion_fn(cls, module_fn_name=None) -> None:
...
+            # show_string_characters = False
+
+            if show_string_characters:
+                return render(r'\text{"%s"}', text[1:-1], in_text=True)
+            else:
+                return render(r"\text{%s}", text[1:-1], in_text=True)
+        elif text and text[0] in "0123456789-.":
+            return render("%s", text)
+        else:
+            # FIXME: this should be done in a better way.
```
rocky, line 274: I believe we now have everything in place from the Mathics-Scanner repository to be able to do the single-character or operator entries in a table driven fashion. 

However this is still a bit wrong since it should be done when extracting text from an operator.

mathics/core/formatter.py
```diff
@@ -119,3 +155,391 @@ def add_conversion_fn(cls, module_fn_name=None) -> None:
...
+
+    def to_expression(self) -> String:
+        return String(self.value)
+
+
+element_formatters = {}
+
+
+# Not sure if this belongs here (in the core side) or to the builtin side.
+# Let's try to put this here
```
rocky, line 326: I think this is a fine place for now.

rocky:
Hmm. It looks like mathicsscript and mathics-django will have to change for this to work. If I understand semantic versioning correctly this will mean 6.0.0 since this is incompatible.

We could probably add the short term a method on BaseElement and BoxElement `.format()` method to make the conversion and then this can be a 5.1.0 release.  

Even though we change the API quickly in two successive releases, I would rather do this than have help up the last release even longer and have even more massive changes. I hope we can spend the time before this goes out to get this roughly right.

rocky:
> Hmm. It looks like mathicsscript and mathics-django will have to change for this to work. If I understand semantic versioning correctly this will mean 6.0.0 since this is incompatible.
> 
> We could probably add the short term a method on BaseElement and BoxElement `.format()` method to make the conversion and then this can be a 5.1.0 release.

I was thinking about this more. Having a `.format()` method for object wasn't a bad thing. What was bad was having to add methods for all of the classes _inside_ each of the classes rather than as part of a module for a particular format.  See [Aspect-Oriented Programming](https://en.wikipedia.org/wiki/Aspect-oriented_programming)

Adding a method on the base class I think is a good thing and the way to go. And it more closely maintains API compatibility, were it not for the fact that the form should be a `Symbol` rather than a `str`. 

But in the short term we can do the same as we did for `Expression` and allow either a `str` or a `Symbol` and do the checking inside the `.format()` routine.  Then eventually we'll remove the check.

mathics/core/formatter.py
```diff
@@ -509,7 +510,9 @@ def do_format_complex(element, evaluation, form) -> "Expression":
     return do_format_expression(Expression(SymbolHoldForm, result), evaluation, form)
 
 
-def do_format_expression(element, evaluation, form):
+def do_format_expression(
+    element: BaseElement, evaluation: Evaluation, form: Symbol
+) -> BaseElement:
     # not sure how much useful is this format_cache
```
rocky, line 516: Let's simply for now then and if we find it needed, then add caching back in via one of the more conventional Python methods like 2526d0a29c7e28ccae99c4cb2ae0bedef22c1a6f

mmatera:
> Hmm. It looks like mathicsscript and mathics-django will have to change for this to work. If I understand semantic versioning correctly this will mean 6.0.0 since this is incompatible.
> 
> We could probably add the short term a method on BaseElement and BoxElement `.format()` method to make the conversion and then this can be a 5.1.0 release.
> 
> Even though we change the API quickly in two successive releases, I would rather do this than have help up the last release even longer and have even more massive changes. I hope we can spend the time before this goes out to get this roughly right.

OK, now I had added a method in ``BaseExpression``, which calls the stand-alone function.

mathics/core/formatter.py
```diff
@@ -119,3 +151,386 @@ def add_conversion_fn(cls, module_fn_name=None) -> None:
...
+        }
+        self.box_options.update(options)
+
+    def __repr__(self):
+        return self.value
+
+    def __str__(self):
+        return self.value
+
+    def boxes_to_text(self, **options):
```
rocky, line 180: I wonder if we could put all out all of the `to_text()`, methods here and in other modules and create that in a  `mathics.format.text` module similar for what we have done for JSON, Asymptote, and SVG. Likewise for `MathML`, and `TeX`. 

Having all of the format-related conversions in one module allows adding another kind of backend-format module easier because we can inspect other similar modules and adapt those.  Personally, I have found it helpful that all of the asymptote routines are in one module rather than each of the classes that could potentially produce asymptote output. 

What do you think?

mathics/core/formatter.py
mmatera: OK, that could work. The question is if we gain something by caching this kind of functions...

rocky: With respect to this question, what I noticed on the Expression cache where specific examples of things were given is that by reducing unnecessary conversions and evaluations and adding other specific targeted and simply done caching such as we do for sympy or mpmath functions, reduces the need and benefit of this homegrown, hard-to-understand and potentially buggy (in that it may be memory leak forming) caching mechanism. 

There is an should be a lesson here: if you don't describe and document what you are doing, someone may come along in the future and remove it or replace it with something simpler. So when we add caches and stuff like this, we also need to describe _why_ with specific examples and tests. 

Currently we are or should be benchmarking performance as we go a long. When there is a regression, we should undertake to fix that.

mathics/core/formatter.py
```diff
@@ -119,3 +151,386 @@ def add_conversion_fn(cls, module_fn_name=None) -> None:
...
+            if include_form:
+                expr = Expression(form, expr)
+            return expr
+        # Repeated and RepeatedNull confuse the formatter,
+        # so we need to hardlink their format rules:
+        if head is SymbolRepeated:
+            if len(elements) == 1:
+                return Expression(
+                    SymbolHoldForm,
+                    Expression(
```
rocky, line 385: This is not a valid Expression. Use `Integer() `and `String()` to convert arguments or use `to_expression()`

mathics/core/formatter.py
```diff
@@ -119,3 +151,386 @@ def add_conversion_fn(cls, module_fn_name=None) -> None:
...
+                        SymbolPostfix,
+                        Expression(SymbolList, elements[0]),
```
rocky, line 387: This is not a valid Expression. Use Integer() and String() to convert arguments or use to_expression()

mathics/core/formatter.py
```diff
@@ -119,3 +151,386 @@ def add_conversion_fn(cls, module_fn_name=None) -> None:
...
+            else:
+                return Expression(SymbolHoldForm, expr)
+        elif head is SymbolRepeatedNull:
+            if len(elements) == 1:
+                return Expression(
+                    SymbolHoldForm,
+                    Expression(
+                        SymbolPostfix,
+                        Expression(SymbolList, elements[0]),
+                        "...",
```
rocky, line 401: Same here.

mathics/core/formatter.py
```diff
@@ -119,3 +151,386 @@ def add_conversion_fn(cls, module_fn_name=None) -> None:
...
+            form,
+        )
+
+    parts: typing.List[Any] = []
+    if element.is_machine_precision() or not element.real.is_zero:
+        parts.append(element.real)
+    if element.imag.sameQ(Integer(1)):
+        parts.append(SymbolI)
+    else:
+        parts.append(Expression(SymbolTimes, element.imag, SymbolI))
```
rocky, line 503: Just an idle curiosity, wonder if we can use that fancy infix notation stuff like `element.imag * SymbollI`.  I have noticed other places that are kind of like this too.  It is just something I noticed and wondered about. Of course, this isn't wrong.

test/format/test_format.py
```diff
@@ -702,7 +704,7 @@ def test_makeboxes_text(str_expr, str_expected, form, msg):
     @pytest.mark.xfail
     def test_makeboxes_tex_fragile(str_expr, str_expected, form, msg):
         result = session.evaluate(str_expr)
-        format_result = result.format(session.evaluation, form)
+        format_result = format_element(result, session.evaluation, Symbol(form))
```
rocky, line 707: Do we still need this? It might be good to leave at least one or two `result.format()` around since in my view this is the way I think it will be typically used.

rocky:
I have been thinking about this some more. I think this is pretty close for merging or could be merged and we might be able to get it done for a 5.0.1 release. But what I'd like to see is due diligence in benchmarking just to check that things haven't gotten worse in terms of performance.  

And then I would definitely like to see more done such as getting all of the conversion code for TeX in a single module and MathML in a single module.

mmatera:
>  get it done for a 5.0.1 release. But what I'd like to see is due diligence in benchmarking just to check that things haven't gotten worse in terms of performance.

I was thinking about how to benchmark this. As a very roughly way to measure the impact in performance, I have to say that these changes didn´t have a measurable impact in the time taken by the doctests.

Most of the advantages in performances came before the release, when we start to use the new BoxExpression instead of `Expression` to store formatted expressions. Since these objects become inert in evaluation, and since something formatted can not be a very long and nested expression, I am not sure how much performance is going to be gained/lost here: I can check how much it takes to format a (let's say) 100x100 array, but it is not going to be a significant fraction of the time required to build (evaluate) its elements. On the other hand, I think it does not make too much sense to consider how much time is required to built larger or complicated expressions (let's say, 10^6 times 10^6 arrays) because in the end, we can not show them on screen.

> And then, later, I would definitely like to see more done such as getting all of the conversion code for TeX in a single module and MathML in a single module.
>
> My own personal thing (not for a 5.0.1 release but soonish) is to start using MathicsScanner tables instead of that hacky string substitution that is used here.

I think that these are the subsequent steps. 









> I have been thinking about this some more. I think this is pretty close for merging or could be merged and we might be able to get it done for a 5.0.1 release. But what I'd like to see is due diligence in benchmarking just to check that things haven't gotten worse in terms of performance.
> 
> And then, later, I would definitely like to see more done such as getting all of the conversion code for TeX in a single module and MathML in a single module.
> 
> My own personal thing (not for a 5.0.1 release but soonish) is to start using MathicsScanner tables instead of that hacky string substitution that is used here.

rocky:
@mmatera The due diligence in benchmarking was two fold: If we were going to include a PR with this functionality (and no more) then we should do something even if not that specific.  The other part though which is more in terms of looking to do more is that we could use something more specific to this aspect to establish a baseline. (We could also do that on 5.0.0 and 4.1.0). 

Do you want to try to get this into the 5.0.1 release for this weekend or leave it for some other release?

There are some small things for this PR, most notably the invalid Expression which also should be a ListExpression. (Yes,, I know that bug was in there from before).

rocky:
I should have mentioned that with respect to _size_ of PRs, this is about the right size and doing more that changes more would in my opinion be better done in a follow-on PR.

mathics/core/element.py
```diff
@@ -369,41 +245,14 @@ def equal2(self, rhs: Any) -> Optional[bool]:
...
-        Applies formats associated to the expression, and then calls Makeboxes
-        """
-        from mathics.core.symbols import (
-            Symbol,
-            SymbolMakeBoxes,
-        )
-        from mathics.core.atoms import String
-        from mathics.builtin.base import BoxExpression
-        from mathics.builtin.box.inout import _BoxedString
+    # Deprecated: use mathics.core.formatter.format_element
```
rocky, line 248: Is using `element.format()` deprecated?

mathics/core/systemsymbols.py
```diff
@@ -70,7 +70,6 @@
 SymbolFailed = Symbol("System`$Failed")
 SymbolFloor = Symbol("System`Floor")
 SymbolFormat = Symbol("System`Format")
-SymbolFullForm = Symbol("System`FullForm")
```
rocky, line 73: Why was this removed here? Removing it makes the current mathicsscript break because this is currently where it expects it.

I do see it also defined in `mathics/core/symbols.py` but it is not _required_ to be there. And current thinking is that we should be explicit about adding the context "System" if that is what we mean rather because it is more clear, rather than assuming the context is "System" when the symbol is defined as it is in `mathics/core/symbols.py`

mathics/core/element.py
mmatera: At the beginning of this, I thought so. Now, I am not sure.

mmatera:
> @mmatera The due diligence in benchmarking was two fold: If we were going to include a PR with this functionality (and no more) then we should do something even if not that specific. The other part though which is more in terms of looking to do more is that we could use something more specific to this aspect to establish a baseline. (We could also do that on 5.0.0 and 4.1.0).

OK, but the point is that I am not sure about what would be a good benchmark for this. I am trying to write something basic checking tables, nested expressions, and plots, but I didn't finish yet. Any suggestions?

mathics/core/systemsymbols.py
mmatera: Actually, these symbols appear in `mathics.core.symbols` because the formatting was inside the module, and required them. Now probably can be removed from there and moved to mathics.core.systemsymbols.

mathics/core/formatter.py
mmatera: Thanks for noticing this!

mathics/core/systemsymbols.py
rocky: Please move. Thanks.

mathics/core/element.py
rocky: I'd say remove the comment.

mathics/core/formatter.py
mmatera: This way of doing it (first building a `Times` expression, and then asking for formatting it, instead of building an Infix notation) gives some flexibility in defining different ways to represent a product in different formats. For example, the expression `Times[a,I]`, in `InputForm` is represented as  `a*I`, but in `StandardForm` is `a I`. This behavior can also be modified by the user inside a session, so I guess that this is the right way to implement it.

mathics/core/systemsymbols.py
mmatera: Done here: [2bca2aa](https://github.com/Mathics3/mathics-core/pull/482/commits/2bca2aaac08de5b20f510ded69bcb8bb733e29dc)

mathics/core/formatter.py
mmatera: done here [d546404](https://github.com/Mathics3/mathics-core/pull/482/commits/d54640411bde3a03c43e7fff86c5f1bbbc489a99)

mathics/core/formatter.py
mmatera: Done here [d546404](https://github.com/Mathics3/mathics-core/pull/482/commits/d54640411bde3a03c43e7fff86c5f1bbbc489a99)

test/format/test_format.py
mmatera: The aim of removing `format` from the interface of ``Element`` was to make it simpler, and avoid circular dependences. 

Conceptually, also I think that "formatting" is one of the possible many things you could do with an Expression. ("Analytical evaluation", "Numerical evaluation", or "Compilation" are other possible things). For this reason, I think that any of these things should be necessarily attached to the objects as methods. On the other hand, it does not harm to have them as methods if they are just called at a very high level. For example, having a complicated expression `expr`, it is fine to have 
a method `format`, that calls `format_element`, which internally works recursively. What I would like to avoid is having calls to the `format` method inside `format_element`, due that in some builtins that method is called. @rocky, makes this sense for you?

mmatera:
Here there are a couple a simple benchmarks, against master and 4.0.0.

1: `Table[a^i/b^j,{i,10},{j,10}]`
2: `Table[a^a^a^a^a^a^a,{i,10},{j,10}]`
3: `Integrate[Integrate[F[x,y],{x,c,d}],{y,a,b}]`
4: `Plot[Sin[x],{x,0,10.}]`

These are the benchmarks for formatting (the evaluation was performed before the benchmark, so in the case of the plot,
what I am timing is to format the `GraphicsBox` that was already generated in evaluation.)
imagen

(times are in ms)

So, in performance, this branch does not introduce a measurable gain/loss in performance regarding master. On the other hand, regarding 4.0.0 is a definitive win.

Code:

```
import mathics
print(mathics.version.__version__)

session=MathicsSession()


expr1=session.evaluate("Table[a^i/b^j,{i,10},{j,10}]")
expr2=session.evaluate("Table[a^a^a^a^a^a^a,{i,10},{j,10}]")
expr3=session.evaluate("Integrate[Integrate[F[x,y],{x,c,d}],{y,a,b}]")
expr4=session.evaluate("Plot[Sin[x],{x,0,10.}]")


print(mathics.version.__version__)

print("\n\n## text")


%timeit expr1.format(session.evaluation, "OutputForm").boxes_to_text()
%timeit expr2.format(session.evaluation, "OutputForm").boxes_to_text()
%timeit expr3.format(session.evaluation, "OutputForm").boxes_to_text()
%timeit expr4.format(session.evaluation, "OutputForm").boxes_to_text()

print("\n\n## mathml")

%timeit expr1.format(session.evaluation, "OutputForm").boxes_to_mathml()
%timeit expr2.format(session.evaluation, "OutputForm").boxes_to_mathml()
%timeit expr3.format(session.evaluation, "OutputForm").boxes_to_mathml()
%timeit expr4.format(session.evaluation, "OutputForm").boxes_to_mathml()

print("\n\n## tex")

%timeit expr1.format(session.evaluation, "OutputForm").boxes_to_tex()
%timeit expr2.format(session.evaluation, "OutputForm").boxes_to_tex()
%timeit expr3.format(session.evaluation, "OutputForm").boxes_to_tex()
%timeit expr4.format(session.evaluation, "OutputForm").boxes_to_tex()
```

mathics/core/formatter.py
rocky: I am not sure I understand because the infix notation is supposed to do the same thing: it is just syntactic sugar. 

But no matter - never was a big deal either way.

test/format/test_format.py
rocky: I agree with what you write. To clarify a bit further. Wen we have something like `eval_N`, the top-level function is needed down the line because in an interpreter you wan a list of operations or instructions that are run, not methods in a class.  

In the unit test though, the purpose should be to test both forms of access, especially the method access because that's what will be used more often. 

In the code itself, it is easier to use the method from because that is "natural" and saves an import. It trades method lookup on an object for a import instruction. (Modern JITing pythons are supposed to drive down the cost of the method lookup.

Some other small things. Why is this test marked to fail? 

Personally, I would have iterated over Symbols rather than call the Symbol function inside the loop.  But these are minor things.

rocky:
LGTM  -  I noted some small stuff mostly in test_format.py but you have been a good sport about all of the other little quibbles. 

Thanks for doing the benchmark. We will drive down time needed for Table and Plot by numerous other things that we have yet to do.

test/format/test_format.py
mmatera: > Some other small things. Why is this test marked to fail?

See the comment at the beginning of the test [/test/format/test_format.py L23](https://github.com/Mathics3/mathics-core/blob/30a5d875770e972304ac5bc410554d226907937c/test/format/test_format.py#L23). I can extend this comment in another PR if it is needed.


> 
> Personally, I would have iterated over Symbols rather than call the Symbol function inside the loop. But these are minor things.

OK

mathics/core/formatter.py
mmatera: The question is: how often do we call more than once a formatting function over the same expression? My guess is that almost never. This is why removing this cache didn´t have any impact on performance.

PR #491, opened 2022-08-05
Split off hyperbolic functions from trig fns..

rocky:
Add Gudermannian from autoload Guidermannian.m

Exponential functions should also be split from Trig functions. And both gone over. But that is for some other time. 

All of this could have the URLs gone over.

This is going to be the last of the url and doc improvements before 5.0.1

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/491"></a>

### DIFF
--- a/mathics/builtin/numbers/hyperbolic.py
+++ b/mathics/builtin/numbers/hyperbolic.py
@@ -241,7 +241,7 @@ class Coth(_MPMathFunction):
 
 class Gudermannian(Builtin):
     """
-    <url>:Gudermannian function: https://en.wikipedia.org/wiki/Gudermannian_function</url> (<url>:WMA: https://reference.wolfram.com/language/ref/Indeterminate.html<url></url>)
+    <url>:Gudermannian function: https://en.wikipedia.org/wiki/Gudermannian_function</url> (<url>:WMA: https://reference.wolfram.com/language/ref/Gudermannian.html</url>, <url>:MathWorld: https://mathworld.wolfram.com/Gudermannian.html</url>)
     <dl>
       <dt>'Gudermannian[$z$]'
        <dd>returns the Gudermannian function $gd$($z$).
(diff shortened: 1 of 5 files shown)


mathics/builtin/numbers/hyperbolic.py
```diff
@@ -0,0 +1,348 @@
...
+    rules = {
+        "ArcCsch[0]": "ComplexInfinity",
+        "ArcCsch[0.]": "ComplexInfinity",
+        "Derivative[1][ArcCsch]": "-1 / (Sqrt[1+1/#^2] * #^2) &",
+    }
+
+    summary_text = "inverse hyperbolic cosecant function"
+    sympy_name = ""
+
+    def to_sympy(self, expr, **kwargs) -> Optional[SympyExpression]:
```
TiagoCavalcante, line 106: Why this is necessary here (and in the next built-in) and not in the other?

TiagoCavalcante (approved):
LGTM

mathics/builtin/numbers/hyperbolic.py
rocky: I don't understand. Please be more specific.

TiagoCavalcante: I mean, why ArcCoth don't need to_sympy?

rocky: ArcCoth is implemented as purely term rewriting. There is no evaluation method. Hey - if we can't fix the backwards function naming convention because it is pervasive, let's please agree _not_ to call them "apply" methods but "evaluation" or "eval" methods since that is what they are and that is what the rest of the informed world calls it.

rocky: The above is not quite correct - sorry. `ArcCoth` uses the generic SymPy evaluation method and that presumably handles `to_sympy`, for ArcCsc someone decided to code in a transformation rule as python code right before passing it off to sympy.

TiagoCavalcante: That makes sense - thanks.

PR #492, opened 2022-08-05
tydyng up inout module

mmatera:
This PR implements the changes proposed in https://github.com/Mathics3/mathics-core/discussions/317#discussioncomment-2865937


mathics/builtin/box/inout.py
```diff
@@ -1,9 +1,13 @@
 # -*- coding: utf-8 -*-
 
-from mathics.builtin.base import BoxExpression
+"""
+Low level notation structures.
```
rocky, line 4: I don't understand what this means. I would appreciate a long and as detailed as you can give. I don't mind care about English grammar, write in Spanish if that helps. What I am interested in is are the idea about what you think is or should be going on in this module.

mathics/core/formatter.py
```diff
@@ -157,6 +167,215 @@ def add_conversion_fn(cls, module_fn_name=None) -> None:
...
+            _BoxedString(base, number_as_text=True),
+            _BoxedString(pexp, number_as_text=True),
+            options,
+        )
+    else:
+        return method(_BoxedString(s), _BoxedString(base), _BoxedString(pexp), options)
+
+
 #
 #   Comment from Rocky:
```
rocky, line 380: As far as I can tell, this comment is (now?) no longer near anything that seems relevant to what it is saying. For example, I don't see where we test for str-ness and convert to Symbol. I think I saw that somewhere the last time I saw this comment.

rocky: Ah. Maybe it is supposed to go before `format_element()` ?

rocky: Below, `_BoxedString()` doesn't belong here. In fact it is the only class where everything else is a function

The boxing class belongs in the box module like all of the others boxing routines. And I don't think the box module should be under `builtin`. 

Here is a suggested reorganization.  Under core, we have a `form` (both `convert` and `form` are intended to be imperative verbs not nouns as we did in summary text, as in: you (we, they, or I) convert, you (or we, they, or I) `form`. 

`parser` is a noun, but if things were done from scratch I would have suggested `parse` instead of `parser`.  If you think it worth doing, then please do. 

Under format we have "box" (also a verb), and  low-level or final format (as opposed to the top level form). Right now this is low-level format is not under core. But I think it should be moved into core.

The core module then has parse, evaluate (not currently a module though) and form modules. The form module has box and format modules.

 If you wanted to write a backend to handle some new `Form` you  would model it on the core form module. 

What do you think?

mathics/builtin/box/inout.py
mmatera: This module contains almost the same that we found in https://reference.wolfram.com/language/guide/LowLevelNotebookStructure.html. I didn´t go beyond on the description because what I have in mind is probably not accurate enough to be useful. What is true is that all these structures are "Builtin symbols" which are the heads of formatted (boxed) expressions.

mathics/core/formatter.py
rocky: Also thoughts about putting `form` up  at the same level as `core` and `builtin` (replacing `format` which gets pushed down a level into `form`)?

mmatera: @rocky, actually, all this PR was just about following the idea in https://github.com/Mathics3/mathics-core/discussions/317#discussioncomment-2865937 about following WR  to split and reorganize `inout.py`, and move as far as possible all the very low-level formatting code to `mathics.core`.  What I am trying to do now (in a private branch) is to try figuring out how all these parts are working now, and how they should really work. Once I get that, it is going to make more sense to write more extensive documentation. If you prefer, I can put this as a draft, or just remove the PR. However, as always that we try to split and reorganize code at this level, we arein the risk of making the effort useless because of merge conflicts.

mathics/builtin/box/inout.py
rocky: Ah. Thanks - I now understand.

Here is a rewording of the information there adjusted in a way that is more with the focus of something _implementing_ tor trying to understand and implementation as opposed to a user sees things at the "what does this do for me" level (used in Notebooks). 

_Formatting constructs are represented as a hierarchy of low-level symbolic "boxes"._ 

_The routines here assist in boxing at the bottom of the hierarchy. At the other end, the top level, we have a Notebook which is just a collection of Expressions usually contained in boxes._

mathics/core/formatter.py
rocky: Thanks for communicating how you are thinking about this.  It helps me a great deal and I appreciate the effort here. 

Let's make this a draft.  And continue privately if that helps.

Just so you know, there were two things that in the back of my mind I am holding off on right now, but would like to see as part of going over the Form/Box/Formatter reworking. 

The first is to get the for function operators to use MathicsScanner tables and use the unicode Unicode  symbol from those tables when that is appropriate. Likewise for (ams)TeX symbols. This has been something we've both kind of known about for a long time, and I have just been putting it off for when we get around to Form and formatting work which is sometime in this phase (although it doesn't have to be right now and cut be done later).

The second thing that i wanted to do was write [DiscretePlot](https://reference.wolfram.com/language/ref/DiscretePlot.html) which seems pretty simple to do. 

However since that also has an arm in formatting where I expect things to get moved around, I was holding off on that as well.

For me, I think the next non-cleanup thing to do is to redo how `Part` works so that it uses `ListExpression` and works in a more functional way.

mathics/builtin/box/inout.py
rocky: Given what i now understand is in `inout`,  I think that is not a great name.  Something like "base" or "core" would make more sense.

mathics/builtin/layout.py
```diff
@@ -0,0 +1,846 @@
...
+
+SymbolNumberForm = Symbol("System`NumberForm")
+SymbolSuperscriptBox = Symbol("System`SuperscriptBox")
+SymbolSubscriptBox = Symbol("System`SubscriptBox")
+
+
+class Row(Builtin):
+    """
+    <dl>
+    <dt>'Row[{$expr$, ...}]'
```
rocky, line 57: I would appreciate it if you'd add some spaces, like two when nesting XML tags. Not doing this was yet another one of those bone-headed weirdnesses of the original code.

mathics/builtin/layout.py
```diff
@@ -0,0 +1,846 @@
...
+                ),
+            ),
+            *options_to_rules(options),
+        )
+
+
+SymbolTableDepth = Symbol("TableDepth")
+
+
+class TableForm(Builtin):
```
rocky, line 122: WIth respect to XXXForm. https://reference.wolfram.com/search/?q=Form lists a little under 6K forms 

If we are aspire to do something coming anywhere close to that. We can't have an organization like this where we put in one module a bunch of individual forms. Probably more like it would be a module to hold various kinds of forms would be a start. And a straightforward way in which we and other can add more Forms.

mmatera: At this moment, what I am trying to do is to disentangle the interface of the `BaseElement` subclasses from the `MakeBoxes` loop, without producing another thing highly entangled with the core. Once we have that, we could add more formats as pymathics pluggins.

rocky: That makes sense. Just keep it in mind, though.

mathics/core/formatter.py
mmatera: > 
> Just so you know, there were two things that in the back of my mind I am holding off on right now, but would like to see as part of going over the Form/Box/Formatter reworking.
> 
> The first is to get the for function operators to use MathicsScanner tables and use the Unicode symbols from those tables when that is appropriate. Likewise for (ams)TeX symbols. This has been something we've both kind of known about for a long time, and I have just been putting it off for when we get around to Form and formatting work which is sometime in this phase (although it doesn't have to be right now and can be done later).
> 


> The second thing that i wanted to do was write [DiscretePlot](https://reference.wolfram.com/language/ref/DiscretePlot.html) which seems pretty simple to do.
> 

How is this related to the formatting refactor?

> However since that also has an arm in formatting where I expect things to get moved around, I was holding off on that as well.
> 
> For me, I think the next non-cleanup thing to do is to redo how `Part` works so that it uses `ListExpression` and works in a more functional way.

Same question.

mathics/builtin/layout.py
rocky: And also note that we will have to provide the basic set of Forms. And even for that, this we should not try to list all of the built-in Forms in one module (mixed with other stuff). 

Should also mention with respect to pymathics that it's namespacing scoping should be fixed up.  And down the line we will probably want a real "autoload" like GNU Emacs has.

mathics/core/formatter.py
rocky: `Part` has nothing to do with formatting refactor. It was just a heads up. (You have a tendency to work scattered so declaring what I intend to do I am hoping make sure we don't collide)

As for `DiscretePlot` I looked again given where this is going I see that it isn't related at all. I don't remember the last time I looked (before release) why I thought it was likely to be an areas where there was going to be a lot of motion.

mathics/builtin/layout.py
mmatera: This is a boring thing that could be automatized at some point. Also, as soon as I finish with this refactor, I would like to start refactoring the documentation using a more canonical system, like sphinx. So, I am not sure how useful is to do this now.

mathics/core/formatter.py
mmatera: > Below, `_BoxedString()` doesn't belong here. In fact it is the only class where everything else is a function
> 
> The boxing class belongs in the box module like all of the others boxing routines. And I don't think the box module should be under `builtin`.
> 
> Here is a suggested reorganization. Under core, we have a `form` (both `convert` and `form` are intended to be imperative verbs not nouns as we did in summary text, as in: you (we, they, or I) convert, you (or we, they, or I) `form`.
> 
> `parser` is a noun, but if things were done from scratch I would have suggested `parse` instead of `parser`. If you think it worth doing, then please do.
> 
> Under form we have "box" (also a verb), and low-level or final format (as opposed to the top level form). Right now this lower-level `format` is not under anything but if we create a `form` i I think it should be moved under that.
> 
> The core module then has parse, evaluate (not currently a module though) and form modules. The form module has `box` and `format` modules.
> 
> If you wanted to write a backend to handle some new `Form` you would model it on the this new form module.
> 
> What do you think?

Regarding `_BoxedString`, this is something that should go away before the next release. I put it in order to start the format decoupling without doing a very drastic change in the interface, but each time I look at it, I feel more regret.

mathics/builtin/layout.py
rocky: > So, I am not sure how useful is to do this now.

This kind of tradeoff  has been pestering me from the beginning. The more we delay the more work we will have to do in conversion, _unless_ there is automation to do it. 

As you recall from the earliest days of getting involved I have been advocating for using something like sphinx. (It or something like that was around from the days that this code was written. It was boneheaded to not build on it but rather ignore it and write something from scratch). 

However we should not kid ourselves that replacing it even had we done this earlier is a lot of work. And in the meantime I do feel strongly that our docs should be useful and we should continue to improve them.

But right now, I think we agree that the priorities have to be improving performance. Sort of up there would also be increasing contributors,  involvement, and hooking into the community. 

 So for the homegrown documentation system, I hold my nose (and probably complain _a_ _lot_.

mathics/core/formatter.py
rocky: Great! Something like this is what I was hoping to hear. Yes, I understand we can't do everything all at once.

mathics/builtin/layout.py
rocky: A little side digression about 
>  increasing contributors

When you have a code base that calls all tree nodes "leaves", and describes "eval" methods as "apply", doesn't seem to understand what a list literal is, but writes some homegrown (and possibly wrong) custom caching and its own autodoc system, (among numerous other examples) you are not going to attract the kind of experienced programmers that are needed here. And that has been in the back of my mind.

rocky:
@mmatera I accidentally force pushed my old branch in updating the adding-KroneckerProduct branch. (I should have narrowed the force push just to that branch). 

Please force push what you have on top of this to revert that last force push. Thanks.

rocky:
The same thing with the  `improve_install_check_in_docs` branch.

mmatera:
@rocky, no problem: probably I need to rebase and split this up.

mmatera:
Superseeded by #515

PR #495, opened 2022-08-06
Move timing out of `mathics.core.crunch_util`

rocky:
* `robust_min()`: is used only one place. Move this function there
<strike>*  `IS_PYPY`: move to `mathics.system_info`</strike>
* timing routines: move to `mathics.timing` module
* `function_arguments()`: We had more complex to support older Pythons which we don't support anymore. Move code to `rules.py` where it is the only place it is used.

The parts removed will be separated into another PR.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/495"></a>

### DIFF
--- a/mathics/algorithm/clusters.py
+++ b/mathics/algorithm/clusters.py
@@ -7,9 +7,17 @@ import bisect
 import math
 
 from mpmath import fsum
-from mathics.core.util import robust_min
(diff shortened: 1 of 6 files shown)


mmatera:
LGTM

PR #496, opened 2022-08-07
Add Angular Momentum builtins...

rocky:
which starts a Quantum Mechanics section.

<strike>TODO:

* Unit tests for errors.
* Ensure function attributes are correct
* Ensure we error on what we are supposed to and not on what we do not</strike>

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/496"></a>

### DIFF
--- a/mathics/builtin/quantum_mechanics/angular.py
+++ b/mathics/builtin/quantum_mechanics/angular.py
@@ -175,6 +175,11 @@ class ThreeJSymbol(SympyFunction):
     >> ThreeJSymbol[{2, 0}, {6, 0}, {4, 0}]
      = Sqrt[715] / 143
 
+    Symmetric under permutations:
+
+    >> % == ThreeJSymbol[{2, 0}, {4, 0}, {6, 0}] == ThreeJSymbol[{4, 0}, {2, 0}, {6, 0}]
+     = True
+
     >> ThreeJSymbol[{2, 0}, {6, 0}, {4, 1}]
      = 0
 
@@ -186,12 +191,14 @@ class ThreeJSymbol(SympyFunction):
      = True
 
     Result 0 returned for unphysical cases:
-    >>ThreeJSymbol[{1, 2, 3}, {4, 5, 12}]
+    >> ThreeJSymbol[{1, 2}, {3, 4}, {5, 12}]
+     = 0
     """
 
     attributes = A_PROTECTED | A_READ_PROTECTED
     messages = {
-        "threejsymbol": "Parameter `` of `` has value ``; ThreeJSymbol cannot handle symbols yet.",
+        "3jsymbol_symbol": "Parameter `` of `` has value ``; ThreeJSymbol cannot handle symbols yet.",
+        "3jsymbol_value": "ThreJSymbol values ``, ``, `` must be integer or half integer and fulfill the triangle relation",
     }
     summary_text = "values of the Wigner 3-j symbol"
     sympy_name = "physics.wigner.wigner_3j"
@@ -217,4 +224,10 @@ class ThreeJSymbol(SympyFunction):
                 # rather than pairs if (j, m).
                 sympy_js[j * 3 + i] = py_element
 
-        return from_sympy(wigner_3j(*sympy_js))
+        try:
+            result = wigner_3j(*sympy_js)
+        except ValueError:
+            evaluation.message("ThreeJSymbol", "3jsymbol_value", j12, j34, j56)
+            return
+
+        return from_sympy(result)
--- a/mathics/builtin/quantum_mechanics/angular.py
+++ b/mathics/builtin/quantum_mechanics/angular.py
@@ -27,11 +27,11 @@ from mathics.core.attributes import (
 
 class ClebschGordan(SympyFunction):
     """
-    <url>:Clebsch Gordan coefficients matrices: https://en.wikipedia.org/wiki/Clebsch%E2%80%93Gordan_coefficients</url> (<url>:SymPy: https://docs.sympy.org/latest/modules/physics/quantum/cg.html</url>, <url>:WMA: https://reference.wolfram.com/language/ref/ClebschGordan</url>)
+    <url>:Clebsch-Gordan coefficients matrices: https://en.wikipedia.org/wiki/Clebsch%E2%80%93Gordan_coefficients</url> (<url>:SymPy: https://docs.sympy.org/latest/modules/physics/quantum/cg.html</url>, <url>:WMA: https://reference.wolfram.com/language/ref/ClebschGordan</url>)
 
     <dl>
       <dt>'ClebschGordan[{$j1$, $m1$}, {$j2$, $m2$}, {$j$ $m$}]'
-      <dd>returns the Clebsch-Gordan coefficient for the decomposition of in terms of |$j$, $m$> in terms of $j2$, $m2$.
+      <dd>returns the Clebsch-Gordan coefficient for the decomposition of in terms of |$j$, $m$|$j2$, $m2$>.
     </dl>
 
     >> ClebschGordan[{3 / 2, 3 / 2}, {1 / 2, -1 / 2}, {1, 1}]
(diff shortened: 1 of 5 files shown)


mathics/builtin/quantum_mechanics/angular.py
```diff
@@ -0,0 +1,220 @@
...
+)
+
+
+class ClebschGordan(SympyFunction):
+    """
+    <url>:Clebsch-Gordan coefficients matrices: https://en.wikipedia.org/wiki/Clebsch%E2%80%93Gordan_coefficients</url> (<url>:SymPy: https://docs.sympy.org/latest/modules/physics/quantum/cg.html</url>, <url>:WMA: https://reference.wolfram.com/language/ref/ClebschGordan</url>)
+
+    <dl>
+      <dt>'ClebschGordan[{$j1$, $m1$}, {$j2$, $m2$}, {$j$ $m$}]'
+      <dd>returns the Clebsch-Gordan coefficient for the decomposition of in terms of |$j$, $m$|$j2$, $m2$>.
```
mmatera, line 34: To be precise,   * ... in terms of |$j_1, m_1$>|j_2 m_2$>.*

Maybe $\langle$ instead of $>$.

mathics/builtin/quantum_mechanics/angular.py
```diff
@@ -0,0 +1,220 @@
...
+
+        sympy_jms: List[int] = []
+
+        for pair in (j1m1, j2m2, jm):
+            if len(pair.elements) != 2:
+                return
+            sympy_jms += [p.to_sympy() for p in pair.elements]
+        return from_sympy(CG(*sympy_jms).doit())
+
+
```
mmatera, line 67: At some place, it would be good to implement also `KroneckerProduct` (https://reference.wolfram.com/language/ref/KroneckerProduct.html)

mathics/builtin/quantum_mechanics/angular.py
```diff
@@ -0,0 +1,220 @@
...
+    <dl>
+      <dt>'PauliMatrix[$k$]'
+      <dd>returns the $k$th Pauli spin matrix).
+    </dl>
+
+    >> PauliMatrix[Range[3]]
+     = {{{0, 1}, {1, 0}}, {{0, -I}, {I, 0}}, {{1, 0}, {0, -1}}}
+    """
+
+    attributes = A_LISTABLE | A_PROTECTED | A_READ_PROTECTED
```
mmatera, line 81: `PauliMatrix` is not `Listable` in WL. However, I have to say that this would be an improvement...

mmatera:
Something that I am not sure about this is whether these functions belong to a `quantum mechanics` section, or to other modules (`PauliMatrix` to linear algebra, `ClebschGordan`, `ThreeJSymbol` and `SixJSymbol` to "Combinatorial Functions" (https://reference.wolfram.com/language/guide/CombinatorialFunctions.html) , etc. Otherwise, we should put the special polynomials also under this category.

mathics/builtin/quantum_mechanics/angular.py
rocky: The persistent problem we have is that we the homegrown doc system really can't handle symbols reliably. We have seen this time and time again with things like Phi. 

Do we use the ket symbols anywhere else? If you know that when we go to print this it will work, then we can use.

mathics/builtin/quantum_mechanics/angular.py
rocky: I have been going by reference guide section. That seems to belong to Matrix or Tensor things even though everything is related. Feel free to add it, but I am not sure where we would add it.

rocky:
> Something that I am not sure about this is whether these functions belong to a `quantum mechanics` section, or to other modules (`PauliMatrix` to linear algebra, `ClebschGordan`, `ThreeJSymbol` and `SixJSymbol` to "Combinatorial Functions" (https://reference.wolfram.com/language/guide/CombinatorialFunctions.html) , etc. Otherwise, we should put the special polynomials also under this category.

I am following https://reference.wolfram.com/language/guide/FunctionsUsedInQuantumMechanics.html

We will also have to deal with the duplication somehow. But that is yet another battle with autodoc for some other time.

mathics/builtin/quantum_mechanics/angular.py
rocky: Thanks. Should be addressed by 270738c7

mathics/builtin/quantum_mechanics/angular.py
rocky: > I am not sure where we would add it.

It looks like it goes under "Vector Space Operations" which we already have. See https://reference.wolfram.com/language/guide/OperationsOnVectors.html

rocky: As I was doing these, it occurred to me that there was a set not intuitive process I used that makes doing this kind of thing more automatic. 

* Find a Builtin that hasn't been implemented. This PR comes from the bottom this [this issue](https://github.com/mathics/Mathics/issues/1483), TODO's at the bottom of newer sections/modules list missing functions. 
* Decide where in the code/doc to put it. As noted,  I first noticed it as a missing Combinatorial Function. However on more recent reflection, those last three are more specifically isolated as Angular Momentum functions (and in Combinatorial Functions it is not split out or noted as such). 
* See if it is a SymPy function. And the rest will assume yes. But if not try to scipi and mpmath. I suppose numpy might be tried too.  
* Copy one of the newer existing Sympy functions and adapt. Newer existing Sympy functions have Wiki, SyPy, and WMA links. Probably in the WMA link, all you will have to do is change the copied WMA name into the new name. 
* Use WMA link to adapt the docstring. It might be consulted to get the summary text too. 
* Go into mathics and see that you can get "info" or "?" on the changed doc. 
* Adapt the pattern match rule and parameters. This is probably the hardest part. `Cases` in mathics can help here.
* Adapt the eval method parameters and body. 
* Make sure the function attributes are correct. 
* Adapt some doctests and test. Examples from WMA and SymPy are a good source, but try to make the examples meaningful
* Run the tests using `python mathics/docpipeline -s <builtin name>` 
* Go over error conditions and error messages
* Write unit tests for all of these edge cases and uninteresting cases to try

The above even though long is still kind of sketchy. So I will go through this process on KroneckerProdect as a detailed example and put this in the developer guide.

mathics/builtin/quantum_mechanics/angular.py
mmatera: OK, let's keep ">", but then the description should read
*returns the Clebsch-Gordan coefficient for the decomposition of in terms of $|j_1, m_1>|j_2, m_2>$.*

mathics/builtin/quantum_mechanics/angular.py
rocky: I started down this path. See https://github.com/Mathics3/mathics-development-guide/tree/adding-KroneckerProduct and https://mathics-development-guide.readthedocs.io/en/adding-kroneckerproduct/extending/developing-code/extending/adding-new-mathics-function.html. 

I stopped however when I see that there is this is not going to make for a simple example, because of the matrix checking and conversion in SymPy that we probably need to do. 

I will pick a simpler example. As for that KroneckerProduct branch, I don't intend on going further there.

mmatera: In that case, I would say that this is almost ready.

rocky: (I never was my intention to add it here - these are the angular momentum functions.)

Possibly over the weekend I may add pytests.. However redoing how Part works and changes ListExpression is more important.

mathics/builtin/quantum_mechanics/angular.py
rocky: Thanks for finding the mistake. I think it is fixed now in bb73c516

rocky: Speaking of >, the ket symbol and our homegrown docsystem fragility. When I was working on the guide section for the process of adding Curl as an example,  I wanted to write:

$$     ( \partial f_3 / \partial x_2 - \partial f_2 / \partial x_3, \ \ %
      \partial f_1 / \partial x_3 - \partial f_3 / \partial x_1, \ \ %
      \partial f_2 / \partial x_1 - \partial f_1 / \partial x_2 )
$$

in the Sphinx document. and I was pleasantly surprised that I could do this without fuss, and that there was a world of flexibility available rather that this kind of constrained system where we are working around things and getting less-good results in the end.

So there _is_ hope if we can get over the mountain of moving this to sphinx.

rocky:
This weekday's activity will probably be finish the development guide for Curl, and I might also add KroneckerProduct since I now see what is up with sympy Matrices.

PR #503, opened 2022-08-13
Correct D[] bug in passing wrong parameter type

### LINKED ISSUE #502: `D[]` error message uses wrong parameter type
## Description

`D[]` is passing a Symbol parameter to `evaluation.message()` which it should be passing a Symbol name, a string. 
## Output Given

```
$ mathics
...
In[1]:= D[{y, -x}[2], {x, y}]
  File "mathics", line 33, in <module>
    sys.exit(load_entry_point('Mathics3', 'console_scripts', 'mathics')())
    symbol = ensure_context(symbol)
  ...
  File "mathics/core/element.py", line 15, in ensure_context
    assert isinstance(name, str)
AssertionError
```

## Expected behavior

Message: 
```
D::dvar: Multiple derivative specifier {x, y} does not have the form {variable, n}, where n is a non-negative machine integer.
```
returned

## Your Environment

All environments.

## Additional context

A PR will be put in shortly.

rocky:
Also:

remove overloaded type-changing variable
reduce vaguess in vargueness "msg" - is what message message back from
evaluation? message given on assertion failure? Some other kind of message?

Fixes #502 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/503"></a>

### DIFF
--- a/mathics/builtin/numbers/calculus.py
+++ b/mathics/builtin/numbers/calculus.py
@@ -228,9 +228,9 @@ class D(SympyFunction):
         elif f == x:
             return Integer1
         elif isinstance(f, Atom):  # Shouldn't happen
-            1 / 0
-            return
-        # So, this is not an atom...
+            raise TypeError(f"D[] function slot {f} is an instance of an Atom")
+
+        # f is not a pattern, nor an Atom, and f != x
 
         head = f.get_head()
         if head is SymbolPlus:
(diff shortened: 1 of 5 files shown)


mathics/builtin/numbers/calculus.py
```diff
@@ -228,9 +228,9 @@ def apply(self, f, x, evaluation):
         elif f == x:
             return Integer1
         elif isinstance(f, Atom):  # Shouldn't happen
-            1 / 0
```
mmatera, line 231: Not sure even if the `elif` block is necessary.  That `f` is not an atom should be handled by the first `if` clause where we check that `f` is free of the derivative parameter.

rocky: Ok - thanks. All of this has been rewritten in 16ad3908

mmatera (approved):
LGTM

PR #504, opened 2022-08-13
Go over Cross and Norm

rocky:
This is in preparation ofr adding Builtin Curl which is going to be used
as a tutorial example.

Norm:
* Remove vector restriction on parameter m when parameter p not specified.
* Pull out its unit tests, and expand on the doctess
* Start pulling out eval methods here too.
* Allow "Frobenius" in parameter p

Some small test doc typos corrected.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/504"></a>

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,15 @@
 CHANGES
 =======
 
+5.0.3dev0
+----------
+
+Enhancements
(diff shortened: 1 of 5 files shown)


mmatera (approved):
LGTM

PR #507, opened 2022-08-14
Revise `set_part()` to be more functional

rocky:
When I was revising ListExpression  a while ago, this code looked like it was preventing simpler ListExpression tests. (See below)

Now I am not totally sure. It fixes a potential bug if Part somehow is not able to transform an `Expression` into `ListExpression`, and if in the future we need to transform `Expression` into some other kind of specialized expression.  

Via testing t am not seeing this right now. 

I have made some slight changes to the code so as to hopefully make it easier to understand and modify should we need to in the future. 

Hopefully this will pave the way for simplifying  and replacing `expr.has_form("List", None)` with `isinstance(expr, ListExpression)` possibly after a few other parts of the code are gone over.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/507"></a>

### DIFF
--- a/mathics/algorithm/parts.py
+++ b/mathics/algorithm/parts.py
@@ -25,83 +25,95 @@ from mathics.builtin.exceptions import (
 SymbolNothing = Symbol("Nothing")
 
 
-def get_part(varlist: BaseElement, indices: List[int]) -> BaseElement:
-    """Extract part of ``varlist`` specified by ``indicies`` and
+def get_part(expression: BaseElement, indices: List[int]) -> BaseElement:
+    """Extract part of ``expression`` specified by ``indicies`` and
     return that.
     """
 
-    def get_subpart(varlist: BaseElement, indices: List[int]) -> BaseElement:
+    def get_subpart(expression: BaseElement, indices: List[int]) -> BaseElement:
         """Recursive work-horse portion of ``get_part()`` that extracts pieces
-        of ``varlist`` based on ``indices``. ``varlist`` and ``indices``
+        of ``expression`` based on ``indices``. ``expression`` and ``indices``
         are smaller parts of the corresponding variables at the outer level.
         """
         if not indices:
-            return varlist
+            return expression
 
-        if isinstance(varlist, Atom):
+        if isinstance(expression, Atom):
             raise PartDepthError(indices[0])
 
         pos = indices[0]
-        elements = varlist.elements
+        elements = expression.elements
         try:
             if pos > 0:
                 part = elements[pos - 1]
             elif pos == 0:
-                part = varlist.get_head()
+                part = expression.get_head()
             else:
                 part = elements[pos]
         except IndexError:
             raise PartRangeError
         return get_subpart(part, indices[1:])
 
-    return get_subpart(varlist, indices).copy()
+    return get_subpart(expression, indices).copy()
 
 
-def set_part(varlist, indices: List[int], newval) -> BaseElement:
-    """Replace the part of ``varlist`` specified by ``indicies`` with the single
+def set_part(expression, indices: List[int], newval) -> BaseElement:
+    """Replace all parts of ``expression`` specified by ``indicies`` with the
     ``newval`` value. Return the modified compound expression.
     """
 
-    def set_subpart(varlist, indices: List[int]) -> BaseElement:
+    def set_subpart(expression, indices: List[int]) -> BaseElement:
         """Recursive work-horse portion of ``set_part()`` that replaces pieces
-        of ``varlist`` with outer variable ``newval`` based on ``indices``.
-        ``varlist`` and ``indices`` are smaller parts of the corresponding variables at the outer level.
+        of ``expression`` with outer variable ``newval`` based on ``indices``.
+        ``expression`` and ``indices`` are smaller parts of the corresponding variables at the outer level.
         """
         if len(indices) > 1:
             pos = indices[0]
-            if isinstance(varlist, Atom):
+            if isinstance(expression, Atom):
                 raise PartDepthError
             try:
                 if pos > 0:
-                    part = varlist.elements[pos - 1]
+                    part = expression.elements[pos - 1]
                 elif pos == 0:
-                    part = varlist.get_head()
+                    part = expression.get_head()
                 else:
-                    part = varlist.elements[pos]
+                    part = expression.elements[pos]
             except IndexError:
                 raise PartRangeError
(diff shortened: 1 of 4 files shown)


mathics/algorithm/parts.py
```diff
@@ -21,78 +24,84 @@
...
                 if pos > 0:
-                    cur.set_element(pos - 1, newval)
+                    varlist.set_element(pos - 1, newval)
                 elif pos == 0:
-                    cur.set_head(newval)
+                    if newval == SymbolList:
+                        varlist = ListExpression(varlist._elements)
+                    # If we have other types of specialized compound expressions
+                    # this might be detected as "elif" here.
+                    varlist.set_head(newval)
```
mmatera, line 97: `else` ?

rocky: I don't understand, please elaborate.

Right now, the only customized kind of (compountd) Expression is ListExpression so that is the only place where we need to redo the entire kind of compound Expression. As we add more that can go through this code we'll have to find and adapt those. At some point some more automated way of replacing these without if's will be needed. But we are not there yet. So the comment serves to mark and note such possibility.

rocky: In a follow-on PR, I see that this code was faulty here: we also need to handle the case for converting from ListExpression to Expression. (Here we convert from Expression to ListExpression only).

The follow-on PR handles creating ListExpression from the outset we go from parser AST to Mathics Expression and this will open up many more opportunities for more streamlined (i.e. faster)  operations. After that I believe we will be able to not have to need `has_form("System`List", None)`, but I believe it opens the possibility of eliminating conversions for list literals in some places.

I am keeping these two PR's separate (at the expense of more work on my side in juggling between versions), so as to keep each PR a little more focused and more intelligible. 

With respect the speedups, I hope there is a lesson getting learned here. We won't get to where we need by low-level tweaks  like changing the order of expressions, or using slots instead of local variables, using Cython, or rewriting the code (where a number of parts are a incomprehensible mess) in C or Rust mirroring the incomprehensible mess.

Instead we'll get there by making this look more like an normal interpreter _first_ and using the long-established practice and techniques there _first_.  Most of the time Slots, homegrown caching, Cython, and stuff like this comes with a lack of readability which was a problem even _before_ any of the low-level speedups which add complexity and confusion were added. 

To me this is doing a final painting the outside or interior of a house before you've added the roof or finished constructing most of the walls interior. 

_After_ we get this to look like a normal interpreter, all of those things can be done. Rewriting in Rust might give a 10 times improvement, I wouldn't undertake doing something like the in advance of cleaning up the mess that is here now.

mmatera:
```
if newval == SymbolList:
     varlist = ListExpression(varlist._elements)
# else:
     varlist.set_head(newval)
```

You shouldn't modify the head of ListExpression

rocky: That's true. 89e97ea5 adds some checks for this.

mathics/algorithm/parts.py
```diff
@@ -21,78 +24,99 @@
...
-    def rec(cur, rest):
-        if len(rest) > 1:
-            pos = rest[0]
-            if isinstance(cur, Atom):
+def set_part(expression, indices: List[int], newval) -> BaseElement:
+    """Replace all parts of ``expression`` specified by ``indicies`` with the
+    ``newval`` value. Return the modified compound expression.
+    """
+
+    def set_subpart(expression, indices: List[int]) -> BaseElement:
```
mmatera, line 65: in order to make this super-clear: maybe the parameters of `set_subpart` should be `sub_expression` and  `remaining_indices`.

mathics/algorithm/parts.py
```diff
@@ -21,78 +24,99 @@
...
                 elif pos == 0:
-                    cur.set_head(newval)
+                    # We may have to replace the entire ``expression``
+                    # variable when changing position 0 or Head. This
+                    # happens when the before and after are
+                    # class objects are different.
+
+                    # Right now, we need to only worry about
+                    # converting between ``Expression`` and
+                    # ``ListExpression`` or vice vera.  In the code
```
mmatera, line 100: I would think that if you change the head, you would like to build a new expression of the generic type `Expression`. Suppose for instance that you have a `List`, and then you want to change the head to be `Raster`. A simple way to do this is to do `Expression(SymbolRaster, *cur._elements, cur.elements_properties)`. Then, in the next evaluation, if `Raster` has associated a  `BuiltinElement` object, the expression can be converted into a `Raster(*elements, cur.elements_properties)`

mathics/algorithm/parts.py
rocky: Ok, will do on next commit. And I suspect pylint (which in my opinion is by default too noisy) will like this better. I think it doesn't like hiding names this way in nested functions. (I don't totally agree with this, but I don't strongly object either.)

mathics/algorithm/parts.py
rocky: In my opinion change the Head element is kind of a dodgy thing to do. 

Overall though if you want a performant interpreter, we need to give up on the notion that you there are several equivalent forms that a compound expression can be in and the interpreter is silently checking and converting forms. 

Right now, it is true that we need to be able to handle Expression(SymbolList, ...) as well as ListExpression(...) but this is transitory in the same way that you could write Expression("Plus", 1, 2) or Expression(SymbolPlus, Integer1, Integer2) and have these be interchangeable. 

In the future though I don't think we can or should allow this. In my view you are overthinking and over complicating things from the outset. Let the complexity come to you rather than seek it out. 

Related but in my view very very different is something like the recently changed addition where you can now take a ListExpression and index into it eg. `list_expression[0]` for the (non-head) first element. At some point I was going to also add the ability to do `len(list_expression)`. And we have this feature ware you can write `Integer0 + Integer1` which is equivalent to `Expression(Symbol Plus, Integer0, Integer1)` .  

These are different things, because in the latter, everything really is strongly typed and there _aren't_ conversions and checking going on: the `+` and `len()` are more syntactic conveniences but underneath the types that they work on are not malleable.

mathics/algorithm/parts.py
rocky: Done via 65269755

rocky:
I believe all of the concerns have been addressed and we have gone over this with probably too fine tooth comb. (Do we worry about line 112 using `==` and `!=` vs `in` and `not in` worry about that when the need comes up because we have another specialization?  - later.

PR #509, opened 2022-08-15
Go over Abs[]

rocky:
I was looking at https://github.com/axkr/symja_android_library/blob/master/symja_android_library/rules/AbsRules.m to see if we have things covered that are listed there. Nope.

This PR doesn't do it either, but at least I've noted what is needed.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/509"></a>

### DIFF
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -24,6 +24,7 @@ from mathics.core.attributes import (
     protected as A_PROTECTED,
 )
 
+from mathics.core.element import BaseElement
 from mathics.core.evaluators import eval_N
 
 from mathics.builtin.base import (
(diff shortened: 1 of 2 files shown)


TiagoCavalcante (approved):
LGTM

mmatera:
The point is that we delegate the evaluation in sympy. Something we could try is to check whether adding rules for typical cases speeds up or slow down the evaluation .

rocky:
> The point is that we delegate the evaluation in sympy. Something we could try is to check whether adding rules for typical cases speeds up or slow down the evaluation .

SymPy can has no idea what `Undefined` means, and I am not sure how it is represente din Mathics either. So obviously for `Abs[Undefined]` is not something it could possibly understand. 

As for  `Abs[ ArcTan[ComplexInfinity] ] == Pi/2`, that I've opened #510  so we don't foget about this.

PR #511, opened 2022-08-15
KroneckerProduct for developer's tutorial

rocky:
Initially I thought I wouldn't be able do this because I was not aware of `to_sympy_matrix()`.

This currently has problems handling vectors, and doesn't have a unit test. But this is possibly good enough for a phase 1 introduction for how to write a Mathics builtin, 

See https://mathics-development-guide.readthedocs.io/en/adding-kroneckerproduct/extending/developing-code/extending/KroneckerProduct/basic-skeleton.html  which will be revised soon. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/511"></a>

### DIFF
--- a/mathics/builtin/vectors/vector_space_operations.py
+++ b/mathics/builtin/vectors/vector_space_operations.py
@@ -4,12 +4,63 @@
 Vector Space Operations
 """
 
-from mathics.builtin.base import Builtin
+from sympy.physics.quantum import TensorProduct
+
+from mathics.builtin.base import Builtin, SympyFunction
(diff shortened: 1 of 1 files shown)


TiagoCavalcante (approved):
LGTM

rocky:
Even though we handle the situation for Matrix, in looking for how to handle the situation for Vectors, it looks to me like we may need to beef up `Array[]` . See #512.

mmatera (approved):
Lgtm

PR #515, opened 2022-08-16
 Tyding up formatting. Second round.

mmatera:
This is the second cut of #492. 
In this round, 
* the interface of boxes_to_* was reformulated to be more scalable. 
* The `inout.py` module was splitter in parts closer to the WR organization.
*  `_BoxedString` was removed in favor of using a `StyleBox` construct, which is simpler.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/515"></a>

### DIFF
--- a/mathics/core/element.py
+++ b/mathics/core/element.py
@@ -469,4 +469,17 @@ class BoxElement(ImmutableValueMixin, BaseElement):
     elements
     """
 
-    pass
+    def boxes_to_text(self, **options) -> str:
+        from mathics.core.formatter import lookup_method
+
+        return lookup_method(self, "text")(self, **options)
+
+    def boxes_to_tex(self, **options) -> str:
+        from mathics.core.formatter import lookup_method
+
+        return lookup_method(self, "tex")(self, **options)
+
+    def boxes_to_mathml(self, **options) -> str:
+        from mathics.core.formatter import lookup_method
+
+        return lookup_method(self, "mathml")(self, **options)
--- a/mathics/core/element.py
+++ b/mathics/core/element.py
@@ -245,7 +245,7 @@ class BaseElement(KeyComparable):
             return self == rhs
         return None
 
-    def format(self, evaluation, form, **kwargs) -> "BoxElement":
+    def format(self, evaluation, form, **kwargs) -> "BoxElementMixin":
         from mathics.core.formatter import format_element
         from mathics.core.symbols import Symbol
 
@@ -463,7 +463,7 @@ class EvalMixin:
         raise NotImplementedError
 
 
-class BoxElement(ImmutableValueMixin, BaseElement):
+class BoxElementMixin(ImmutableValueMixin):
     """
     The base class for all the boxed
     elements
--- a/mathics/core/element.py
+++ b/mathics/core/element.py
@@ -469,17 +469,19 @@ class BoxElementMixin(ImmutableValueMixin):
     elements
     """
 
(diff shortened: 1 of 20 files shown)


rocky:
Looks even better than before. Would like to spend some time over the weekend trying out. 

The module name `inout` on the boxing seems either vague or misleading.

mmatera:
> Looks even better than before. Would like to spend some time over the weekend trying out.
> 

The bad side is that it is longer... In any case, most of the changes are just moving things around.

> The module name `inout` on the boxing seems either vague or misleading.

Indeed, is not a good name. However, I would keep it in this round, to make it easier to compare with master.

rocky:
As @mmatera informs, this should be revisited and merged after #518 and #519 get merged.

mmatera:
@rocky, I think now this is ready for review. There are changes in many files, but most of them are trivial removal of `_BoxedString` in favor of `String` and `StyleBox(...)`.

mathics/core/element.py
```diff
@@ -469,4 +469,17 @@ class BoxElement(ImmutableValueMixin, BaseElement):
     elements
     """
 
-    pass
+    def boxes_to_text(self, **options) -> str:
```
rocky, line 472: It seems to me that instead of having three specific box formatters, there should be one  "box" routine that is passed the format which is either "text", "text", "mathml", or one an unlimited number of formatters.  

Yes, this would mean places were "boxes_to_x" is done, but I wonder how onerous that would be.

mmatera: Yes, the idea is to remove these `boxes_to_x` routines in favor of `mathics.format.[format].[method that takes a boxed expression and produces an string or bitstring of the appropriate format ]`

I kept this just for backward compatibility, and to make the PR a little bit smaller...

rocky: The backwards compatibility I had forgotten about, so good catch. Yes, this is used for example in pymathics-graph and mathics-django.  

pymathics-graph that has to be revised, anyway. 

But still, there are two things that would be beneficial here:

* split out BoxElementMixin and ImmutableValueMixin into a separate module, say "mixin.py"
* Write a single box routine and have those specific box routines call that.

By doing this, you don't have to do an import inside each def. And we DRY the code.

For example:


```
from mathics.core.formatter import lookup_method

class ImmutableValueMixin:
    @property
    def is_literal(self) -> bool:
        """
        The value value can't change once it is set.
        """
        return True

class BoxElementMixin(ImmutableValueMixin):
    """
    The base class for all the boxed
    elements
    """

    def box(self, format: str, **options: dict) -> str:
        return lookup_method(self, format)(self, **options)

    
    def boxes_to_mathml(self, **options:dict) -> str:
        """For compatibility deprecated"""
        return self.box("mathml", **options)

    def boxes_to_tex(self, **options: dict) -> str:
        """For compatibility deprecated"""
        return self.box("tex", **options)

    def boxes_to_text(self, **options: dict) -> str:
        return self.box("text, **options)

```

mmatera: OK with the DRY part. However, I am not sure about the part about splitting mixings into a module. I checked it and even doing it,   it is not possible to avoid the circular import without doing it locally.  On the other hand, I do not understand the logic of having different mixins in one module from a conceptual point of view. Sometimes it could make sense to import `ImmutableValueMixin` (or `EvalMixn`) in contexts where `BoxElementMixin` is not involved.

rocky:
LGTM - it would be good to use "Squash and merge" when going to master.

PR #517, opened 2022-08-18
Add Undefined value

### LINKED ISSUE #514: Add Undefined in System
When mathics starts there is an `Undefined` symbol for it, but it does not appear as in the System context, nor do we have Info on it.

```
$ mathics --full-form

Mathics 5.0.3dev0
.. 
In[1]:= Undefined
Global`Undefined
Out[1]= Undefined

In[2]:= ?Undefined
Out[2]= Null

In[3]:= ??Undefined
Out[3]= Null

In[4]:= Attributes[Undefined]
Out[4]:= {}
```

We expect: 

```
In[1]:= Undefined
System`Undefined
Out[1]:= Undefined

In[2]:= ?? Undefined                                                            

Out[2]= Undefined is a symbol that represents a quantity with no defined value.


        Attributes[Undefined]={Protected}
```

rocky:
Fixes #514

The first commit in the PR will be another example in the developer's doc for how to add a builtin. 

This one (even though it was done last) is in fact the is the simplest and simplest to explain. 

After that in the doc will be KroneckerProduct (for Arrays not vectors). This too will link to a specific commit in the code for reference. 

and finally there will be a Curl example (not finished yet even in code). 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/517"></a>

### DIFF
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -24,7 +24,6 @@ from mathics.core.attributes import (
     protected as A_PROTECTED,
 )
 
-from mathics.core.element import BaseElement
 from mathics.core.evaluators import eval_N
 
 from mathics.builtin.base import (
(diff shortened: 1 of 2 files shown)


mmatera (approved):
LGTM

PR #520, opened 2022-08-18
BoxExpression instead of BoxConstruct

mmatera:
This PR replaces trailing uses of BoxConstruct in the code. `BoxConstruct` definition is kept for backward compatibility.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/520"></a>


TiagoCavalcante (approved):
LGTM

rocky:
LGTM - I don't see that BoxConstruct is kept for compatibility, and I don't see the need either. I see that BoxConstructErorr hasn't been renamed. Is that what you mean?

mmatera:
The definition of BoxConstruct   is here: https://github.com/Mathics3/mathics-core/blob/5bac51cbafea9ec63560da00df45c367ccfb8d8d/mathics/builtin/base.py#L1053
I keep it because it is maybe used on some of the other projects. Once we ensere that it is not used anymore, we can just remove that line .

mmatera:
Regarding BoxConstructError, it was there before I introduced the BoxConstruct class. Maybe it also deserves a new name.

rocky:
> Regarding BoxConstructError, it was there before I introduced the BoxConstruct class. Maybe it also deserves a new name.

If you are up for it, I think it would be good to do and make things more uniform and less confusing. 

Thanks for pointing the assignment ot BoxConstruct. I am pretty sure it is not used in outside packages, but this assignment is small enough that it doesn't do harm and could help.

mathics/builtin/base.py
```diff
@@ -10,7 +10,7 @@
 
 
 from mathics.builtin.exceptions import (
-    BoxConstructError,
+    BoxExpressionError,
```
rocky, line 13: Thanks again for doing this.

rocky:
@mmatera Ok to merge this in?

PR #522, opened 2022-08-19
Controlling symbol duplication

mmatera:
During the last months, we devoted some time to reorganizing and splitting the modules under builtin. One consequence was that accidentally, some symbols could result in duplications along different modules. This PR fixes the cases I found, and adds a pytest to check that subsequent reorganizations do not produce more duplications.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/522"></a>

### DIFF
--- a/test/test_duplicated.py
+++ b/test/test_duplicated.py
@@ -1,0 +1,44 @@
+import pytest
+import os
+from mathics.builtin import modules, sanity_check, is_builtin, Builtin
+
+
+@pytest.mark.skipif(
+    os.environ.get("LINT"), reason="Lint checking done only when specified"
+)
+def test_check_duplicated():
+    msg = ""
+    builtins_by_name = {}
+    for module in modules:
+        vars = dir(module)
+        for name in vars:
+            var = getattr(module, name)
+            if (
+                hasattr(var, "__module__")
+                and var.__module__.startswith("mathics.builtin.")
+                and var.__module__ != "mathics.builtin.base"
+                and is_builtin(var)
+                and not name.startswith("_")
+                and var.__module__ == module.__name__
+            ):  # nopep8
+                instance = var(expression=False)
+                if isinstance(instance, Builtin):
+                    # This set the default context for symbols in mathics.builtins
+                    if not type(instance).context:
+                        type(instance).context = "System`"
+                    name = instance.get_name()
+                    """
+                    assert (
+                        builtins_by_name.get(name, None) is None
+                        ), f"{name} defined in {module} already defined in {builtins_by_name[name]}."
+                    """
+                    if builtins_by_name.get(name, None) is not None:
+                        print(
+                            f"\n{name} defined in {module} already defined in {builtins_by_name[name]}."
+                        )
+                        msg = (
+                            msg
+                            + f"\n{name} defined in {module} already defined in {builtins_by_name[name]}."
+                        )
+                    builtins_by_name[name] = module
+    assert msg == "", msg
--- a/test/test_duplicated.py
+++ b/test/test_duplicated.py
@@ -4,7 +4,7 @@ from mathics.builtin import modules, sanity_check, is_builtin, Builtin
 
 
 @pytest.mark.skipif(
-    os.environ.get("LINT"), reason="Lint checking done only when specified"
+    not os.environ.get("LINT"), reason="Lint checking done only when specified"
 )
 def test_check_duplicated():
     msg = ""


mathics/builtin/base.py
```diff
@@ -216,24 +216,10 @@ def contribute(self, definitions, is_pymodule=False):
         # Set the default context
         if not self.context:
             self.context = "Pymathics`" if is_pymodule else "System`"
-
         name = self.get_name()
         options = {}
         option_syntax = "Warn"
 
-        for option, value in self.options.items():
```
mmatera, line 224: This is a duplicated piece of code, which appears again below.

mathics/builtin/lists.py
```diff
@@ -1012,64 +1012,6 @@ def apply_multi(self, expr, first, sequ, evaluation):
         return to_expression(name, to_expression(name, expr, *sequ), first)
 
 
-class Join(Builtin):
```
mmatera, line 1015: Duplicated in `mathics.builtin.list.rearrange`

mathics/builtin/numbers/integer.py
```diff
@@ -234,6 +234,92 @@ def _valid_base(self, b, evaluation):
             return base
 
 
+# This class is duplicated. Check what is the right one, or merge...
```
mmatera, line 237: This was in `mathics.builtin.atomic.numbers.py`. Now are in the same module, so the second definition overwrites this one. Since both definitions are different, this requires a merging.

mathics/builtin/base.py
```diff
@@ -393,6 +379,15 @@ def contextify_form_name(f):
         if is_pymodule:
             definitions.pymathics[name] = definition
         else:
+            # Then we move around builtins among modules, sometimes happens
+            # that the same builtin appears in more that one module. These
+            # lines can warn us about these duplications.
+            already_loaded = definitions.builtin.get(name, None)
+            if already_loaded is not None and already_loaded.builtin is not None:
+                print(f"Builtin `{name}` already loaded ({definitions.builtin[name]})")
```
rocky, line 387: Is this supposed to be here?

mmatera: Probably not...

rocky:
it feels like it would be better to do the test for redefining symbols as a git commit hook or maybe a separate CI test. 

I would like to keep local testing fast which encourages more local testing, and something like this needs to be done only at commit time. 

A simple way to have keep in pytest but not run it all of the time would be to put run it only if an environment variable is set.

mmatera:
> it feels like it would be better to do the test for redefining symbols as a git commit hook or maybe a separate CI test.
> 
> I would like to keep local testing fast which encourages more local testing, and something like this needs to be done only at commit time.
> 
> A simple way to have keep in pytest but not run it all of the time would be to put run it only if an environment variable is set.

This is not a very time-consuming test, but OK.

mmatera:
Something else that I also want to check is if some symbols could be lost in the reorganization, but this is going to be in another PR.

rocky:
> > it feels like it would be better to do the test for redefining symbols as a git commit hook or maybe a separate CI test.
> > I would like to keep local testing fast which encourages more local testing, and something like this needs to be done only at commit time.
> > A simple way to have keep in pytest but not run it all of the time would be to put run it only if an environment variable is set.
> 
> This is not a very time-consuming test, but OK.

I totally agree that is is probably not time consuming and I hesitated to mention. But in the end I did because there has been this habit do doing things like this that are really more commit orCI kinds of tests than code tests. For example we don't run black or flake8 as a unit test although it is easy to write something that does this.

mmatera:
> Something else that I also want to check is if some symbols could be lost in the reorganization, but this is going to be in another PR.

Done in #523

mmatera:
> > This is not a very time-consuming test, but OK.
> 
> I totally agree that is is probably not time consuming and I hesitated to mention. But in the end I did because there has been this habit do doing things like this that are really more commit orCI kinds of tests than code tests. For example we don't run black or flake8 as a unit test although it is easy to write something that does this.

The question is: when do we want to run this test? I guess that by now, when we are often doing this kind of reorganization, the best is to keep it as a pytest, or as a github action. @rocky, what do you think is the best option?

rocky:
> > > This is not a very time-consuming test, but OK.
> > 
> > 
> > I totally agree that is is probably not time consuming and I hesitated to mention. But in the end I did because there has been this habit do doing things like this that are really more commit orCI kinds of tests than code tests. For example we don't run black or flake8 as a unit test although it is easy to write something that does this.
> 
> The question is: when do we want to run this test? I guess that by now, when we are often doing this kind of reorganization, the best is to keep it as a pytest, or as a github action. @rocky, what do you think is the best option?

In my opinion I'd like to run this only on commit which is also where black and flake8 are run locally. Running it as a workflows test is fine too or in conjunction with _one_ of the existing ones, e.g. cythonization is fine too. 

A common practice of some projects is to separate tests into day-to-day tests and "full" tests. That is, some (non-python older GPL) projects use the convention: "make check" vs "make check-full". This test would be in the "check-full" category.  A simple mechanism here is to use `pytest.mark.skipif(os.environ.get("LINT"), reason="Lint checking done only when specified")` or something like that. 

Also it would be great if we would create a directory like "lint" and push tests of this kind down into that.

rocky:
@mmatera I've been thinking about this PR and it would be good to split this into two PRs, one with all the little fixups and one with just the pytest. 

We all have a tendency to lump independent things into a single PR, I think partly because this is how things occur as one is working on things, and also because of I guess the overhead in the review and CI pipeline.

However lumping in different things (even if they are related in some way) can also keep good uncontroversial changes from getting in while the other parts that are subject to debate are debated. 

In the long view of things keeping separate PR ideas separate is also good from a historical perspective.

mmatera:
I put the changes together to illustrate the motivation of this pr. But of course, I can split this up.

rocky:
@mmatera LGTM - sorry for the delay.

PR #523, opened 2022-08-19
Control builtin manifest

mmatera:
This continues #522, by adding a check against involuntary erasing of already defined symbols.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/523"></a>


mmatera:
This also implements an alternative check on duplicated symbols, which could be an alternative to #522. Of course, it is not going to pass until #524 be merged.

rocky:
LGTM - sorry for the delay.

PR #526, opened 2022-08-22
fix docstring in mathics.format.tex

mmatera:


TiagoCavalcante (approved):
LGTM

rocky:
The more I think about this more I am in favor of changing the module name to LaTeX. 

About the name TeX - it has in same sense become generic that is true. Originally it is what is or was quickly called "Plain TeX".
Other variants are
*  the form used in GNU documents "info TeX", 
* LaTeX which now is the most popular
*  AMS TeX which later originally was its own but then got rewritten to be a variant of LaTeX.

Now onto Wolfram Alpha terminology and Mathics. For the most part we should use that. However there are quirks and we sometimes deliberately deviate from that. 

In the documentation for builtin functions we use the Wikipedia name to describe the function. So under the WMA "Glaisher" function, in Mathics the title is "Glaisher-Kinklen" and for WMA EulerPhi the Mathics title is Euler's totient function. 

TeXForm is a _Form_ not a format. The Form and format are loosely coupled. In graphics we see that  sometime certain forms can go to PNG, SVG, threejs, matplotlib or something else.  Similarly TeXForm could go to MathML or LaTeX, the frontend may decide that it prefers Plain TeX for either MathMLForm, or TeXForm.

I think part of the reason this bothers me so much is that if we can't start to think clearly about concepts the code will be forever confused and someone trying to understand this will have a harder time.

Expect another PR when I get a chance, to revise all of this.

mmatera:
@rocky, if it is OK for you, I'm going to merge this as it is (a typo fixing). Then, if you are up for it, add a better description. Otherwise, I am going to leave this part for when I finish with the refactor (there are still several things to do).

rocky:
@mmatera You need not worry about this. Thanks for making the correction. At some point in the future when I am frustrated again with the vagueness and lack of clarity of thought, I'll probably put in a PR.

PR #534, opened 2022-08-26
Rename lower-level tex formatters to latex...

rocky:
since that is what it is, and we are striving for better clarity.

I am still being a little loose here since LaTeX extensions are ".tex" instead of ".ltx" or ".latex". And I not insisting on LaTeX instead of latex. (But, who knows that may happen someday).

For compatibility with outside projects, e.g. Django, a symbolic link from mathics/doc/tex is to mathics/doc/latex has been added.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/534"></a>


TiagoCavalcante (approved):
LGTM

mmatera:
I'm ok with the changes. However, regarding the question about TeX/LaTeX , I think it could be enlighten to read the Wikipedia article:

https://en.m.wikipedia.org/wiki/TeX

The reason why a LaTeX file has extension .tex is because the content of the file is TeX code, which loads certain packages which defines a LaTeX style. Is like a .py file that loads, let's say, the TK library for handling a graphics interface.

rocky:
> I'm ok with the changes. However, regarding the question about TeX/LaTeX , I think it could be enlighten to read the Wikipedia article:
> 
> https://en.m.wikipedia.org/wiki/TeX
> 
> The reason why a LaTeX file has extension .tex is because the content of the file is TeX code, which loads certain packages which defines a LaTeX style. Is like a .py file that loads, let's say, the TK library for handling a graphics interface.

At some level I understood that.  (It has been a couple of decades since I seriously used TeX or LaTeX.)

A distinction could and sometimes is made between the stuff implementing LaTeX, for example the styles which have a `.sty`. And you do find books and papers using the `.ltx` or `.latex` extension at the top level.

I totally understand that in Human language and communication there will always be some ambiguity. In of itself, it doesn't bother me that in the conventional use, "TeX" refers to: 

* a particular program written by Donald Knuth, 
* a Typesetting System (which includes the program and all the surrounding support stuff), 
* the original set of macros that were used in the initial implementation, and so on. 

Ditto for the fact that a `.tex` extension could mean several things as well. 

I should say that it doesn't bother me that this happens in the world at large. With respect to code that I work on and write, I am okay using one of these multifaceted meaning terms, _provided as one shows that one understands the difference, points that out and defines which TeX (or multi-faceted term) is meant_. 

_Begin Rant:_

Too often in Mathics we have vagueness which I think borders on illiteracy or over-bloatedness. Saying that the homegrown markup is "XML" is kind of laughable. Real XML comes with a schema definition. And handles arbitrary tags. The system in Mathics have everything hacked in as somewhat fragile regular expressions, and uses XML largely as though it were stripped down HTML. It is probably a quirk of my behavior, but kind of false puffery annoys me. For those who believe the hype, it also hinders getting rid of this and getting swapped out with something better. 

_End Rant_

WIth respect to the `.tex` extension, if there is some place that is natural  to  `.tex` extension is the way it is because that is a common convention and may have historical reason, please suggest a place and some wording. I am happy to leave it out. 

I recall when I first looked at Mathics wondering what was up with TeX versus LaTeX. With these changes, I suspect someone like me wouldn't get confused on looking at this the first time.

mmatera (approved):
Said that, LGTM

PR #535, opened 2022-08-27
Add 2d and 3d vector forms of Curl

rocky:
This is intended to match text in the development guide.

It might be interesting to expand Curl to handle those all four forms, but this is right, but this requires more skill than I can easily handle.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/535"></a>


TiagoCavalcante (approved):
LGTM

PR #539, opened 2022-08-29
boxes_to_ -> boxes_to_format.  Improving support for encodings.

mmatera:
This PR continues the refactor of the formatting methods, with some of the changes proposed here: https://github.com/Mathics3/mathics-core/discussions/317#discussioncomment-3450105

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/539"></a>

### DIFF
--- a/mathics/builtin/atomic/strings.py
+++ b/mathics/builtin/atomic/strings.py
@@ -28,7 +28,7 @@ from mathics.core.attributes import listable, protected
 from mathics.core.expression import Expression
 from mathics.core.convert.expression import to_mathics_list
 from mathics.core.convert.python import from_bool
-from mathics.core.formatter import format_element
+from mathics.core.formatter import boxes_to_format, format_element
 from mathics.core.list import ListExpression
 from mathics.core.parser import MathicsFileLineFeeder, parse
 from mathics.core.symbols import (
@@ -42,6 +42,8 @@ from mathics.core.systemsymbols import (
     SymbolDirectedInfinity,
     SymbolInputForm,
     SymbolOutputForm,
+    SymbolRule,
+    SymbolRuleDelayed,
 )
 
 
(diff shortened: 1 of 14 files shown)


mathics/builtin/atomic/strings.py
```diff
@@ -857,10 +859,33 @@ def apply_default(self, value, evaluation, options):
 
     def apply_form(self, value, form, evaluation, options):
         "ToString[value_, form_, OptionsPattern[ToString]]"
-        encoding = options["System`CharacterEncoding"]
+
+        # For some reason, the rule for `apply_default` is never used.
```
TiagoCavalcante, line 863: I added a print in `apply_default` and it is getting called.

TiagoCavalcante (changes requested):
LGTM - just a small thing

mathics/builtin/atomic/strings.py
mmatera: Exactly. This is about the order in which rules are tried. We can go back to this when the pattern matching routines be refactored.

TiagoCavalcante:
```py
def apply_default(self, value, evaluation, options):
        "ToString[value_, OptionsPattern[ToString]]"
        print(1)
        return self.apply_form(value, SymbolOutputForm, evaluation, options)

def apply_form(self, value, form, evaluation, options):
        "ToString[value_, form_, OptionsPattern[ToString]]"
        print(2)
        ...
```

```wl
In[1]:= ToString[1]
1
2
Out[1]= "1"

In[2]:= ToString[1, OutputForm]
2
Out[2]= "1"
```

mmatera: @TiagoCavalcante, do you want to add here an `asserter` line or something like that to detect when the `apply_default` start to be called?

mmatera: What is wrong is the following case:
```
ToString["hola", CharacterEncoding->"Unicode"]
```
for which I would expect that calls `apply_default`. However, I just check it, and the behavior in WL is the same than here:
```
In[1]:= F[a_, b_, OptionsPattern[]]:=ThreeParams[a, b, OptionValue[l]]

In[2]:=F[a_, OptionsPattern[]]:=TwoParams[a, OptionValue[l]]

In[3]:=F[r,l->2] 
```
produces `Out[3]= TwoParams[r, 2]`

So what is wrong here is the pattern of `apply_default` (should not have a parameter OptionsPattern).

TiagoCavalcante: @mmatera here is what I get:

```
In[1]:= ToString["hola", CharacterEncoding->"Unicode"]
1
2
Out[1]= "hola"
```

apply_default gets called first, as I expected.

In the case above what do you think it's the expected behavior?

About the assert, if it works as expected it isn't necessary, anyway, the just print works well.

mmatera: @TiagoCavalcante, the problem was with something like `ToString[expr, CharacterEncoding->"ASCII"]`. With the change in the rule form `apply_form`, now it works as it should.

mmatera: > @mmatera here is what I get:
> 
> ```
> In[1]:= ToString["hola", CharacterEncoding->"Unicode"]
> 1
> 2
> Out[1]= "hola"
> ```
> 
> apply_default gets called first, as I expected.
> 
> In the case above what do you think it's the expected behavior?
> 
> About the assert, if it works as expected it isn't necessary, anyway, the just print works well.

Interesting. Locally, and before the last change in the rule, I got just a call to `apply_form` ...

mathics/builtin/atomic/strings.py
```diff
@@ -42,6 +42,8 @@
     SymbolDirectedInfinity,
     SymbolInputForm,
     SymbolOutputForm,
+    SymbolRule,
```
TiagoCavalcante, line 45: I guess these can be removed now.

mathics/builtin/atomic/strings.py
TiagoCavalcante: @mmatera I checked the last version, and it is good now. It only calls `apply_default` if it is required to `form` be a `Symbol` in my tests.

TiagoCavalcante (approved):
LGTM - just remove the unused importes

TiagoCavalcante:

mmatera:
Outdated...

PR #540, opened 2022-08-29
Improving backward compatibility support for boxes_to_ methods in `BoxExpression` subclasses.

mmatera:
Improves the implementation for supporting `boxes_to_` methods in derived classes. Now, instead of looking for them online, the methods are loaded inside the `contribute` method of the class. This allows also overwritting the formatters from pymathics modules.


mathics/format/mathml.py
```diff
@@ -276,7 +277,7 @@ def stylebox(self, **options) -> str:
 def graphicsbox(self, elements=None, **options) -> str:
     # FIXME: SVG is the only thing we can convert MathML into.
     # Handle other graphics formats.
-    svg_body = self.boxes_to_svg(elements, **options)
+    svg_body = boxes_to_format(self, "svg", elements=elements, **options)
```
TiagoCavalcante, line 280: Why this change? I don't understand why this is done only in graphicsbox and not in other "boxes".

mathics/core/formatter.py
```diff
@@ -113,7 +113,15 @@ def boxes_to_format(boxes, format, **options) -> str:  # Maybe Union[str, bytear
     Translates a box structure ``boxes`` to a file format ``format``.
 
     """
-    return lookup_method(boxes, format)(boxes, **options)
+    try:
+        return lookup_method(boxes, format)(boxes, **options)
+    except RecursionError:
+        # This is just for debugging.
```
TiagoCavalcante, line 119: Is this supposed to be kept?

TiagoCavalcante (approved):
LGTM

mathics/format/mathml.py
mmatera: The idea is to replace all the calls of the form `self.boxes_to_[format](...)` by `boxes_to_format(self, "format", ...)`.  Have I  left anyone?

mathics/core/formatter.py
mmatera: For some reason, in mathics-Django, I got this exception. However, I was not able to determine why it was raised, and the traceback crashes my browser. Do you have a better idea to have the diagnostic?

TiagoCavalcante: Maybe a print?

mathics/format/mathml.py
TiagoCavalcante: Yes:
mathics/builtin/box/graphics3d.py line 337
mathics/core/element.py line 479, 483, 487
mathics/format/mathml.py line 322

mmatera: I guess all these changes were addressed in #539. I didn't include them here to avoid making this PR longer.

TiagoCavalcante:

mmatera:
I think this comes after the work that @rocky is doing on the loading mechanism.

mmatera:
Deprecated

PR #541, opened 2022-08-30
Character encoding in ToString

mmatera:
In this PR, basic support for different encodings in the `text` formatter is provided, using the functions in `mathics_scanner`. The support for the `CharacterEncoding` option in `ToString` is also improved.


mathics/builtin/atomic/strings.py
```diff
@@ -835,7 +835,16 @@ class ToString(Builtin):
...
+     = \int f\left[x\right] \, dx
+
+    By means of the option `CharacterEncoding` we can control
+    the way in which characters are encoding:
+    >> ToString["\[Integral]f(x)\[DifferentialD]x", CharacterEncoding->"UTF-8"]
+     = ∫f(x)𝑑x
+    When the option is set to "ASCII", Unicode characters are mapped
+    to their named form.
+    >> ToString["\[Integral]f(x)\[DifferentialD]x", CharacterEncoding->"ASCII"]
+     = \[Integral]f(x)𝑑x
```
mmatera, line 847: Notice that `𝑑` is not an ASCII character, and should be (in this line) `\[DifferentialD]`. This seems to be an issue from mathics_scanner.

rocky: Do a git blame and look at the commit log. (This is a reason why I spend time writing meaningful commit messages). 

I recall there was some back and forth on this - it was changed for a _reason_. I don't recall why and don't have the time to look it up.

mathics/format/text.py
```diff
@@ -31,7 +31,18 @@ def boxes_to_text(boxes, **options) -> str:
...
 def string(self, **options) -> str:
+    from mathics_scanner.characters import replace_wl_with_plain_text
+
     value = self.value
+    encoding = options.get("encoding", "Unicode")
+
+    # TODO: check if it makes sense to generalize this to
+    # other codepages...
+    if encoding in ("UTF8", "UTF-8"):
+        value = replace_wl_with_plain_text(value, True)
```
rocky, line 42: Note that `replace_with_plain_text()` from mathics_scanner is deprecated. If we are using this there is something missing.

mmatera: @rocky, yes, I know. What I tried here was to have something that works with the current version of (1.2.4) mathics_scanner. Later I am going to try writing a version that works with both. If you have an idea about how to do it, I would appreciate the help.

rocky: I have vague thoughts but I won't have time for a while to check or work things out for myself. 

I'll leaving to go hiking on Friday and will be back Monday.

mmatera: OK, there is no rush on solving this. I will continue with the other branches, which are more or less independent.

TiagoCavalcante:

mmatera:
Deprecated

PR #542, opened 2022-08-30
Improve latex formatter

mmatera:
This PR improves a little bit the latex formatter. Instead of a long if, now a dictionary is used. This is one step in the direction of using a table from mathics_scanner to assign the LaTeX representation to each operator. Right now, the corresponding table is not generated by MathicsScanner.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/542"></a>


test/format/test_format.py
```diff
@@ -387,8 +387,8 @@
             "System`OutputForm": "<mrow><mi>Integrate</mi> <mo>[</mo> <mrow><mrow><mi>F</mi> <mo>[</mo> <mi>x</mi> <mo>]</mo></mrow> <mtext>,&nbsp;</mtext> <mrow><mo>{</mo> <mrow><mi>x</mi> <mtext>,&nbsp;</mtext> <mi>a</mi> <mtext>,&nbsp;</mtext> <mrow><mi>g</mi> <mo>[</mo> <mi>b</mi> <mo>]</mo></mrow></mrow> <mo>}</mo></mrow></mrow> <mo>]</mo></mrow>",
         },
         "latex": {
-            "System`StandardForm": "\\int_a^{g\\left[b\\right]} F\\left[x\\right] \uf74cx",
-            "System`TraditionalForm": "\\int_a^{g\\left(b\\right)} F\\left(x\\right) \uf74cx",
+            "System`StandardForm": "\\int_a^{g\\left[b\\right]} F\\left[x\\right] \, dx",
+            "System`TraditionalForm": "\\int_a^{g\\left(b\\right)} F\\left(x\\right) \, dx",
```
TiagoCavalcante, line 391: I believe this should be `\U0001D451x`

https://github.com/Mathics3/mathics-scanner/blob/be4d6de65386e181dd68be3558bee23efdc29b43/mathics_scanner/data/named-characters.yml#L1607-L1615

TiagoCavalcante (approved):
LGTM - just a small thing

test/format/test_format.py
mmatera: There is a problem with it: the ci still uses mathics-scanner 1.2.4 so if we use that character, we get an error. Once mathics-scanner get released, we should adjust it.

TiagoCavalcante: Ok, sure

rocky:
LGTM as a stepping stone. Obviously though what we want is reading mathic-scanner tables which handles many more character symbols that we have here. But that is intertwined a little bit with what "unicode-equivalent" means.

mmatera:
> LGTM as a stepping stone. Obviously though what we want is reading mathic-scanner tables which handles many more character symbols that we have here. But that is intertwined a little bit with what "unicode-equivalent" means.

For this, I think there are two possibilities: one is to build the table in mathics-core, and when we get convinced about what is the best for the inner encoding, then we can move that table to mathics-scanner. On the other hand, as it is now if we add the "inner_to_ams_table"  to mathics_scanner, this should use it with the current code.

rocky:
> For this, I think there are two possibilities: one is to build the table in mathics-core, and when we get convinced about what is the best for the inner encoding, then we can move that table to mathics-scanner.

For mathics-core WL-unicode is fine. It just needs canonicalization when standard unicode appears as it usually does. How does one specify WL-unicode input versus a normal Unicode in WMA? 

Also, When I set unprotect and set `$CharacterEncoding = ASCII"` I still get Unicode equivalences. What's up here? 

>  On the other hand, as it is now if we add the "inner_to_ams_table" to mathics_scanner, this should use it with the current code.

Personally, I don't see the benefit of making more work here to complicating things, and duplicating information to have it possibly ripped out.

PR #544, opened 2022-09-05
List to python+roundtrip quotes

rocky:
In the Mathics python code, there are a number of places where we need to strip off double quotes. This should have been a red flag that something was wrong in the design of Mathics. Around Mathics 2.2.0, we added a "string_quotes" parameter to `to_python()` so that we could specify the behavior. 

However, more recently in getting rid of `get_string_value()` and replacing it with `value`, I see that `.get_string_value()`/ `.value` is another way that could have been used to not add quotes around the string in many of the offending cases that this occurs.

Note that  still need `to_python()`  when there are things like lists (or other composite structures) of strings and you want to specify whether or not to include quotes. 

This PR, simplifies string quote removal which is most noticeable in Mathics Builtins involving filesystem operations. It does this by simply using `.value` in place of `.to_python()`. 

The other thing it does, make from LIstExpressions  of literals have a round-trip behavior:  

```python
x = ['a', 'b', 'c']
assert to_mathics_list(*x).value == x
```

--- 

A long while ago I had said that we shouldn't worry about using `to_python()` over `get_string_value()` or `get_int_value()` and let Python figure that out. 

Here, I realize I have to amend that. Especially in light of unifying the `get_...value()` routines. (This PR adds `.value` on list expression literals). 

What I hadn't realized before is that `to_python()` is not invertible unless you specify `string_quotes=False` and there may be other situations like this.  Also `.value` is a common interpreter idiom.  And third (but in my mind _least_ important) is that fact that yes, it is faster too.  

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/544"></a>

### DIFF
--- a/mathics/core/element.py
+++ b/mathics/core/element.py
@@ -338,9 +338,7 @@ class BaseElement(KeyComparable):
         return rules
 
     def get_sequence(self) -> Union[tuple, list]:
-        """If self has return that, otherwise turn self into a
-        tuple() and return that"""
-
+        """Convert's a Mathics Sequence into a Python's list of elements"""
         from mathics.core.symbols import SymbolSequence
 
         # FIXME: using the below test causes:
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -261,9 +261,12 @@ class Expression(BaseElement, NumericOperators, EvalMixin):
         last_element = None
         values = []
         for element in self._elements:
-            # Test for the three properties mentioned above.
+            # Test for the literalness, and the three properties mentioned above
+            if element.is_literal:
+                values.append(element.value)
+            else:
+                self.elements_properties.elements_fully_evaluated = False
 
-            # First we test for not flatness, and elements_fully_evaluated
             if isinstance(element, Expression):
 
                 # "self" can't be flat.


mathics/builtin/files_io/filesystem.py
```diff
@@ -538,17 +534,16 @@ def apply_with_n(self, name, n, evaluation, options):
             py_n = 1
         else:
             expr = to_expression("DirectoryName", name, n)
-            py_n = n.to_python()
+            py_n = n.to_python(string_quotes=False)
```
TiagoCavalcante, line 537: Isn't `py_n` a Integer?

rocky: I just looked again at. When it is valid, yes it is an integer. However when it is invalid it could be a string. That said, ths keyword parameter doesn't do anything, so I have removed. it in f44e8ae2589cee

mathics/builtin/arithmetic.py
```diff
@@ -82,7 +82,7 @@
 )
 
 
-@lru_cache(maxsize=1024)
+@lru_cache(maxsize=4096)
```
mmatera, line 85: Why this number?

mathics/builtin/base.py
```diff
@@ -763,10 +775,8 @@ def is_literal(self) -> bool:
         depend on definition bindings. That is why, in contrast to
         `is_uncertain_final_definitions()` we don't need a `definitions`
         parameter.
-
-        Think about: We will say that a BoxExpression can't change.
         """
-        return True
+        return False
```
mmatera, line 779: Why a `BoxExpression` shouldn´t be handled as a literal?

mathics/builtin/arithmetic.py
rocky: Experiments show that we are exceeding 1024.

mathics/builtin/base.py
rocky: From line 187 in this PR of element.py: 

> - *literal_values - optional: if this is not None, then all elements are (Python) literal values and literal_value  contains these literals.

mmatera: OK, I see.

mathics/builtin/files_io/importexport.py
```diff
@@ -1366,10 +1368,10 @@ def apply_element(self, filename, element, evaluation, options={}):
         )
 
     def apply_elements(self, filename, elements, evaluation, options={}):
-        "Import[filename_, elements_List?(AllTrue[#, NotOptionQ]&), OptionsPattern[]]"
+        "Import[filename_String, elements_List?(AllTrue[#, NotOptionQ]&), OptionsPattern[]]"
         # Check filename
-        path = filename.to_python()
-        if not (isinstance(path, str) and path[0] == path[-1] == '"'):
+        path = filename.value
+        if not isinstance(path, str):
```
TiagoCavalcante, line 1374: I don't think this can happen anymore

mathics/core/element.py
```diff
@@ -337,10 +337,16 @@ def get_rules_list(self):
...
-    def get_sequence(self):
-        """Convert's a Mathics Sequence into a Python's list of elements"""
+    def get_sequence(self) -> Union[tuple, list]:
+        """If self has return that, otherwise turn self into a
+        tuple() and return that"""
+
         from mathics.core.symbols import SymbolSequence
 
+        # FIXME: using the below test causes:
+        # TypeError: boxes_to_text() takes 1 positional argument but 2 were given
```
mmatera, line 347: which test?

mathics/builtin/files_io/importexport.py
rocky: You are correct that something isn't needed. But it is `filename.value`.  We should be checking to see that `filename` is a `String`. 

And I have changed the docstring signature to remove `_String` as well. The purpose of the `if not isinstance(path, str):` is to give an error message if the first argument is not a string. By removing that, we would remove the check altogether and silently return. That apparently was not desired. 

Also this method can be called from another method, and when that happens it defeats the signature check. 

Down the line, I have in mind revamping how all of this should work to again make it more scalable and performant. 
But this is another big change.

mathics/core/element.py
rocky: > which test?

```
pyston -m pytest test -x
=========================================== test session starts ============================================
platform linux -- Python 3.8.12, pytest-7.1.2, pluggy-1.0.0
rootdir: /src/external-vcs/github/Mathics3/mathics-core
collected 0 items / 1 error                                                                                

================================================== ERRORS ==================================================
__________________________________ ERROR collecting test/test_calculus.py __________________________________
test/test_calculus.py:6: in <module>
    from .helper import check_evaluation, session
test/helper.py:6: in <module>
    session = MathicsSession()
mathics/session.py:58: in __init__
    self.reset(add_builtin, catch_interrupt)
mathics/session.py:64: in reset
    self.definitions = Definitions(add_builtin)
mathics/core/definitions.py:137: in __init__
    autoload_files(self, ROOT_DIR, "autoload")
mathics/core/definitions.py:56: in autoload_files
    Expression(SymbolGet, String(path)).evaluate(Evaluation(defs))
mathics/core/expression.py:491: in evaluate
    expr, reevaluate = expr.rewrite_apply_eval_step(evaluation)
mathics/core/expression.py:1242: in rewrite_apply_eval_step
    result = rule.apply(new, evaluation, fully=False)
mathics/core/rules.py:90: in apply
    self.pattern.match(yield_match, expression, {}, evaluation, fully=fully)
mathics/core/pattern.py:351: in match
    self.head.match(yield_head, expression.get_head(), vars, evaluation)
mathics/core/pattern.py:193: in match_symbol
    yield_func(vars, None)
mathics/core/pattern.py:341: in yield_head
    self.get_pre_choices(
mathics/core/pattern.py:489: in get_pre_choices
    yield_func(vars)
mathics/core/pattern.py:318: in yield_choice
    self.match_element(
mathics/core/pattern.py:702: in match_element
    self.get_wrappings(
mathics/core/pattern.py:518: in get_wrappings
    yield_func(items[0])
mathics/core/pattern.py:690: in yield_wrapping
    element.match(
mathics/builtin/patterns.py:1023: in match
    self.pattern.match(yield_func, expression, new_vars, evaluation)
mathics/builtin/patterns.py:1228: in match
    yield_func(vars, None)
mathics/core/pattern.py:687: in match_yield
    yield_func(new_vars, items_rest)
mathics/core/rules.py:62: in yield_match
    new_expression = self.do_replace(expression, vars, options, evaluation)
mathics/core/rules.py:223: in do_replace
    return self.function(evaluation=evaluation, **vars_noctx)
mathics/builtin/files_io/files.py:453: in apply_default
    evaluation.message("General", "stream", filename)
mathics/core/evaluation.py:535: in message
    text = self.format_output(
mathics/core/evaluation.py:476: in format_output
    boxes = result.boxes_to_text(evaluation=self)
mathics/core/element.py:493: in boxes_to_text
    return self.boxes_to_format("text", **options)
mathics/core/element.py:481: in boxes_to_format
    return boxes_to_format(self, format, **options)
mathics/core/formatter.py:116: in boxes_to_format
    return lookup_method(boxes, format)(boxes, **options)
mathics/core/formatter.py:135: in ret_fn
    return boxes_to_method(elements, **opts)
E   TypeError: boxes_to_text() takes 1 positional argument but 2 were given
============================================= warnings summary =============================================
```

mmatera:
> In the Mathics python code, there are a number of places where we need to strip off double quotes. This should have been a red flag that something was wrong in the design of Mathics. Around Mathics 2.2.0, we added a "string_quotes" parameter to `to_python()` so that we could specify the behavior.
> 


makes a lot of sense...

mathics/core/element.py
mmatera: The error actually happens because another error was raised before formatters are loaded. Check it by applying the patch

```
diff --git a/mathics/core/definitions.py b/mathics/core/definitions.py
index 47d0d0e9..87970609 100644
--- a/mathics/core/definitions.py
+++ b/mathics/core/definitions.py
@@ -108,6 +108,11 @@ class Definitions:
         )
         self.trace_evaluation = False
         self.timing_trace_evaluation = False
+
+        # Inport formats before builtins, in case
+        # we need to show a message.
+        import mathics.format
+
         if add_builtin:
             from mathics.builtin import modules, contribute
             from mathics.settings import ROOT_DIR
@@ -153,7 +158,6 @@ class Definitions:
             self.clear_cache()
 
         # This loads all the formatting functions
-        import mathics.format
 
         # FIXME load dynamically as we do other things
         # import mathics.format.asy  # noqa
```

rocky: Thanks for tracking this down. My understanding is that `mathics.format` was one of the main things that was getting worked on right now. 

(Not scanner stuff, Precision, or Box formatting of which two of these areas stray a bit outside of the kinds of things you are good at.) 

I would be grateful if you would do the needful here. Thanks.

mmatera: > Thanks for tracking this down. My understanding is that `mathics.format` was one of the main things that was getting worked on right now.
> 
> (Not scanner stuff, Precision, or Box formatting of which two of these areas stray a bit outside of the kinds of things you are good at.)
> 

Again, I am aware of my limitations. However, in order to make work some simple and appealing tests, I need also that these parts work properly. 

> I would be grateful if you would do the needful here. Thanks.

I am on the way. I also put some other PRs in that direction.

PR #545, opened 2022-09-08
Fixing parsing for FormBox

mmatera:
This PR starts to fix the issues enumerated (right after) the docstring of `MakeBoxes` (https://github.com/Mathics3/mathics-core/blob/a90d0ad493c7324f2c797b811e7375e802609220/mathics/builtin/makeboxes.py#L374)

In this first round, the behavior of `FormBox` is fixed in a way to  make it closer to the WL expected behavior:

* ``\(a_(String|Real|Integer) ` b_ \)`` is parsed as `FormBox[ToBoxes[b], Removed[$$Failure]]` (notice that the argument of `Removed` is a `Symbol` not a `String`).
* ``\(a_Symbol \` b_ \)`` is parsed as `FormBox[ToBoxes[b], a]`
* Otherwise, ``\(a_ \` b_ \)`` is parsed as `FormBox[b, ToBoxes[a]]`

Tests were adjusted  accordingly.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/545"></a>

### DIFF
--- a/mathics/builtin/makeboxes.py
+++ b/mathics/builtin/makeboxes.py
@@ -369,13 +369,10 @@ class MakeBoxes(Builtin):
     #> (a <> b)[x]
      : String expected.
      = (a <> b)[x]
-    """
-
-    # TODO: Convert operators to appropriate representations e.g. 'Plus' to '+'
-    """
     >> \\(a + b\\)
      = RowBox[{a, +, b}]
 
+    This fails because is not properly parsed. Check in Mathics-scanner.
     >> \\(TraditionalForm \\` a + b\\)
      = FormBox[RowBox[{a, +, b}], TraditionalForm]
 
--- a/mathics/builtin/makeboxes.py
+++ b/mathics/builtin/makeboxes.py
@@ -372,7 +372,6 @@ class MakeBoxes(Builtin):
     >> \\(a + b\\)
      = RowBox[{a, +, b}]
 
-    This fails because is not properly parsed. Check in Mathics-scanner.
     >> \\(TraditionalForm \\` a + b\\)
      = FormBox[RowBox[{a, +, b}], TraditionalForm]
 
--- a/mathics/builtin/makeboxes.py
+++ b/mathics/builtin/makeboxes.py
@@ -327,38 +327,38 @@ class BoxForms_(Predefined):
 
 
 class MakeBoxes(Builtin):
-    """
+    r"""
     <dl>
       <dt>'MakeBoxes[$expr$]'
-      <dd>is a low-level formatting primitive that converts $expr$
-        to box form, without evaluating it.
-      <dt>'\\( ... \\)'
+      <dd>is a low-level formatting primitive that converts $expr$ to box form, without evaluating it.
+
+      <dt>'\( ... \)'
       <dd>directly inputs box objects.
     </dl>
 
     String representation of boxes
-    >> \\(x \\^ 2\\)
+    >> \(x \^ 2\)
      = SuperscriptBox[x, 2]
 
-    >> \\(x \\_ 2\\)
+    >> \(x \_ 2\)
      = SubscriptBox[x, 2]
 
-    >> \\( a \\+ b \\% c\\)
+    >> \( a \+ b \% c\)
      = UnderoverscriptBox[a, b, c]
 
-    >> \\( a \\& b \\% c\\)
+    >> \( a \& b \% c\)
      = UnderoverscriptBox[a, c, b]
 
-    #> \\( \\@ 5 \\)
+    #> \( \@ 5 \)
      = SqrtBox[5]
 
-    >> \\(x \\& y \\)
+    >> \(x \& y \)
      = OverscriptBox[x, y]
 
-    >> \\(x \\+ y \\)
+    >> \(x \+ y \)
      = UnderscriptBox[x, y]
 
-    #> \\( x \\^ 2 \\_ 4 \\)
+    #> \( x \^ 2 \_ 4 \)
      = SuperscriptBox[x, SubscriptBox[2, 4]]
 
     ## Tests for issue 151 (infix operators in heads)
@@ -369,13 +369,13 @@ class MakeBoxes(Builtin):
     #> (a <> b)[x]
      : String expected.
      = (a <> b)[x]
-    >> \\(a + b\\)
+    >> \(a + b\)
      = RowBox[{a, +, b}]
 
-    >> \\(TraditionalForm \\` a + b\\)
+    >> \(TraditionalForm \` a + b\)
      = FormBox[RowBox[{a, +, b}], TraditionalForm]
 
-    >> \\(x \\/ \\(y + z\\)\\)
+    >> \(x \/ \(y + z\)\)
      =  FractionBox[x, RowBox[{y, +, z}]]
     """
 
--- a/mathics/builtin/makeboxes.py
+++ b/mathics/builtin/makeboxes.py
@@ -309,7 +309,7 @@ def number_form(expr, n, f, evaluation, options):
 
 # TODO: Differently from the current implementation, MakeBoxes should only
 # accept as its format field the symbols in `$BoxForms`. This is something to
-# fix in a following step.
+# fix in a following step, changing the way in which Format and MakeBoxes work.
 
 
 class BoxForms_(Predefined):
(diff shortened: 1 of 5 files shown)


mathics/core/parser/parser.py
```diff
@@ -308,13 +308,35 @@ def p_RawLeftAssociation(self, token):
...
-        self.bracket_depth += 1
+        # If this does not happend, it would be because
+        # it was called when a `FormBox` was found.
+        if token.tag == "LeftRowBox":
+            self.box_depth += 1
+            self.bracket_depth += 1
+
         token = self.next()
         while token.tag not in ("RightRowBox", "OtherscriptBox"):
+            if token.tag in "FormBox":
```
TiagoCavalcante, line 319: Should this be in the `while` condition?

TiagoCavalcante (approved):
LGTM

mathics/core/parser/parser.py
mmatera: True!

mathics/builtin/makeboxes.py
```diff
@@ -369,10 +369,6 @@ class MakeBoxes(Builtin):
     #> (a <> b)[x]
      : String expected.
      = (a <> b)[x]
-    """
-
-    # TODO: Convert operators to appropriate representations e.g. 'Plus' to '+'
-    """
     >> \\(a + b\\)
```
rocky, line 372: I have been looking at this and notice a couple of small formatting changes. If this docstring were marked as a raw string, e.g. `r""" ... """` we would need all of the ugly and confusing double backslashes `\\` everywhere..

On line 333: 

```
      <dd>is a low-level formatting primitive that converts $expr$
        to box form, without evaluating it.
```

has a hard line break which the current homegrown formatter doesn't ignore. And it would be nice to add an extra space between and end `</dt>` and the next starting `<dd>`. 

These aren't strictly about this PR, but I happen to notice since you happen to be in the area.

mathics/core/parser/parser.py
```diff
@@ -308,13 +308,33 @@ def p_RawLeftAssociation(self, token):
     def p_LeftRowBox(self, token):
         self.consume()
         children = []
-        self.box_depth += 1
-        self.bracket_depth += 1
+        # If this does not happend, it would be because
```
rocky, line 311: happend -> happen

mathics/core/parser/parser.py
```diff
@@ -308,13 +308,33 @@ def p_RawLeftAssociation(self, token):
...
+        # it was called when a `FormBox` was found.
+        if token.tag == "LeftRowBox":
+            self.box_depth += 1
+            self.bracket_depth += 1
```
rocky, line 315: Every place there is some depth that increases by one, there should be another place that decreases the depth by one. 

Where is that?

test/core/parser/test_parser.py
```diff
@@ -694,10 +694,14 @@ def testFraction(self):
         self.check("\\(   \\/   \\)", 'FractionBox["", ""]')
 
     def testFormBox(self):
-        self.check("\\( 1 \\` b \\)", 'FormBox["b", Removed["$$Failure"]]')
         self.check("\\( \\` b \\)", 'FormBox["b", StandardForm]')
         self.check("\\( a \\` b \\)", 'FormBox["b", a]')
         self.check("\\( a \\` \\)", 'FormBox["", a]')
+        self.check("\\( a \\` b + c \\)", 'FormBox[RowBox[{"b", "+", "c"}], a]')
```
rocky, line 700: A small thing but I would appreciate it if you' use raw strings here as well. Thanks.

rocky:
I am not seeing this `Removed[$$Failure]` thing described above. I see [`$Failed`]( https://reference.wolfram.com/language/ref/$Failed.html) though.  Please educate me here.

mathics/core/parser/parser.py
```diff
@@ -840,20 +860,6 @@ def b_FractionBox(self, box1, token, p):
         box2 = self.parse_box(q + 1)
         return Node("FractionBox", box1, box2)
 
-    def b_FormBox(self, box1, token, p):
```
rocky, line 843: Why was this removed?

rocky:
I am having trepidations about this, and worry that a myopic approach like this, if it can work may lead to a mess, if it can be done. It may be like trying to solve a Rubik's cube face by face instead of understanding the natural subgroups imposed by turns of the cube and reducing those instead.

mathics/core/parser/parser.py
mmatera: If the tat is not ```LeftRowBox```  then we never reach the decrement.

mathics/core/parser/parser.py
mmatera: Because with the current logic, to implement the behavior of `FormBox` is not possible using this mechanism, so in my proposal, this method is never used.

mathics/core/parser/parser.py
```diff
@@ -308,13 +308,33 @@ def p_RawLeftAssociation(self, token):
...
+            if len(children) == 0:
+                fmt = Symbol("StandardForm")
+            elif len(children) == 1:
+                fmt = children[0]
+                if type(fmt) is String:
+                    fmt_name = fmt.value
+                    if is_symbol_name(fmt_name):
+                        fmt = Symbol(fmt_name)
+                    else:
+                        fmt = Node("Removed", Symbol("$$Failure"))
```
mmatera, line 333: @rocky,you are right: it should be 
```
fmt = Node("Removed", String("$$Failure"))
```
to be compatible with WMA.

For example:

```
In WMA, if `FormBox` receives wrong arguments (a Number as a format for instance) the interpreter produces
```
In[1]:= \(3 \` + b\)                                                            

Out[1]= FormBox[RowBox[{+, b}], Removed[$$Failure]]

In[2]:= %//FullForm                                                             

Out[2]//FullForm= FormBox[RowBox[List["+", "b"]], Removed["$$Failure"]]
```

File: mathics/core/parser/parser.py (hunk 6 above)
Comment (rocky) [reply mathics/core/parser/parser.py:843]: Yes, this is what I mean about the myopic kind of thing. FormBox needs to worry about precedence and the `b_` routines are where this kind of thing is done. So if you are defeating that, then maybe the thinking around this is flawed.

File: mathics/core/parser/parser.py (hunk 4 above)
Comment (rocky) [reply mathics/core/parser/parser.py:315]: You are saying something more or less like what the code says. Why is the decrement not reached. What is the depth used for?

File: mathics/core/parser/parser.py (hunk 3 above)
Comment (mmatera) [reply mathics/core/parser/parser.py:311]: I try to explain the problem in the comments, but it seems it is not easy. Here we go again: the problem is that
```\( a b \` c + d \)```
should be parsed as 
```FormBox[RowBox[{"c", "+", "d"}, RowBox[{"a", "b"}]]```

Then, to build the expression, we need to know all the tokens before ```\` ```, to build the "Format" expression, and then collect the following elements in the expression as the first argument. With the current parsing algorithm, I cannot make to `b_FormBox` what was the previous tokens. 
The alternative I found without changing the algorithm was to hack just `p_RawLeftAssociation`. If a ``` \` ```  token is found, then call again `p_RawLeftAssociation` but with a different tag. In this case, `self.box_deep` and `self.bracket_deep` does not change their values, unless a nested `Raw(Left|Right)Association` pair is found.

Comment (mmatera) [reply mathics/core/parser/parser.py:311]: As far as I could see, the unique case with this special behavior was `FormBox`. If there were more tags with a similar behavior, where to parsing the next token we need to look at several previous ones and reinterpret them, 
 I would consider rework the parser.

**Comment by mmatera** [comment]
> I am having trepidations about this, and worry that a myopic approach like this, if it can work may lead to a mess, if it can be done. It may be like trying to solve a Rubik's cube face by face instead of understanding the natural subgroups imposed by turns of the cube and reducing those instead.

Indeed, I have a similar feeling. This is why I tried to implement this in the most locally possible way.  A broader approach would require to share with the `p_*` methods all the previous tokens, or some other more elaborated mechanism.

File: mathics/core/parser/parser.py (hunk 4 above)
Comment (mmatera) [reply mathics/core/parser/parser.py:315]: > You are saying something more or less like what the code says. Why is the decrement not reached. What is the depth used for?

The deep was used to handle nested expressions like ```\(  a b  \( InputForm \`   \(OutputForm \` c d\) e  \) \) ```

When a ``` \` ``` is found, then the previous tokens are collected, and evaluated as a `Symbol`, a  Box expression or `StandardForm` if there is no token before ``` \` ```. The result of this step is stored in a variable, to be used as the second argument of `FormBox`. 
 Then, a new call to the method is done, with `FormBox` as a tag.  In the new call, the next elements are processed until the bracket and box deeps reach 0. Then, these elements are collected into a `String` or a `RowBox`. Then, the instance where ``` \` ``` was found takes the format and the other collected elements, builds a ```FormBox[]```, and returns it (before reaching the line where the decrements happen). 


I tried to find a way to put the increment and the decrement together, but it just was more involved than this approach: I should have allowed the increment of the deep variables, and then doing the decrement before returning the FormBox. I didn't think that it was clearer.

File: mathics/core/parser/parser.py (hunk 6 above)
Comment (mmatera) [reply mathics/core/parser/parser.py:843]: Yes, I was aware about the possible issue with the precedence. So I checked it against WMA, and it also seems to neglect the operator precedence of ``` \` ```, so it gives the maximum precedence to it. On the other hand, the tokens that follows ``` \` ``` are processed using all the precedence rules. 


What makes me think that this is the right approach (or at least a better approach than the current one) is that it does not break the existing tests, and it just makes a difference if a ``` \`   ``` token  is found inside a ```LeftRowBracket```/```RightRowBracket``` expression.

File: test/core/parser/test_parser.py (hunk 5 above)
Comment (mmatera) [reply test/core/parser/test_parser.py:700]: I just followed the convention used in this file. But I agree, it would be clearer with raw strings.

File: mathics/builtin/makeboxes.py (hunk 2 above)
Comment (mmatera) [reply mathics/builtin/makeboxes.py:372]: Indeed. Also, I was thinking about to patching the homegrown formatter to handle these line breaks, and then avoiding that flake8 complains about long lines. But one step at the time...

Comment (mmatera) [reply mathics/builtin/makeboxes.py]: Fixed in [f6f3d1f](https://github.com/Mathics3/mathics-core/pull/545/commits/f6f3d1fc29ed338d92829377f77d6a521e099129)

File: mathics/core/parser/parser.py (hunk 3 above)
Comment (mmatera) [reply mathics/core/parser/parser.py]: I tried to make this more explicit in an extended comment  in
[f6f3d1f](https://github.com/Mathics3/mathics-core/pull/545/commits/f6f3d1fc29ed338d92829377f77d6a521e099129)

Also, I changed the comparison as @rocky suggested.

File: mathics/core/parser/parser.py (hunk 4 above)
Comment (mmatera) [reply mathics/core/parser/parser.py:320]: I hope the new comment helps to understand the logic.

File: test/core/parser/test_parser.py (hunk 5 above)
Comment (mmatera) [reply test/core/parser/test_parser.py]: Done in [f6f3d1f](https://github.com/Mathics3/mathics-core/pull/545/commits/f6f3d1fc29ed338d92829377f77d6a521e099129)

File: mathics/core/parser/parser.py (hunk 7 above)
Comment (mmatera) [reply mathics/core/parser/parser.py]: Fixed in [f6f3d1f](https://github.com/Mathics3/mathics-core/pull/545/commits/f6f3d1fc29ed338d92829377f77d6a521e099129)

File: mathics/core/parser/parser.py (hunk 4 above)
Comment (rocky) [reply mathics/core/parser/parser.py:320]: Thanks for the description. Something about this seems a little to complicated and that there should be something that follows the specification in a little more direct way: 

>  * In [InputForm](https://reference.wolfram.com/language/ref/InputForm.html) and [StandardForm](https://reference.wolfram.com/language/ref/StandardForm.html), \(form∖`input∖) yields [FormBox](https://reference.wolfram.com/language/ref/FormBox.html)[input,form].
> * ∖(∖`input∖) yields [FormBox](https://reference.wolfram.com/language/ref/FormBox.html)[input,RawForm].

I would like to think about and reflect on this some more. I think we can make this cleaner, clearer and simpler.

What would be ideal is if we could do a cleanup step first - no changes to behavior yet. 
Then add the list of tests that are wrong and that we want to fix. 

And after this, _then_ come up with a way to implement this.

Comment (mmatera) [reply mathics/core/parser/parser.py:320]: > > * ∖(∖`input∖) yields [FormBox](https://reference.wolfram.com/language/ref/FormBox.html)[input,RawForm].
> 

Actually, in WMA,
```
In[1]:= \(\`input\)                                                             

Out[1]= FormBox[input, StandardForm]
```



> I would like to think about and reflect on this some more. I think we can make this cleaner, clearer and simpler.
> 

I am sure it is possible, just that after trying for a while, I didn't find a better way. 

> What would be ideal is if we could do a cleanup step first - no changes to behavior yet. Then add the list of tests that are wrong and that we want to fix.
> 

What would be a cleanup step? The list of (simple) tests to check are the tests that I added in test/core/parser/test_parser.py. 


> And after this, _then_ come up with a way to implement this.

OK, in that case, I am going to put this as a draft for a while.

rocky:
> What would be a cleanup step? The list of (simple) tests to check are the tests that I added in test/core/parser/test_parser.py.

The changes to `mathics/builtin/makeboxes.py`, `test/core/parser/test_parser.py` with the tests that are currently broken marked with `pytest.skip.markif` and the change to `CHANGES.rst` that is a mistake from a prior commit, not the line that mentions adding a new feature. 

All changes to `mathics/core/parser/parser.py` I would not include but keep here in the branch. This is the code that should be isolated, understood better and possibly reorganized. 

Thanks for your patience and understanding.

mmatera:
@rocky, this is the PR that I remember.

rocky:
> @rocky, this is the PR that I remember.

I looked at this again, and I my opinion and view of this has not changed. I am not convinced there is a parsing problem but rather a problem somewhere outside of parsing.

PR #546, opened 2022-09-09
Use Integer 0 when we have a relative precision of the value 0

mmatera:
In WL, ```n`p``` represents a number `n` with precision  `p`, which means that the number is defined up to an error `dn ~ n * 10^-p`. In particular, in WMA and in Mathics,

```
In[1]:=0.`5
```
is parsed as ```Out[1]=0.x 10^-5```. On the other hand, in WMA,  ```0`5``` (without the decimal point) is parsed as `0` (an `Integer`), but in the current master of Mathics, it is parsed also as   `0. x 10^-5`. This PR changes this behavior to make it compatible with the expected in WMA. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/546"></a>

### DIFF
--- a/mathics/core/parser/convert.py
+++ b/mathics/core/parser/convert.py
@@ -114,6 +114,9 @@ class GenericConverter:
                     prec10,
                 )
             else:
+                # "0`3"->0
+                if node.value == "0":
+                    return "Integer", 0
                 return (
                     "PrecisionReal",
                     ("DecimalString", str("-" + s if sign == -1 else s)),
--- a/mathics/core/parser/convert.py
+++ b/mathics/core/parser/convert.py
@@ -102,8 +102,13 @@ class GenericConverter:
             elif suffix == "":
                 return "MachineReal", sign * float(s)
             elif suffix.startswith("`"):
+                # A double Reversed Prime ("``") represents a fixed accuracy
+                # (absolute uncertainty).
                 acc = float(suffix[1:])
                 x = float(s)
+                # For 0, a finite absolute precision even if
+                # the number is an integer, it is stored as a
+                # PrecisionReal number.
                 if x == 0:
                     prec10 = acc
                 else:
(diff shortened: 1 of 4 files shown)


test/core/parser/test_convert.py
```diff
@@ -78,7 +78,10 @@ def testReal(self):
         self.check("1.5`", Real("1.5"))
         self.check("0.0", Real(0))
         self.check("-1.5`", Real("-1.5"))
-
+        self.check("0`3", Integer(0))
+        self.check("0``3", "0.000`3")
```
TiagoCavalcante, line 82: Why this one contains 2 "`"?

mmatera: 2 "`" represents an accuracy on 2, meaning that the number has an absolute uncertainty dx=0.01

mmatera: An `Integer` 0 with an absolute uncertainty is represented as a `PrecisionReal` with that uncertainty, while if the uncertainty is relative, the absolute uncertainty vanish, and hence, it is an Integer.

mathics/core/parser/convert.py
```diff
@@ -114,6 +114,9 @@ def convert_Number(self, node: AST_Number) -> tuple:
                     prec10,
                 )
             else:
+                # "0`3"->0
```
rocky, line 117: This comment is cryptic. The explanation given in the example  is clearer:, a relative uncertainty value times zero is zero.  

Personally, I stay away from words like more mysterious and complicated sounding words "vanish" when I want to convey something simple and definite like the product of zero and a number is zero. 

In research and academia you come across this feature where instead of saying something simply that everyone can easily understand, it is couched in more complicated words. This is why I think Richard Feynman was so popular - he could explain things in a simple manner.

rocky:
LGTM - see comment though, and you may want to add something to CHANGES.rst.

mathics/core/parser/convert.py
mmatera: Actually, I left this because I supposed that there was a place in the documentation where this notation for absolute and relative errors was established, but no, it wasn't.

rocky: Ok - I am sorry. But I'd appreciate it if would remove the crypticness in this code when it appears. Thanks.

mmatera: Now I improved the comments and added a new section in the documentation. Sadly, nice examples like

```
Pi == 3.1416`4
```
does not work by now...

mmatera: When you have time, please check the redaction of the new section in the documentation.

rocky:
@mmatera - greatly improved. This is greatly improved.

PR #548, opened 2022-09-12
earlier load of formatters

mmatera:
This PR moves the loading of formatter functions at an early position in the `mathics.core.definitions` module. This fixes the issue mentioned in https://github.com/Mathics3/mathics-core/blob/2a716c9f332555032fc7aafc3855dcfc9b6a809f/mathics/core/element.py#L344

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/548"></a>

### DIFF
--- a/mathics/core/definitions.py
+++ b/mathics/core/definitions.py
@@ -108,6 +108,12 @@ class Definitions:
         )
         self.trace_evaluation = False
         self.timing_trace_evaluation = False
+
+        # This loads all the formatting functions.
+        # Needs to be early because it can be used in
+        # messages during the builtins loading.
+        import mathics.format
+
         if add_builtin:
             from mathics.builtin import modules, contribute
             from mathics.settings import ROOT_DIR
@@ -151,14 +157,6 @@ class Definitions:
             self.builtin.update(self.user)
             self.user = {}
             self.clear_cache()
-
-        # This loads all the formatting functions
-        import mathics.format
-
-        # FIXME load dynamically as we do other things
-        # import mathics.format.asy  # noqa
-        # import mathics.format.json  # noqa
-        # import mathics.format.svg  # noqa
 
     def load_pymathics_module(self, module, remove_on_quit=True):
         """
--- a/mathics/core/element.py
+++ b/mathics/core/element.py
@@ -341,10 +341,6 @@ class BaseElement(KeyComparable):
         """Convert's a Mathics Sequence into a Python's list of elements"""
         from mathics.core.symbols import SymbolSequence
 
-        # FIXME: using the below test causes:
-        # TypeError: boxes_to_text() takes 1 positional argument but 2 were given
-        # Why?
-        # if hasattr(self, "element"):
         if self.get_head() is SymbolSequence:
             return self.elements
         else:


rocky:
It would be neat to make the change in https://github.com/Mathics3/mathics-core/blob/2a716c9f332555032fc7aafc3855dcfc9b6a809f/mathics/core/element.py#L344

where the commented-out `if` expression is used instead of the one that is used (and remove the comment about how this is broken).  Thanks.

mmatera:
> It would be neat to make the change in
> 
> https://github.com/Mathics3/mathics-core/blob/2a716c9f332555032fc7aafc3855dcfc9b6a809f/mathics/core/element.py#L344
> 
> where the commented-out `if` expression is used instead of the one that is used (and remove the comment about how this is broken). Thanks.

OK. However, I also find why this raise an error once the formatting issue is fixed: `get_sequence` should return a list of the elements of the expression just if the head of the expression is a `Sequence`. Otherwise, it should return a list with a single element: self.

Consider for example the expressions `List[x,1,2]`  and `Sequence[x,1,2]`. In the first case, `get_sequence` should return
```
[ListExpression(Symbol("Global`x), Integer1, Integer2)]
``` 
while in the second case

```
[Symbol("Global`x), Integer1, Integer2]
```
The difference is important for functions that accept an arbitrary number of parameters. For example,

```
    def apply(x, evaluation):
        """F[x__]"""
        args = x.get_sequence()
```

Then if the expression ```F[a,b,c]``` is evaluated, the `apply` method get in `args`  `[Symbol(a),Symbol(b),Symbol(c)]`.
On the other hand, if we evaluate ```F[{a,b,c}]```, `apply` receives in `args` `[ListExpression[Symbol(a),Symbol(b),Symbol(c)]]`.

mmatera:
Probably, `Sequence` should be also handled as a subclass of `Expression`, just like `ListExpression`, and this method (`get_sequence`) should belong just to this subclass.

rocky:
> > It would be neat to make the change in
> > https://github.com/Mathics3/mathics-core/blob/2a716c9f332555032fc7aafc3855dcfc9b6a809f/mathics/core/element.py#L344
> > 
> > where the commented-out `if` expression is used instead of the one that is used (and remove the comment about how this is broken). Thanks.
> 
> OK. However, I also find why this raise an error once the formatting issue is fixed: `get_sequence` should return a list of the elements of the expression just if the head of the expression is a `Sequence`. Otherwise, it should return a list with a single element: self.
> 
> Consider for example the expressions `List[x,1,2]` and `Sequence[x,1,2]`. In the first case, `get_sequence` should return
> 
> ```
> [ListExpression(Symbol("Global`x), Integer1, Integer2)]
> ```
> 
> while in the second case
> 
> ```
> [Symbol("Global`x), Integer1, Integer2]
> ```
> 
> The difference is important for functions that accept an arbitrary number of parameters. For example,
> 
> ```
>     def apply(x, evaluation):
>         """F[x__]"""
>         args = x.get_sequence()
> ```
> 
> Then if the expression `F[a,b,c]` is evaluated, the `apply` method get in `args` `[Symbol(a),Symbol(b),Symbol(c)]`. On the other hand, if we evaluate `F[{a,b,c}]`, `apply` receives in `args` `[ListExpression[Symbol(a),Symbol(b),Symbol(c)]]`.

I am grateful for the detailed explanation. Given this, the docstring for `get_sequence()` should be updated to include this important and useful information.

rocky:
> Probably, `Sequence` should be also handled as a subclass of `Expression`, just like `ListExpression`, and this method (`get_sequence`) should belong just to this subclass.

If I understand correctly, there would be a new `Sequence` class.  What are the important methods of this class? 

As I look at Expression right now, it has operations that are not needed in ListExpression, like `has_form` and probably much of that caching stuff (and where it is needed, it probably could be done simpler). 

I imagine this is true even more so for such as Sequence class. 

What I suspect is needed is to split off of Expression a CompoundExpression base base class with the minimual set of things that are needed for compound expressions. Then Expression and ListExpression (and possibly Sequence) would be based off of this. 

Your thoughts?

rocky:
Given the complexities of Sequence, I suggest merging this (with possibly a small set of additional changes), and then deal with the `Sequence` class issue later.

mmatera:
> Given the complexities of Sequence, I suggest merging this (with possibly a small set of additional changes), and then deal with the `Sequence` class issue later.

Yes, my comment here was about what to do with the comment in the `get_sequence` code. I merge this because it does not essentially change anything, except the place in which the import for formatters is placed.  Let's work out the other part in a subsequent PR.

PR #549, opened 2022-09-12
Symbol arguments

mmatera:
This PR implements the idea that I had yesterday about how to handle the parameter ``value`` for ``Symbol`` and ``PredefinedSymbol``.

### DIFF
--- a/mathics/core/symbols.py
+++ b/mathics/core/symbols.py
@@ -369,7 +369,7 @@ class Symbol(Atom, NumericOperators, EvalMixin):
 
     # __new__ instead of __init__ is used here because we want
     # to return the same object for a given "name" value.
-    def __new__(cls, name, sympy_dummy=None, value=None):
+    def __new__(cls, name, **opts):
         """
         Allocate an object ensuring that for a given `name` we get back the same object.
         """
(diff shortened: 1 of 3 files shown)


mathics/core/symbols.py
```diff
@@ -369,7 +369,7 @@ class Symbol(Atom, NumericOperators, EvalMixin):
 
     # __new__ instead of __init__ is used here because we want
     # to return the same object for a given "name" value.
-    def __new__(cls, name, sympy_dummy=None, value=None):
+    def __new__(cls, name, **opts):
```
rocky, line 372: Using a grab bag `**opts` parameter instead of named parameters is frowned upon because you no longer get any sort of checking when the wrong parameter is used. 

In fact this occurred a little while ago when I passed `use_string_quotes` in `to_python()` when the parameter was `string_quotes`.

mmatera: The point is that `Symbol.__new__` shouldn't use another parameter than `name:str`. By now I left the use of `sympy_dummy`, but I think that it is wrong.

The need for having the `opts` bag is that subclasses could have other optional parameters (like `value`) for the initialization, and then, if we do not let this slot for optional parameters, a call to `__new__` would produce an exception. 
Another approach would be to add a `__new__` method in the subclasses, but I found it uglier...

rocky: I don't doubt that there is a cleaner implementation out there. However this isn't it. And this like so many other times when you wander into areas that are not you field of expertise - Python programming idioms, parsing, interpreters, to name a few, they just bog things down .From my standpoint this leads me to spend far more time on distractions that I would like to or have time for. 

Mark your objection as a comment and let's just leave this for now until we find someone who has a better sense of  or more expert in programming style.

rocky:
LGTM

PR #551, opened 2022-09-13
Revision of Symbol.to_python()

mmatera:
For some reason, in the past, it was defined that the Python equivalent of a `mathics.core.Symbol` object was a `str` object. Following the discussion I started in #550, I propose here a change to this behavior, which looks to be more concise and consistent with the current implementation (i.e. instead of compare names, compare directly the objects). 

CHANGES:

* Symbol.to_python() now returns a `Symbol`  instead an string
* the parameter n_evaluation shows a `DeprecationWarning` in favor of `eval_N(expr, evaluation).to_python()`
* some adjustments on other modules to make this work.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/551"></a>

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,6 +35,7 @@ Internals
 #. `from_mpmath` conversion now supports a new parameter ``acc`` to set the accuracy of the number.
 
 
+
 Bugs
 ++++
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,7 @@ Internals
 
 #. ``boxes_to_`` methods are now optional for ``BoxElement`` subclasses. Most of the code is now moved to the ``mathics.format`` submodule, and implemented in a more scalable way.
 #. ``mathics.builtin.inout`` was splitted in several modules (``inout``, ``messages``, ``layout``, ``makeboxes``) in order to improve the documentation.
+#. ``Symbol.to_python`` now returns `self` except for `True`, `False` and `None`. The use of the parameter  `n_evaluation` was marked as deprecated. 
 
 Bugs
 ++++
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,7 @@ CHANGES
 New Builtins
 +++++++++++
 
+#. ``Accuracy``
 #. ``$BoxForms``
 #. ``ClebschGordan``
 #. ``Curl`` (2-D and 3-D vector forms only)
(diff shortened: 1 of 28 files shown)


rocky:
I think this is the wrong direction. `to_python()` should be returning a Python object. Period. 

The callers want a symbol back should be changed so they don't call to_python at all but instead work with symbols and not call to_python at all.

rocky:
I suggest breaking this up into changing those smaller things that can just be converted to using Symbol without changing the behavior of to_python() or in other words, without having to  call to_python. 

This is another way of saying what I said before, but applied to the particular changes here.

rocky:
It may be that in the future we want to come up with a more compact M-expression representation, sort of analogous to what was done in JSON for threejs. But this is design work and please, it is not something that you have shown strength at and don't do it by futzing around here and there.  There is a ton of clean-up work that could go on. And were that to happen, it would make the case clearer and cleaner to do.

rocky:
> [3., "System`Automatic", "System`Automatic"] (with the version in Master) or to
> [3., SymbolAutomatic, SymbolAutomatic] with the version in this PR.

Is `3.` a Python float? If so then that is generally not how Python reports it. It would be 3.0. If more likely `3.` is some sort of Mathics representation, then it sounds like whatever is creating that should have created this as a ListExpression and then everything would go through properly. 

Not that there is never a `to_python()` used in creating a ListExpression because we aren't going to Python.

mmatera:
> > [3., "System`Automatic", "System`Automatic"] (with the version in Master) or to
> > [3., SymbolAutomatic, SymbolAutomatic] with the version in this PR.
> 
> Is `3.` a Python float? 

Yes, thanks. I fixed it in the comment. What I meant was that whatever is the internal representation of `expr`, if it represents 
the WL expression ```{3.`, Automatic, Automatic}```,  then `expr.to_python()` could have one of these two outputs. 

> >  ``` [3.0, "System`Automatic", "System`Automatic"]```  (Python code), with the version in Master or to
> >  ``` [3.0, SymbolAutomatic, SymbolAutomatic]``` (Python code)  with the version in this PR.

I am going to use it to add a comment in the code.

> Note that there is never a `to_python()` used in creating a ListExpression because we aren't going to Python.

Yes, but why it would be used to build a `ListExpression`? In `ListExpression` I would do something like adding an specialized `to_python` definition like


```
    def to_python(self, *args,**kwargs):
         if self.value is None:
             self.value = [el.value if el.is_literal() else el.to_python(*args, **kwargs) for el in self.elements]
         return self.value
```
or something like that...

mmatera:
@rocky, BTW, another question. What do you think about the change `expr.to_python(n_evaluation=evaluation)` by `eval_N(expr, evaluation).to_python()`. If this makes sense, I can move this part to another PR.

rocky:
> @rocky, BTW, another question. What do you think about the change `expr.to_python(n_evaluation=evaluation)` by `eval_N(expr, evaluation).to_python()`. If this makes sense, I can move this part to another PR.

This is good. And small PR's are good.

TiagoCavalcante:

mmatera:
Again, just another proposal that is waiting for a more clear design.

PR #552, opened 2022-09-13
Removing the n_evaluation parameter from the `to_python` method

mmatera:
Following the line of attempts to clarify, document, and remove flaky practices from the code, I collected in this PR the part of #551 in which @rocky seems to agree.

* adding the prototype of `to_python` in `BaseElement`.
* adding type annotations to its parameters.
* removing the use of the parameter n_evaluation in ``expr.to_python(n_evaluation=evaluation)`` in favor of ``eval_N(expr, evaluation).to_python()``.
* in the modified functions, make changes to avoid the same variable to represent objects of different classes from one line to the next one.
* adding several comments.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/552"></a>

### DIFF
--- a/mathics/builtin/numbers/randomnumbers.py
+++ b/mathics/builtin/numbers/randomnumbers.py
@@ -17,9 +17,11 @@ from functools import reduce
 from mathics.builtin.base import Builtin
 from mathics.builtin.numpy_utils import instantiate_elements, stack
 from mathics.core.atoms import Integer, String, Real, Complex
+from mathics.core.evaluators import eval_N
 from mathics.core.expression import Expression
 from mathics.core.list import ListExpression
 from mathics.core.symbols import Symbol, SymbolDivide, SymbolNull
+
 
 SymbolRandomComplex = Symbol("RandomComplex")
 SymbolRandomReal = Symbol("RandomReal")
@@ -497,11 +499,21 @@ class RandomComplex(Builtin):
 
     @staticmethod
     def to_complex(value, evaluation):
-        result = value.to_python(n_evaluation=evaluation)
-        if isinstance(result, (float, int)):
-            result = complex(result)
-        if isinstance(result, complex):
-            return result
+        result = eval_N(value, evaluation)
+
+        if hasattr(result, "value"):
+            result_value = result.value
+        else:
+            # TODO: result.value does not work, because
+            # Complex does not have a ``value`` attribute.
+            # Otherwise, we could return here ``None``.
+            result_value = result.to_python()
+
+        if isinstance(result_value, (float, int)):
+            return complex(result_value)
+        if isinstance(result_value, complex):
+            return result_value
+
         return None
 
     def apply(self, zmin, zmax, evaluation):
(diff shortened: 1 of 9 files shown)


mathics/builtin/numbers/randomnumbers.py
```diff
@@ -497,11 +499,21 @@ class RandomComplex(Builtin):
...
-            result = complex(result)
-        if isinstance(result, complex):
-            return result
+        result = eval_N(value, evaluation)
+
+        if hasattr(result, "value"):
+            result_value = result.value
+        else:
+            # TODO: result.value does not work, because
+            # Complex does not have a ``value`` attribute.
```
rocky, line 508: I have been thinking about this. When the complex value is known to be exactly the same, for example we have an integer imaginary number that is not too large, or the "float" representations of the real and imaginary parts fit exactly in the float number, then Complex could have a Python value which could be Python datatype complex or integer (and possibly float if the float representation is exact). 

However this kind of thing, especially at this stage of development is to involved seems like a rather custom kind of things and probably not the main source of slowdowns, that I don't we should undertake this activity just yet.

mathics/core/element.py
```diff
@@ -422,12 +422,28 @@ def user_hash(self, update) -> None:
...
+        # something in (int, float, complex, str, tuple, list or dict.).
+        # (See discussions in
+        #  https://github.com/Mathics3/mathics-core/discussions/550
+        # and
+        # https://github.com/Mathics3/mathics-core/pull/551
+        #
+        #
+        # if n_evaluation is an Evaluation object, then the expression
+        # is passed by an eval_N().
+        # If python_form is True, the standard behaviour is changed,
```
rocky, line 436: Please let's find and remove any use of to_python() with `python_form=True` and remove this parameter.

mmatera: Sure!

rocky:
LGTM - rather than remove `python_form=False` in this PR, let us do this in a following one.

mmatera:
> LGTM - rather than remove `python_form=False` in this PR, let us do this in a following one.

Just in time! I finished removing it, and I was at 30s to push it :)

mathics/builtin/numbers/randomnumbers.py
mmatera: Indeed, this is why I just left a comment...

rocky: It is not just Complex though. And it is not clear that adding value to complex would handle everything either. Python3 has Factions as well, and fixed-point numbers. And even with this there may be other Atoms that don't have values. 

So from my point of view this is a little vague. It is not that Complex _alone_ doesn't have a value so much as right now value is optional. And that is what the test says.

mmatera: The comment was about `Complex` in a very particular context. If this class had a `value` property (that could even be calculated on the fly) then the `if` and the call to `to_python` would not be necessary.

PR #554, opened 2022-09-13
Bye-bye python_form parameter.

mmatera:
This continues #552. Now I removed the parameter `python_form` from `to_python` and replaced it with something local into the unique place where it is needed.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/554"></a>

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -1382,7 +1382,9 @@ class Expression(BaseElement, NumericOperators, EvalMixin):
         """
         from mathics.builtin.base import mathics_to_python
 
-        n_evaluation = kwargs.get("n_evaluation")
+        n_evaluation = kwargs.get("n_evaluation", None)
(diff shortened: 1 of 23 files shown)


TiagoCavalcante (approved):
LGTM

mathics/core/expression.py
```diff
@@ -1416,6 +1418,9 @@ def to_python(self, *args, **kwargs):
             #         keywords=[],
             #         )
             return py_obj
+
+        # Notice that in this case, `to_python` returns a Mathics Expression object,
+        # instead of a builtin native object.
```
rocky, line 1423: I expect over time that more iterations over the code will be done and the thing that seem unusual will be addressed or there will be explanations of _why_ they are the way they are.

rocky:
LGTM

PR #558, opened 2022-09-17
moving pending doctests in MakeBoxes to pytests

mmatera:
This PR moves several doctests (mostly about broken behavior) associated to MakeBoxes to (xfailed) pytest tests.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/558"></a>

### DIFF
--- a/test/builtin/test_makeboxes.py
+++ b/test/builtin/test_makeboxes.py
@@ -3,6 +3,29 @@ import pytest
 from test.helper import check_evaluation, session
 from mathics_scanner.errors import IncompleteSyntaxError
 
+
+# 15 tests
+@pytest.mark.parametrize(
+    ("str_expr", "str_expected", "msg"),
+    [
+        (r"MakeBoxes[0`3]", r"0", None),
+        (r"MakeBoxes[14]", r"14", None),
+    ],
+)
+def test_makeboxes_real(str_expr, str_expected, msg):
+    """
+    # TODO: Constructing boxes from Real
+    """
+    check_evaluation(
+        str_expr,
+        str_expected,
+        to_string_expr=True,
+        to_string_expected=True,
+        hold_expected=True,
+        failure_message=msg,
+    )
+
+
 # 15 tests
 @pytest.mark.parametrize(
     ("str_expr", "str_expected", "msg"),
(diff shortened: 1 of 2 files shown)


test/builtin/test_makeboxes.py
```diff
@@ -0,0 +1,127 @@
...
+        (r"MakeBoxes[0`3]", r"0", None),
+        (r"MakeBoxes[0``30]", r"0.``30.", None),
+        (r"MakeBoxes[0.`]", r"0.`", None),
+        (r"MakeBoxes[0.`3]", r"0.`", None),
+        (r"MakeBoxes[0.``30]", r"0.``30.", None),
+        (r"MakeBoxes[14]", r"14", None),
+        (r"MakeBoxes[-14]", r'RowBox[{"-", 14}]', None),
+    ],
+)
+@pytest.mark.xfail
```
rocky, line 27: This is a skipped test. xfail is for testing properties that  _properly_ fail. Like divide by zero. Not things that currently happen to be bugs.k Thanks.

test/builtin/test_makeboxes.py
```diff
@@ -0,0 +1,127 @@
...
+            r"\(x \/ (y + z)\)",
+            r'FractionBox["x", RowBox[{"(", RowBox[{"y", "+", "z"}], ")"}]]',
+            None,
+        ),
+        (r"\( \@ a + b \)", r'RowBox[{SqrtBox["a"], "+", "b"}]', None),
+    ],
+)
+@pytest.mark.xfail
+def test_makeboxes_precedence(str_expr, str_expected, msg):
+    """ """
```
TiagoCavalcante, line 57: Blank comment here and bellow

rocky:
Thanks for moving this out of doctest.

TiagoCavalcante (approved):
LGTM - I looked at time some time ago and thought it was confusing, now it makes sense

test/builtin/test_makeboxes.py
rocky: @mmatera I am wrong. I checked https://docs.pytest.org/en/7.1.x/how-to/skipping.html and it does mention bugs. 

(That said, my own use of of skip is for bugs not expected failures.

rocky:
LGTM

test/builtin/test_makeboxes.py
mmatera: @rocky, I used `xfail` just to have a notion about the progress. However, it is true that it takes time to check that something still does not work. So, now I make it optional and controlled by the "DEBUG" environment variable.

mmatera: BTW, thanks for the reference. I got some ideas to improve the implementation  of some of the tests.

mmatera:
Let's merge this and iterate. The plan for the following week is to make progress fixing the bugs one by one.

rocky:
>  The plan for the following week is to make progress fixing the bugs one by one.

I very much like coming up with the Boxing behavior _before_ starting coding. And that these are segregated. I'd ask though that you tread very lightly here though - start _very small_ and  easy to understand; pick something where we can somehow show that the kind of thing that we are doing follows _exactly_ how WMA thinks of this.  

Perhaps there is a way to trace WMA's boxing routines by augmenting them to show trace information, or maybe by adding `Print`'s various places? 

One service that Mathics can provide to the community is to explain in a clear and coherent fashion the behavior of WMA that is vague (possibly sometimes deliberately). 

Our past history though has been make these myopic low-level changes without fully understanding the consequences or the bigger picture. By not doing so, we make our code base more fragile and complicated when the bigger picture is understood. 

Thanks.

PR #559, opened 2022-09-18
removing n_evaluation

mmatera:
This finished the changes from `expr.to_python(n_evaluation=evaluation)` to  `eval_N(expr, evaluation).to_python()`

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/559"></a>


TiagoCavalcante (approved):
LGTM

mathics/core/expression.py
```diff
@@ -1394,7 +1394,17 @@ def to_python(self, *args, **kwargs):
             return expression_to_callable_and_args(expr_fn, vars, n_evaluation)
 
         if n_evaluation is not None:
```
rocky, line 1396: Line 1386 or thereabouts has an assert that `n_evaluation is None`. So this isn't needed. `n_evaluation`, a relatively new feature, is a hack that should not have been added in the first place.

mmatera: Indeed, this line should be removed. Regarding `n_evaluation`, it is a feature that I proposed to add before we have the `eval_N` function, and I think that was a useful intermediate step. In any case, if we still want that `Expression(SymbolFunction,...).to_python(n_evaluation=evaluation)` still produce a callable Python object, we can not remove the parameter completely.

mmatera: BTW, if we want that `to_python` always returns a "native" Python object, one possibility for Expressions that do not have a clear Python equivalent would be to return a callable object, like

```
    def result_callable():
        res = expr.evaluate(evaluation)
        if res.sameQ(expr):
             return expr
        else:
             expr.to_python()
    return result_callable
```

Then, if at some point the expression becomes evaluable and produces a Python native object, then the result would be a Python object...

rocky: > BTW, if we want that `to_python` always returns a "native" Python object, one possibility for Expressions that do not have a clear Python equivalent would be to return a callable object, like
> 
> ```
>     def result_callable():
>         res = expr.evaluate(evaluation)
>         if res.sameQ(expr):
>              return expr
>         else:
>              expr.to_python()
>     return result_callable
> ```
> 
> Then, if at some point the expression becomes evaluable and produces a Python native object, then the result would be a Python object...

This is a possibility but there would be a different method called. As I have said in the past there was this pervasive vaguess which assists a muddling of concepts. `to_python()` is used for different kinds of situations which haven't been clearly specified. And that is why we have weird or hybrid behavior such as returning a string when conversion fails instead of raising an exception. (And letting some other routine convert this to a string then if that is what it needs to do). 

As I have alluded to before `to_python_literal()` would be a better name for at least one of the disentangled functions.

mmatera: This 
```
Expression(SymbolFunction,...).to_python(n_evaluation=evaluation)
```
was used for a while to convert WL functions into Python code, to be passed to numerical routines like integrators or optimizers in `scipy`. 

The history was similar than the one that leads us to remove the idiom 
```Expression(SymbolN, expr).evaluate(evaluation).to_python()```
in favor of `eval_N(expr, evaluation)`.

Before of this, we have to do something like `Expression(SymbolCompile, vars, expr).evaluate(evaluation).elements[1].cfunc` (among other idioms) in order to get a Python callable. At some point, we agree that having this kind of idiom spread on a so large code was not a good idea. Also, as in many other cases, we didn't have a clear big picture about how this should be implemented. 
Then, it comes up with the idea of using `to_python` to produce that callable. Eventually, when we have a clearer picture, finding and replacing a small chunk of code in `to_python`, or by localizing and replacing the calls.

I think that workflow is not so bad:  it allows us to disentangle the code, understand better the different pieces, and eventually clean up the parts that are not necessary anymore, without stopping all the development.

Comment (mmatera) [reply mathics/core/expression.py:1395]: > 
> As I have alluded to before, `to_python_literal()` would be a better name for at least one of the disentangled functions.

Regarding this, I think that it would be a great idea.

Comment (rocky) [reply mathics/core/expression.py:1395]: > This
> 
> ```
> Expression(SymbolFunction,...).to_python(n_evaluation=evaluation)
> ```
> 
> was used for a while to convert WL functions into Python code, to be passed to numerical routines like integrators or optimizers in `scipy`.
> 
> The history was similar than the one that leads us to remove the idiom `Expression(SymbolN, expr).evaluate(evaluation).to_python()` in favor of `eval_N(expr, evaluation)`.
> 
> Before of this, we have to do something like `Expression(SymbolCompile, vars, expr).evaluate(evaluation).elements[1].cfunc` (among other idioms) in order to get a Python callable. 
> ...

I don't remember any of this and wasn't involved in the discussion of any of this. Basically you are saying that the way you had to learn this isn't so bad from your standpoint. From my standpoint, I'd wish recognize when you are going off into an area that is not your core competence and maybe even not go there. It puts a lot more work on me, to try to figure out what you understand (previously based on PR code) and what you don't.

Comment (mmatera) [reply mathics/core/expression.py:1395]: @rocky, the change was introduced in March 14th 2021,  in this PR: https://github.com/mathics/Mathics/pull/1161
Please, have a look over the discussion.

What I am saying is that even having 15 years of experience using WL, and having an expert in computer science in the team, (and even some developers that are actually developing WMA), in seven years we were not able to have a modular, efficient and clear framework, or even a clear big picture of how all these pieces should look and work together. 

So I think this is not just about me trying to learn something, but as a sinuous way in which we are trying to reformulate a long, complicated and monolithic code into something that we can analyze and gradually improve and contribute, with the knowledge and tools that we have.

Comment (rocky) [reply mathics/core/expression.py:1395]: Ok - point taken. I think you are correct. That PR was way too big to easily grasp all that is going on. Especially the relevant part here where this interface was changed. (Github says the diff is too large to display)

I should say that I now understand a lot about how this code needs to change to get it to the point where it seems like a normal non-toy interpreter. And I think to some degree we all understand some of this. 

It is going to be a long haul. I'd ask then to please stay away from the parser and scanner and compile methods for now. Pattern matching needs also going over and that is on the list.

Comment (mmatera) [reply mathics/core/expression.py:1395]: > Ok - point taken. I think you are correct. That PR was way too big to easily grasp all that is going on. Especially the relevant part here where this interface was changed. (Github says the diff is too large to display)
> 

OK. That was because I was much less aware than now about how to make a readable PR. This is something that I learn in the hard way during this time...

> I should say that I now understand a lot about how this code needs to change to get it to the point where it seems like a normal non-toy interpreter. And I think to some degree we all understand some of this.
> 
> It is going to be a long haul. I'd ask then to please stay away from the parser and scanner and compile methods for now. Pattern matching needs also going over and that is on the list.
> 
> The way symbols are bound to variables is another thing that is going to be a mess.
> 
> Evaluation we've really only started to go over. It is nowhere near gone over though.

Yes, I agree with all that. But then, we can take two positions: leave the code as it is, diagram a master plan, then rework all the existing pieces, and then (let's say, in a couple of years) see if we can start improving the functionality. 

The other way (the one in which I think I could be useful now) is to fix peripherical parts with (maybe suboptimal) code, and to add tests to reinforce the compatibility with WMA. This is what I was trying to do during the last months.

Comment (rocky) [reply mathics/core/expression.py:1395]: > Yes, I agree with all that. But then, we can take two positions: leave the code as it is, diagram a master plan, 

There has been a rough plan. I was making my way over the rewrite-apply-eval process. I had gotten to the top-level evaluation.  That was blocking Formatting and Boxing work you wanted to do. So I stopped that process so that you would have free reign over Form, Boxing, and Format. While you are doing that I am looking to expand the use of ListExpression and customized kinds of expressions. Association is a natural for this as well. 

The Boxing and Formatting work quickly spilled into scanning and parsing - rightly or wrongly. I'll handle issues that spill over there. 

> The other way (the one in which I think I could be useful now) is to fix peripherical parts with (maybe suboptimal) code, and to add tests to reinforce the compatibility with WMA. This is what I was trying to do during the last months.

In the places where we have no choice, okay. But in the areas where I have a clearer idea of what needs to be done, let's wait on that. With respect specifically to Forms, Boxing, and Format, this is a clear area where you know more about this than I can even hope to understand, so focus on that. The plan that you have put together seems good so far. So just continue with that. 

As for _this_ PR, let's put back that `assert` and remove the more complex code that deprecating something that we now know to be ill-defined.

Comment (mmatera) [reply mathics/core/expression.py:1395]: OK. The assert was removed, and I left the code for converting  `Expression(SymbolFunction, ...)` as it was before, except by the name of the parameter (I have removed the "n_"). If a call is done with the "n_evaluation" parameter, a deprecation warning is shown, and then continue almost as it was before, except that instead of using the `Expression(SymbolN,...).evaluate(n_evaluation)` idiom, a `eval_N()` call is used.

File: mathics/builtin/scipy_utils/integrators.py [hunk 2]
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 import sys
-from mathics.builtin import check_requires_list
-from mathics.core.utils import IS_PYPY
+from mathics.builtin.base import check_requires_list
+from mathics.core.util import IS_PYPY
```
mmatera, line 4: Also, I fixed a typo here, which was avoiding that this module gets loaded.

mathics/core/expression.py
rocky: Please no deprecation warning for something that should not have been added. 

This PR I thought (and actually the previous one) were  supposed to _remove_  uses, not deprecate them.  The "assert" checks that.  In a little while the assert can be removed as well. Thanks.

In a little while

mathics/builtin/scipy_utils/integrators.py
mmatera: Fixed also in #560.

mathics/core/expression.py
mmatera: I commented out both the assert and the code that shows the deprecation. I am going to leave this as a draft, and eventually use this discussion (and the changes) when we want to go over this method.

TiagoCavalcante:

PR #561, opened 2022-09-20
fix and test expressions of the form BoxExpression[[...]]

mmatera:
Another small fix to something that I found working on MakeBoxes. When a ```BoxExpression``` is passed as the first element to ```Part```, since it does not have in general an attribute `_elements`, the evaluation ends with an unhandled exception. In particular, this happens to `RowBox`. This PR fixes it by providing an implementation for that method.

### DIFF
--- a/mathics/builtin/box/layout.py
+++ b/mathics/builtin/box/layout.py
@@ -305,6 +305,13 @@ class RowBox(BoxExpression):
         result = RowBox(*items)
         return result
 
+    @property
+    def elements(self):
+        return self.to_expression().elements
+
+    def get_head(self):
+        return SymbolRowBox
+
     def init(self, *items, **kwargs):
         # TODO: check that each element is an string or a BoxElementMixin
         self.box_options = {}
@@ -352,8 +359,8 @@ class RowBox(BoxExpression):
                 for item in self.items
             )
 
-            self._elements = Expression(SymbolRowBox, ListExpression(*items))
-        return self._elements
+            self._elements = ListExpression(*items)
+        return Expression(SymbolRowBox, self._elements)
 
 
 class ShowStringCharacters(Builtin):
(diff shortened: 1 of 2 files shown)


TiagoCavalcante (approved):
LGTM

mathics/builtin/box/layout.py
```diff
@@ -305,6 +305,13 @@ def apply_list(self, boxes, evaluation):
         result = RowBox(*items)
         return result
 
+    @property
+    def elements(self):
```
rocky, line 309: It feels like something in the code or the API is wrong here. 

I get nervous about "properties" doing possibly unbounded calculations such as `self.to_expressions()`.  

If I understand correctly, the caller somehow knows this is a compound Expression/Element and assumes that therefore it has an `.element` attribute which is of type `Expression`? 

It feels like what should be happening that the _caller_ should be asking instead for  a _conversion_, that is, should be calling `to_expression().elements` 

I need to think about this more.

rocky:
The _what_ is most appreciated. However the _how_ may need some more thought or discussion.

mathics/builtin/box/layout.py
mmatera: > It feels like something in the code or the API is wrong here.
> 

Probably it is, but I tried to build over the existing API. At least, this is "localized" in this class.

> I get nervous about "properties" doing possibly unbounded calculations such as `self.to_expressions()`.
> 
> If I understand correctly, the caller somehow knows this is a compound Expression/Element and assumes that therefore it has an `.element` attribute which is of type `Expression`?
> 

The problem is that  `Part` assumes that an Expression has a property `elements`. This just provides the property by creating an Expression. A different possibility would be to implement a different rule for working with RowBoxes. I thought this change was simpler, but maybe there is a better approach.

> It feels like what should be happening that the _caller_ should be asking instead for a _conversion_, that is, should be calling `to_expression().elements`
> 
> I need to think about this more.

Sure.

TiagoCavalcante:

mmatera:
> @mmatera what is the status of this PR?

@TiagoCavalcante, by now, it is a reminder to something we need to fix.

mmatera:
Deprecated

PR #562, opened 2022-09-21
Fix RandomSample and add test cases

### LINKED ISSUE #468: RandomSample on a list returns one item.
Again an exmple from https://www.wolfram.com/language/elementary-introduction/2nd-ed/30-rearranging-lists.html

```
RandomSample[Range[10]]
```

always produces one item when it should produce more.  A revised version of chapter 30 will go out with the next docker update (which will not be done with the 5.0.0 release but probably a week later.)

Kalsofar:
Fixes #468
 
RandomSample with one list argument should return a random ordering of the items in the given list, not just one item from the list.

Before:
```Mathematica
In[1]:= SeedRandom[42]; RandomSample[Range[10]]
Out[1]= {9}
```

After:
```Mathematica
In[1]:= SeedRandom[42]; RandomSample[Range[10]]
Out[1]= {9, 2, 6, 1, 8, 3, 10, 5, 4, 7}
```

### DIFF
--- a/mathics/builtin/numbers/randomnumbers.py
+++ b/mathics/builtin/numbers/randomnumbers.py
@@ -745,7 +745,7 @@ class RandomSample(_RandomSelection):
      = {{1, 2}, {}, a}
     >> SeedRandom[42]
     >> RandomSample[Range[10]]
-     = {9, 2, 6, 1, 8, 3, 10, 5, 4, 7
+     = {9, 2, 6, 1, 8, 3, 10, 5, 4, 7}
     >> SeedRandom[42]
     >> RandomSample[Range[100], {2, 3}]
      = {{84, 54, 71}, {46, 45, 40}}
--- a/mathics/builtin/numbers/randomnumbers.py
+++ b/mathics/builtin/numbers/randomnumbers.py
@@ -253,7 +253,7 @@ class _RandomBase(Builtin):
         ),
     }
     rules = {
-        "%(name)s[spec_]": "%(name)s[spec, {Length[spec]}]",
+        "%(name)s[spec_]": "%(name)s[spec, {1}]",
         "%(name)s[spec_, n_Integer]": "%(name)s[spec, {n}]",
     }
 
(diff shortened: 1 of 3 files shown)


mathics/builtin/numbers/randomnumbers.py
```diff
@@ -735,22 +735,30 @@ class RandomSample(_RandomSelection):
     </dl>
 
     >> SeedRandom[42]
-    >> RandomSample[{a, b, c}]
-     = {a}
+    >> RandomSample[{a, b, c, d}]
+     = {b, d, a, c}
```
rocky, line 739: We probably shouldn't hard-code an order here. The way to give an example but allow variation is using `...`: 

```
>> RandomSample[{a, b, c, d}]
     = {...}
```

test/builtin/numbers/test_randomnumbers.py
```diff
@@ -0,0 +1,40 @@
...
+from test.helper import check_evaluation
+
+
+@pytest.mark.parametrize(
+    ("str_expr", "str_expected"),
+    [
+        ("SeedRandom[42]; RandomSample[{a, b, c, d}]", "{b, d, a, c}"),
+        (
+            "SeedRandom[42]; RandomSample[{a, b, c, d, e, f, g, h}, 7]",
+            "{b, f, a, h, c, e, d}",
```
rocky, line 16: I am kind of curious. If SeedRandom is a different number, do we get a different order?

Kalsofar: In my tests, yes, changing the SeedRandom value results in a different order:
```Mathematica
In[1]:= SeedRandom[42]; RandomSample[{a, b, c, d}]
Out[1]= {b, d, a, c}

In[2]:= SeedRandom[137]; RandomSample[{a, b, c, d}]
Out[2]= {a, b, d, c}

In[3]:= SeedRandom[137]; RandomSample[{a, b, c, d}]
Out[3]= {a, b, d, c}

In[4]:= SeedRandom[962]; RandomSample[{a, b, c, d}]
Out[4]= {d, a, c, b}
```

And I also assume that changing the random number generator will also change the order.

rocky:
Thanks for doing this. Sorry for the delay.

PR #566, opened 2022-09-29
Fixing and improving  tests on docstring format

mmatera:
This is a long PR, but most of it is formatting changes in the docstrings.

* Improving consistency tests for docstrings
* Adding a tool to fix indentation and like breaks in docstrings automatically
* Fix all the cases.


mathics/builtin/arithmetic.py
```diff
@@ -418,7 +418,6 @@ class Complex_(Builtin):
     <dl>
       <dt>'Complex'
       <dd>is the head of complex numbers.
-
```
rocky, line 421: Space between groups of <dt> was deliberate.  Is it too hard to preserve this?

mathics/builtin/comparison.py
```diff
@@ -599,15 +594,7 @@ class Equal(_EqualityOperator, _SympyComparison):
     <dl>
       <dt>'Equal[$x$, $y$]'
       <dt>'$x$ == $y$'
-      <dd>is 'True' if $x$ and $y$ are known to be equal, or
-        'False' if $x$ and $y$ are known to be unequal, in which case
-        case, 'Not[$x$ == $y$]' will be 'True'.
-
```
rocky, line 605: These line breaks are deliberate.

mathics/builtin/arithmetic.py
mmatera: OK. It is almost trivial to restore them. Please, when you have time, check the rules I have used, to see if they need to be adjusted.

mathics/builtin/comparison.py
mmatera: Why here we can have here the line breaks?

rocky: I may have mistagged  lines. I am seeing  _paragraph_  spacing around

"For any expression $x$ .." 

And "Commutative properties apply so"   (there should be a "," after apply). 

And in the replaced text I am not seeing that paragraph breaks are preserved.

rocky:
When doing something that makes massive changes like this, here is a technique I use that helps. 

First try this on a small example, and check it very carefully. Here for example, go into Django and look, and produced a PDF and look. If things look fine, do that in a PR. Then try something more ambitious.

mathics/builtin/arithmetic.py
rocky: Sure - will do! Probably this weekend after some of the other backlogged items.

mathics/builtin/comparison.py
mmatera: I was looking at this case. The problem here is that I expected that the `<dd>` tag contains a single paragraph, which follows the way in which (I think) WR is written. I proposed a change here to move the extra paragraphs after the `<dl>...</dl>` block.

mmatera:
> When doing something that makes massive changes like this, here is a technique I use that helps.
> 
> First try this on a small example, and check it very carefully. Here for example, go into Django and look, and produced a PDF and look. If things look fine, do that in a PR. Then try something more ambitious.

Yes, I did it locally, and check that basic stuff. However, what I am looking for here is to determine a global pattern to normalize the documentation. So, I need to agree with you both on what is that normalized pattern, over all the cases.

TiagoCavalcante:

rocky:
> @mmatera what is the status of this PR?

Part of PR introduces code to format everything. It doesn't completely work. The rest of the PR is used over the entire code.

mmatera:
Deprecated

PR #568, opened 2022-10-04
Change where do we build documentation...

rocky:
settings.py:
  documentation.tex needs to be in the code source directory, not installed Mathics

sed-hack.sh:
  If documentation.tex does not exist, we don't need to make a backup
  copy of it.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/568"></a>

### DIFF
--- a/mathics/doc/latex/sed-hack.sh
+++ b/mathics/doc/latex/sed-hack.sh
@@ -1,6 +1,9 @@
 #!/bin/bash
+set -x
 # Brute force convert Unicode characters in LaTeX that it can't handle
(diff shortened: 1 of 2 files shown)


rocky:
Note that when things are fully sorted out, we should no longer need "sed-hack.sh" since translations to LaTeX symbols should work out properly.  However getting this probably going to take several iterations and PR's to get right.

mmatera (approved):
LGTM

TiagoCavalcante (approved):
LGTM

PR #569, opened 2022-10-10
Infix formatting

rocky:
This revises the work done in the `ascii-op-to-Unicode` branch. 

I had a hard time understanding the flow of what's going on, and this process was both painful and made me a bit annoyed. 

I was pleased though at @mmatera start to move the makeboxes into its own module, and we have a _little_ bit of segregating boxing routines, so that I suppose helps a little.

It is not that the code isn't sophisticated or complicated. It is just that it is not organized and changes haven't been coordinated. 

Let me describe a little of the history around this PR and the code I see, and reflect. The aim show of this reflection shows how the code has become of poorer quality, is in constant need of serious refactoring, and can get worse with uncoordinated bug fixes or feature improvements. Functionally the code does a quite a bit; but it is at the cost of lots of code with undue complexity. I assume that much of the complicated code is to ensure correct behavior.

However, where we find that the code is not correct or lacking, we have a hard time figuring out where to go and what to change that isn't going to impact a lot of other things.

mmatera notes a problem with infix operator behavior in the `ascii-op-to-Unicode` branch. I had sort of seen this when I wrote that branch initially. One thing that I noticed that was was wrong in my PR is that the place where the change occurs felt wrong: it is code inside a "MakeBox" _rule_. But this rule is defined on class creation or registering each of the infix operators. These rules are is never changed. Something like `$CharacterEncoding` can change at will, so this location is too static.

Therefore I had to give up (temporarily) on making this work for `$CharacterEncoding`. Instead, I used the less dynamic `$SystemCharacterEncoding` instead.

Looking over this in a second pass, I see now that  a  rule like this is boneheaded:

```
formatted_output = 'MakeBoxes[Infix[{%s}," %s ",%d,%s], form]' % (
                replace_items,
                operator,
                self.precedence,
                self.grouping,
            )
default_rules = {
  # ...
  "MakeBoxes[{0}, form:InputForm|OutputForm]".format(op_pattern):
  formatted_output
}
```

Do you see why? ...

Hint: it is in the `" %s "` part.

This is supposed to be a rule that is called when boxing infix operators. And already it is deciding that operators should be surrounded by spaces for "InputForm" and "OutputForm".
In other words, right here in part which dictates rules to Box Infix expressions, we are already making decisions about the low-level formatting: that there should be spaces around the operator and what characters to use to represent the operator.

And when we actually get to the low-level format to final string, we have lost structure. Basically we have violated the principle that you evaluate to get an M-expression, then you Box the result, and then after Boxing then a low-level formatting is done.

If we want to write dozens more forms, the above isn't scalable. And we make high-quality formatting harder. 

At some level this was probably noticed, at least implicitly; when there are such few comments it is hard to know what was noticed and what was just random programming by reacting to problems that come up.
Since everything can't be done by MakeBox rules, we have `do_format_xxx` routines in `mathics-core/mathics/core/formatter.py` as well as special routines in `builtin/arithfns/basic.py`, `builtin/pympler/asizeof.py`, some that are done on inside `MakeBox()` internal functions.

In short, since principles if they were decided or defined, they were not communicated anywhere I can tell; so naturally formatting codes is scattering  at many of places in the code at several conceptual levels. Possibly some of the code is redundant or worse, works cross purposes.

If discussion of high-level principle were lacking, so are basic description code. Things like docstrings on functions. Or making an effort to _name_ what a function does. For example take the `get_op()` an interanal function inside _evaluation_ method `apply_infix()` leaving aside the missing `apply_` prefix that I have mentioned many times before.

What is `get_op()` "getting"? You already pass it some sort of operator. If you look at the function you realize it is not getting or accessing something, but rather converiting or _formatting_. 

And then once you realize that you are then in a position to ask _why_ is this routine formatting inside a Boxing routine? In the overall architecture isn't a separate step. Or should low-level formatting routines, be put  together?

Vagueness in code, lack of description and discussion around the code has led to very haphazard code that, in the end result, does not seem fully thought out.

With all the effort spent in just figuring out what he code _does_; by the time I understand that, I am exhausted and often not in a mood to think about how it should be designed or whether it follows a design or how to best write this.

Let me come back to adding spaces around the operator and losing the operator structure (the `" %s "` part) one more time. Because of this, the "remedy" was put in place that makes things worse. A routine was added that scans strings and unconditionally changes some strings (e.g. ASCII-formatted operators) into Unicode characters. I doesn't  matter if this was done by ignorance, willful frustration: in the end, it makes the code a bigger mess.

It is, as I say, like trying to solve a Rubik's cube by getting adjacent faces in line one at a time. In the beginning there is a certain satisfaction because you can "make progress". However getting to the end this way is much harder than understanding what is going on performing operations that work in conjunction with the principles and groups of the Rubik's cube.

Given all this, my inclination right now is to hold off on adding new Forms, or correcting the existing ones to be more correct. Instead if we just make things work they do but in a logical sensible and extensible way, I think this would buy us the most to then correct the existing behavior to match the specifications more closely and to add more Forms.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/569"></a>

### DIFF
--- a/mathics/builtin/makeboxes.py
+++ b/mathics/builtin/makeboxes.py
@@ -5,11 +5,13 @@
 Low level Format definitions
 """
 
+from typing import Union
 import mpmath
 
 
 from mathics.builtin.base import Builtin, Predefined
 from mathics.builtin.box.layout import _boxed_string, RowBox, to_boxes
+from mathics.core.convert.op import operator_to_unicode, operator_to_ascii
 from mathics.core.atoms import (
     Integer,
     Integer1,
@@ -44,6 +46,7 @@ from mathics.core.symbols import (
 from mathics.core.systemsymbols import (
     SymbolAutomatic,
     SymbolInfinity,
+    SymbolInputForm,
     SymbolMakeBoxes,
     SymbolOutputForm,
     SymbolRowBox,
@@ -80,11 +83,17 @@ def parenthesize(precedence, element, element_boxes, when_equal):
     return element_boxes
 
 
-def make_boxes_infix(elements, ops, precedence, grouping, form):
+# FIXME: op should be a string, so remove the Union.
+def make_boxes_infix(
+    elements, op: Union[str, list], precedence: int, grouping, form: Symbol
+):
     result = []
     for index, element in enumerate(elements):
         if index > 0:
-            result.append(ops[index - 1])
+            if isinstance(op, list):
+                result.append(op[index - 1])
+            else:
+                result.append(op)
         parenthesized = False
         if grouping == "System`NonAssociative":
             parenthesized = True
@@ -474,26 +483,26 @@ class MakeBoxes(Builtin):
         else:
             return MakeBoxes(expr, f).evaluate(evaluation)
 
-    def apply_infix(self, expr, h, prec, grouping, f, evaluation):
-        """MakeBoxes[Infix[expr_, h_, prec_:None, grouping_:None],
-        f:StandardForm|TraditionalForm|OutputForm|InputForm]"""
+    def apply_infix(self, expr, operator, prec, grouping, form: Symbol, evaluation):
+        """MakeBoxes[Infix[expr_, operator_, prec_:None, grouping_:None],
+        form:StandardForm|TraditionalForm|OutputForm|InputForm]"""
 
         def get_op(op):
             if not isinstance(op, String):
-                op = MakeBoxes(op, f)
+                op = MakeBoxes(op, form)
             else:
                 op_value = op.get_string_value()
-                if f.get_name() == "System`InputForm" and op_value in ["*", "^"]:
+                if form.get_name() == "System`InputForm" and op_value in ["*", "^"]:
                     pass
                 elif (
-                    f.get_name() in ("System`InputForm", "System`OutputForm")
+                    form in (SymbolInputForm, SymbolOutputForm)
                     and not op_value.startswith(" ")
                     and not op_value.endswith(" ")
                 ):
                     op = String(" " + op_value + " ")
             return op
 
-        precedence = prec.get_int_value()
+        precedence = prec.value
         grouping = grouping.get_name()
 
         if isinstance(expr, Atom):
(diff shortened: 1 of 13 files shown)


mathics/builtin/makeboxes.py
```diff
@@ -502,15 +526,36 @@ def get_op(op):
...
+            if operator.has_form("List", len(elements) - 1):
+                operator = [format_operator(op) for op in operator.elements]
+                return make_boxes_infix(elements, operator, precedence, grouping, form)
             else:
-                ops = [get_op(h)] * (len(elements) - 1)
-            return make_boxes_infix(elements, ops, precedence, grouping, f)
+                encoding_rule = evaluation.definitions.get_ownvalue(
+                    "$CharacterEncoding"
+                )
+                encoding = (
```
mmatera, line 536: Again, according to the experiments I have run, `MakeBoxes` rules should be agnostic about the encoding. The encoding should be take into account when the Boxes are "rendered" by a consumer (the command line, the function that implements ToString, or the exporter.)

rocky: Right now, we don't have a formatting phas for this for infix operators and probably not for any operators in general.

What was done here was to push this back one small level from making formatting decisions on infix operators inside static MakeBox rules to the evaluation of the MakeBox. 

Furthermore, moving this to a separate formatting phase looks like it will be major work. 

What I have done here though was to _mark_ with FIXME the places that violate the Eval, Box, Format process for later.
So in that sense this is a step and direction towards where this needs to go.

As I said, given the size of this PR, and the amount of work that still needs to be done, not just here but across the board. my take is to go with this for now, and give up on further Form work, until we have better isolated Boxing routines, and can start isolating more formatting routines on something simpler.

rocky: Also, I would like to note that the other big thing of these PRs is that we have started to separate properties of characters and operators and pull these in from mathics scanner.   In particular using character tables we can start to represent translations of operators to Unicode, and ASCII. (WL-Unicode, and AMSLaTeX could easily follow).

There is  more work to do here too. For example operator precedence should come from mathics scanner. However this is a start.

In sum, the amount of refactoring work is quote huge and I do not think can be done in one step. However, there is a vision or direction outlined here, a step made towards that, and FIXME's for where there is more to do.

mmatera: OK, then go ahead. Merge it when you feel ready.

PR #574, opened 2022-10-12
Parse string arguments in ToString

rocky:
This is a little closer to the correct behavior when `ToString[]` is passed a string argument. 

We currently can't handle passing an "encoding" parameter to `ToString[]`, for example:

```mathematica
ToString['1->3', OutputForm, CharacterEncoding->"Unicode"]
```

The main obstacle here is understanding how go get the desired encoding passed down into `MakeBoxes`. Docs do not explicitly list this as an option to `MakeBoxes` that can be specified. Does it exist?  If not how does WMA pass this information down. 

However it is done, in our current code base, we do not have such a feature (sort of changing `$CharacterEncoding` which we should avoid doing.


mmatera:
> This is a little closer to the correct behavior when `ToString[]` is passed a string argument.
> 
> We currently can't handle passing an "encoding" parameter to `ToString[]`, for example:
> 
> ```mathematica
> ToString['1->3', OutputForm, CharacterEncoding->"Unicode"]
> ```
> 
> The main obstacle here is understanding how go get the desired encoding passed down via the direct call to `MakeBoxes` down to formatting where this should be used.
> 
> WMA do not explicitly list `CharacterEncoding` as an option to `MakeBoxes`. Does it exist and is not documented? Assuming it does not exist, how would WMA pass this information down?
> 

No, it does not exists (I tested it). And, as I showed before, MakeBoxes does not use that information to build a Boxed expression. It happens afterwards (when Boxes are converted into text strings) 

> However it is done, in our current code base, we do not have such a feature (short of changing `$CharacterEncoding` which we should avoid doing.
> 
> [](https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/574)

Also, this does not happen in WMA. Otherwise, it would come up in my tests.  In some way, `ToString` produces Boxed expressions, and then apply the encoding when the boxed expression is translated into a String.

mathics/builtin/atomic/strings.py
```diff
@@ -851,6 +869,16 @@ def apply_default(self, value, evaluation, options):
     def apply_form(self, expr, form, evaluation, options):
         "ToString[expr_, form_, OptionsPattern[ToString]]"
         encoding = options["System`CharacterEncoding"]
+
+        if isinstance(expr, String):
+            # ToString[] of a string parses the string into an M-expression and then
+            # runs ToString[] on the resulting M-expression.
+            session = MathicsSession()
+            try:
+                expr = parse(session.definitions, MathicsSingleLineFeeder(expr.value))
```
mmatera, line 878: really do you want to create a session to convert a string?

rocky: I suppose it is better to just Definitions. I am not sure though if we want `add_builtin` set `True` or `False`. You tell me.

The thought here was to use a  clean slate of the definitions, and not modification made to the built-in functions. 

But you tell me, does `ToString` behavior of an expression change if there has been an altered the behavior of functions?

rocky: > MakeBoxes does not use that information to build a Boxed expression. 

I wasn't intending to imply that. Instead there is a function that accepts an encoding parameter. In the course of doing its thing it triggers a `MakeBox`  

Right now `MakeBoxes` performs some formatter before it returns  We need to get the information through to the formatter that `MakeBoxes` is invoking .  In a functional approach which is what would be preferred here, this would passed via the function call arguments somehow. 

> It happens afterwards (when Boxes are converted into text strings)

Currently, in Mathics Boxing and Formatting can be interleaved. So right now there is no absolute "afterwards".  

In Wolfram Alpha are there ever situations where Boxing and Formatting are interleaved? 

Removing the interleaving would require a major refactoring of our code. If that needs to be done, I think we need to reduce our goals for the time being and set a more achievable bar to reach.

rocky:
In order to dig deeper into `ToString[]` and what's up with encoding, it looks the `encode_tex` and `encode_mathml` functions should be replaced, and the  `do_format_xxx` routines `mathics/core/formatter.py` needs to be rethought.  Sigh.

mathics/builtin/atomic/strings.py
```diff
@@ -998,26 +774,140 @@ def apply_truncated(self, s, n, m, expression, evaluation):
...
+    >> "U" <> 2
+     : String expected.
+     = U <> 2
+    >> "U" <> ToString[2]
+     = U2
+
+    >> ToString[Integrate[f[x],x], TeXForm]
+     = \\int f\\left[x\\right] \\, dx
+
+    When 'ToString' is passed a string, the string is first parsed to an expression.
```
mmatera, line 838: This does not happen in WMA:
```
In[1]:= ToString["x  ->   7", StandardForm]                                     

Out[1]= x  ->   7

In[2]:= ToString["F[2"]                                                        

Out[2]= F[2

```
In the first case, parsing to an expression and converting back to `String` would normalize the form of the output, removing the extra spaces. In the second case, we should get a syntax error.

mmatera:
> This is a little closer to the correct behavior when `ToString[]` is passed a string argument.
> 
> We currently can't handle passing an "encoding" parameter to `ToString[]`, for example:
> 
> ```mathematica
> ToString['1->3', OutputForm, CharacterEncoding->"Unicode"]
> ```
> 
> The main obstacle here is understanding how go get the desired encoding passed down via the direct call to `MakeBoxes` down to formatting where this should be used.
> 
> WMA does not explicitly list `CharacterEncoding` as an option to `MakeBoxes`. Does it exist and is not documented? Assuming it does not exist, how would WMA pass this information down?
> 

Now this seems to be more clear: ToString calls `MakeBoxes` to produce a boxed version of the Expression. Then the Box Expression is rendered as Text. In that step, we can ask the render method to use a specific encoding as an option (this part is not currently implemented, but it seems straightforward).

mathics/builtin/atomic/strings.py
```diff
@@ -998,26 +774,140 @@ def apply_truncated(self, s, n, m, expression, evaluation):
...
+
+    >> ToString["1->2", OutputForm]
      = ...
-    """
 
-    name = "$SystemCharacterEncoding"
+    >> ToString["1->2"] == ToString[1 -> 2]
```
mmatera, line 843: In WMA,
```
In[1]:= ToString["1->2"] == ToString[1 -> 2]                                    

Out[1]= False
```

rocky: Ok. Thanks - mystery solved. In the back of my mind, I sort of expected this kind of thing was going on.

Closing this PR, since I think we now have a better handle on this.

PR #575, opened 2022-10-14
Improving the way in which outputs are presented in the interpreter

mmatera:
This PR modifies the interpreter to show (like in WMA) the output of formatted expressions like

```
In[1]:=InputForm[expr]
Out[1]//InputForm = expr
```

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/575"></a>

### DIFF
--- a/mathics/main.py
+++ b/mathics/main.py
@@ -126,12 +126,21 @@ class TerminalShell(MathicsLineFeeder):
         else:
             return "{1}In[{2}{0}{3}]:= {4}".format(next_line_number, *self.incolors)
 
-    def get_out_prompt(self):
+    def get_out_prompt(self, format=None):
         line_number = self.get_last_line_number()
-        return "{1}Out[{2}{0}{3}]= {4}".format(line_number, *self.outcolors)
+        if format:
+            return "{2}Out[{3}{0}{4}]//{1}= {5}".format(
+                line_number, format, *self.outcolors
+            )
+        else:
+            return "{1}Out[{2}{0}{3}]= {4}".format(line_number, *self.outcolors)
 
     def to_output(self, text):
         line_number = self.get_last_line_number()
+        # TODO: check if this method still works when //`Format`
+        # is appended.
+        # Also, consider to call `get_out_prompt` to build
+        # the string...
         newline = "\n" + " " * len("Out[{0}]= ".format(line_number))
         return newline.join(text.splitlines())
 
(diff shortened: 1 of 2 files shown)


mmatera:
Similar changes should be implemented in Mathics-Django an Mathicsscript

mathics/main.py
```diff
@@ -126,12 +126,21 @@ def get_in_prompt(self):
         else:
             return "{1}In[{2}{0}{3}]:= {4}".format(next_line_number, *self.incolors)
 
-    def get_out_prompt(self):
+    def get_out_prompt(self, format=None):
         line_number = self.get_last_line_number()
-        return "{1}Out[{2}{0}{3}]= {4}".format(line_number, *self.outcolors)
```
rocky, line 131: A very small thing, but nowadays f-strings are nicer than `.format()`.

mathics/main.py
```diff
@@ -451,7 +462,15 @@ def dump_tracing_stats():
                 print(query)
             result = evaluation.evaluate(query, timeout=settings.TIMEOUT)
             if result is not None:
-                shell.print_result(result, strict_wl_output=args.strict_wl_output)
+                head = result.last_eval.get_head()
+                form = ""
+                if head in evaluation.definitions.printforms:
```
rocky, line 467: It appears that there are other forms not in `$PrintForms` that till get rendered `//XXXForm=`. In particular `QuantityForm` or `ScientificForm`.  But others like `StringForm` don't work that way. 

Although this PR came first, it  should come after and build on #567 .  With that the separation of Form classes like those in `$PrintForms` versus those like `StringForm` can be marked inside the code, say as a class variable. And then the BaseForm class could gather those marked into something like `printforms` but is more appropriate to front-end marking `//XXXForm`. 

Is there something that distinguishes `StringForm` kinds of forms from those that appear in as `//`?  Is there something that distinguishes ``$PrintForm` items from those things that can appear after `//`?

mmatera: You are right. It should be for items of `$OutputForms` instead of `$PrintForms`.
Both should be updated when a new format rule is set.

mmatera: About `StringForm`, maybe it is because there is not a clear "underlying" expression associated to the form, like in the other Forms. For example, `BaseForm[10,2]` means `10`, and it is shown in binary representation. In `StringForm["``+``", a,b]`,
`"``+``"` does not mean anything without the other elements.

mmatera:
#576 provides a more accurate solution.

PR #576, opened 2022-10-14
Split Forms from inout

mmatera:
This PR splits the `mathics.builitin.inout` module into   `mathics.builitin.inout` and `mathics.builitin.forms`.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/576"></a>


mathics/builtin/forms.py
```diff
@@ -0,0 +1,520 @@
+"""
+Print Forms
```
rocky, line 2: "Print Forms" seems wrong. I wouldn't call `InputForm`  a print form - although it can be used like that, its main purpose is for _input_.  `StringForm` is included here, and it too is not a print form, or not something listed in `$PrintForms`. 

Also, at the top it would be good to describe _something_ about what a form is.

mathics/builtin/forms.py
```diff
@@ -0,0 +1,520 @@
...
+
+        result = super(MatrixForm, self).apply_makeboxes(table, f, evaluation, options)
+        if result.get_head_name() == "System`GridBox":
+            return Expression(
+                SymbolRowBox, ListExpression(String("("), result, String(")"))
+            )
+        return result
+
+
+class StringForm(Builtin):
```
rocky, line 207: On new or revised code, let's please put classes in alphabetic order unless there is a need not to. Thanks.

mathics/builtin/forms.py
```diff
@@ -0,0 +1,520 @@
...
+
+    attributes = A_LOCKED | A_PROTECTED
+    name = "$PrintForms"
+    summary_text = "list the print formats"
+
+    def evaluate(self, evaluation):
+        return ListExpression(*evaluation.definitions.printforms)
+
+
+class TableForm(Builtin):
```
rocky, line 77: Let's please create a `BaseForm` class which is starts out: 

```
class BaseForm(Builtin):
    """Base class which all Forms are derived from"""
```

I suspect we will also soon want to add methods to this.  formatting (which could be called "evaluation") might be such a method which might come soon.

mathics/builtin/forms.py
```diff
@@ -0,0 +1,520 @@
...
+     = True
+
+    """
+
+    attributes = A_LOCKED | A_PROTECTED
+    name = "$PrintForms"
+    summary_text = "list the print formats"
+
+    def evaluate(self, evaluation):
+        return ListExpression(*evaluation.definitions.printforms)
```
rocky, line 74: I realize `printforms` was from a prior PR, as I follow this. The setup of printforms in `systemsymbols`  is wrong. 

It doesn't allow the user to create more printforms. 

Instead there should be a class variable indicating whether a Form belowns in `definitions.printforms` and the base class should add that.

mathics/builtin/forms.py
mmatera: Please, notice that `BaseForm` is a builtin (https://reference.wolfram.com/language/ref/BaseForm.html)
That base class should be named something like `_BaseForm`

mathics/builtin/forms.py
mmatera: Actually, if you want to add `MyForm`, you can do something like
```
In[1]:= Format[F[x_],MyForm]:=MyForm[x_]
```
then, `MyForm` is appended to `$PrintForms` (and also should be appended to  `$OutputForms`).

rocky: Ok  - would you just make the change.

mathics/builtin/forms.py
rocky: Good - would you please just make the change?

mathics/builtin/forms.py
```diff
@@ -34,224 +40,47 @@
...
-        result = super(MatrixForm, self).apply_makeboxes(table, f, evaluation, options)
-        if result.get_head_name() == "System`GridBox":
-            return Expression(
-                SymbolRowBox, ListExpression(String("("), result, String(")"))
-            )
-        return result
-
-
-class StringForm(Builtin):
+class _FormBaseClass(Builtin):
```
rocky, line 53: This should not be private. Users writing a Form extension should be able to use this without guilt, same as they would for `Builtin` or `Predefined`.

mmatera: OK, but then this should be moved to `mathics.builtin.base`

rocky: It doesn't feel like that to me. This may grow with all sorts of Form specific methods and that would be better put closer to its use, here. 

I have seen this kind of thing say with symbols. I am not sure Format symbols should be in symbols. What happens is that we get these modules that are large and require or touch stuff that isn't that related to them. 

I had considered adding a new directory here, forms, and then adding this as `base` inside that. And also the non `$OutputForms` like `StringForm`, `HornerForm`, `ScientificForm` would go there, and the things in `$OutputForm` would be in another module too. 

But right now I thought to tread lightly and not cause too much disruption since this doesn't fully work yet.

mmatera: OK, I think we can do it using #584

rocky:
@mmatera Thanks - I'll look at this and #584 and #575 over the weekend.

mmatera:
@rocky, just to know, is this ready?

rocky:
Yes

mmatera:
Then, let's merge this and iterate.

PR #578, opened 2022-10-15
eval as a synonym of apply...

mmatera:
We have methods that start `apply_` that are really evaluation methods. This allows us to migrate the code to start using `eval_` like everyone else does.


rocky:
LGTM - and thanks!

PR #591, opened 2022-11-04
Add rudimentary SetOptions

rocky:
I have been looking at "Computer Science with Mathematica_ Theory and Practice for Science, Mathematics, and Engineering" by Maeder. It uses `SetOptions` so this adds that. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/591"></a>

### DIFF
--- a/mathics/builtin/options.py
+++ b/mathics/builtin/options.py
@@ -1,9 +1,15 @@
 # -*- coding: utf-8 -*-
 
 """
-Options and Default Arguments
-"""
+Options Management
 
+A number of functions have various options which control the behavior or the default behavior that function.
+Default options can be queried or set.
+
+<url>
+:WMA link:
+https://reference.wolfram.com/language/guide/OptionsManagement.html</url>
+"""
 
 from mathics.builtin.base import Builtin, Test, get_option
 from mathics.builtin.drawing.image import Image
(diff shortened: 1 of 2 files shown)


mathics/builtin/options.py
```diff
@@ -113,12 +130,13 @@ def matched():
...
 
+# Has this been removed from WL? I cannot find a WMA link.
 class NotOptionQ(Test):
     """
     <dl>
-    <dt>'NotOptionQ[$expr$]'
-        <dd>returns 'True' if $expr$ does not have the form of a valid
-        option specification.
+      <dt>'NotOptionQ[$expr$]'
+      <dd>returns 'True' if $expr$ does not have the form of a valid \
```
rocky, line 138: Note that `\` is the way to add a soft line break.

mmatera: this works if the docstring is not a raw docstring, isn't it?

mmatera (approved):
LGTM

mathics/builtin/options.py
rocky: Yes, you are correct, not raw When I start out with `r"""`  the backslash appears in the Django doc output .

mathics/builtin/options.py
```diff
@@ -401,6 +429,49 @@ def apply_3(self, f, optvals, optname, evaluation):
             return Symbol(name)
         return val
 
+    # FIXME until we figure out what test/test_evaluation.py fails
+    apply_3 = eval_3
```
rocky, line 433: @mmatera there is something incomplete regarding `def eval..` when there is an (error) message printed. This can be seen by removing this line and running pytest on `test/test_evaluation.py` https://github.com/Mathics3/mathics-core/actions/runs/3390463097/jobs/5634637319#step:6:4793 shows this failure.

PR #592, opened 2022-11-04
my suggestions

mmatera:
@rocky, these are my suggestions for your PR.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/592"></a>

### DIFF
--- a/mathics/builtin/assignments/internals.py
+++ b/mathics/builtin/assignments/internals.py
@@ -558,6 +558,21 @@ def process_assign_format(self, lhs, rhs, evaluation, tags, upset):
     return count > 0
 
 
+def process_assign_makeboxes(self, lhs, rhs, evaluation, tags, upset):
+    # FIXME: the below is a big hack.
+    # Currently MakeBoxes boxing is implemented as a bunch of rules.
+    # See mathics.builtin.base contribute().
+    # I think we want to change this so it works like normal SetDelayed
+    # That is:
+    #   MakeBoxes[CubeRoot, StandardForm] := RadicalBox[3, StandardForm]
+    # rather than:
+    #   MakeBoxes[CubeRoot, StandardForm] -> RadicalBox[3, StandardForm]
+    makeboxes_rule = Rule(lhs, rhs, system=True)
+    makeboxes_defs = evaluation.definitions.builtin["System`MakeBoxes"]
+    makeboxes_defs.add_rule(makeboxes_rule)
+    return True
+
+
 def process_assign_messagename(self, lhs, rhs, evaluation, tags, upset):
     lhs, condition = unroll_conditions(lhs)
     lhs, rhs = unroll_patterns(lhs, rhs, evaluation)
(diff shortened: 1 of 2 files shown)


rocky:
LGTM - merge or continue as you choose.

mathics/builtin/assignments/internals.py
```diff
@@ -685,23 +700,24 @@ def process_tags_and_upset_allow_custom(tags, upset, self, lhs, evaluation):
...
-        "System`Messages": process_assign_definition_values,
-        "System`Attributes": process_assign_attributes,
-        "System`Options": process_assign_options,
-        "System`$RandomState": process_assign_random_state,
         "System`$Context": process_assign_context,
         "System`$ContextPath": process_assign_context_path,
-        "System`N": process_assign_n,
-        "System`NumericQ": process_assign_numericq,
-        "System`MessageName": process_assign_messagename,
+        "System`$RandomState": process_assign_random_state,
```
rocky, line 705: Something  to keep in mind for when the time is right...

The key should be a `Symbol` not a `str`.  This will probably change the code a bit though (but for the better).

mmatera: I have been oscillating on what is the best. Last year I had the idea that the key for `Definitions` should be a `Symbol`, instead of a str`. However, for the way in which Python `dict` works, each time we call the `get` method of the dict, we have to compute a hash value. But then, to evaluate a hash value for a native type is much faster than doing it from a custom class. This is why here and in several other places of the code, the keys are `str` instead of `Symbol`s.

axkr: If the `Symbol`s are unique, can you use an approach like this?
https://stackoverflow.com/a/17054416/24819

mmatera: Symbols should not be a problem, because are singletons and are not being destroyed during a session. However, in any case, my only concern here is not to go back to the spaghetti version of this code...

rocky: > I have been oscillating on what is the best. Last year I had the idea that the key for `Definitions` should be a `Symbol`, instead of a `str`. However, for the way in which Python `dict` works, each time we call the `get` method of the dict, we have to compute a hash value. But then, to evaluate a hash value for a native type is much faster than doing it from a custom class. This is why here and in several other places of the code, the keys are `str` instead of `Symbol`s.

This is penny wise and pound foolish.

For now, the overriding concern should be clarity, not efficiency.

A lot of damage to this code base has been done in the name of efficiency, and the code is neither clear nor efficient.

So let us start out with what is the intent. The intent is that the action is based on the the kind of builtin (Symbol in the WL sense, Function, or Expression LHS) rather than some arbitrary string.

As I am coming to learn though, we shouldn't be going down this path at all because the intent is that an assignment changes depending on the kind of LHS or head item and its underlying type. 

One problem timing `str` and our current `Symbol` class for `get` access from a `dict`, is that Python is constantly changing. 

Here are timing results for Python 3.6.15 versus 3.10.8

First 3.6.15:

```python
$ python
Python 3.6.15 (default, Sep 21 2021, 05:10:37)
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import timeit
>>> from mathics.core.systemsymbols import SymbolContext
>>> special_cases1 = {"System`Context": 1}
>>> special_cases2 = {SymbolContext: 1}
>>> timeit.timeit(lambda: special_cases1.get("System`Context"))
0.23206797917373478
>>> timeit.timeit(lambda: special_cases2.get(SymbolContext))
0.6045385750476271
>>> timeit.timeit(lambda: SymbolContext.get_name())
0.27853861986659467
>>>
```

Now Python 3.10.8: 

```python
>>> import timeit
>>> from mathics.core.systemsymbols import SymbolContext
>>> special_cases1 = {"System`Context": 1}
>>> special_cases2 = {SymbolContext: 1}
>>> timeit.timeit(lambda: special_cases1.get("System`Context"))
0.13024636614136398
>>> timeit.timeit(lambda: special_cases2.get(SymbolContext))
0.3871238350402564
>>> timeit.timeit(lambda: SymbolContext.get_name())
0.22654549591243267
```

There are a couple of things to notice. Yes, the way the code _as written now_ would be slightly slower. But notice how that gap decreases as we get to more modern Python. And this I don't think is an accident: we should let Python worry about the implementation of `dict.get`.  

One of our problems in the code is that there are a lot of conversions into and out of strings. And this is part of a general looseness in how things are represented. Since too often have to go with the simple but very inefficient Expression representation we have a lot of conversions into that as well. 

So in the timings, note that you just can't compare get of a Symbol with get of a str, but you also need to count the time to do the conversion to a string which happens all over the place and makes for conceptually more difficult code.

But the much more important thing is that the reason get with a symbol is slow is because there is this custom `__hash()__`  function.  The code is "too clever by a half".  If we were to have a Symbol object that functioned like a Lisp Symbol object you'd find that a `get` on that is _faster_ than a get on a string.

For Python 3.10.8:
```python
>>> class Context(): pass
...
>>> special_cases3 = {Context: 1}
>>> timeit.timeit(lambda: special_cases3.get(Context))
0.12829051516018808
```

This takes _less_ time that `dict.get` on a `str` (0.13024636614136398) _and_ it doesn't require the conversion. 

In sum, we have this very unfortunate sequence of events. 

The code is written is a vague, simplistic and sometimes sloppy way and its intent isn't made clear. The next generation comes around and says I'll just try to speed it up without trying to make sense of it.  That is not only not working out but in some cases makes the situation worse. 

Instead of trying to figure out how to make `dict.get` run faster, we should be trying to understand how to get the `Symbol` class just be about Lisp-like Symbols and nothing else. Things like computing sort order on Atomic Elements, or the Mathics "==" shouldn't exist on Lisp-like Symbols, but instead some other class should do that.

PR #602, opened 2022-11-10
improve `OneIdentity` builtin

mmatera:
OneIdentity is an attribute that allows to implement of the property of Idempotency in the pattern matching mechanism.

https://reference.wolfram.com/language/ref/OneIdentity.html
https://mathematica.stackexchange.com/questions/124372/about-oneidentity

Documentation of this attribute in WR is not very complete, so the previous implementation was the result of a poor understanding of the expected behavior. This also forced us to put a lot of hacks in order to make basic things like assignments to work.

This PR simplifies a lot this behavior, removing unnecessary parameters in the `match` methods, and provides a behavior closer to the expected WL. Also, several new pytests were included.

In the process, I started to add comments on the `mathics.core.pattern` module to make it easier to understand and follow (I still do not do it fully). 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/602"></a>


mathics/builtin/assignments/internals.py
```diff
@@ -611,9 +611,7 @@ def process_tags_and_upset_dont_allow_custom(tags, upset, self, lhs, focus, eval
     # TODO: the following provides a hacky fix for 1259. I know @rocky loves
     # this kind of things, but otherwise we need to work on rebuild the pattern
     # matching mechanism...
-    flag_ioi, evaluation.ignore_oneidentity = evaluation.ignore_oneidentity, True
```
mmatera, line 614: Finally, I could get ride of this...

mathics/builtin/attributes.py
```diff
@@ -476,9 +476,14 @@ class OneIdentity(Predefined):
 
     'OneIdentity' affects pattern matching:
     >> SetAttributes[f, OneIdentity]
-    >> a /. f[args___] -> {args}
+    >> a /. f[x_:0, u_] -> {u}
```
mmatera, line 479: Reading the documentation, I realize that this test was wrong...

mathics/builtin/files_io/files.py
```diff
@@ -296,7 +296,7 @@ class FilePrint(Builtin):
     }
 
     def apply(self, path, evaluation, options):
-        "FilePrint[path_ OptionsPattern[FilePrint]]"
+        "FilePrint[path_, OptionsPattern[FilePrint]]"
```
mmatera, line 299: Fixing OneIdentity, also come up this other bug...

mathics/core/evaluation.py
```diff
@@ -262,9 +262,6 @@ def __init__(
         # status of last evaluate
         self.exc_result = self.SymbolNull
         self.last_eval = None
-        # Necesary to handle OneIdentity on
-        # lhs in assignment
-        self.ignore_oneidentity = False
```
mmatera, line 267: bye bye...

mathics/core/pattern.py
```diff
@@ -327,7 +347,6 @@ def yield_choice(pre_vars):
                     first=True,
                     fully=fully,
                     element_count=len(self.elements),
-                    wrap_oneid=expression.get_head_name() != "System`MakeBoxes",
```
mmatera, line 330: neither this another hack...

test/test_context.py
```diff
@@ -60,37 +62,37 @@ def test_context1():
     check_evaluation("bar[]", "42", to_string_expr=False, to_string_expected=False)
 
 
-def test_context2():
-    nomessage = tuple([])
-    for expr, expected, lst_messages, msg in [
+@pytest.mark.parametrize(
```
mmatera, line 65: I reformulated this test as a parametrized test because it was easier to debug...

test/test_rules_patterns.py
```diff
@@ -1,7 +1,133 @@
 # -*- coding: utf-8 -*-
+import os
+
+import pytest
+
 from .helper import check_evaluation
 
 
+DEBUGRULESPAT = int(os.environ.get("DEBUGRULESPAT", "0")) == 1
```
mmatera, line 9: Still there are a part of the behaviour that I did not cover. This allows to choose mark the corresponding tests as xfail or skip, as in other unit tests.

test/test_rules_patterns.py
```diff
@@ -1,7 +1,133 @@
...
+        (
+            "s[a,b,b,c]/.s[x_,x_]->sp[x]",
+            "s[a, rp[b], c]",
+            "https://reference.wolfram.com/language/tutorial/Patterns.html",
+        ),
+        (None, None, None),
+    ],
+)
+@skip_or_fail
+def test_one_identity_stil_failing(str_expr, str_expected, msg):
```
mmatera, line 120: To pass these tests, we should fix the `Rule.apply` method. I left it for another round.

TiagoCavalcante (approved):
Looks great!

mathics/builtin/attributes.py
```diff
@@ -466,9 +466,14 @@ class OneIdentity(Predefined):
...
      = {a}
-    It does not affect evaluation:
+
+    However, without a default argument, the pattern does not match:
+    >> a /. f[u_] -> {u}
+     = a
+
+    Also, it does not affect evaluation:
```
rocky, line 476: I find these examples and the documentation to be extremely confusing. 

Do you mind if I just rewrite the examples and doc here? 

A small change is to replace "Also, it does not" with "'OneIdentity' does not" 

Page 272 of the Mathematica 5 book motivates OneIdentity: 

> ... in the case where `x_` matches a single argument in a flat function, the question comes up as to whether the object it matches is really just and argument _a_ itself or _f[a]_. Mathematica choses the first of these cases if the function caries the attribute `OneIdentity` and chooses the second case otherwise.  

"In" 16-20" make this clear.

mathics/builtin/assignments/internals.py
```diff
@@ -608,12 +608,7 @@ def process_rhs_conditions(lhs, rhs, condition, evaluation):
 
 
 def process_tags_and_upset_dont_allow_custom(tags, upset, self, lhs, focus, evaluation):
-    # TODO: the following provides a hacky fix for 1259. I know @rocky loves
-    # this kind of things, but otherwise we need to work on rebuild the pattern
-    # matching mechanism...
```
rocky, line 613: I note we didn't need to "work on rebuild the pattern matching mechanism". I point out partly because there is a tendency to exaggerate.

Instead, my take is that we needed to think about this more like it might be written if this code were written in WL itself and understand the WL OneIdentity better.

I am glad to see this hack get removed.

mathics/builtin/attributes.py
rocky: Good catch! This bug goes back to the very first Mathics version.

mathics/builtin/files_io/files.py
rocky: Another good catch!

test/test_rules_patterns.py
```diff
@@ -1,7 +1,133 @@
...
+            "https://reference.wolfram.com/language/tutorial/Patterns.html",
+        ),
+        (
+            "s[a,b,b,c]/.s[x_,x_]->sp[x]",
+            "s[a, rp[b], c]",
+            "https://reference.wolfram.com/language/tutorial/Patterns.html",
+        ),
+        (None, None, None),
+    ],
+)
```
rocky, line 118: Thanks for adding these tests. The ones from Patterns.html I think should go in a doctests with suitable commentary; they were written with the intention of describing how `OneIdentity` works.

mathics/builtin/attributes.py
mmatera: Sure, go ahead.

mathics/core/pattern.py
```diff
@@ -3,8 +3,9 @@
 # -*- coding: utf-8 -*-
 
 
+from mathics.core.atoms import Integer
 from mathics.core.element import ensure_context
-from mathics.core.expression import Expression
+from mathics.core.expression import Expression, SymbolDefault
 from mathics.core.symbols import Atom, Symbol, system_symbols
 from mathics.core.systemsymbols import SymbolSequence
```
rocky, line 10: I realize this was code from before. If you are up to it: 

`SymbolAlternatives`, `SymbolBlank`, `SymbolCondition`, `SymbolPattern` can be imported and removed from the initialization list on line 21  and the remaining items sorted.

mathics/builtin/assignments/internals.py
mmatera: It was not so bad as I thought then, but nevertheless I had to make changes on the mathics.core.pattern module. Maybe now we can rewrite this part better.

mathics/core/pattern.py
```diff
@@ -33,6 +34,11 @@
 
 
 def Pattern_create(expr):
```
rocky, line 36: Again, I realize this is legacy stuff, but we shouldn't have functions that start with capital letters. 

Suggested replacement signature: 

```
def pattern_create(expr: BaseElement) -> Union[AtomPattern | ExpressionPattern]:
       """
       If ``expr`` is listed in ``pattern_object``  return the pattern found there. 
       Otherwise, if ``expr`` is an ``Atom``, create and return  ``AtomPattern`` for ``expr``.
       Otherwise, create and return and ``ExpressionPattern`` for ``expr``.
       """

Comment (rocky) [reply mathics/core/pattern.py:36]: Also, do you know why we do on line 86:

```
   create = staticmethod(Pattern_create)
```

rather than use:

```
    @staticmethod
    def pattern_create(...)
``` 
?

File: test/test_rules_patterns.py (hunk 8 above)
Comment (rocky) [inline test/test_rules_patterns.py:120]: If you are up for it, let's move the builtin tests under `test/builtin`.  A lot of this would be `test/builtin/test_patterns.py`

File: mathics/core/pattern.py [hunk 14]
```diff
@@ -82,22 +95,33 @@ def match(
...
         #raise NotImplementedError
         result = []
         def yield_func(vars, rest):
             result.append(vars, rest)
         self._match(yield_func, expression, vars, evaluation, head,
-                    element_index, element_count, fully, wrap_oneid)
+                    element_index, element_count, fully)
         return result"""
 
     def does_match(self, expression, evaluation, vars=None, fully=True):
```
rocky, line 121: Suggested signature change 

```
def does_match(self, expression: BaseElement, evaluation: Evaluation, vars=Optional[dict], fully: bool=True) -> bool: 
```

rocky:
@mmatera In trying to come up with examples, I am not seeing this follow the examples shown in https://reference.wolfram.com/language/tutorial/Patterns.html

```

In[1]:= SetAttributes[r, Flat]
Out[1]= None

In[2]:= r[a, b, a, b] /. r[x_, x_] -> rp[x]
Out[2]= rp[r[a, b]]

In[3]:= SetAttributes[r, OneIdentity]
Out[3]= None

In[4]:= r[a, b, a, b] /. r[x_, x_] -> rp[x]
Out[4]= rp[r[a, b]]
```

Thoughts?

mmatera:
> @mmatera In trying to come up with examples, I am not seeing this follow the examples shown in https://reference.wolfram.com/language/tutorial/Patterns.html
> 
> ```
> 
> In[1]:= SetAttributes[r, Flat]
> Out[1]= None
> 
> In[2]:= r[a, b, b, c] /. r[x_, x_] -> rp[x]
> Out[2]= r[a, b, b, c]  (* Wrong: should be r[a, rp[r[b]], c] *)
> 
> In[3]:= SetAttributes[r, OneIdentity]
> Out[3]= None
> 
> In[4]:= r[a, b, b, c] /. r[x_, x_] -> rp[x]
> Out[4]= r[a, b, b, c]  (* Wrong: should be r[a, rp[b], c] *)
> ```
> 
> Thoughts?
> 
> What is an example that shows the difference between Flat versus Flat and OneIdentity?
> 
> Also, I notice that `SetAttributes[OneIdentity]` on WMA should give the error:
> 
> ```
> SetAttributes::argr: 
>    SetAttributes called with 1 argument; 2 arguments are expected.
> ```

This is something that I detected, but I didn't try to fix yet. I have put a test for this marked as xfail

mathics/core/pattern.py
rocky: #613 attempts to do this.

mathics/core/pattern.py
rocky: #613 attempts to do this. This is a holdover from the Jan days.

mathics/core/pattern.py
rocky: #613 adds this

PR #603, opened 2022-11-10
Correct LHS evaluation in SetDelayed assignment

mmatera:
This PR follows #602. After fixing OneIdentity, it was possible to fix a little bit the behavior of `Set*`, in a way that in expressions like
```
F[x_]:=G[x]

H[F[x_]]:= x^2
```

the assignment is done to `H[G[x_]]` instead of `H[F[x_]]`, i.e., the argument of the LHS is evaluated before the assignment. Also, more comments on the routines of the `mathics.builtin.assignment.internals` is provided.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/603"></a>


mathics/core/list.py
```diff
@@ -98,16 +102,17 @@ def evaluate_elements(self, evaluation):
...
-            self._build_elements_properties()
-            self.value = None
+        new_list = ListExpression(*elements)
+        # TODO: we could have a specialized version of this
+        # that keeps self.value up to date when that is
+        # easy to do. That is left of some future time to
+        # decide whether doing this this is warranted.
+        new_list._build_elements_properties()
+        new_list.value = None
+        return new_list
```
mmatera, line 115: Notice that in `Expression.evaluate_elements`, a new expression is returned.

rocky:
@mmatera I believe we are getting conflicts because this brange duplicates commits in other branches and assumed that the other branches weren't going to change before commit. 

If this follows #602 I think a better strategy is to base the branch off of #602 and create the PR to go back into #602.

mmatera:
> @mmatera I believe we are getting conflicts because this brange duplicates commits in other branches and assumed that the other branches weren't going to change before commit.
> 
> If this follows #602 I think a better strategy is to base the branch off of #602 and create the PR to go back into #602.

Yep, to make this work I had to merge the previous version of the oneidentity pr. This needs a rebase now, and yes, that strategy sound reasonable.

mathics/builtin/assignments/internals.py
```diff
@@ -122,6 +124,32 @@ def is_protected(tag, defin):
     return A_PROTECTED & defin.get_attributes(tag)
 
 
+def normalize_lhs(lhs, evaluation):
```
rocky, line 127: If seems to me that this function and, more generally, the entire module should be moved to `mathics.core`.

mathics/builtin/assignments/internals.py
```diff
@@ -122,6 +124,32 @@ def is_protected(tag, defin):
...
+    """
+    Process the lhs in a way that
+    * if it is a conditional expression, reduce it to
+      a shallow conditional expression
+      ( Conditional[Conditional[...],tst] -> Conditional[uncondlhs, tst])
+      with `uncondlhs` the result of strip all the conditions from lhs.
+    * if `uncondlhs` is not a `List` or a `Part` expression, evaluate the
```
rocky, line 134: A small note here. I have been using the RsT tagging inside docstrings like this. The reason id this that this is what Sphinx uses by default when told to create system-level documentation.

rocky: Small thing, as a variable name `lhs` is fine, when not a variable symbol, I think caps is better: Process the LHS because it is in fact an acronym like IBM.

rocky: the name "uncondlhs" feels awkward. Please consider using underscores to separate words, e.g. "uncond_lhs" or the maybe "stripped_lhs".

mathics/builtin/assignments/internals.py
rocky: Suggested function signature with type annotation: 

```
def normalize_lhs(lhs: Expression, evaluation: Evaluation) -> Tuple(Expression, str):
```

rocky: By the way, although there are lots of little comments, this style of function is a _great_ improvement over the style that we were left with. Thanks!

mathics/core/expression.py
```diff
@@ -538,6 +538,10 @@ def evaluate(
         return expr
 
     def evaluate_elements(self, evaluation) -> "Expression":
+        """
+        return a new expression with the same head, and the
+        evaluable elements evaluated.
```
rocky, line 543: "evaluable" is jarring. Maybe we can just remove this word. 

From an old IBM manual: 

> Depressing the Power On key initiates the power on sequence. 
> Depressing the Power Off key initiates the power off sequence.
> Depressing the Power On key when the power is already on has no effect.
> Depressing the Power Off key when the power is already off has no effect.

Similarly trying to evaluation an element where there is nothing to evaluate has no effect, so suspect does not need to be spelled out.

mathics/core/list.py
```diff
@@ -86,6 +86,10 @@ def __str__(self) -> str:
 
     # @timeit
     def evaluate_elements(self, evaluation):
```
rocky, line 88: Suggested signature: 

```
    def evaluate_elements(self, evaluation: Evaluation) -> Expression:
```

Yes, I realize this was from before. I am trying to up our game in terms of quality incrementally as we improve things.

mathics/core/list.py
rocky: Thanks for catching. I guess this was a bug fixed then. If this was causing a problem, I am curious as to where it was causing a problem.

test/builtin/test_assignment.py
```diff
@@ -158,6 +158,58 @@ def test_setdelayed_oneidentity():
...
+        (
+            "F[x_]:=G[x]; H[F[y_]]:=Q[y]; ClearAll[F]; {H[G[5]],H[F[5]]}",
+            "{Q[5], H[F[5]]}",
+            "The arguments on the LHS are evaluated before the assignment",
+        ),
+        (None, None, None),
+        (
+            "F[x_]:=G[x]; H[F[y_]]^:=Q[y]; ClearAll[F]; {H[G[5]],H[F[5]]}",
+            "{Q[5], H[F[5]]}",
+            "The arguments on the LHS are evaluated before the assignment",
```
rocky, line 183: It would be nice to change the failure message to something _different_ than the others above and below. In particular, _how_ is this test different from the others?

rocky:
@mmatera All very good and much needed. Thanks!

The coding style in this new code base has been greatly improved over what has been there before (although if it is not too much I am pushing for a tad more).

rocky:
LGTM - merge when satisifying.

mmatera:
Ok, when the ci finish, let's merge this and then we can continue with the other parts

PR #604, opened 2022-11-10
basic implementation of DisplayForm

mmatera:
This PR provides a basic / preliminar implementation for  `DisplayForm`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,7 @@ New Builtins
 #. ``Accuracy``
 #. ``ClebschGordan``
 #. ``Curl`` (2-D and 3-D vector forms only)
+#. ``DisplayForm``
 #. ``Kurtosis``
 #. ``PauliMatrix``
 #. ``Remove``
@@ -57,7 +58,8 @@ Enhancements
 #. Better handling of comparisons with finite precision numbers.
 #. Improved implementation for  ``Precision``.
 #. Infix operators, like ``->`` render with their Unicode symbol when ``$CharacterEncoding`` is not "ASCII".
-
+#. Basic implementation of ``DisplayForm``
+   
 5.0.2
 -----
 
--- a/mathics/builtin/forms/output.py
+++ b/mathics/builtin/forms/output.py
@@ -149,6 +149,50 @@ class BaseForm(Builtin):
             )
 
 
+class DisplayForm(FormBaseClass):
+    """
+    <dl>
+      <dt>'DisplayForm[$obj$]'
+      <dd>is a form that shows low-level box structures as the correspoding
+      print form.
+    </dl>
+
+    By default, box structures are shown without a format:
+    >> box = MakeBoxes[Infix[{a, b}," + "], StandardForm]
+     = RowBox[{a,  + , b}]
+    Wrapping the box structure in a $DisplayForm$ ensures to render the expression:
+    >> DisplayForm[box]
+     = a + b
+
+    The same is valid if the box is nested into an expression:
+    >> DisplayForm[F[box]]
+     = F[a + b]
+
+    If the argument does not have a box structure,  `DisplayForm` is ignored:
+    >> DisplayForm[a + b]
+     = a + b
+
+    """
+
+    in_outputforms = True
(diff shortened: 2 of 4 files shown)


TiagoCavalcante (approved):
LGTM

mathics/builtin/forms/output.py
```diff
@@ -149,6 +149,50 @@ def apply_makeboxes(self, expr, n, f, evaluation):
             )
 
 
+class DisplayForm(FormBaseClass):
+    """
+    <dl>
+      <dt>'DisplayForm[$obj$]'
+      <dd>is a form that shows low-level box structures as the correspoding
```
rocky, line 156: correspoding -> corresponding

mathics/builtin/forms/output.py
```diff
@@ -149,6 +149,50 @@ def apply_makeboxes(self, expr, n, f, evaluation):
             )
 
 
+class DisplayForm(FormBaseClass):
+    """
```
rocky, line 153: Please add a <url> link here. See the next class below for an example. Thanks.

CHANGES.rst
```diff
@@ -15,6 +15,7 @@ New Builtins
 #. ``Accuracy``
 #. ``ClebschGordan``
 #. ``Curl`` (2-D and 3-D vector forms only)
+#. ``DisplayForm``
```
rocky, line 18: Also there is that MANIFEST file you added that needs updating.

mmatera: I expected that the CI complains about this...

mathics/builtin/forms/output.py
```diff
@@ -149,6 +149,54 @@ def apply_makeboxes(self, expr, n, f, evaluation):
...
+    If the argument does not have a box structure,  `DisplayForm` is ignored:
+    >> DisplayForm[a + b]
+     = a + b
+
+    """
+
+    in_outputforms = True
+    summary_text = "render low-level box structures"
+
+    def apply_makeboxes(self, expr, f, evaluation):
```
rocky, line 184: apply -> eval in new code please.

CHANGES.rst
rocky: We should see why it doesn't.

mathics/builtin/makeboxes.py
```diff
@@ -394,6 +394,10 @@ def apply_general(self, expr, f, evaluation):
         """MakeBoxes[expr_,
         f:TraditionalForm|StandardForm|OutputForm|InputForm|FullForm]"""
         if isinstance(expr, BoxElementMixin):
+            # If we are inside a DisplayForm block,
+            # BoxElementMixin are not processed.
+            if evaluation.in_display_form:
```
rocky, line 399: Can this be handled by some sort of `HoldForm` attribute?

mmatera: I don't think so. In the first place, I guess that in WMA ,  it does not have that attribute for a reason. But, I also tried to avoid the new attribute, and I could not figure out how to do that without changing more parts of the code.

rocky: Ok - I'd like to check on this. Doing this kind of thing feels weird.

mmatera: Thinking about it again, one alternative would be to set a symbol in a private context, and then instead of looking at that property, look at the value of the symbol. This could have the advantage to make easy to replace the Python code that implements MakeBoxes by WL code. But in the end, it would be the same idea.

rocky: To start out, I believe we should be following the way that can be written in WL would do that. This keeps us from inventing mechanisms, and flags and so on. Also, since WL follows a certain logic, makes it more likely we will go with the natural flow here. 

If later we need to get things faster then starting with the logical flow is better than some guessed version.

rocky: As I think I mentioned on the other PR, `TagBox` is used by `DisplayForm` and that is probably a mechanism by which we get special behavior. 

See https://reference.wolfram.com/language/ref/TraditionalForm.html

mmatera:
Deprecated

PR #601, opened 2022-11-08
Adding comments and tests for Assignment

mmatera:
This PR adds some comments on the internal implementation of the assignment methods. Also, adds tests to show some behaviors that must be fixed in order to improve the compatibility with WMA.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/601"></a>

PR #605, opened 2022-11-10
Fix set ops another round

mmatera:
With this PR, that comes after #601, #602, #603 and #604, allows to load the mymakeboxes.m package, and hopefully, should help to load other packages like KnotTheory, Feyncalc and others.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/605"></a>


rocky:
@mmatera - This is and #601 to #604 are interesting and I would like to go through these and understand in more detail. I haven't had the time to do so, but will very soon. Thanks for doing this. 

I am excited to hear that Mathics package loading may now be fixed. This is very important.

rocky:
> With this PR, that comes after #601, #602, #603 and #604, 

Does this really depend on #604?  At any rate perhaps the merge branch should be #603 (or #604) instead of master. If either of these lands onto master, _then_  this can be merged to master directly. What we've seen is that when there are lot of changes to the dependent branches (and PRs here have to go in the order they were done), then we are likely to create merge conflicts here where none were needed.

mmatera:
> .

#604 and this are independent. What I was saying it that I need these PR to make mymakeboxes.m work.

mmatera:
reformulated in #616 as a PR over #603

PR #606, opened 2022-11-10
fix a bug that makes that URLSave always fails

mmatera:
Another random bug came up trying to see the progress in loading packages. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/606"></a>


TiagoCavalcante (approved):
LGTM

PR #607, opened 2022-11-11
Update 'attributes' for current standards

rocky:
`mathics.builtin.attributes`:

* `def apply(` -> `def eval(`
* Add WMA links 
* Class names ordered alphabetically
* Hard breaks in docstring removed
* Some incorrect references to "leaves" changed to "attributes"

`mathics.core.attributes`:

* Add short comments above flag value

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/607"></a>

PR #608, opened 2022-11-11
catch not known attributes in ClearAttributes and SetAttributes

mmatera:
@rocky, please before merging #607, add this patch, to a bug that came up trying to load FeynCalc

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/608"></a>

### DIFF
--- a/mathics/builtin/attributes.py
+++ b/mathics/builtin/attributes.py
@@ -72,6 +72,9 @@ class Attributes(Builtin):
     """
 
     attributes = A_HOLD_ALL | A_LISTABLE | A_PROTECTED
+    messages = {
+        "attnf": "`1` is not a known attribute.",
+    }
     summary_text = "find the attributes of a symbol"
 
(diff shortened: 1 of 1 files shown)


rocky:
@mmatera ok - after this passes CI let's merge and I will rebase off of this.

rocky:
@mmatera  If you have it, what is the exact error that occurs in FeynCalc?

rocky:
This PR would suggest that Feyncalc has a bug which is hard to believe. So what changed in Mathematica?

mmatera:
> @mmatera If you have it, what is the exact error that occurs in FeynCalc?

FeynCalc has a typo, that tries to Clear an attribute that do not exist. This patch just catch the exception.

Regarding the bug, it just tries to set ReadProtect attribute to some symbols, but there is a typo (something like ReadProtetc), that in WMA does not have any consequence, but here kills the kernel.

PR #612, opened 2022-11-12
Put test_rules_patterns tests where they belong

rocky:
Move tests into more specific tests for the module that the tests are in.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/612"></a>

### DIFF
--- a/test/builtin/atomic/test_symbols.py
+++ b/test/builtin/atomic/test_symbols.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+"""
+Unit tests from mathics.builtin.atomic.symbols.
+"""
+
+from test.helper import check_evaluation
+
+
+def test_downvalues():
+    for str_expr, str_expected, message in (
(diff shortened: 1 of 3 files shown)


mmatera (approved):
LGTM

mmatera:
Just one comment. The test are about how `Pattern` which is a class in mathics.core behaves over certain expressions, more than about how the mathics.builin.attributes module does. My guess is that the right place would be test.core.pattern.

rocky:
> Just one comment. The test are about how `Pattern` which is a class in mathics.core behaves over certain expressions, more than about how the mathics.builin.attributes module does. My guess is that the right place would be test.core.pattern.

Ok - I will make the change. But I am not exactly sure which function(s) you are referring to. Please be more specific. Or if you want, feel free to commit on top of this.

rocky:
I encourage moving tests around further later as we see fit. This is just a start.

PR #613, opened 2022-11-12
Dry pattern symbols.py

rocky:
Pattern_create -> Pattern.create (if you dig into the git history, it is informative how this got that way, and suggests what we shouldn't do in the future)
  
More pervasive use of Symbols
    
symbols.py:

system_symbols() -> symbol_set(); "systems_symbols" name is too close Symbol( System`  to module systemsymbols. Also, we now require symbols as parameters), not strings.
    
systemsymbols.py: more system symbols

Some annotations to functions in the Pattern class were added, and some imports made less circular by moving a module-level variable.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/613"></a>


mmatera (approved):
LGTM

PR #614, opened 2022-11-12
Dry pattern symbols.py

rocky:
More pervasive use of Symbols
    
symbols.py:

system_symbols() -> symbol_set(); "systems_symbols" name is too close  to module systemsymbols. Also, we now require symbols as parameters, not strings.
    
systemsymbols.py: more system symbols

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/614"></a>


rocky:
A mistake - there is already another  PR with a closer base branch.

PR #617, opened 2022-11-13
using a private symbol to store the state.

mmatera:
This is the alternative version. This makes the control variable "hackable" from inside the mathics session (we can read it) but avoids that the user overwrite the value (because it is protected and locked).

### DIFF
--- a/mathics/builtin/forms/output.py
+++ b/mathics/builtin/forms/output.py
@@ -30,6 +30,7 @@ from mathics.core.atoms import (
     StringFromPython,
 )
 
+from mathics.core.attributes import A_LOCKED, A_PROTECTED
 from mathics.core.expression import Expression, BoxError
 from mathics.core.formatter import format_element
 from mathics.core.list import ListExpression
@@ -194,7 +195,7 @@ class DisplayForm(FormBaseClass):
         try:
             result = MakeBoxes(expr, f).evaluate(evaluation)
         except Exception:
-            return
+            result = None
         evaluation.definitions.set_ownvalue(
             SYMBOL_FORMATBOXES.name, old_value_in_display_form
         )
@@ -1141,9 +1142,10 @@ class PrivateSymbolThatControlsTheStateOfDisplayForm(Builtin):
     </dl>
     """
 
-    name = "PrivateSymbolThatControlsTheStateOfDisplayForm"
+    attributes = A_LOCKED | A_PROTECTED
     context = "System`Private`MakeBoxes`"
-    summary_text = "control if boxes are reevaluated."
+    name = "PrivateSymbolThatControlsTheStateOfDisplayForm"
     rules = {
         "System`Private`MakeBoxes`PrivateSymbolThatControlsTheStateOfDisplayForm": "False",
     }
+    summary_text = "control if boxes are reevaluated."
--- a/mathics/builtin/forms/output.py
+++ b/mathics/builtin/forms/output.py
@@ -18,7 +18,7 @@ from mathics.builtin.base import Builtin
 from mathics.builtin.box.layout import GridBox, RowBox, to_boxes
 from mathics.builtin.comparison import expr_min
 from mathics.builtin.forms.base import FormBaseClass
-from mathics.builtin.makeboxes import MakeBoxes, number_form
+from mathics.builtin.makeboxes import MakeBoxes, SYMBOL_FORMATBOXES, number_form
 from mathics.builtin.tensors import get_dimensions
 
 from mathics.core.atoms import (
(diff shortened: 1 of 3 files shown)


mathics/builtin/forms/output.py
```diff
@@ -188,12 +189,16 @@ def eval_makeboxes(self, expr, f, evaluation):
         # MakeBoxes is formatting an expression inside a DisplayForm.
         # Hopefully, this is temporal and it is not going to be
         # needed after the Format/MakeBoxes refactor.
-        previous_df, evaluation.in_display_form = evaluation.in_display_form, True
+
+        old_value_in_display_form = SYMBOL_FORMATBOXES.evaluate(evaluation)
+        evaluation.definitions.set_ownvalue(SYMBOL_FORMATBOXES.name, SymbolTrue)
```
mmatera: Notice that here we can not use the higher level way
```
Expression(SymbolSet, SYMBOL_NO_FORMATBOXES, SymbolTrue).evaluate(evaluation)
```
because the attributes.

mmatera: The drawbacks of this approach are that the access to the attribute is slower, and the code to set the context is larger and weirder than just setting an attribute.

mathics/builtin/forms/output.py
```diff
@@ -1124,3 +1129,23 @@ def eval_makeboxes_matrix(self, table, f, evaluation, options):
...
 
         return result
+
+
+# Custom private symbol that helps DisplayForm to control if
+# MakeBoxes must process BoxElements as expressions, or keep as they are.
+class PrivateSymbolThatControlsTheStateOfDisplayForm(Builtin):
+    """
+    <dl>
+    <dt>'PrivateSymbolThatControlsTheStateOfDisplayForm'
```
rocky, line 1139: Small thing - please indent nested `<dt>`.

mathics/builtin/makeboxes.py
```diff
@@ -396,7 +403,7 @@ def apply_general(self, expr, f, evaluation):
         if isinstance(expr, BoxElementMixin):
             # If we are inside a DisplayForm block,
```
rocky, line 404: Is this really just DisplayForm that could do this? There are no other forms that are like this. Or this isn't something Form creator might want to do?

mmatera: By now, this is the only case. This could also be a way to implement `InputForm/OutputForm` and the other non-standard print forms but setting a different control symbol.  In any case, I am not sure that this way is better than the one I proposed before.

mathics/builtin/forms/output.py
mmatera: @rocky, the question is, do you think that this is a better approach?

rocky: @mmatera good question. I have to think about it. What I'd be looking for is clues in the WMA code. 

Think of this like the way a physicist would. We have two hypotheses for how to code. Is there a test that distinguishes these? Suppose we hypothesize that that WMA follows the WMA way to do this, what test could we come up with that would distinguish the implementations?

mathics/builtin/makeboxes.py
rocky: > and the other non-standard print forms

What makes a print form "non-standard"?

mathics/builtin/forms/output.py
mmatera: Ja, it is a curious request! Because it is exactly as this physicist is trying to distinguish what is the implementation in WMA :) 

Look at this, for example,

imagen

If I believe in this output, I have to assume that there is not a private hidden symbol but an internal variable inside the kernel.

mathics/builtin/makeboxes.py
mmatera: MakeBoxes accepts `StandardForm|TraditionalForm` as the second parameter. In some way, when you evaluate

```
MakeBoxes[OutputForm[expr], StandardForm]
```

Makeboxes first applies the format rules associated with `OutputForm` to  `expr` and its subexpressions. Then, the reference to `OutputForm` are lost in the expression, but in some way, MakeBoxes knows that has to apply the `OutputForm` makeboxes rules to the inner expressions. One way to simulate this behavior would be to set a state variable that remembers that we are using OutputForm rules. This is not the way in which I think we should implement this, but it is a possibility.

mathics/builtin/forms/output.py
rocky: Ok - I trust your conclusion, but I don't follow it. The output above doesn't indicate anything to me one way or another. Please break it down for a novice. 

If it were implemented as a private hidden symbol what then would you expect to see?

rocky: We should probably implement TracePrint so that we can check correspondence easier.

mmatera: I would expect an evaluation for the corresponding hidden symbol. This is an example in which internal variables are used:

imagen

(I just cut the first lines)

Notice the use of symbols in hidden contexts (BoxForm`ls, IntegrateDump`TestLimits, BoxForm`opt). Tracing `MakeBoxes[DisplayForm[...]]`  I do not see such kind of symbols in hidden contexts.

mathics/builtin/makeboxes.py
rocky: > but in some way, MakeBoxes knows that has to apply the OutputForm makeboxes rules to the inner expressions. 

https://reference.wolfram.com/language/ref/TraditionalForm.html says:

> TraditionalForm inserts invisible TagBox and InterpretationBox constructs into the box form of output it generates, to allow unique interpretation.tionalForm

Does this provide a mechanism for how to do this. If so, possibly the same is could done for StandardForm (and is done but just not mentioned). Sometimes it happens that only in implementing the second thing of a kind the general mechanism, and the first is done in a more hacky way is not retrofitted.

mmatera:
Deprecated

PR #616, opened 2022-11-13
Fix set eval 2rd round

mmatera:
#605 but over fix_set_eval 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/616"></a>

PR #618, opened 2022-11-13
fix get_sort_key second argument and BoxExpression.sameQ. 

mmatera:
With these changes, over #616, https://github.com/jkuczm/MathematicaSyntaxAnnotations/blob/master/SyntaxAnnotations/SyntaxAnnotations.m finally loads without killing the session.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/618"></a>

### DIFF
--- a/mathics/builtin/base.py
+++ b/mathics/builtin/base.py
@@ -818,7 +818,7 @@ class BoxExpression(BuiltinElement, BoxElementMixin):
     def get_lookup_name(self):
         return self.get_name()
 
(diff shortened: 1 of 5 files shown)


rocky:
What is a test that shows this bug?

mmatera:
Just download the file and try to load it. It will raise an unhanded exception.

mmatera:
The bug comes from a consistency issue in the signature of this method. At some point , we need to think about if this is correct, or if we need to split the method in two different methods.

PR #619, opened 2022-11-14
Fix set 3rd round

mmatera:
This completes the fixes needed to pass all the failing tests added in the first round.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/619"></a>


mathics/builtin/assignments/internals.py
```diff
@@ -779,6 +779,27 @@ def process_assign_messagename(self, lhs, rhs, evaluation, tags, upset):
...
+
+    while ``f[-4]/.(f[x_]:> (p[x]/;x>0))``
+    would result in ``f[-4]``
+
+    In Mathics, the first case results in
+    ``p[4]/; 4>0``
+    while the second,
+    ``p[-4]/; -4>0``
+
+    This should be handled in the apply method of the BaseRule class.
```
rocky, line 802: This was going to be my point in general. Too often it feels that when we try to run to run some WL code and that fails, the immediate solution is well, let's just add a bunch of flags and tests on symbols in Python code address the specific case we see. 

This is done instead of seeing the problem as a manifestation of some larger structural or problem feature lacking in the way particular built-in functions should work. 

I would prefer to see fixing the root cause issues in favor of the workaround coding. 



(You mean eval, but okay.)

mmatera: > This was going to be my point as well. Too often it feels that when we try to run to run some WL code and that fails, the immediate solution is well, let's just add a bunch of flags and tests on symbols in Python code address the specific case we see.
> 

No, the idea is to identify these issues, and then try to understand how it works and how it should work, and then fix it. 



> This is done instead of seeing the problem as a manifestation of some larger structural problem, or feature lacking in the way particular built-in functions should work.
> 

A manifestation of a large structural problem is the performance issue, and the fact that we can not load wl modules. All of this is about to narrow the different issues and try to determine what should be the pattern.


> I would prefer to see fixing the root cause issues in favor of the workaround coding.
> 

The workaround was there for a while. What I did here is to determine why this workaround was needed to reproduce the WMA behavior.  As the picture gets clear, we can continue removing these issues, like the one with oneidentity. What I am trying to avoid is to pretend to fix many things in the same PR.


> (You probably mean eval method, but okay.)
I said apply, because is the current name of the method. Then, if you think that is accurate to say "evaluate a replacement rule over an expression" over "apply a replacement rule over an expression" let's change the name of the method.

rocky: > The workaround was there for a while. What I did here is to determine why this workaround was needed to reproduce the WMA behavior. As the picture gets clear, we can continue removing these issues, like the one with oneidentity. What I am trying to avoid is to pretend to fix many things in the same PR.

So what I am thinking is let's focus on removing the workarounds that we can fix rather than changing workarounds. 

I know of some on the horizon but that could be fixed now, like #597 and the box precedence thing. Stuff like this keeps rolling in though which from my standpoint isn't helping as much as it is delaying getting more root issues fixed.

rocky: > I said apply, because is the current name of the method. Then, if you think that is more clear/ accurate to say "evaluate a replacement rule over an expression" over "apply a replacement rule over an expression" let's change the name of the method.

I just looked this up in the code. You are correct - this he method named `apply` _is_ doing application, not evaluation. And this reinforces why it would be better to have terminology cleaned up - I don't if a term is used in the right sense or the wrong sense and conceptually they mean different things so I get more confused.

rocky:
> No, the idea is to identify these issues, 

Ok - it would be helpful create detailed issues for these.

mmatera:
Superseded by  #625

PR #621, opened 2022-11-15
Fix pat->Condition[...] apply rules

mmatera:
This PR fixed the bug described in #619 regarding the result of applying this kind of rule. The problem was not in the `apply` method itself but in the `do_replace` method.


test/builtin/test_patterns.py
```diff
@@ -26,3 +26,17 @@ def test_replace_all():
         ),
     ):
         check_evaluation(str_expr, str_expected, message)
+
+
+def test_rule_repl_cond():
+    for str_expr, str_expected, message in (
+        ("f[x]/.(f[u_]->u^2/; u>3/; u>2)", "f[x]", "conditions are not evaluated"),
```
rocky, line 33: I just tried this on galatea and I don't see these results. Here is what I see:  

```
>> f[x]/.(f[u_]->u^2/; u>3/; u>2)
 = x ^ 2 /; x > 3 /; x > 2
>> f[1.]/.(f[u_]->u^2/; u>3/; u>2)
 = 1. /; 1. > 3 /; 1. > 2
```

mmatera: You are right. When I tested it, I did it with delayed rules instead. There is something that I still am missing.

mathics/core/rules.py
```diff
@@ -131,12 +131,22 @@ class Rule(BaseRule):
     ``G[1.^2, a^2]``
     """
 
-    def __init__(self, pattern, replace, system=False) -> None:
+    def __init__(self, pattern, replace, delayed=True, system=False) -> None:
         super(Rule, self).__init__(pattern, system=system)
         self.replace = replace
+        self.delayed = delayed
```
mmatera, line 137: This property helps to distinguish if the Rule comes from a Rule expression or a RuleDelayed expression, and it is needed to handle the different behavior observed in WMA. I do not see a reason for this different behavior, but maybe it is,  and maybe there is a different behavior in other contexts.

mmatera:
Still, there is something that is not covered with this: consider the following code:
```
F[k_] := 1 /; (k == 1);

F[k_] := 2  /; (k > 1);

F[k_] := A[k] ;

F[k_] := B[k] ;

DownValues[F]
```
In Mathics, the result is
```
{HoldPattern[F[k_]] ⧴ B[k]}
```
because the four lines sets over the same LHS pattern. On the other hand, the same code in WMA results in
```
{HoldPattern[F[k_]] :> 1 /; k == 1, HoldPattern[F[k_]] :> 2 /; k > 1, HoldPattern[F[k_]] :> B[k]}
```
so rules with the same LHS but different RHS is also overwritten, except if the RHS is a `Condition` expression.

rocky:
> Still, there is something that is not covered with this: consider the following code:
> 
> ```
> F[k_] := 1 /; (k == 1);
> 
> F[k_] := 2  /; (k > 1);
> 
> F[k_] := A[k] ;
> 
> F[k_] := B[k] ;
> 
> DownValues[F]
> ```
> 
> In Mathics, the result is
> 
> ```
> {HoldPattern[F[k_]] ⧴ B[k]}
> ```
> 
> because the four lines sets over the same LHS pattern. On the other hand, the same code in WMA results in
> 
> ```
> {HoldPattern[F[k_]] :> 1 /; k == 1, HoldPattern[F[k_]] :> 2 /; k > 1, HoldPattern[F[k_]] :> B[k]}
> ```
> 
> so rules with the same LHS but different RHS is also overwritten, except if the RHS is a `Condition` expression.

Maybe ask on https://mathematica.stackexchange.com/  what is up here and more generally how to think about how WL groups assignment categories?

mmatera:
> 

Actually, I am looking there, but at existing related questions.

mmatera:
Deprecated

PR #622, opened 2022-11-15
Changes suggested in PR review

rocky:
Move core-like assignment interals out of builtins and into core. (We may want to split up core more in the future though)

Better sort long list of assignment methods alphabetically

Some small RsT tagging in a docstring

Remove nonexistent word "evaluable"

Add yet another type annotation to a signature

Make test assert failure messages unique

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/622"></a>

### DIFF
--- a/mathics/builtin/assignments/assignment.py
+++ b/mathics/builtin/assignments/assignment.py
@@ -4,8 +4,27 @@ Forms of Assignment
 """
 
 
-from mathics.builtin.assignments.internals import _SetOperator
 from mathics.builtin.base import BinaryOperator, Builtin
+from mathics.core.assignment import (
+    AssignmentException,
+    assign_store_rules_by_tag,
+    normalize_lhs,
+    process_assign_attributes,
+    process_assign_context,
+    process_assign_context_path,
+    process_assign_default,
+    process_assign_definition_values,
+    process_assign_format,
+    process_assign_list,
+    process_assign_makeboxes,
+    process_assign_messagename,
+    process_assign_n,
+    process_assign_numericq,
+    process_assign_options,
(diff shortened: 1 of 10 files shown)


mathics/builtin/assignments/assignment.py
```diff
@@ -18,6 +37,62 @@
...
+    # FIXME:
+    # Assigment is determined by the LHS.
+    # Is there a larger patterns or natural grouping that we are missing?
+    # For example it might be that
+    # there are some attributes or other properties of of the
+    # LHS of a builtin that we should be targetting.
+    # Below, we key on a string, but Symbol is more correct.
+    #
+
+    special_cases = {
```
mmatera, line 58: Maybe this dictionary could be moved also to marhics.core.assignment

rocky: Good idea! Will do next.

mmatera (approved):
LGTM. The only problem with this is that I am going to have to rework the other three PR. In any case, ranch one just comprised small changes over this.

rocky:
> LGTM. The only problem with this is that I am going to have to rework the other three PR. In any case, ranch one just comprised small changes over this.

Please let's hold off on the others for now. It is not my intention to make additional work and I don't like to see this happen.

Make detailed issues describing the problem. 

Something we keep running into is that basically the code doesn't follow more modern standards of having signatures, ordering imports. (There is a python module called [isort](https://pycqa.github.io/isort/) that will do this). Naming eval methods with `eval` (when indeed that is what it is), adding WMA links and so on. 

So maybe before modifying a module we should make a pass over that first.

PR #623, opened 2022-11-15
Small function rename process->eval ...

rocky:
for someday in the future when we have an instruction interpreter rather than a tree evaluator.

We should also change "assign" into "set" or "set_delayed" or "set_..."  as appropriate. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/623"></a>


rocky:
Previously I had mentioned that maybe core should be split up. I now think that down the line we should create a `mathics.eval` module and move all of the `eval_` functions (those that are not inside a class) into this module similar to how things are done with tests.

mmatera:
LGTM. Also, have been thinking in the same way.

rocky:
> LGTM. Also, have been thinking in the same way.

Ok Great - would you like to sort the eval functions into SetDelayed versus Set versus ... (we have some in each category, right) and create the `mathics.eval` and start moving stuff there?

TiagoCavalcante (approved):
LGTM - I also agree with assign -> set

For me (without any base on CS :smiling_face_with_tear:) assign should be used when the assigned object changes its value, but assign doesn't change the "value" of a symbol, only changes one of its values (which now I don't remember the name), but we could, for example, set a numeric value and a different representation for a numerical constant.

mmatera:
> > LGTM. Also, have been thinking in the same way.
> 
> Ok Great - would you like to sort the eval functions into SetDelayed versus Set versus ... (we have some in each category, right?) and create the `mathics.eval` and start moving stuff there?

yes, why not? At this point, most of the code inside mathics.core.assignment is more or less isolated from the rest, so it shouldn´t interfere too much with the rest of the PRs.

mmatera:
I am trying to merge this over the last round of fixes. When it be ready, I am going to put a  PR to this brach

mmatera:
I am trying to merge this over the last round of fixes. When it be ready, I am going to put a  PR to this brach.

PR #626, opened 2022-11-16
Fix a crash produced when a pattern contains a BoxExpression as an element. 

mmatera:
This PR fixes a crash produced when a pattern contains a BoxExpression. Suppose for example that
we want to evaluate the expression
```
expr/.F[RowBox[list___]]->G[list] 
``` 
The pattern in the rule contains a RowBox, so in the evaluation, `F[RowBox[list___]]` is evaluated. Then, the `RowBox[]` expression is replaced by a `RowBox` object. Then, when `Rule` is evaluated,   Pattern.create is called, with ```Expression(Symbol("Global`F"), RowBox(...))``` as a parameter. It is not an Expression,  so `Patter.create` fails raising an exception.

In this PR this behavior is fixed by catching this case and converting `RowBox`  into an `Expression` before creating the pattern.

* Fix Pattern.create in a way that handles input expr that are not Atom or Expression, but have a to_expression method.
* Fix BoxExpression.sameQ to avoid infinite loops.
* adds the pattern_sort parameter to the ``BuiltinExpression.get_sort_key`` method.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/626"></a>


TiagoCavalcante (approved):
LGTM

rocky:
LGTM but a couple of comments. First it would be good to add a an example involving 
```
expr/.F[RowBox[list___]]->G[list] 
``` 

so that it is clearer what is getting fixed. 

Second there are these places where we don't have a `pattern_sort` parameter. Will any of these be a problem? 
```
mathics/builtin/patterns.py:1670:    def get_sort_key(self) -> tuple:
mathics/core/element.py:150:    def get_sort_key(self) -> list:
mathics/core/evaluation.py:160:    def get_sort_key(self) -> Tuple[bool, bool, str]:
mathics/core/rules.py:110:    def get_sort_key(self) -> tuple:
```

mmatera:
This seems to have been fixed in some other way, so let's close this.

PR #627, opened 2022-11-16
Adjust attribute test from the time when attributes  were string names instead of bit values

mmatera:
This is another small but found testing CellsToTeX. The new tests kills the session if they are run in master.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/627"></a>


test/builtin/test_attributes.py
```diff
@@ -226,3 +226,16 @@ def test_Attributes_wrong_args(str_expr, arg_count):
             f"SetAttributes called with {arg_count} arguments; 2 arguments are expected.",
         ),
     )
+
+
+def test_anonymous_func_attributes():
+    check_evaluation(
+        str_expr="""
+        a=Function[x, Hold[x]];
```
rocky, line 234: Simplicity and clarity don't seem to come naturally.  The problem fixed here has to do with  `b=Function[x, Hold[x], HoldAllComplete]`;  as far as I can tell, it doesn't have anything to do with `a=Function[x, Hold[x]]`. 

If that is in there for comparison of something, the person reading the test has to figure this out. By using the idiom of running two tests inside a list (with no comment and with no assert failure message), left this person minutes studying this to figure out what was actually getting tested.

rocky:
LGTM

test/builtin/test_attributes.py
mmatera: No, `Hold` is there to avoid that `1+1` evaluates to `2`.

mmatera: If you run this in master, you get an exception, because tries to evaluate a bitwise operator between an integer and a list of str.

mmatera: (But OK, I can add a more detailed description in the test).

rocky: I did run this in master. I did not get a crash on just the first test: 

```
In[1]:= a=Function[x, Hold[x]]
Out[1]= Function[x, Hold[x]]

In[2]:= a[1+1]
Out[2]= Hold[2]
```

And this is the same thing as you get in master.

A simpler example to show the problem is: 

```
In[1]:= b=Function[x, Hold[x], HoldAllComplete]
Out[1]= Function[x, Hold[x], HoldAllComplete]

In[2]:= b[1+1]
Out[2]= Hold[1 + 1]
```

PR #629, opened 2022-11-20
Init builtins

rocky:
This starts to address the mess that is in `mathics.builtins`, `mathics.core.definitions` and `pymathics`

It barely scratches the surface though. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/629"></a>


mathics/builtin/__init__.py
```diff
@@ -5,34 +5,25 @@
...
+    SympyObject,
+    mathics_to_python,
 )
-
 from mathics.core.pattern import pattern_objects
-
 from mathics.settings import ENABLE_FILES_MODULE
 from mathics.version import __version__  # noqa used in loading to check consistency.
 
-from typing import List
```
rocky, line 28: Although this is moved later below right now. All of this should not be in in `__init__.py` 

`__init__.py` should be small.

mathics/builtin/__init__.py
```diff
@@ -88,23 +79,6 @@ def contribute(definitions):
...
-    else:
-        title = module.__name__
-        for prefix in ("mathics.builtin.", "mathics.optional."):
-            if title.startswith(prefix):
-                title = title[len(prefix) :]
-        title = title.capitalize()
-        text = ""
-    return title, text
-
-
```
rocky, line 107: This is going to require a small import change in mathics_django. This has been moved into common_doc.py since this is used in generating documentation or doctests.

mathics/builtin/__init__.py
```diff
@@ -202,7 +161,7 @@ def sanity_check(cls, module):
 modules = []
 import_builtins(module_names)
 
-_builtins = []
+_builtins_list = []
```
rocky, line 164: `_builtins` is defined previously as a dictionary.  We should remove the bad coding style of picking a name and using type-morphing variable for whatever happens to be needed. 

Also both `_builtins`  (and `_builtins_list`)  are really bad names because they are generic. Another bad habit that is found in the code that should be removed.

Finally, instead of this global variable getting used in various functions in this module, it should be passed as a parameter.

mathics/builtin/__init__.py
```diff
@@ -262,16 +248,15 @@ def sanity_check(cls, module):
...
 mathics_to_sympy = {}  # here we have: name -> sympy object
-mathics_to_python = {}  # here we have: name -> string
 sympy_to_mathics = {}
 
 builtins_precedence = {}
 
-new_builtins = _builtins
+new_builtins = _builtins_list
 
 # FIXME: some magic is going on here..
```
rocky, line 261: I now understand a little that what was going on, as described above, was that there are global variables with changing meaning, that type shift and now it is getting  reset here for possibly for some different purpose.

mathics/builtin/base.py
```diff
@@ -3,24 +3,15 @@
 
 import importlib
 import re
-import sympy
```
rocky, line 6: I used isort to sort imports.

mathics/builtin/base.py
```diff
@@ -103,7 +97,7 @@ def split_name(name: str) -> str:
     return result.lower()
 
 
-mathics_to_python = {}
+mathics_to_python = {}  # here we have: name -> string
```
rocky, line 100: This appeared also in `mathics.builtins` . That use was removed and replaced by an import to this use.

rocky:
I suspect pymathics modules are broken in master. 

It feels like there is a lot of over generalization with this pymathics stuff, like being able to unload a pymathics module for a particular definition object.  It looks to me like this is adding complexity for little to no effective gain.

mmatera:
> I suspect pymathics modules are broken in master.
> 
 
We have some tests that seems to indicate that it works at least in a minimal level. But probably we should need to go over this after the API refactor.

rocky:
> > I suspect pymathics modules are broken in master.
> 
> We have some tests that seems to indicate that it works at least in a minimal level. But probably we should need to go over this after the API refactor.

The last time I went over the nltk packages they needed work independent of pymathics to make it work with more modern versions of the Python nltk package. 

And in addition to that there are fundamental questions about how the context namespacing should work that need to be resolved.

So in general, yes, the way pymathics works and the specific modules need going over.  And yes, this is something that should be put off for later.

rocky:
If there are no objections to this, I will move closer to what I was originally intending to do - one time initialization of builtins. 

There are several options... I can 

1. add new commits on top of this, 
1. add a new branch to merge on this
1. merge this and create another branch to continue

Suggest how you would like me to proceed. 

Until we decide, I think I will create a new branch locally since that allows me to accommodate all possibilities above.

mmatera:
Maybe, let's merge this and continue over it.

rocky:
Great - thanks!

PR #631, opened 2022-11-20
Move code out of definitions

rocky:
This moves builtin-creation out of mathics.builtin and core.definitions and into a new module.

I thing this is the last step before we actually move builtin creation of the Definitions initialization and into some one-time code.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/631"></a>


mmatera (approved):
LGTM

TiagoCavalcante (approved):
LGTM - I'm just wondering now why MakeBoxes has to contribute first

rocky:
> LGTM - I'm just wondering now why MakeBoxes has to contribute first

It is not _now_  - it always was that way. And I am wondering that too. @mmatera had one of his long explanations, given the current code  in master and process if follows,  why this makes things work. 

You can probably find that in a recent PR that about "fixing" Makeboxes. 

I wonder, though, if there isn't a simpler organization and way to do things that does not require this kind of special logic.  We have seen this kind of thing over and over. 

But for now, I am struggling just to pull stuff out of `mathics.builtin` and then the  Definitions object. And we've seen this kind of behavior too - we can't begin to do what we want because the code is a bit complicated. 

The wrongness of rereading builtin module over and over goes back to the earliest days V 0.5. But since then instead of not reading it once, a complicated custom module cache code was added, then pymathics was added, and then this MakeBoxes fix was added (or is in the process of being added?) 

Fixing the V 0.5 version is much easier than fixing the code as it is now. This is what happens when you build on a features on a foundation that isn't solid to begin with.

mmatera:
It is because other symbols can add rules to the MakeBoxes definition. Hopefully, we are going to get ride it this behavior

PR #633, opened 2022-11-21
OuterPrecenceForm->PrecedenceForm

mmatera:
The change is to use the WL name.


TiagoCavalcante:
This is already in master now

PR #636, opened 2022-11-21
Two bugs found via another PR

rocky:
`rules.py __setstate__()`:
   `mathics.builtin.builtins` was renamed to `mathics.builtin._builtins`
   I guess on the assumption it was private.

It turns out it wasn't. And we were using the older (more correct) name "builtins" instead of "_builtins". Use "_builtins" to make this work. `__setstate__` apparently isn't used right now, but would be if  we used `deepcopy()`. Note that in #635, this variable is renamed `system_builtins_dict`.

rest:
   Change imports of Builtin from `mathics.builtin` to  `mathics.builtin.base`. `mathics.builtin only accidentially imports `Builtin`.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/636"></a>

### DIFF
--- a/mathics/core/rules.py
+++ b/mathics/core/rules.py
@@ -232,9 +232,9 @@ class BuiltinRule(BaseRule):
         return odict
 
     def __setstate__(self, dict):
-        from mathics.builtin import builtins
+        from mathics.builtin import _builtins
 
         self.__dict__.update(dict)  # update attributes
-        cls, name = dict["function_"]
+        class_name, name = dict["function_"]
 
-        self.function = getattr(builtins[cls], name)
+        self.function = getattr(_builtins[class_name], name)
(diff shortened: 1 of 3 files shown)


rocky:
(After this goes in there will be a small merge conflict with #635 but that should be easy to fix)

TiagoCavalcante (approved):
LGTM

mathics/core/rules.py
```diff
@@ -232,9 +232,9 @@ def __getstate__(self):
...
     def __setstate__(self, dict):
-        from mathics.builtin import builtins
+        from mathics.builtin import _builtins
 
         self.__dict__.update(dict)  # update attributes
-        cls, name = dict["function_"]
+        class_name, name = dict["function_"]
 
-        self.function = getattr(builtins[cls], name)
+        self.function = getattr(_builtins[class_name], name)
```
mmatera, line 240: I Just wonder, is it OK to keep the name ``_builtins`` for something that is not private to the module?

rocky: This is a temporary change. #635 or a stripped down version of that has the more complete fix.

mmatera: In that case, let's merge and iterate.

PR #639, opened 2022-11-21
patch

mmatera:
This is a patch that makes that the test passes. However, I think in means that the scheme proposed is not working as it should.

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/639"></a>


rocky:
> This is a patch that makes that the test passes. However, I think in means that the scheme proposed is not working as it should.
> 
> [](https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/639)

Right now I am using `copy.copy` to initialize the definitions `builtins` field.  We may need `copy.deepcopy` instead. And there are bugs in the code that prevent that. #636 fixes one of the bugs found. But there is another one out there.

mmatera:
> > This is a patch that makes that the test passes. However, I think in means that the scheme proposed is not working as it should.
> > [](https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/639)
> 
> Right now I am using `copy.copy` to initialize the definitions `builtins` field. We may need `copy.deepcopy` instead. And there are bugs in the code that prevent that. #636 fixes one of the bugs found in running `deepcopy`. But there is another one out there.

I added a comment in that direction.

mmatera:
@rocky,
I guess this is ready now: the problem was in mathics.builtin.base.Builtin.contribute:  when a pymathics module got loaded, system_definitions.builtin was overwriten. This caused that if one test (test/test_custom_boxexpression) contributes with rules to in System`MakeBoxes, the following tests can not get ride of the new definitions. Now it seems it is working properly.

mmatera:
Also, following  #642, some imports in `mathics.core`  from `mathics.builtin`  were avoided.

rocky:
> @rocky, I guess this is ready now: the problem was in mathics.builtin.base.Builtin.contribute: when a pymathics module got loaded, system_definitions.builtin was overwriten. This caused that if one test (test/test_custom_boxexpression) contributes with rules to in System`MakeBoxes, the following tests can not get ride of the new definitions. Now it seems it is working properly.

Thanks for isolating what's wrong and working around/fixing this.  I'd like to understand why when a pymathics module gets loaded system_definitons.builtin is overwritten.  That seems wrong, no? 

All that code with a `try` block to work around this seems like a workaround for a bug or misfeature somewhere else.

mmatera:
> Thanks for isolating what's wrong and working around/fixing this. I'd like to understand why when a pymathics module gets loaded system_definitons.builtin is overwritten. That seems wrong, no?
> 
The problem comes with MakeBoxes rules. When a pymathics module "contributes" with a MakeBoxes rule, before this patch,  a rule is added to  ``definitions.builtin["System`MakeBoxes"]``. Since we didn't make a deep copy of `definitions.builtin`, the added rule ends in ```system_definitions.builtin["System`MakeBoxes"]```.

With the workaround, if a pymathics module wants to add a rule to MakeBoxes, now it adds it to a definition in ```definitions.pymathics```. 



> All that code with a `try` block to work around this seems like a workaround for a bug or misfeature somewhere else.

The `try`  block is one way to handle the case in which pymathics does not have already a definition for ```System`MakeBoxes```. I could also have implemented with something as
```
if "System`MakeBoxes" not in definitions.pymathics:
   ....
```
But let's say that this is something "exceptional" (it happens at most once in the life of a `Definitions` object). In any case, I hope to get rid of all the related to the MakeBoxes special case as soon as possible.

rocky:
> The problem comes with MakeBoxes rules. When a pymathics module "contributes" with a MakeBoxes rule, before this patch, a rule is added to `` definitions.builtin["System`MakeBoxes"] ``. Since we didn't make a deep copy of `definitions.builtin`, the added rule ends in `` system_definitions.builtin["System`MakeBoxes"] ``.

Ok - so let see if we can get deepcopy to work or narrow the copy so something like this doesn't happen.

mmatera:
The question is, do we want wo modify Definitions.builtin?  If not, even the copy is not necessary. On the other hand if what bothers you is the try block, it can be easily replaced with a conditional structure.

rocky:
> The question is, do we want [t]o modify Definitions.builtin? If not, even the copy is not necessary. On the other hand if what bothers you is the try block, it can be easily replaced with a conditional structure.

What  bothers me is the layer upon layer of workaround, code and complexity because we aren't understanding the problem at the root cause or at a logical conceptual level. 

Let me look at this over the weekend so I understand what's going on at a higher level.

mmatera:
> > The question is, do we want [t]o modify Definitions.builtin? If not, even the copy is not necessary. On the other hand if what bothers you is the try block, it can be easily replaced with a conditional structure.
> 
> What bothers me is the layer upon layer of workaround, code and complexity because we aren't understanding the problem at the root cause or at a logical conceptual level.

At the higher level, what is wrong is how we process and store MakeBoxes rules.  

Notice also that to copy (or even deep copy) an object that should be immutable, to have several immutable copies of the same thing is already a workaround. 

I have a couple of ideas about how to fix the way in which we handle MakeBoxes, but that would imply a lot of other changes that would make this harder to review.

For this reason, the strategy that I am trying to propose since some time ago is to try to replace unlocalized workarounds (the ones that imply changing something in one module to fix something in another module) with workarounds localized inside a module, or even better, inside the function that should be fixed, which would be  easier to document and correct
afterwards.
 
> 
> Let me look at this over the weekend so I understand what's going on at a higher level.

Sure.

> 
> You've given a lot of details as to what's up at an operational level, and that he helpful for understanding what is wrong at that level.

I give many details because I think they are needed to understand what is actually going on.

mathics/builtin/base.py
```diff
@@ -384,13 +396,32 @@ def contextify_form_name(f):
...
         for rule in box_rules:
             makeboxes_def.add_rule(rule)
 
+        # If a pymathics module was loaded, then there are at least two
+        # definitions for MakeBoxes, the builtin, the pymathics, and the
+        # cached from combining the former two. Rules are added to one of
+        # both, which are not going to be in the cached version that
+        # Definitions.get_definition provides. To make the new rules
+        # accesible, we need to clear the definitions cache.
+        definitions.clear_cache("System`MakeBoxes")
```
rocky, line 423: If this is only needed when pymathics modules are loaded, then why isn't this only in the branch of the `if is_pymathics_module` branch?

mathics/builtin/base.py
```diff
@@ -191,7 +192,12 @@ def __new__(cls, *args, **kwargs):
         # mathics.builtin.inout
 
         if kwargs.get("expression", None) is not False:
-            return to_expression(cls.get_name(), *args)
+            try:
```
rocky, line 195: This was my bad  - should be removed.

mathics/builtin/base.py
mmatera: But then, I should duplicate lines 414 and 415 in both branches.

mmatera:
outdated

PR #644, opened 2022-11-26
leaves->elements & Makefile...

rocky:
* Correct one more incorrect use of leaves.
* In pytest be explicit about character encoding

The first change is split off from #639 .  
The second change is split off from #631. With that change some tests attempt to start using the unicode symbols, presumably because some MakeBox rules are no longer wiped out (or vice versa?).  Either way this should be there. 

<a href="https://gitpod.io/#https://github.com/Mathics3/mathics-core/pull/644"></a>


mmatera (approved):
LGTM

PR #648, opened 2022-12-03
Move `mathics.core.builtin` initialization code into `system_initialize()`

rocky:
@mmatera and @TiagoCavalcante : this is kind of important to understand. 

Although we now require a frontend to *explicitly* call ``initialize_system()`` and this may seem more awkward, it is helpful and important for a couple of reasons.

**First**: it provides better modularization and reduces circular imports.

``mathics.builtin``  is close to top-level import and, in contrast to ``mathics.core.system_init`` there are a lot of modules underneath ``mathics.builtin``. (There are none under ``mathics.core.system_init``).

_In general, it is considered bad Python coding practice to put a lot of code in **``__init__.py ``** files._

As a result, we can have more ``import mathics.core.system_init`` from a module where we can't do the same for ``import mathics.builtin``.

**Second**:  by placing initialization of structures which do not change outside of development we are in a better position to write an administrative tool to load all of this and then dump it as a system image file, which can be read in rather than built it up from scratch as happens now. This new function ``initialize_system`` might have a parameter to read the image file rather than do the work it does now.

In its current form though ``initialize_system()`` does as not encapsulate system loading as much as it could.

### DIFF
--- a/examples/symbolic_logic/gries_schneider/test_gs.py
+++ b/examples/symbolic_logic/gries_schneider/test_gs.py
@@ -2,10 +2,12 @@
 # -*- coding: utf-8 -*-
 
 
+from mathics.builtin.system_init import initialize_system
 from mathics.core.definitions import Definitions
 from mathics.core.evaluation import Evaluation
 from mathics.core.parser import MathicsSingleLineFeeder, parse
 
+initialize_system()
 definitions = Definitions(add_builtin=True)
 
 for i in range(0, 4):
--- a/examples/symbolic_logic/gries_schneider/test_gs.py
+++ b/examples/symbolic_logic/gries_schneider/test_gs.py
@@ -2,10 +2,10 @@
 # -*- coding: utf-8 -*-
 
 
-from mathics.builtin.system_init import initialize_system
 from mathics.core.definitions import Definitions
 from mathics.core.evaluation import Evaluation
 from mathics.core.parser import MathicsSingleLineFeeder, parse
+from mathics.core.system_init import initialize_system
 
 initialize_system()
 definitions = Definitions(add_builtin=True)
--- a/mathics/core/system_init.py
+++ b/mathics/core/system_init.py
@@ -3,15 +3,9 @@ One-time initialization of Mathics module loading, and creation of built-in func
 """
 
 import glob
+import inspect
 import os.path as osp
-
-from mathics.builtin.base import (
-    Builtin,
-    Operator,
-    PatternObject,
-    SympyObject,
-    mathics_to_python,
-)
+from typing import Optional
 
 from mathics.core.pattern import pattern_objects
 
@@ -37,6 +31,13 @@ system_builtins = {}
 
 
 def add_builtins(new_builtins):
+    from mathics.builtin.base import (
+        Operator,
+        PatternObject,
+        SympyObject,
+        mathics_to_python,
+    )
+
     for var_name, builtin in new_builtins:
         name = builtin.get_name()
         if hasattr(builtin, "python_equivalent"):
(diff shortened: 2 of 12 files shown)


mmatera:
@rocky,  In the proposal I made in #639, there is no need to modify the front-ends. What is the improvement here over that branch?

rocky:
> @rocky, In the proposal I made in #639, there is no need to modify the front-ends. What is the improvement here over that branch?
> 
> (in that branch, there is no need to call `_initialize_system_definitions()` from outside of the `mathics.builtin.system_init` module [mathics/builtin/system_init.py](https://github.com/Mathics3/mathics-core/pull/639/files#diff-a2d6a84f5787e277f90b6736be5b7132b788c868d619cb0c6dbdd50404b5b896) )

There is a principle [Explicit is better than Implicit](https://towardsdatascience.com/the-zen-of-python-a-guide-to-pythons-design-principles-93f3f76d088a).

In #639  (and master) magic initialization happens when the module is loaded. _Exactly_ when is hard to determine, since it relies on what module imports what where and in what order.  

The order of module loading for purposes of what gets initialize when is not something a programer a programmer generally keeps track of nor wants to. 

But since the intent here is to serve as a place for performing initialization a a base image, this is something that I believe we want developers to know about and be able to know very definitely when where this is triggered - the call to `system_initialize`.  Again, this is something that I believe developers should  aware of.

rocky:
BTW The main other problem I have with #639 is that tries to do too much and doesn't succeed in either activity as well as if we isolate the two concerns of reducing `mathics.builtin` (which does a bit of system initialization) and pulling out module loading from inside Definitions object creation. 

That second aspect we have come to learn is a bit thorny and needs more attention.  This PR (when done) lays a better foundation for  that to occur.

rocky:
@mmatera this is about all the time I have this week for this. I don't think I was completely successful here since I'd like to move to move out _all_ of the code from `mathics.builtin.__init__`  and some still remains. 

The "sanity check" which looks for that Builtins have a "summary_text" should be turned into a standalone program outside of the code, and either made a separate program or added to one of the other consistency checks. 

That there are this many changes and this many commits (with squashing down a number of commits into one) is  a testament to the messiness we still have in or code and the problems with adding features before cleaning up a base that is wanting. 

If you have are willing and have time to move this forward more, please do. Otherwise we can merge I will iterate next week.

examples/symbolic_logic/gries_schneider/test_gs.py
```diff
@@ -5,7 +5,9 @@
 from mathics.core.definitions import Definitions
 from mathics.core.evaluation import Evaluation
 from mathics.core.parser import MathicsSingleLineFeeder, parse
+from mathics.core.system_init import initialize_system
 
+initialize_system()
```
mmatera, line 10: This routine is called each time ```Definitions(add_builtins=True)``` is called. And what it essentially does is to initialize (a part of) the ```mathics.builtins``` module, so if you do not call it,  ```Definitions(add_builtins=True)``` can not do what it is expected. So, should not be better either
a) call it in ```mathics.buitlin.__init__``` module.
b) import it (locally) and call it  in ```mathics.core.definitions.Definitions.__init__```, if the `add_builtins==True`?

mathics/core/system_init.py
```diff
@@ -0,0 +1,184 @@
...
+    osp.basename(f[0:-3])
+    for f in glob.glob(osp.join(osp.dirname(__file__), "[a-z]*.py"))
+]
+
+builtins_precedence = {}
+
+system_builtins = {}
+
+
+def add_builtins(new_builtins):
```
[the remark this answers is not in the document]
rocky: What I have is a still a mess still. 

I'd like to remove `add_builtin` from `mathics/builtin/__init__.py` and the call at the end to of that module. When I try though I am getting errors in evaluation. So right now duplication of code is as close as I can get.

examples/symbolic_logic/gries_schneider/test_gs.py
rocky: I don't know how to resolve. I don't like a) because we want to _remove_ stuff from `mathics.builtin.__init__` - that is the whole _point_ of this. 

We want a function, `initialize_system()` that is explicitly called that will load will initialize all predefined Builtins. 

Once that is done, we are in a position to write a routine to dump out all of this to say a pickle file and then we have the possibility  to replace that code with code that reads in the dumped information instead. 

Equally important,  this is conceptually simpler, and coding-wise we don't have critical stuff or magic code inside the an  `__init__`  module.

PR #650, opened 2022-12-04
Compute Symbol.__hash__() once

rocky:
Also note *why* we need `Symbol.__hash__()`

A while back, but I forget where, the justification for why we using Strings in function instead of  Symbols in the code was because `is` or equality testing was slow.  This simple change addresses that. 

I forget where the benchmarking data was posted for previous comparisons. I'd be interested though in what the change (if any) is.

### DIFF
--- a/mathics/core/symbols.py
+++ b/mathics/core/symbols.py
@@ -1,9 +1,10 @@
 # cython: language_level=3
 # -*- coding: utf-8 -*-
 
-import sympy
 import time
 from typing import Any, FrozenSet, List, Optional, Tuple
+
+import sympy
 
 from mathics.core.element import (
     BaseElement,
@@ -359,13 +360,14 @@ class Symbol(Atom, NumericOperators, EvalMixin):
     """
 
     name: str
+    hash: str
     sympy_dummy: Any
     defined_symbols = {}
     class_head_name = "System`Symbol"
 
     # __new__ instead of __init__ is used here because we want
     # to return the same object for a given "name" value.
-    def __new__(cls, name, sympy_dummy=None, value=None):
+    def __new__(cls, name: str, sympy_dummy=None, value=None):
         """
         Allocate an object ensuring that for a given `name` we get back the same object.
         """
@@ -374,6 +376,14 @@ class Symbol(Atom, NumericOperators, EvalMixin):
         if self is None:
             self = super(Symbol, cls).__new__(cls)
             self.name = name
+
+            # Set a value for self.__hash__() once so that every time
+            # it is used this is fast.
+            # This tuple with "Symbol" is used to give a different hash
+            # than the hash that would be returned if just name were
+            # used.
+            self.hash = hash(("Symbol", name))
+
             # TODO: revise how we convert sympy.Dummy
             # symbols.
             #
@@ -412,8 +422,11 @@ class Symbol(Atom, NumericOperators, EvalMixin):
     def __getnewargs__(self):
         return (self.name, self.sympy_dummy)
 
-    def __hash__(self):
-        return hash(("Symbol", self.name))  # to distinguish from String
+    def __hash__(self) -> int:
+        """
+        We need self.__hash__() so that we can use Symbols as keys in dictionaries.
+        """
+        return self.hash
 
     def __ne__(self, other) -> bool:
         return self is not other


mmatera (approved):
LGTM

rocky:
> In any case, I am not sure if this makes the comparison faster: `is` compares memory locations, while `__equal__` compares the result of `get_sort_key()` method. In any case, go cal try to benchmark this.

In any event, let's please phase out the use of strings when symbols are meant.

mmatera:
Here we go with some basic benchmarks. The point here is  ```hash(key)``` is called each time a dictionary is accessed using the key ```key```. 

Now, let's compare the time required for computing ```hash``` for  a ```str```, for a ```tuple``` of two ```str```, and for a reduced implementation implementation of the old and new ```Symbol``` class:

```
class SymbolOld:
    def __init__(self, name):
        self.name = name

    def __hash__(self):
        return hash(tuple(("Symbol", self.name)))


class SymbolNew:
    def __init__(self, name):
        self.name = name
        self.hashval = hash(tuple(("Symbol", self.name)))
        
    def __hash__(self):
        return self.hashval
```


| `str` | `tuple` | `SymbolOld` | `SymbolNew`  |
|----      | ----    |  ----        | -----       |
| 21.3  |  43.4   |   87.9      |   51.9         |

(times are in ns)

Times were estimated using timeit in Pyston 2.3.4, called from the command line:

```
python -m timeit   -s 'from atestclass import SymbolNew; a=SymbolNew("System`ASymbolName")'  'hash(a)'
```

Notice that the time required for calling ```hash``` in the cached version (i.e., which does not have into account the time required for actual evaluation of the hash of a tuple)  requires 51.9ns, against the 21ns required to hash a ```str```.

rocky:
To be clear here . The difference we care about, or that is significant in overall execution between SymbolOld and SymbolNew  is operationally the difference between accessing a class variable (new) versus calling hash on a tuple of strings.

The object creation time isn't that important - it is a one-time thing.

rocky:
Access using the symbol object, not then name.

rocky:
I spent a little time trying benchmarking the difference. What I am seeing is not much difference between either old, new, or using a string  in pyston. In 3.10.8, I can see a little difference but not much to warrant perverting and complicating the code. And as soon as a conversion is used to get a string from a Symbol, the negligible performance "benefit" is lost. 

A big problem I notice with the code above is that it defines  `__init__()`  and that is wrong - it should use `__new__()`. 

```
$ pyenv local pyston-2.3.5
$ python s2-timing.py 
'func_old'  0.36 ms
'func_new'  0.34 ms
'func_str'  0.32 ms
$ python s2-timing.py 
'func_old'  0.36 ms
'func_new'  0.35 ms
'func_str'  0.32 ms
$pyenv local python-3.10.8
$ python s2-timing.py 
'func_old'  1.61 ms
'func_new'  1.44 ms
'func_str'  1.06 ms
$ python s2-timing.py 
'func_old'  1.82 ms
'func_new'  1.19 ms
'func_str'  1.03 ms
```

Here is the code for s2-timing.py

```python
from mathics.timing import timeit

class SymbolOld:
    def __new__(self, name):
        self.name = name

    def __hash__(self):
        return hash(tuple(("Symbol", self.name)))


class SymbolNew:
    def __new__(self, name):
        self.name = name
        self.hashval = hash(tuple(("Symbol", self.name)))

    def __hash__(self):
        return self.hashval

old = SymbolOld("old")
new = SymbolOld("new")

d = {old: 1, new: 2, "string": 3}

ITERATIONS=10000
@timeit
def func_old():
    x = 0
    for _ in range(ITERATIONS):
        x |= d[old]

@timeit
def func_new():
    x = 0
    for i in range(ITERATIONS):
        x |= d[new]

@timeit
def func_str():
    x = 0
    for i in range(ITERATIONS):
        x |= d["string"]

func_old()
func_new()
func_str()
```

Trying another way in pyston 2.3.5 (output is old, new, str): 

```$ python symbol-timing.py
0.0005529960617423058
0.0005363970994949341
0.0006319154053926468
$ python symbol-timing.py
0.0005638459697365761
0.0005182567983865738
0.0005195159465074539
$ python symbol-timing.py
0.0005540568381547928
0.0005348557606339455
0.0004916153848171234
```

and the code for this: 

```
import timeit
class SymbolOld:
    def __new__(self, name):
        self.name = name

    def __hash__(self):
        return hash(tuple(("Symbol", self.name)))


class SymbolNew:
    def __new__(self, name):
        self.name = name
        self.hashval = hash(tuple(("Symbol", self.name)))

    def __hash__(self):
        return self.hashval

old = SymbolOld("old")
new = SymbolOld("new")

d = {old: 1, new: 2, "string": 3}

def test(key): d[key]

ITERATIONS=10000
print(timeit.timeit("test(old)", setup="from __main__ import test, old", number=ITERATIONS))
print(timeit.timeit("test(new)", setup="from __main__ import test, new", number=ITERATIONS))
print(timeit.timeit("test('string')", setup="from __main__ import test", number=ITERATIONS))
```

PR #654, opened 2022-12-09
Fix ENABLE_FILES_MODULE

### LINKED ISSUE #653: ENABLE_FILES_MODULE flag seems to be ignored
## Description

Running `mathicsserver` with `ENABLE_FILES_MODULE` set to `False` inside of `mathics/settings.py` seems to not disable the files module properly (for me at least).

## Expected behavior

Commands like
```
CreateFile["test"]
```
shouldn't work when `ENABLE_FILES_MODULE` is false, but instead the file `test` is created anyway.

## Your Environment

The modules involved are:
```
Mathics-Django==5.0.0
Mathics3==5.0.2
```
from either pip or installed manually with `python3 setup.py install` (plus `Mathics-Scanner==1.2.4` but that was installed automatically).

I tried this on Manjaro and inside an Ubuntu Kinetic Docker container, both with Python 3.10.

## Workarounds

Apparently when `subdir` inside `mathics/builtin/__init__.py` is `files_io`, `import_name` is `mathics.builtin.files_io`.

Because this won't match any of the strings inside `disable_file_module_names`, the following code won't prevent the module from being loaded:
```py
if import_name in disable_file_module_names:
        continue
```

I guess this could be fixed by replacing the if statement with:
```py
if subdir in disable_file_module_names:
        continue
```
and setting:
```py
disable_file_module_names = (
    [] if ENABLE_FILES_MODULE else ["files_io"]
)
```
since you probably wouldn't want other functions like `CopyDirectory` either.

alek3y:
This should fix #653.

Disabling the entire `files_io` module wasn't feasible because `mainloop.py` contains definitions like `$Line` which are needed by the REPL.

However, loading individual modules like `mainloop` while excluding others would crash `mathicsserver`, because (from what I understand) it [automatically searches for the children modules](https://github.com/Mathics3/mathics-django/blob/0fd3ffd9f9693ff45628b1f105f797f10383afce/mathics_django/doc/django_doc.py#L326-L330) and [tries to add them](https://github.com/Mathics3/mathics-django/blob/0fd3ffd9f9693ff45628b1f105f797f10383afce/mathics_django/doc/django_doc.py#L349) to the documentation even if they're not loaded.

My solution (which avoids the `mathicsserver` crash) was to not load the parent (like `files_io`) but to still load the required children (like `mainloop`).

### DIFF
--- a/mathics/builtin/__init__.py
+++ b/mathics/builtin/__init__.py
@@ -94,7 +94,7 @@ def contribute(definitions):
             definitions.builtin[op] = Definition(name=op)
 
 
-def import_builtins(module_names: List[str], submodule_name=None) -> None:
+def import_builtins(module_names: List[str], submodule_name=None, import_parent=True) -> None:
     """
     Imports the list of Mathics Built-in modules so that inside
     Mathics we have these Builtin Functions, like Plus[], List[] are defined.
@@ -113,7 +113,7 @@ def import_builtins(module_names: List[str], submodule_name=None) -> None:
         if module:
             modules.append(module)
 
-    if submodule_name:
+    if submodule_name and import_parent:
         import_module(submodule_name, f"mathics.builtin.{submodule_name}")
 
     for module_name in module_names:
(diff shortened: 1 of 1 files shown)


rocky:
I don't want to make the code more complicated only to rip it out later.  I will look at this over the weekend and let see if we can come up with a cleaner fix.

alek3y:
I guess this draft can be closed now as it was fixed in #655.

PR #655, opened 2022-12-11
Respect ENABLE_FILES_MODULE environment variable

### LINKED ISSUE #653: ENABLE_FILES_MODULE flag seems to be ignored

rocky:
Basically we split out the "mainloop" module from `files_io` and add it to builtins. This way, everything in the `files_io` module can be ignored when desired.

This basically follows the suggestion by alek3y in https://github.com/Mathics3/mathics-core/issues/653

Fixes #653


mmatera (approved):
LGTM

PR #657, opened 2022-12-14
Add discrete plot

rocky:


mathics/eval/plot.py
```diff
@@ -120,6 +124,149 @@ def quiet_f(*args):
...
+                plot_points = [
+                    [[float(i + 1), l] for i, l in enumerate(line)]
+                    for line in plot_points
+                ]
+            else:
+                return
+        else:
+            return
+    else:
+        return
```
TiagoCavalcante, line 167: Maybe use early returns here?

mathics/builtin/drawing/discrete_plot.py
```diff
@@ -0,0 +1,179 @@
...
+class _DiscretePlot(Plot):
+
+    attributes = A_HOLD_ALL | A_PROTECTED
+
+    expect_list = False
+
+    messages = {
+        "prng": (
+            "Value of option PlotRange -> `1` is not All, Automatic or "
+            "an appropriate list of range specifications."
```
TiagoCavalcante, line 37: This is a bit confusing, but I don't know how to reformulate it

mathics/builtin/drawing/discrete_plot.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def eval(self, function, x, start, nmax, step, evaluation, options):
+        """DiscretePlot[function_, {x_Symbol, start_Integer, nmax_Integer, step_Integer},
+        OptionsPattern[DiscretePlot]]"""
+        if isinstance(function, Symbol) and function.name is not x.get_name():
+            rules = evaluation.definitions.get_ownvalues(function.name)
+            for rule in rules:
+                functions = rule.apply(function, evaluation, fully=True)
+
+        if function.get_head_name() == "List":
```
TiagoCavalcante, line 70: Can't this be converted to a get_head() == SymbolList?

TiagoCavalcante (approved):
LGTM

mathics/builtin/drawing/discrete_plot.py
```diff
@@ -0,0 +1,179 @@
...
+
+        base_plot_points = []  # list of points in base subdivision
+        plot_points = []  # list of all plotted points
+
+        compiled_fn = compile_quiet_function(
+            function[0], [x_name], evaluation, self.expect_list
+        )
+
+        @lru_cache
+        def apply_fn(fn, x_value: int) -> Optional[float]:
```
rocky, line 125: One of the things I find amusing is that in the old code this is called "eval_fn". So not only did it sometimes use "apply" when "eval" was meant, but it also used "eval" when "apply" was meant.

mathics/builtin/drawing/discrete_plot.py
rocky: What aspect is confusing. Describing the problem well sometimes is harder than solving it once the problem is pinpointed.

mathics/builtin/drawing/discrete_plot.py
rocky: Good catch. Even better is `function.head is SymbolList`. Fixed now (and in the code where this was copied from).

It is amazing how much rot we find in the code when a new section is looked at.

mathics/eval/plot.py
```diff
@@ -1,13 +1,16 @@
 """
 Evaluation routines for 2D plotting.
+
+Note this is distinct from formatting/rendering, e.g. to SVG..
```
mmatera, line 4: Maybe "These routines build Expressions that describe plots as Graphics in WL. Note this is distinct from formatting/rendering e.g. to SVG."

rocky: Ok - I had changed this in the interim (also thinking it might be deficient) but what you write is also good, so I will change it to this.

rocky: Changed, but I'd be interested in the parts that were deficient. I had in the interim that another pass is made for formatting. Should that be stated? And I think I added in the interim M-Expressions instead of Expressions. With that we can remove the WL part because it is not clear that we are 100% WL (we always be 100% Mathics though.)

mathics/eval/plot.py
rocky: Good point. I did one level. Doing other levels is trickier.

mmatera:
Something that is not working not (and I think is did) is plotting more than a single function. For example,

```
DiscretePlot[{Sin[Pi x/20],Cos[Pi x/20]},{x,1,40}]
```
Should draw points for the two functions with different colors:
imagen


For some reason, this line raises an unhandled exception in Mathics-django:

```
Exception raised: 'tuple' object does not support item assignment Traceback (most recent call last):  File "/home/mauricio/Projects/mathics-django/mathics_django/web/views.py", line 588, in ...
```

rocky:
I think this is straightforward to address. I reduced the setup code in DiscretePlot over that in Plot, not realizing that it too handled multiple fns. 

What probably is needed though is to  DRY the duplicate setup code by creating common routine for the two uses. 

The tuple vs list thing is also easily fixed when I get a chance to address. Or feel free to commit onto this. 

LogPlot, and ListStepPlot are also easily added.

mathics/eval/plot.py
mmatera: With WL I mean WL M-Expressions. However, it is true that in WMA the form of the resulting expression could be a little different from what we get in Mathics. 
Regarding the whole process, I guess that this deserves an entry in the documentation, and then we can add a link here to explain how a Plot expression is evaluated to a Graphics Expression, and then formatter to a Boxed Expression, to finally be rendered to a (for example) an SVG encoded picture.

mmatera:
> I think this is straightforward to address. I reduced the setup code in DiscretePlot over that in Plot, not realizing that it too handled multiple fns.
> 
> What probably is needed though is to DRY the duplicate setup code by creating common routine for the two uses.
> 
> The tuple vs list thing is also easily fixed when I get a chance to address. Or feel free to commit onto this.
> 
> LogPlot, and ListStepPlot are also easily added.

According to the exception message, the problem seems to be that at some point we are trying to assign to an element of a set, that should be a list. I can try to handle it later.

mathics/eval/plot.py
rocky: I agree the developer docs need going over.  In the meanwhile have adjusted the comment yet again go put back some of the stuff I mentioned in parallel with your comment.

rocky:
@mmatera DiscretePlot no longer crashes. It doesn't handle multiple functions yet, but it is closer, and it uses a common routine shared with Plot. 

In the process, ListPlot origins are now more often (0, 0). 

This is all the time I have for. If you can get DiscretePlot handling multiple plots, that would be great. Some work on detecting the form and setting hue colors still needs doing. 

And I've run across miscellaneous other bugs. But I am more inclined to get it good enough for now and come back to a lot of the cleanup that is needed: bug fixing and documenting interfaces, changes to the developer guide, etc.

It is amazing that even for something that I had hoped would be small uncovers so much stuff. (The compiled function evaluation that is used here is another mess for later. Being able to handle large lists via list literals is later as well.)

mathics/builtin/drawing/discrete_plot.py
TiagoCavalcante: What is "range specifications"? "Specifications" isn't used as much as "options" in other places, so maybe "range options"?

rocky:
> Something that is not working not (and I think is did) is plotting more than a single function. For example,
> 
> ```
> DiscretePlot[{Sin[Pi x/20],Cos[Pi x/20]},{x,1,40}]
> ```
> 
> Should draw points for the two functions with different colors: imagen
> 
> For some reason, this line raises an unhandled exception in Mathics-django:
> 
> ```
> Exception raised: 'tuple' object does not support item assignment Traceback (most recent call last):  File "/home/mauricio/Projects/mathics-django/mathics_django/web/views.py", line 588, in ...
> ```
> 
> The same happens with
> 
> ```
> Plot[{Sin[Pi x/20],Cos[Pi x/20]},{x,1,40}]
> ```
> 
> which is working as expected in the master branch.

I think all is fixed now.  The code changes here are getting rather largish. We should consider merging and iterating.

mmatera:
Sure, go ahead.

mathics/builtin/drawing/discrete_plot.py
rocky: I just had a chance to look at this. The wording of this message comes from WMA (it allows more options than we do, but it ends in "or an appropriate list of range specifications".   

Given that this is the way WMA words this, we shouldn't try to change the wording.

PR #658, opened 2022-12-15
Move eval function from _Plot[] to mathics.eval...

rocky:
Also clean up mathics.drawing.plot module.

My initial motivation was to add DiscretePlot[] (and some of that is in another unfinished branch). 

As too often happens when I want to do something, there is so much existing rot that I find that I first need to do a cleanup step. And even that can be huge. 

So this is a separate PR.


rocky:
Superceded by #659

PR #662, opened 2022-12-17
Compute hash inside __new__ ...

rocky:
rather than do that in `__hash__()` each time that `__hash()__` is called. 

Also: 

* run isort on changed modules.  
* make a small summary text for regularization: should start with active verb
* Remove duplication of Integer and Rational class inside `mathics.builtin.atoms`
* Complex atoms now store a tuple of the real and imaginary parts when the complex does not degenerate to a Real.

In doing this I noticed that we weren't always creating unique objects for things like Integer even though the objects were functionally equivalent. So it was possible some Python `is` testing was returning `False` when it should have returned `True` failing.   

As a result, our code may be using `==` in places where it should be possible to use `is` were it not for this misfeature.
And this means `@lru_cache` decorator is now superfluous on `Integer.__init__()`. 

Aside: 
LRU cache provides a temptation for use in that it allows putting a limit on the caching to some value; that is the LRU-ness. However we have to create _all_ objects anyway, and except in somewhat pathological situations, these objects don't disappear. 

_I haven't checked, but we should be using less memory overall with this change. PR_

Finally, I've revised `test/core/test_atoms.py` to be more rigorous, understandable, and use pytest.


mmatera:
Does it makes sense to pre compute the hash for numeric atoms? I guess that __hash__ is not called so often in these classes, and it could produce an unnecessary overhead.

rocky:
> Does it makes sense to pre compute the hash for numeric atoms? 

Let's leave that for later when considering how to make a preloaded image that contains the static objects.

> I guess that **hash** is not called so often in these classes, and it could produce an unnecessary overhead.

The overhead is repaid on the first use of a use; the "overhead" is not that large anyway. 

Here is an experiment  (I have tried by accident). Try removing `__hash__` from one of these "little-used" and see if you can run the tests?

rocky:
https://github.com/Mathics3/mathics-core/commit/e33affb3ab19649566f6490a1d0becc09eb7ab0d has benchark timings since the last run. 

Overall things have sped up a little using Python 2.3.5. The line below that uses this branch. 

Load timings are: 

Version                                    | timing    
 ---------------------------------------- |  ---------------
old code and  pyston 2 3.4  | 361.94 ms 
master and     pyston 2.3.5| 366.05 ms 
 this branch & pyston 2.35  | 369.06 ms

3 ms is not worrying about.

mmatera:
The problem is not the start up time, but the overhead in building large arrays of numbers.

rocky:
> The problem is not the start up time, but the overhead in building large arrays of numbers.

Ok - so let's see a benchmark and code that shows this clearly. 

I suspect though that has nothing to do with this.

rocky:
There is this old music joke about how violinists spend half their time tuning and the other half playing out of tune. 

It too often feels like something like this is going on here. We spend a lot of time in tuning code that is way lacking in performance because none of the big stuff has been addressed.

mmatera:
Ok, but this is the opposite: to precompute and store hashes in numbers seems to be added in the sake of the performance. And yes, I agree about to decide what is the best we need benchmarks.

rocky:
> Ok, but this is the opposite: to precompute and store hashes in numbers seems to be added in the sake of the performance. And yes, I agree about to decide what is the best we need benchmarks.

Of course I realize this.  But when there is this thing where we can do something once rather than an unbounded number of times (with some constant startup time) and it is easy to do, just do it. 

I was hoping this would have slipped in without the discussion it has already caused and we'd move on.

mmatera:
OK, here I go with my test:
```
Timing[a=Table[i,{i,1,2,.0001}];]
```
Using this branch, (in pyston) I got   
```{0.785044, Null}```


In master, 
```
{1.21356, Null}
```

so, this is a success!

```

mathics/core/expression.py
```diff
@@ -270,6 +270,12 @@ def _build_elements_properties(self):
         values = []
         for element in self._elements:
             # Test for the literalness, and the three properties mentioned above
+            try:
```
mmatera, line 273: Is this something that is going to stay here?

mathics/core/symbols.py
```diff
@@ -410,7 +410,7 @@ def __new__(cls, name: str, sympy_dummy=None, value=None):
             # self.sympy_dummy, for which I have to dig into the
             # code to see even what type of value should be expected
             # for it.
-            self.value = value
+            self._value = value
```
mmatera, line 413: Except for some immutable symbols, it does not make sense to attach a value to a Symbol. Values of Symbols are always obtained from a Definitions object.

rocky: There is confusion here. The symbol value is its name. It is not the value that the symbolic variable holds. That should be in the definitions object.

mathics/builtin/numbers/randomnumbers.py
```diff
@@ -145,9 +143,11 @@ def randchoice(self, n, size, replace, p):
 
 class RandomState(Builtin):
     """
+    <url>:WMA: https://reference.wolfram.com/language/ref/RandomState.html</url>)
```
mmatera, line 146: Should not be a colon before the closing URL tag?

rocky: There is a stray ")" at the end. That should be fixed now.

mathics/core/expression.py
rocky: Thanks for catching!  removed now.

mathics/core/list.py
```diff
@@ -55,6 +55,12 @@ def __init__(
             self._is_literal = True
             values = []
             for element in elements:
+                try:
```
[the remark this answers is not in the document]
rocky: And thanks again!  Removed also.

mathics/core/symbols.py
mmatera: So, it is the same than it's name?

mathics/core/atoms.py
```diff
@@ -387,29 +326,71 @@ class MachineReal(Real):
     Stored internally as a python float.
     """
 
+    # Collection of MachineReals defined so far.
```
mmatera, line 329: Suppose you evaluate something like this:
```
Min[Eigenvalues[Table[Random[], {10^6},{10^6}]]]
```
Then, after that, we are storing 10^12 useless random numbers, plus the corresponding hashes. 
(Usually in my work I have to compute things like this, but with a more complicated function instead of Random[], but the idea is the same....

rocky: There may be some confusion here as well.  The only things that are stored in that dictionary are Python objects that were created anyway. And except in somewhat pathological situations, Python doesn't garbage collect these objects.

So in sum this is a problem we have independent of what we have here. If you run that code twice in the current system if it allocates 10^12 objects, it will allocate 10^12 _more_ the next time it is run without garbage collectiong the first 10^12. 

So what is needed independent of this is a better way to represent list literals and ensuring not to create Mathics objects like Integer except on demand.

mmatera: So, do you say that in master, the numbers in the array are not collected after the evaluation?

rocky: Yes, I believe numbers are not generally garbage collected by Python. Prove me wrong though!

I believe that with this branch our memory consumption may go down. 

However, this situation even with this PR is far from ideal.  I know that there are some situations where we don't have to allocate integers. I recall making a change to the `Range()` class and I recall that was a noticeable win. 

And I think we do something like this in Plotting. But in general we need to regularize this kind of thing and do more of it.

mmatera: Good, that is something that I can check. In any case, to remove the singleton pattern in numbers should not be too hard.

mathics/core/symbols.py
rocky: I just looked at the code, and this is confusing, and could use a bit of cleaning up, now that I understand things better. 

I started doing this, but I think it best to do this as a new single isolated PR. 

But let me try to explain first...

In short `value` is the same as its `name` (with _possibly_ one exception) and the two fields should have the same value in the future. 

In contrast to the comments in `Symbol.__init__()`  I believe this should work like Lisp's Symbol or Python's `system.intern()` .

And I think that the pattern used in the PR for numbers like Integer, Complex, and Real should be followed more closely in `Symbol.__init__()`.   Here, I think I was bootstrapping code from one onto the other. But in the revision, here are places I have improved things in my view. So that should be backported to  `Symbol.__init()`.

So now let's compare and contrast say ``Integer(0)`` from ``Symbol("System`N")``.  The part that is similar is that we want singletons based on a parameter for these. For Integer,  a parameter is 0. for Symbol, a parameter is ``System`N``. But for either of these, there should just be one object. 

For numbers, the value is constant. We never need to go through a Definitions object for that.  For Symbols, for most of them, values can change and so we have to go through a Definitions object. 

But there _are_ a couple of Symbols like True, False, List which we want to treat as constant - a user is not allowed to redefine these. And there is therefore no need to consult Definitions for these in order to get their values. 

So a hack I added was for this add True and False was to add `value` parameter which served in place of the name ``System`True``. 

So this should get cleaned up and a lot of comments and docstrings in the existing code simplified.  As I said, I think I understand better what's up here, based on @mmatera 's remark that Definition entries acted like bindings in other interpreters.

It might be that PredefinedSymbol's init looks more like Integer and that value parameter removed from `Symbol`'s init and `name` and `value` made identical. 

We'll see.

rocky:
> OK, here I go with my test:
> 
> ```
> Timing[a=Table[i,{i,1,2,.0001}];]
> ```
> 
> Using this branch, (in pyston) I got `{0.785044, Null}`
> 
> In master,
> 
> ```
> {1.21356, Null}
> ```
> 
> so, this is a success!

Thanks for checking!

mathics/core/atoms.py
mmatera: Here is the check. I hacked this branch to see what are the references for each element in the MachineReal instances dict:

https://github.com/Mathics3/mathics-core/tree/debug_MachineReal_references

Then I run this:

```
mathics -e 'Sqrt[124/743.]' -e 'Print["---"]'   -e '1.'
```

This is the last (and relevant) part of the output:

```

list of entries and their references for the dict (<class 'dict'>, 140479948858880)
0.0:              ['MachineReals._machine_reals', "dict: {'expr', 'atom'}", "dict: {'replace', 'system', 'pattern'}", "dict: {'expr', 'atom'}", "dict: {'expr', 'atom'}", "dict: {'expr', 'atom'}", "dict: {'expr', 'atom'}", "dict: {'expr', 'atom'}", "dict: {'expr', 'atom'}", 'class / module', 'class / module', 'class / module: \nRepresentation of Numbers\n\nIntegers and Real numb', 'tuple: (<MachineReal: 0.0>,)', 'tuple: (<MachineReal: 0.0>,)', 'tuple: (<MachineReal: 0.0>,)', 'tuple: (<MachineReal: 0.0>,)', 'tuple: (<Symbol: System`z>, <MachineReal: 0.0>)', 'tuple: (<MachineReal: 0.0>,)', 'tuple: (<MachineReal: 0.0>,)']
-3.0:             ['MachineReals._machine_reals', "dict: {'original', 'hash', '_value'}", "dict: {'replace', 'system', 'pattern'}", "class / module: \n    <dl>\n      <dt>'$SystemTimeZone'\n      <dd> g", "class / module: \n    <dl>\n      <dt>'$TimeZone'\n      <dd> gives t"]
0.6:              ['MachineReals._machine_reals', 'class / module: \nEvaluation routines for 2D plotting.\n\nThese routi', 'tuple: (<MachineReal: 0.6>, <MachineReal: 0.4>, <MachineReal: 0.2>)']
0.2:              ['MachineReals._machine_reals', 'class / module: \nEvaluation routines for 2D plotting.\n\nThese routi', 'tuple: (<MachineReal: 0.6>, <MachineReal: 0.4>, <MachineReal: 0.2>)', 'tuple: (<Symbol: System`i>, <Integer: 2>, <MachineReal: 0.2>)', 'tuple: (<Symbol: System`i>, <Symbol: System`r>, <MachineReal: 0.2>)']
-10.0:            ['MachineReals._machine_reals']
-1.0:             ['MachineReals._machine_reals']
0.5:              ['MachineReals._machine_reals', 'tuple: (<MachineReal: 0.5>,)', 'tuple: (<MachineReal: 0.5>,)', 'tuple: (<Integer: 1>, <MachineReal: 0.5>, <Integer: 0>)', 'tuple: (<MachineReal: 1.0>, <MachineReal: 0.5>, <MachineReal: 0.5>)', 'tuple: (<MachineReal: 0.5>, <Integer: 0>, <MachineReal: 0.5>)', 'tuple: (<Symbol: System`n>, <MachineReal: 0.5>)', 'tuple: (<Symbol: System`n>, <MachineReal: 0.5>)', 'tuple: (<Symbol: System`n>, <MachineReal: 0.5>)', 'tuple: (<Symbol: System`n>, <MachineReal: 0.5>)']
0.4:              ['MachineReals._machine_reals', 'tuple: (<MachineReal: 0.6>, <MachineReal: 0.4>, <MachineReal: 0.2>)', 'tuple: (<Integer: 1>, <Integer: 1>, <MachineReal: 0.4>)']
0.87:             ['MachineReals._machine_reals', 'tuple: (<MachineReal: 0.87>, <MachineReal: 0.94>, <Integer: 1>)']
0.94:             ['MachineReals._machine_reals', 'tuple: (<MachineReal: 0.87>, <MachineReal: 0.94>, <Integer: 1>)']
0.85:             ['MachineReals._machine_reals', 'tuple: (<Symbol: System`Brown>, <MachineReal: 0.85>)', 'tuple: (<Symbol: System`Pink>, <MachineReal: 0.85>)', 'tuple: (<Symbol: System`Red>, <MachineReal: 0.85>)']
0.9:              ['MachineReals._machine_reals', 'tuple: (<Symbol: System`Cyan>, <MachineReal: 0.9>)', 'tuple: (<Integer: 1>, <MachineReal: 0.9>, <MachineReal: 0.8>)']
0.88:             ['MachineReals._machine_reals', 'tuple: (<Symbol: System`Green>, <MachineReal: 0.88>)', 'tuple: (<Symbol: System`Purple>, <MachineReal: 0.88>)']
0.8:              ['MachineReals._machine_reals', 'tuple: (<Integer: 1>, <MachineReal: 0.9>, <MachineReal: 0.8>)']
1.0:              ['MachineReals._machine_reals', 'tuple: (<MachineReal: 1.0>, <MachineReal: 0.5>, <MachineReal: 0.5>)']
1.3:              ['MachineReals._machine_reals', 'tuple: (<MachineReal: 1.3>, <MachineReal: -2.4>, <Integer: 2>)']
-2.4:             ['MachineReals._machine_reals', 'tuple: (<MachineReal: 1.3>, <MachineReal: -2.4>, <Integer: 2>)']
2.0:              ['MachineReals._machine_reals']
743.0:            ['MachineReals._machine_reals', 'tuple: (<MachineReal: 743.0>, <Integer: -1>)', 'tuple: (<MachineReal: 743.0>, <Integer: -1>)']
0.0013458950201884253: ["(<class 'functools._lru_cache_wrapper'>, 140479778063360)", 'MachineReals._machine_reals']
0.16689098250336473: ["(<class 'functools._lru_cache_wrapper'>, 140479778063360)", "(<class 'list'>, 140479632004992)", 'MachineReals._machine_reals', "dict: {'System`x'}", "dict: {'System`y', 'System`x'}", "dict: {'x', 'y'}", "dict: {'evaluation', 'y', 'x'}", 'tuple: (<MachineReal: 0.16689098250336473>, <Expression: System`Times[1, System`Power[2, -1]]>)', 'tuple: (<MachineReal: 0.16689098250336473>, <Rational: 1/2>)', 'tuple: (<MachineReal: 0.16689098250336473>,)', 'tuple: (<MachineReal: 0.16689098250336473>, <Rational: 1/2>)', 'tuple: (<MachineReal: 0.16689098250336473>, <Rational: 1/2>)']
0.40852292775726157: ['MachineReals._machine_reals']
0.408523

---
None

list of entries and their references for the dict (<class 'dict'>, 140479948858880)
0.0:              ['MachineReals._machine_reals', "dict: {'expr', 'atom'}", "dict: {'replace', 'system', 'pattern'}", "dict: {'expr', 'atom'}", "dict: {'expr', 'atom'}", "dict: {'expr', 'atom'}", "dict: {'expr', 'atom'}", "dict: {'expr', 'atom'}", "dict: {'expr', 'atom'}", 'class / module', 'class / module', 'class / module: \nRepresentation of Numbers\n\nIntegers and Real numb', 'tuple: (<MachineReal: 0.0>,)', 'tuple: (<MachineReal: 0.0>,)', 'tuple: (<MachineReal: 0.0>,)', 'tuple: (<MachineReal: 0.0>,)', 'tuple: (<Symbol: System`z>, <MachineReal: 0.0>)', 'tuple: (<MachineReal: 0.0>,)', 'tuple: (<MachineReal: 0.0>,)']
-3.0:             ['MachineReals._machine_reals', "dict: {'original', 'hash', '_value'}", "dict: {'replace', 'system', 'pattern'}", "class / module: \n    <dl>\n      <dt>'$SystemTimeZone'\n      <dd> g", "class / module: \n    <dl>\n      <dt>'$TimeZone'\n      <dd> gives t"]
0.6:              ['MachineReals._machine_reals', 'class / module: \nEvaluation routines for 2D plotting.\n\nThese routi', 'tuple: (<MachineReal: 0.6>, <MachineReal: 0.4>, <MachineReal: 0.2>)']
0.2:              ['MachineReals._machine_reals', 'class / module: \nEvaluation routines for 2D plotting.\n\nThese routi', 'tuple: (<MachineReal: 0.6>, <MachineReal: 0.4>, <MachineReal: 0.2>)', 'tuple: (<Symbol: System`i>, <Integer: 2>, <MachineReal: 0.2>)', 'tuple: (<Symbol: System`i>, <Symbol: System`r>, <MachineReal: 0.2>)']
-10.0:            ['MachineReals._machine_reals']
-1.0:             ['MachineReals._machine_reals']
0.5:              ['MachineReals._machine_reals', 'tuple: (<MachineReal: 0.5>,)', 'tuple: (<MachineReal: 0.5>,)', 'tuple: (<Integer: 1>, <MachineReal: 0.5>, <Integer: 0>)', 'tuple: (<MachineReal: 1.0>, <MachineReal: 0.5>, <MachineReal: 0.5>)', 'tuple: (<MachineReal: 0.5>, <Integer: 0>, <MachineReal: 0.5>)', 'tuple: (<Symbol: System`n>, <MachineReal: 0.5>)', 'tuple: (<Symbol: System`n>, <MachineReal: 0.5>)', 'tuple: (<Symbol: System`n>, <MachineReal: 0.5>)', 'tuple: (<Symbol: System`n>, <MachineReal: 0.5>)']
0.4:              ['MachineReals._machine_reals', 'tuple: (<MachineReal: 0.6>, <MachineReal: 0.4>, <MachineReal: 0.2>)', 'tuple: (<Integer: 1>, <Integer: 1>, <MachineReal: 0.4>)']
0.87:             ['MachineReals._machine_reals', 'tuple: (<MachineReal: 0.87>, <MachineReal: 0.94>, <Integer: 1>)']
0.94:             ['MachineReals._machine_reals', 'tuple: (<MachineReal: 0.87>, <MachineReal: 0.94>, <Integer: 1>)']
0.85:             ['MachineReals._machine_reals', 'tuple: (<Symbol: System`Brown>, <MachineReal: 0.85>)', 'tuple: (<Symbol: System`Pink>, <MachineReal: 0.85>)', 'tuple: (<Symbol: System`Red>, <MachineReal: 0.85>)']
0.9:              ['MachineReals._machine_reals', 'tuple: (<Symbol: System`Cyan>, <MachineReal: 0.9>)', 'tuple: (<Integer: 1>, <MachineReal: 0.9>, <MachineReal: 0.8>)']
0.88:             ['MachineReals._machine_reals', 'tuple: (<Symbol: System`Green>, <MachineReal: 0.88>)', 'tuple: (<Symbol: System`Purple>, <MachineReal: 0.88>)']
0.8:              ['MachineReals._machine_reals', 'tuple: (<Integer: 1>, <MachineReal: 0.9>, <MachineReal: 0.8>)']
1.0:              ['MachineReals._machine_reals', 'tuple: (<MachineReal: 1.0>, <MachineReal: 0.5>, <MachineReal: 0.5>)']
1.3:              ['MachineReals._machine_reals', 'tuple: (<MachineReal: 1.3>, <MachineReal: -2.4>, <Integer: 2>)']
-2.4:             ['MachineReals._machine_reals', 'tuple: (<MachineReal: 1.3>, <MachineReal: -2.4>, <Integer: 2>)']
2.0:              ['MachineReals._machine_reals']
743.0:            ['MachineReals._machine_reals', 'tuple: (<MachineReal: 743.0>, <Integer: -1>)', 'tuple: (<MachineReal: 743.0>, <Integer: -1>)']
0.0013458950201884253: ["(<class 'functools._lru_cache_wrapper'>, 140479778063360)", 'MachineReals._machine_reals']
0.16689098250336473: ["(<class 'functools._lru_cache_wrapper'>, 140479778063360)", 'MachineReals._machine_reals', 'tuple: (<MachineReal: 0.16689098250336473>, <Rational: 1/2>)']
0.40852292775726157: ["(<class 'functools._lru_cache_wrapper'>, 140479778063360)", 'MachineReals._machine_reals', "dict: {'replace', 'system', 'pattern'}"]
1.

```

We see that three of the items (```2.```,``` -1.``` and ``` -10`.``) just have instances inside the class instances dict. ```743.``` are stored in the input expression, while the intermediate values (```0.16689098250336473```,```0.0013458950201884253``` ) were stored also in the lru_cache.

mmatera: It would be great to have a handler that allows to flush all the lru_cache. Then, we could use it in ```Share``` to remove all these instances when more memory is needed.

rocky: I am not getting that much insight from the above.  We got into this because there was a fear that there would be all of these items in a dictionary cache and that would cause problems. 

That benchmark doesn't seem to say anything one way or the other about this concern. 

My claim is that all of these items are around in Python's object store any way and are not likely to be garbage collected. 

The lru cache is removed in this branch, because we are going with the home-grown caching approach since we  need `__new__()` to compute object hash values and in some cases are hoping that we can  canonicalize numeric values. 

We could put lru_cache back in and remove the homegrown cache, but unless this is causing new major problems (and I don't see that yet), this just isn't something I want to focus on right now, and I don't think we should be focusing on this.

The point of departure here had to do with seeing that the hash value is computed many times when it needs to only do it once on object creation.  Then there was the question of how often object creation for a given numeric value occurs. 

A more major concern I have is organizing the code base to be more like a conventional interpreter, and one that is better documented and understandable.

mmatera (approved):
LGTM

mathics/core/atoms.py
rocky: > * At some point, what we need to do is to improve `Share` to clean both the lru cache and the unused entries in the class dictionary.
> * It would be also good to introduce classes that handle large arrays without creating MachineReal objects, at least for the intermediate steps.

In https://github.com/Mathics3/mathics-core/pull/664/commits/55ced2c7f408149ff8a425841067f1ba70644a31 I made an attempt to note the connection in this code versus `@lru_cache` and its relation with Builtin `Share[]`. If this isn't sufficient, feel free to add more. 

If you want to add notes or ideas in the ideas section, or open this as a feature for the future, please go ahead.

It is not my intent is to stifle ideas. It is just that right now, we are just a long ways away from handling  something like `Share[]` in any serious - we don't have a handle on normal evaluation.

In my opinion a big part of this is that the code isn't organized in a way where we can reasonably make sense of it and do the kinds of things that are typically done to speed up evaluation.  Let's start along the lines that interpreters typically use to speed things up before inventing or relying on custom tweaking to get there.

It has been noted, and I have noted for a while that yes, we need a way to handle large lists of numeric literals, and be able to operate on such large lists without massive conversions. It is not that I don't have ideas about this, it is again a matter of coming up with a plan of what to tackle next and focusing on that. 

If after the next release, if that's the thing you'd like to focus on instead of the makebox, boxing rules and formatting, I am okay for doing that.

mmatera: > In [55ced2c](https://github.com/Mathics3/mathics-core/commit/55ced2c7f408149ff8a425841067f1ba70644a31) I made an attempt to note the connection in this code versus `@lru_cache` and its relation with Builtin `Share[]`. If this isn't sufficient, feel free to add more.

By now, I think this is enough. My point with the comment was just to note that in the future we need to handle the overhead of storing all the used numbers in some way.


> 
> It is not my intent is to stifle ideas. It is just that right now, we are just a long ways away from handling something like `Share[]` in any serious - we don't have a handle on normal evaluation.
> 

Sure, I think we do not need to worry too much about this until we really are in the position of working with very large sets of numbers.


> In my opinion a big part of this is that the code isn't organized in a way where we can reasonably make sense of it and do the kinds of things that are typically done to speed up evaluation. Let's start along the lines that interpreters typically use to speed things up before inventing or relying on custom tweaking to get there.
> 

> It has been noted, and I have noted for a while that yes, we need a way to handle large lists of numeric literals, and be able to operate on such large lists without massive conversions. It is not that I don't have ideas about this, it is again a matter of coming up with a plan of what to tackle next and focusing on that.
> 
> If after the next release, if that's the thing you'd like to focus on instead of the makebox, boxing rules, and formatting, I am okay for doing that.

What I really want to have is a more clear API pattern, in a way that improving Boxes or the handling of large lists can be done outside Mathics-core. In this sense, I feel more comfortable continuing with the formatting refactor. In the process, we can discuss on a more concrete basis how the API and the organization of the code should be done.

PR #663, opened 2022-12-17
Add rudimentary LogPlot and ListLogPlot

rocky:
There is a bit more that could and should be done for robustness, and showing tick mark values as 10^n.

That kind of thing is left for another iteration.


mmatera:
Something that still is missing here is fixing the distribution of the ticks in the Log axis. For example,
```LogPlot[Exp[x],{x,0,3}]```
produces this plot:
imagen

Notice that in the y-axis, the space between ticks is constant, but it should also be linear on a logarithmic scale. For example, in WMA, matplotlib, and other plot tools you get something like this:

image

@rocky, I think it is OK to merge this as it is, but it would be awesome y if you have in mind how to fix this.

File: mathics/builtin/box/graphics.py [hunk 1]
```diff
@@ -798,12 +811,18 @@ def add_element(element):
                             ),
                         ]
                     )
+
+                    tick_value = 10**x if is_logscale else x
```
rocky, line 815: Here is where better labels for logplots would be added.

rocky: @mmatera Yes, I noticed this. Somewhere in a side comment I think I (or intended to) methion that this was a weakness. 

Also weak is robustness when a bad function is given, for example instead of `Sin[]` try `PartitionQ[]` which we don't implement. Or more generally when a single value can't be evaluated. 

Or more generally, it would have been good to write unit tests for the split out functions, and in also for error conditions. Or start adding docstrings for all of those internal functions that exist.

Please, please any of this interests you by all means continue here!

But this commit is getting large. There is too much to to do continuing here. (Every time we look behind some existing code there is way to much to try to redo, so at least for me, I need to break it up into little chunks and do as a number of PRs.)

But, right now I said that the main focus was thinking about how to get a release out. So we can then continue with the more serious stuff. 

The addition of plot routines like this I had mentioned _months_ ago, but because every week I was looking the more import deeper issues, I never had a chance to think about looking at something that is conceptually simpler.

mmatera: I just meant to fix it if it were a trivial change, or just add a comment to the TODO list. Again, I am OK with merging this just as it is right now.

PR #664, opened 2022-12-19
Symbol init cleanup

rocky:
Here we separate PredefinedSymbol (Symbol constants like True and False) from Symbol better.

Hmm... maybe this should be called SymbolConstant instead? 

PredefinedSymbol doesn't require Definition lookup, and uses a `_value` field instead. 

This doesn't work yet - there is some screwy interaction with List symbols and operator precedence in `mathics.builtin.makeboxes`! 

@mmatera I would be grateful if you could find out what's up here and suggest a fix.


mathics/core/symbols.py
```diff
@@ -395,25 +412,8 @@ def __new__(cls, name: str, sympy_dummy=None, value=None):
             # value attribute.
             self.sympy_dummy = sympy_dummy
 
-            # This is something that still I do not undestand:
```
rocky, line 398: When you don't understand, open a discussion or QA for this, if I created this. Thanks

mathics/core/symbols.py
```diff
@@ -706,7 +733,7 @@ def symbol_set(*symbols: Tuple[Symbol]) -> FrozenSet[Symbol]:
 # PredefineSymbol.
 
 SymbolFalse = PredefinedSymbol("System`False", value=False)
-SymbolList = PredefinedSymbol("System`List")
+SymbolList = PredefinedSymbol("System`List", value=list)
```
rocky, line 736: I am not totally sure about this. SymbolList is something that can't be changed so it is a PredefinedSymbol in that respect.

But it doesn't have a value like True or False do. (In the past I wondered also about Null)

The other option, is to allow the value parameter be optional for Predefined.

rocky:
One other thing I have been meaning to mention is that ByteArrayAtom's and String's `__new__` method should follow the singleton pattern like the other numeric classes do.

mathics/core/symbols.py
mmatera: If I would have do that, then I would be even more lost about what are these attributes about.

mathics/core/atoms.py
```diff
@@ -640,14 +670,26 @@ def __new__(cls, real, imag):
         if isinstance(imag, MachineReal) and not isinstance(real, MachineReal):
             real = real.round()
 
-        self.real = real
-        self.imag = imag
+        value = (real, imag)
```
mmatera, line 673: Why not a class first element here?

mathics/core/atoms.py
```diff
@@ -640,14 +670,26 @@ def __new__(cls, real, imag):
...
+
+            # Cache object so we don't allocate again.
+            self._complex_numbers[value] = self
+
+            # Set a value for self.__hash__() once so that every time
+            # it is used this is fast. Note that in contrast to the
+            # cached object key, the hash key needs to be unique across all
+            # Python objects, so we include the class in the
+            # event that different objects have the same Python value
+            self.hash = hash((cls, value))
```
mmatera, line 691: Notice that the hash has a different form than the key. Is this intentional?

mathics/core/symbols.py
```diff
@@ -333,56 +333,73 @@ def replace_slots(self, slots, evaluation) -> "Atom":
...
+    A Symbol is a kind of Atom that acts as a symbolic variable.
 
-    A Symbol is a kind of Atom that acts as a symbolic variable or
-    symbolic constant.
+    All Symbols have a name that can be converted to string.
 
-    All Symbols have a name that can be converted to string form.
-
-    Inside a session, a Symbol can be associated with a ``Definition``
+    A Variable Symbol is a ``Symbol``` that is associated with a ``Definition``
```
mmatera, line 340: Maybe we should also mention that a Variable is a symbol whose definition has a Downvalue.

mmatera: For example, ```System`Plus``` does not act as a variable, because it does not have an Ownvalue. As it has a Downvalue, it behaves as a "function".

mathics/core/symbols.py
```diff
@@ -395,25 +412,8 @@ def __new__(cls, name: str, sympy_dummy=None, value=None):
...
-            #
-            # My guess is that this property should be set for
-            # ``PredefinedSymbol`` but not for general symbols.
-            #
-            # Like it is now, it looks so misterious as
-            # self.sympy_dummy, for which I have to dig into the
-            # code to see even what type of value should be expected
-            # for it.
-            self._value = value
             self._short_name = strip_context(name)
```
mmatera, line 415: the short name depends on the evaluation context (i.e., on the Definitions object) so _short_name probably does not belongs here.

mathics/core/symbols.py
mmatera: The `value` field refers to a native Python's equivalent. For ```Integer``` the conversion is its value (of type int). For ```System`True``` the conversion is the Python's  `bool` `True`. For ```System`List``` the conversion is the class `list`, which is not too bad. Otherwise, I don't see the need to make ```System`List``` a ```PredefinedSymbol```.

mathics/core/atoms.py
rocky: That is a possibility.  It might also be possible to use a python complex for most of the cases where that works. There may be edge cases where it doesn't and then some other representation could be used in those special cases. 

Right now though, since this is new we aren't using this. So rather than over-engineer, this is something simple and provisional. As the need arises, it can change. But let's do this based on need. 

The reason to add a value was that this is the way all of the other numeric constants work.

mathics/core/atoms.py
rocky: Yes, again simplicity. While object hashes have to be unique across_all_ Mathics objects, keys for cache objects for a particular type, e.g. `String` or `Symbol `only need to be unique across that kind of object. And since the Mathics cache object represent a specific value in for that object, e.g. 5 for the Integer object, that value is be unique (for the Integer cache).

So taking String and Symbol as a good example (but note that right now we aren't caching strings for String), there could be a string called ``System`N`` and there _is_  a Symbol called  ``System`N``.  If we hashed these independent of the class these would come up with the same object hash. However in the type-specific object hash those values without the class name is unique. 

If one is debugging or working with the object hash, it is much nicer to work with the the value as its key rather than something more complicated.

mathics/core/symbols.py
rocky: My understanding is that the short name is the name with all contexts remove. In other words, it is _short_. 

It is like the POSIX `basename` command or `os.path.basename`.   I believe this has always been around. In the back of my mind though, I have been thinking of having print mechanisms that use the short name. 

I don't see this as a big deal one way or another.

mathics/core/symbols.py
rocky: The motivation for ``System`List``  being not redefinable has to do with confidence to create a ListExpression. If a user could change List then we would have to consult the current binding (or "definition") of `List` to see if it were okay to create a ListExpression.

mathics/core/symbols.py
rocky: Ok - in ce9a20fe I made an attempt to mention OwnValue for variables and DownValue for functions. Feel free to suggest further changes.  It feels to me though that more detail about variable symbols and function symbols should go somewhere else that is more specific to each of these aspects.

mathics/core/symbols.py
mmatera: Exactly, ```System`List``` is always interpreted as something used as the head of a List expression. And in Python, a head of a list expression is the ```list``` identifier. Regular ```Symbol``` (not ```PredefinedSymbol```s) should not have a (Python native) *value*. In any case, again the problem is how the Mathics API is (not clearly) defined.

mmatera:
> One other thing I have been meaning to mention is that ByteArrayAtom's and String's `__new__` method should follow the singleton pattern like the other numeric classes do.

In that case, could be possible to DRY the pattern for all Atoms?

mathics/builtin/makeboxes.py
```diff
@@ -470,12 +470,17 @@ def eval_postprefix(self, p, expr, h, prec, f, evaluation):
         else:
             return MakeBoxes(expr, f).evaluate(evaluation)
 
+    # FIXME: prec is sometimes not an Integer for a ListExpession.
+    # And this is recent with respect to PredefinedSymbol revision.
```
mmatera, line 474: In WMA, precision is measured in *decimal* digits. This is why `$MachinePrecision` is a `MachineReal` value. This is different for example in mpmath, where precision is defined in terms of an `int` amount of *binary* digits. This is why we need to accept floating point numbers.

rocky: The code  indicates that "prec" stands for (operator) "precedence", not "precision". Operator precedence should be an integer value.  

Something weird happens though when ListExpression get passed are embedded. Probably they just need to set the precedence to a high value so that they never get parenthesis since a list expression comes with its own outside boundary markers.

rocky:
Actually, I am closing this, and we will merge from Symbol-init-cleanup2. I started squashing commits there.

PR #665, opened 2022-12-20
Make builtin/atomic/symbols.py current...

rocky:
Bring code in line with current standards:

* Add WMA links
* use soft line breaks in doctest lines
* apply -> eval
* dl formatting
* isort imports

This is split off from #664. I probably will try to pull out other parts from that in future PRs in order to reduce the size of that PR and narrow it to the part that still needs work.

PR #667, opened 2022-12-20
Remove @lru_cache in favor of homegrown cache

rocky:
Starts to split off more of #665

We need `__new__` anyway to compute hash values.

In some cases we can canonicalize object by value, e.g. for Rational by reducing the value.

Document better what is going on here.


rocky:
Narrowed version of #665. The Symbol vs PredefinedSymbol doesn't work, although I think it a good idea. So for now the messy combined Symbol version that is in master is retained. 

We should again see a slight performance increase here at the expense of more memory. But `Share[]` handling is a interesting topic of its own for another time.

PR #669, opened 2022-12-21
url in docstrings for modules   string - statistic - specialfns

mmatera:
Here we go with the first bunch from #668. There are 14 in total, covering three submodules.

PR #670, opened 2022-12-21
docstring urls for vectors, numbers and matrices

mmatera:
Continues #669 on the submodules vectors, numbers and matrice


mathics/builtin/numbers/calculus.py
```diff
@@ -1540,6 +1568,9 @@ def apply_D(self, func, domain, var, evaluation, options):
 
 class O_(Builtin):
     """
+
+    <url>:WMA link:https://reference.wolfram.com/language/ref/O_.html</url>
```
TiagoCavalcante, line 1572: This link is broken

mathics/builtin/numbers/calculus.py
```diff
@@ -577,6 +587,8 @@ def apply(self, f, n, n0, evaluation, options={}):
 
 class _BaseFinder(Builtin):
     """
+    <url>:WMA link:https://reference.wolfram.com/language/ref/_BaseFinder.html</url>
```
TiagoCavalcante, line 590: This link is broken

mathics/builtin/numbers/integer.py
```diff
@@ -88,6 +92,8 @@ def apply(self, n, evaluation):
 
 class Ceiling(SympyFunction):
```
TiagoCavalcante, line 93: Not related to this PR, just me curious: do you know why WL have Ceiling and Floor, instead of Ceil and Floor, or Ceiling and Flooring?

mathics/builtin/numbers/integer.py
```diff
@@ -464,6 +484,8 @@ def apply(self, n, base, evaluation, nr_elements=None):
 
 class IntegerDigits(_IntBaseBuiltin):
```
TiagoCavalcante, line 485: This class is repeated, see above

mathics/builtin/vectors/vector_space_operations.py
```diff
@@ -160,6 +164,12 @@ def eval(self, u: ListExpression, v: ListExpression, evaluation):
 
 class UnitVector(Builtin):
     """
+    <url>
+    :Unit vector:
+    https://en.wikipedia.org/wiki/Unit_vector</url> (<url>
+    :WMA:
+    https://reference.wolfram.com/language/ref/UnitVector.html</url>)
```
TiagoCavalcante, line 171: This is split into multiple lines without the "\", isn't this a problem?

TiagoCavalcante (changes requested):
LGTM - just a few things

mathics/builtin/numbers/integer.py
mmatera: Good question!

mathics/builtin/numbers/integer.py
mmatera: This was here for a while. Eventually, we should try to merge both.

mathics/builtin/vectors/vector_space_operations.py
mmatera: I just copy and paste another existing case...

mathics/builtin/numbers/calculus.py
mmatera: This comes from the automatic fixing, and should be removed.

mathics/builtin/numbers/calculus.py
mmatera: This should be ```O``` instead of ```O_```

mathics/builtin/vectors/vector_space_operations.py
rocky: > This is split into multiple lines without the backslash, isn't this a problem?

The home-grown code is copying the nature of HTML and XML (where XML is a more fancy worLd for HTML because XML schemas are not supported; the HTML schema for XML is assumed. So while you will see references to XML in the code, HTML is meant.).

In HTML, recall that newlines are not significant. Recall you'd use `<br/>` to force a new line, or maybe a `<p>` would do that as well. 

The problem with the homegrown approach is, well, that it is _homegrown_. I imagine there are XML/HTML Python libraries out there, but this homegrown code isn't using any of those. 

As with most homegrown code,  it is in incomplete in various ways or has bugs.

(I can't help be reminded between parallels with the flaws of the homegrown HTML code and homegrown code to to change it. Please, let us learn from this mistake. Or, if not that, our constant mistake that whenever homegrown code is written, accept that is going to have bugs. And when code you know is going to have bugs, don't try to shove the whole codebase through it to start out.)

 Back to the homegrown doc system. In particular, it does not reflow paragraphs the way a HTML renderer typically does. When it sees newlines in non-tagged text, it just copies that to output.

Why don't we fix this? Well, the code is a mess and the thing should have never been used in the first place. We should have hooked into a Python doctest system, since that was intended.  My own feeling is that I'd rather put effort into ditching this for something more powerful and that is maintained by others in standard package like sphinx than fixing this homegrown system. 

So sadly, we've been limping along with this crap.  

The homegrown doctest system does however strip extra spaces between words.  And that is the feature we are making use of with adding `\` at the end of a line. Since this inside a Python triple-quoted string --  _Python_ is turning the "\n" into " ".

mmatera: In any case, it seems to work for now. Something that I have back in my mind when I wrote the automatic replacement was to see if it would be possible to write an automatic translation from our homebrew documentation system to some more standard like sphinx (as @rocky suggested several times) without the need of rewriting al the documentation by hand. In any case, even adding the URL field for the proper Builtin subclasses is not an easy task...

mathics/builtin/numbers/integer.py
rocky: 3edfde29 merges the two.

rocky:
@mmatera @TiagoCavalcante I think all of the issues have been resolved. However if there are more or I missed something, just jot them down and a future commit will address these. Thanks.

PR #671, opened 2022-12-21
docstr url for files_io, intfns, list and trace

mmatera:
This continues #670. Here appears for the first time some symbols that are not present in WMA. Just for completeness, I added in these cases a fake URL field of the form
```<url>:trace debug native symbol:</url>```
```<url>:internal native symbol:</url>``
`
The plan is, when this round finishes, to add a test to check that all the ```Builtin``` symbols have an URL field and that the field links to a valid URL. In that step, we can decide if we handle these cases as special cases and just remove the field, or build entries in mathics-devel documentation and link to them there.


rocky:
Pleas add '##' to before <url> so we don't this this show up in docs. Thanks.

mmatera:
> Pleas add '##' before `<url>` so we don't this in the docs. Thanks.

good idea, thanks!

rocky:
@mmatera - To speed things up, I am just committing on top more. If you have questions or issues, please raise them and they will be addressed in future commits. Thanks .

mmatera:
Sure, go ahead

PR #680, opened 2022-12-23
Make ByteArray more WMA compatible...

rocky:
and handle Numpy 1.24


mathics/builtin/drawing/image_internals.py
```diff
@@ -78,7 +78,7 @@ def pixels_as_ubyte(pixels):
         return pixels
     elif dtype == numpy.uint16:
         return (pixels / 256).astype(numpy.uint8)
-    elif dtype == numpy.bool:
+    elif dtype == bool:
         return pixels.astype(numpy.uint8) * 255
```
mmatera, line 82: Why equal instead of is?

mmatera: Also, I guess is a good ocasion to move this to the mathics.eval module.

rocky: > Why equal instead of is?

Fixed now.

> Also, I guess is a good occasion to move this to the mathics.eval module.

Yes, easy enough done.

rocky: And done now.

mathics/core/atoms.py
```diff
@@ -559,10 +559,10 @@ def __hash__(self):
         return self.hash
 
     def __str__(self) -> str:
-        return base64.b64encode(self.value).decode("utf8")
+        return f"<{len(self.value)}>"
```
mmatera, line 562: Not sure about that: __str__ should provide a str representation of the object. One possibility would be to replace line 566 by this one and keep the current behavior. @rocky, what do you think?

mathics/builtin/list/eol.py
```diff
@@ -986,7 +986,7 @@ def eval(self, list, i, evaluation):
             return
         indices = i.get_sequence()
         # How to deal with ByteArrays
-        if list.get_head_name() == "System`ByteArray":
+        if list.get_head() == SymbolByteArray:
```
mmatera, line 989: `is` instead of `==`

mathics/core/atoms.py
rocky: Yes, you are correct that this is better. Will appear in next revision.

mathics/builtin/list/eol.py
rocky: Thanks - will appear in next revision.

mathics/core/atoms.py
mmatera: Notice also that this form is just for the Output form:
In WMA, 
```
In[1]:= a=ByteArray[Table[Mod[i^2,255],{i,1,40}]]                               

Out[1]= ByteArray[<40>]


In[2]:= ToString[a]                                                             

Out[2]= ByteArray[<40>]

In[3]:=%//FullForm

Out[3]//FullForm= "ByteArray[<40>]"
```

However,

```
In[4]:= a//FullForm

Out[4]//FullForm= 
 
>   ByteArray["AQQJEBkkMUBRZHmQqcThASJFapG65RNCc6bbE0yHxARFiM0VXqn2Rg=="]

In[5]:= a//InputForm                                                            

Out[5]//InputForm= 
ByteArray["AQQJEBkkMUBRZHmQqcThASJFapG65RNCc6bbE0yHxARFiM0VXqn2Rg=="]

In[6]:= a//OutputForm                                                          

Out[6]//OutputForm= ByteArray[<40>]
``` 

So, the right thing to do is to implement this different behaviours in `atom_to_boxes`

mathics/core/atoms.py
```diff
@@ -559,10 +559,10 @@ def __hash__(self):
...
 
     def atom_to_boxes(self, f, evaluation) -> "String":
-        res = String('""' + self.__str__() + '""')
+        res = String(self.__str__())
```
mmatera, line 565: Here, we should build the return string conditionated to `f`.

mathics/core/atoms.py
rocky: But it looks like there is second place that also needs to be changed for this.

rocky: Fixed in a0c9e55f

mathics/builtin/list/eol.py
rocky: Fixed in a0c9e55f

mathics/core/atoms.py
rocky: Was done in a0c9e55f

mathics/core/atoms.py
rocky: > 
> So, the right thing to do is to implement this different behaviours in `atom_to_boxes`

I don't like the way this is going..  The Form should be intercepting here not `atom_to_boxes()` testing on the Form. We went over that before. 

So let's leave this for now and address when we go over forms. I will, however make a note of this information.

rocky: And probably the same thing is true for ToString.

mathics/core/atoms.py
```diff
@@ -561,6 +561,12 @@ def __hash__(self):
     def __str__(self) -> str:
         return base64.b64encode(self.value).decode("utf8")
 
+    # FIXME: the below does not use the "f" parameter to
+    # change behavior between FullForm and OutputForm
+    # Below we have the OutputForm behavior.
+    # A refactoring should be done so that this routine
+    # is removed and the form makes decisions, rather than
+    # have this routine know everything about all forms.
```
mmatera, line 569: Yep, I fully agree: `atom_to_boxes` should be removed in favor of formatting rules (`do_format_...`). However, in the meantime, I guess it is OK to implement here the right behavior, and then move the code to the right place.

rocky: The problem here is that we really don't know all the cases and I don't want to find out right now and code in all of the possibilities. You haven't tried for example InputForm  or all of the other forms. 

Were this done this routine would be a mess. And when it is done the proper way it is organized by Form, not by atom it doesn't look like a mess. 

So again, I don't want to create a mess now to just have to clean that up later. My take is acknowledge the deficiency and move on.

rocky: I am happy to revert all of the ByteArray stuff, and just leave the code that is needed to support NumPy 1.24.

mmatera: I just was talking about handling the basic cases. More general cases could be handled by adding rules, and eventually, removing the atom_to_boxes. But OK. Let's keep this simple and improve it later. In any case, I think that the best is that `__str__` returns the b64 encoded version of the bytearray.

mmatera (approved):
LGTM

mathics/core/atoms.py
mmatera: > I am happy to revert all of the ByteArray stuff, and just leave the code that is needed to support NumPy 1.24 as a _separate_ PR.

I do not think it is needed. Let's merge and continue.

PR #683, opened 2022-12-24
Add isort as a pre-commit hook and as a CI check

rocky:
Should we add isort import reformatting and checking? 

Your thoughts?


mmatera:
I like this! anything that helps to canonicalize the code automatically, for me is welcome.

TiagoCavalcante (approved):
I 100% agree - looks great!

PR #686, opened 2022-12-24
The remaining part of e1

mmatera:
Not sure what part of this we want to keep.


mathics/builtin/atomic/numbers.py
```diff
@@ -960,6 +961,7 @@ class MinPrecision(Builtin):
 
 class NumericQ(Builtin):
     """
+
```
rocky, line 964: Why add a line here and remove it from below?

rocky: Let's get rid of these black lines that have been added. I don't see that this is doing anything.

rocky:
LGTM - but please let's get rid of the blank lines at the top of the docstring.

PR #688, opened 2022-12-24
Start refactoring image module

rocky:
Fixes problems loading GIF image basically by doing less and letting PIL do more. This is probably also more robust. (And there is probably a lesson here as well.)

Move out static image function to mathics.eval.  

We need to do something better though with respect to Import/Export Image. In WMA this is simply Import/Export.


mmatera (approved):
LGTM. The only comment is that eventually, I think this should be moved as an external module.

rocky:
> LGTM. The only comment is that eventually, I think this should be moved as an external module.

While I don't disagree necessarily, there may be some misimpressions that I'd like to clarify. 

First, in a sense stuff _is_ getting moved. Basically we are separating/segregating the Builitin Mathics-specific and Python library independent code from the more Python library code. The former is in `mathics.builtin` while the later is in `mathics.eval`. 

Second, more important is to clean up the code or make it more intelligible. In that, issues surface. 

Here is one such issue. I think the code has its own method for doing Gaussian resampling for Image resizing and filtering.  Exactly to what level, I don't know. But I am pretty sure it is nowhere near sophisticated as https://reference.wolfram.com/language/ref/Resampling.html

My own experiments on the existing images that we have is that I am not seeing that much difference in adjusting the sampling methods. So the effort here on this does not seem warranted, especially given all of the other problems in the code base and the limited resources. 

We hook into PIL filters, which I am not sure are exactly comparable to what WMA provides. 

So given this, here and now I ditched the Gaussian resampling (it is commented out) and we are going with just the PIL kind of resampling. 

This is a slight incompatibility, in behavior. But I don't think many will notice. No options does something reasonable.  
"Automatic" is probably also the most used option when an option is given and  this still does something,; "Nearest" is probably the same . "Hamming" and "Lanczos" are probably similar.  

But the principle that I want to raise here is that on a case by case basis we decide when to be compatible. Especially when it is too much work for not enough benefit.  And my own take is reduce the homegrown code which breaks (was probably the main cause of breakage here) in favor of some better maintained library code.

Again, here it looks like Image handling has been broken for a while. My guess is that PIL has changed over the years and we haven't been keeping track of this. And we were using somewhat undocumented low-level features as well to create homegrown features that are available in libraries. In particular PIL has a image crop and an image resize. So using that makes us less fragile - all the incompatibility in PIL is addressed by the PIL folks, they keep the API resize, and crop the same. And when they don't, it is much easier to adapt then adapt the homegrown code. 

So having something working and living with slight differences in resampling methods is far better than offering a non-working "Gaussian" for some images with homegrown code.

Ok, so first to sum up. The work here is a prerequisite for _contemplating_ moving it out. Before moving code out we need to understand what must to be done, unless there happens to be someone magically who has a clear idea and can just ditch all of this and write it from scratch.  In reality, there isn't likely to be someone we will find for that. 

This train of thought reminds me of what the wishful or hopeful thinking that if we were to just rewrite this in C or Rust, all of the performance problems would be fixed.  I don't think so if you try to do a direct translation of the mess we  have right now. If one is thinking that one can write a clean rewrite WMA from scratch in Rust,  _without_ fully understanding this code, then it is likely that the same or similar mistakes might be made in that code base. 

One final aspect with respect to moving this out. There are current problems and fundamental with pymathics.  That would need to be addressed first. Those problems are intertwined with some of the existing problems we have with scoping that we haven't started looking at yet.

rocky:
This PR has gotten a lot large than I had expected. The rot is probably greater than I expected. 

So let's merge and continue iterating. ImageResize and ImageTake have been somewhat fixed. I haven't rigorously gone through all of the options and forms. ImageTake in particular needs more variants taken into account.

mmatera:
@rocky, first, thanks for the work of improving and tiding up this module. Of course, if we ant this functionality here or in a different module this is clearly the first step.
My comment was in the direction that this kind of things, that depends on other non standard libraries, should be moved out of the core, to make them easily maintainable. For it, we also need to fix some of the big problems with Mathis core: to define the API and to fix the way in which we handle external modules. But I guess this is going to come after the next release.

PR #690, opened 2022-12-26
alternative implementation for MakeBoxes[Infix[...],format]

mmatera:
This provides an alternative implementation, that avoids using WL pattern matching to handle the different cases in ```MakeBoxes[Infix[...],format]```

### DIFF
--- a/mathics/builtin/layout.py
+++ b/mathics/builtin/layout.py
@@ -1,11 +1,11 @@
 # -*- coding: utf-8 -*-
 
 """
-This module contains symbols used to define the high level layout for
+This module contains symbols used to define the high level layout for \
 expression formatting.
 
-For instance, to represent a set of consecutive expressions in a row,
-we can use ``Row``
+For instance, to represent a set of consecutive expressions in a row, \
+we can use ``Row``.
 
 """
 from typing import Optional, Union
@@ -18,48 +18,51 @@ from mathics.builtin.options import options_to_rules
 from mathics.core.atoms import Integer, Integer1, Real, String
 from mathics.core.convert.op import operator_to_ascii, operator_to_unicode
 from mathics.core.element import BaseElement
+from mathics.core.evaluation import Evaluation
 from mathics.core.expression import Expression
 from mathics.core.list import ListExpression
 from mathics.core.symbols import Atom, Symbol
 from mathics.core.systemsymbols import (
     SymbolFullForm,
-    SymbolInfix,
     SymbolInputForm,
     SymbolLeft,
     SymbolMakeBoxes,
     SymbolNone,
     SymbolOutputForm,
     SymbolRight,
+    SymbolRowBox,
 )
-from mathics.eval.makeboxes import (
-    eval_fullform_makeboxes,
-    format_element,
-    make_boxes_infix,
-)
+from mathics.eval.makeboxes import eval_fullform_makeboxes, format_element, parenthesize
 
 SymbolNonAssociative = Symbol("System`NonAssociative")
 SymbolSubscriptBox = Symbol("System`SubscriptBox")
 
 
-def eval_makeboxes_infix(
+####################################################################
+# This section might get moved to mathics.box
+#
+# Some of the code below get replace Mathics code or may get put in
+# collection of boxing modules,
+# e.g. ``mathics.box.boxes_operators.box_infix()``.
+#
+####################################################################
+
+
+def box_infix(
     expr: Expression,
     operator: BaseElement,
     form: Symbol,
-    evaluation,
-    precedence_value: Optional[int] = None,
+    evaluation: Evaluation,
+    precedence_value: Optional[int] = 0,
     grouping: Optional[str] = None,
-):
-    """Implements MakeBoxes[Infix[...]]."""
+) -> Optional[Expression]:
+    """Implements MakeBoxes[Infix[...]].
+    This function kicks off boxing for Infix operators.
 
-    # Actually, this function implements one part of the evaluation:
-    # the one that should be done at the "format" level. In that point,
-    # the operators are processed to add spaces and use the right
-    # encoding. This should be done at the level of the formatter, not
-    # in the "MakeBoxes" step.
-    # Then, the true implementation is done
-    # in  ```mathics.eval.makeboxes.make_boxes_infix```.
+    Operators are processed to add spaces and to use the right encoding.
+    """
 
-    ## FIXME: this should go into a some formatter.
+    # FIXME: this should go into a some formatter.
     def format_operator(operator) -> Union[String, BaseElement]:
         """
         Format infix operator `operator`. To do this outside parameter form is used.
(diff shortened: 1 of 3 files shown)


mathics/eval/makeboxes.py
```diff
@@ -102,6 +102,16 @@ def make_boxes_infix(
     return Expression(SymbolRowBox, ListExpression(*result))
 
 
+def eval_fullform_makeboxes(self, expr, evaluation, f=SymbolStandardForm):
+    """
+    This function takes the definitions prodived by the evaluation
```
rocky, line 107: prodived -> provided.

rocky:
LGTM - and much needed. Thanks!

mathics/builtin/layout.py
```diff
@@ -166,68 +170,34 @@ class Infix(Builtin):
...
-    def eval_infix_3(
-        self,
-        expr: Expression,
-        operator: BaseElement,
-        precedence: BaseElement,
-        form: Symbol,
-        evaluation,
-    ):
-        """MakeBoxes[Infix[expr_, operator_, precedence_],
+    def eval_all(self, expression, form, evaluation):
```
rocky, line 173: Why not just call this `eval` ?

mathics/builtin/layout.py
```diff
@@ -166,68 +170,34 @@ class Infix(Builtin):
...
+        elements = list(infix_expr.elements)
+        num_parms = len(elements)
+
+        if num_parms == 0 or num_parms > 4:
+            evaluation.message("Infix", "argb", Integer(len(parms.get_sequence())))
+            return eval_fullform_makeboxes(infix_expr, evaluation, form)
+
+        if num_parms == 1:
+            expr = elements[0]
+            return self.do_eval_infix(expr, String("~"), form, evaluation)
```
rocky, line 186: These `do_eval` methods in a class really should be  functions outside of any class in a boxing module. 

Note that when I say a boxing module maybe say `mathics.box`.  Note that this kind of "eval" an evaluation that is done in Boxing and we should start making distinct from expression evaluation.

While it is true there may be rare cases where in boxing a general evaluation goes on, for the most part it doesn't.  In other words it is rare that I will want to find to prime integer factors of a number in boxing.  In _formatting_ I might need to compute a value like 3/2 and maybe I feel compelled to use the full strength of WL to do that. But in such rare cases that would be triggered by an explicit `Expression().evaluate()`, rather than implicitly by iterating over an M-Expressions in the boxing process.

rocky: One other thing I'd like to point out. Note that renaming `self.do_eval_infix` to `mathics.box.box_infix`  is different than collecting the misguided `atom_to_boxes(<form>)` in one place. 

I think there is a subtle difference between  `do_eval_infix` and some of the `atom_to_boxes()` routines. In`do_eval_infix` should be working at a high-level of boxing (in contrast to the atoms which are turning say an image into a base64-encoded PNG wrapped in a box).   

And as such -  working at a high  level - it could make use of form transformation rules (say collected in a file). 

Right now I do not envision testing on the specific form passed to it `if form is ...` but rather that rules that the form object contains are used.   I think this kind of thing was shown in one of your makeboxes implementations in WMA.

rocky: I just pulled the branch and looked at the code. Maybe gentler than creating a `mathics.box` module for essentially one thing right now (that might be replaced by Mathics rules), would be just to move this out of the class - that is turn it from being a method in a class to a module function - and call it `box_infix()`. 

So what I am also suggesting is that we _not_  use the `eval_` convention for the evaluation aspect of boxing, but instead use a different prefix like `box_`. (Note that "box" is an active verb as well as a noun).

mmatera: What I tried to do here was an small change, keeping the existing structure. But yes, at some point what you suggest is where I am trying to go.

mmatera: To be clear, I think that if the implementation of a `eval_` method is more or less complex, then it should be moved to an external function, in the same module or in another one. I will try to do something like this over this branch, but avoiding changing the code too much. Then, we can try to improve and simplify the internal logic.

rocky: I totally understands. 

Would pulling out `self.do_eval_infix` out and renaming it to `box_infox` and  `eval_all` to `eval`  be disruptive? 

I note that `self.do_eval_infix()` seems to be new in this branch and is not in master.

mmatera: ```Infix.do_eval_infix```  is almost the same code that we have in ```MakeBoxes.eval_infix``` in master.

mmatera: I renamed ```eval_all``` to ```eval_makeboxes``` because I want to be able to find these methods afterward when we can continue with the Format/Box refactor.

rocky: > `Infix.do_eval_infix` is almost the same code that we have in `MakeBoxes.eval_infix` in master.

Right, so what should eventually happen is that all of those `MakeBoxes.eval_...` methods should be pulled out of Makeboxes and turned into `box_x...` methods. In the future these should not be called by the existing Expression.rewrite_apply_eval()`  but instead called via `MakeBoxes.box(form, ...)`

mathics/builtin/layout.py
```diff
@@ -41,6 +41,78 @@
 SymbolSubscriptBox = Symbol("System`SubscriptBox")
 
 
+def eval_makeboxes_infix(
```
rocky, line 44: Small thing, but instead of `eval_makeboxes_infix` how about more simply `box_infix`.  Right now you can search on `def box_` and you will not find anything.

I think it important that we start being more precise about what we call things, and that I hope will lead to clearer thinking and discussion about these things and lead to less confusion.

mmatera: OK. However, notice the comment that I added in https://github.com/Mathics3/mathics-core/pull/690/commits/11fc371d3d31fb7ae7bfd78883fb4f7bf8fcf976

At some point, this function is going to be removed, in favor of ```mathics.eval.makeboxes.make_boxes_infix```. This function is the one that would be renamed as ```box_infix```.

rocky: See other comment about the name  `mathics.eval.makeboxes.make_boxes_infix`

rocky: > See other comment about the name `mathics.eval.makeboxes.make_boxes_infix`

I can't find the other comment right now. Maybe I was interrupted and didn't send. 

Please not `mathics.eval`.  Boxing M-expressions is not the same thing as M-expression evaluation. 

So instead I suggest  `mathics.boxes.box_infix`.

mmatera: Ok, maybe it is going be simpler if you perform the changes over this branch.

rocky: Ok will do. Thanks.

rocky:
@mmatera Ok - here is what I had in mind. Thanks for giving me the opportunity to steer this.

PR #698, opened 2022-12-28
using rules instead of atom_to_boxes

mmatera:
In this PR I removed the `atom_to_boxes` for almost all the atoms, in favor of MakeBoxes rules.


mmatera:
Deprecated

PR #699, opened 2022-12-28
reimplementing eval_makeboxes. step 1

mmatera:
This is the first step in my proposal of refactoring  `MakeBoxes`. The main change here is that MakeBoxes rules are applied in `mathics.eval.makeboxes.eval_makeboxes` instead of the `mathics.core.expression.rewrite_apply_eval_step`. In a next step, makeboxes rules are going to be stored not as `downvalues` of the `MakeBoxes` symbol, but as format values.


rocky:
@mmatera I do not intent to look at this until after the next release. Actually, in some respect this is kind of a distraction right now.

mmatera:
Deprecated

PR #700, opened 2022-12-29
Start to split out Image into its own section

rocky:
Also, at this point numpy and PIL are not optional. So remove `_image_requires` which was numpy and PIL.


mmatera (approved):
LGTM

rocky:
Let's merge after CI finishes.

PR #701, opened 2022-12-31
Make LaTeX doc internal refs work

rocky:


mmatera:
I have contradictory feelings about this: it is great to have better documentation for the next release. On the other hand, we are continuing patching our homebrew documentation system since at least two years ago we want to migrate to something more standard. And as we add more features, the migration is going to be more complex.

rocky:
> I have contradictory feelings about this: it is great to have better documentation for the next release. On the other hand, we are continuing patching our homebrew documentation system since at least two years ago we want to migrate to something more standard. And as we add more features, the migration is going to be more complex.

I agree that I hate this code.  I also hate working on it. It keeps reminding me about how boneheaded the code was in concept, as well as poorly coded.

But replacing it as you have said in the past is going to be major work. And in advance of doing that I'd like to improve our documentation. And that includes cross reference.

rocky:
I would like to elaborate a little more on why it may seem unfair that we aren't doing boxing refactoring but LaTeX doc refactoring now.

Boxing and Formatting code is something that Mathics users are confronted with all of the time. So it has to be right. Little flaws are magnified greatly.  The Boxing work is going to take months before I think we are satisfied with the design and that it is safe. 

On the other hand, the LaTeX code seems to come up only around release time. It is rarely used otherwise.  The stuff I am doing right now which is splitting off LaTeX-specific code from from the common Doc code is not controversial. It was one of those many TODOS that just hasn't been done.  Before release, personally, I would like to see more of the TODOS done. 

This refactor isn't going to take months, but rather days.

mmatera:
To be clear, my concern is not about priorities, but about not increase too much the complexity of the documentation, if we want to implement a massive refactor in the future.

rocky:
> To be clear, my concern is not about priorities, but about not increase too much the complexity of the documentation, if we want to implement a massive refactor in the future.

Actually, this is _reducing_ complexity.  For example, we basically duplicated the documentation code into mathics django and then tried to strip it. Here we are splitting off the LaTeX specific stuff from the common stuff so that we can fix bugs in the LaTeX part. So that reduces confusion about whether this fix is related to Django or no.

And the code is getting better type annotated and documented in general. So I hope there is a benefit that it when revised we have a clearer sense of all the parts that are going on that need to be accomodated. 

Lastly, there are just some basic things about being able to find and how the asymptote relates to the section in the doc. 

As long as we are using this, this has been an annoying aspect of fixing the documentation part of the doc bugs.

I will be adding the ability to render the images in the LaTeX doc which is something that has been missing all along.

However, to the extent that mathics-core is cleaner this should not be so bad.

mmatera:
For some reason, make doc is failing to build the pdf, but it seems unrelated to this PR.

rocky:
This is getting large. Let's merge and iterate.

Issue #681, opened 2022-12-24
Issues to address before 6.0.0 release


rocky:
* [x] `Import["ExampleData/MadTeaParty.gif"]` is broken
* [x] Show SciPy version (or non) under version info (and in Django)
* [x] (Django) Go over gallery examples, possibly add new stuff. color boxes are too small
* [x] Asymptote axes for Plot are misaligned possibly now that they are aligned for SVG
* [x] LaTeX doc rendering of Asymptote graphs is misaligned. Go over LaTeX doc for content
* [x] Remove Combinatorica error in trying to redefine CatalanNumber
* [x] Replace all methods named `apply` which are really `eval` methods?
* [x] Remove acceptance of `apply` in the `contribute()` function?
* [x] Finish removing "miscellaneous" `List` functions.
* [x] Many Mathics3 Modules builtins docstrings don't have WMA links
* [x] Organize Mathics3 Modules graph and natlang into sections like we do for builtins with suitable names
* [x] Add doctests (docpipeline.py) for pymathics modules - sort of along the lines of what we do for Mathics-Django. However testing should only be done of the Mathics3 Modules3 module that is getting added
* [x] Add minimal set of pytests for Mathics3 Modules. I have started some work here in `pymathics.graph`.
* [x] Build LaTeX PDF with Mathics3 Modules natlang and graph included. 
* [x] Add Gallery and Mathics Doc Demo examples for Mathics3 Modules
* [x] One more dockerhub release beforehand to make sure this works.
* [x] Max Recursion with CPython when `Flat` attributes are present #736 
* [x] Go over `CHANGES.rst` (which is really part of the release itself.

For the release, Mathics3 Modules don't have to be perfect, just good enough to show that they exist so that users start to realize that they are there . Mathics3 Modules can be improved after release, but we should ensure that we are not likely to need another API change in this area. 

The docs just should have a reasonable organization and information like we do for Built-ins. Having something that is _obviously_ shoddy when it is a simple matter to correct this, doesn't encourage participation or use of Mathics3 Modules. So again the goal is "not obviously shoddy" using an API that won't likely need changing so soon again.

rocky:
@mmatera - in going over documentation bugs, I am seeing really weird behavior that seems boxing related that has occurred since the last release.  I hope you can help with this:

If you look at the output in Django for: 

```
[...]

PR #703, opened 2022-12-31
fix extra quote in insets

mmatera:
This fixes the issue mentioned in #681 regarding the superfluous quotes in graphics. I just commented out the code that produces the effect because maybe we want to fix this in a different way.


rocky:
LGTM.

rocky:
And I also agree that after release, all of this code will probably be gone over and made easier to understand _and_ more accurate (and probably even faster too).

rocky:
LGTM - and thanks!

PR #705, opened 2023-01-01
Improving tests and examples for FormatMapping

mmatera:
This fix an issue in building the pdf documentation, and improves the documentation of the corresponding symbols.


rocky:
LGTM -  I tried building the PDF and this is works. (We still have other problems but that is for another PR.

PR #708, opened 2023-01-02
Doc updates and add mising morph.py

rocky:
* Document new Django buttons.
* We use Django 4 now
* Rename Mathics to  Mathics3
* Go over history and some spelling corrections


mmatera (approved):
LGTM.  Just run admin-tools/build_and_check_manifest.py to check the symbols manifest.

PR #709, opened 2023-01-02
Allow fontsize on Asymptote pen creation...

rocky:
Use that in asy inset_box in creating labels.

In general, we trying to make SVG and Aysmptote more the same.

In the process add some comments and type annotations.


TiagoCavalcante (approved):
LGTM

PR #710, opened 2023-01-03
Small changes to Result class.

rocky:
This came up when I was considering a larger refactor of testing and doc generation


TiagoCavalcante (approved):
LGTM - named tuple for this is a good idea

PR #691, opened 2022-12-27
Symbol init cleanup 2

rocky:
Merging this into master, and then I'll rebase Symbol-init-cleanup and we are done!

Thanks @mmatera to getting this to the point where I can see the light at the end of the tunnel.

PR #714, opened 2023-01-03
Remove "value" from Symbol; add SymbolConstant...

rocky:
from PredefinedConstant.

Doc strings have been over.

This includes some of the stuff from #691. I think there is a little other stuff from that which should go in. But this narrows the differences.


mathics/builtin/makeboxes.py
```diff
@@ -485,7 +485,7 @@ def format_operator(operator) -> Union[String, BaseElement]:
                 return op
             return operator
 
-        precedence = prec.value
+        precedence = prec.value if hasattr(prec, "value") else 0
```
rocky, line 488: Value has been removed from Symbol, so is a workaround #691 I think handles this better, but other than fixing things so that we don't call this routine when it shouldn't be called I don't recall the details

@mmatera I think you had the key insight here.

rocky:
I will let this sit for a day, and if there are no comments or objections, I'll merge.

mathics/builtin/makeboxes.py
mmatera: @rocky, maybe it would be better to set the default to 670 (which is the default for precedence).

mmatera:
Regarding this PR, I agree that Symbols should not have a "value" property.

mmatera (approved):
apart from the comment, LGTM

mathics/builtin/makeboxes.py
rocky: 0 means no parenthesis ever.  670 adds a parenthesis for everything except Get `<<`. 

So restating are you sure we want parenthesis most of the time?

mmatera: OK, in that case, let's keep 0...

rocky: In another PR we may add a constant named "NEVER_ADD_PARENTHESIS".  There was more refactoring oef the parenthesize function going on in #691

PR #716, opened 2023-01-03
Handle degenerate LineBox when there are no points

rocky:
I encountered this when building the docs and something along the lines of

```
ParametricPlot[{Sin[u], Cos[3 u]}, {u, 0, 2 Pi}]
```
which creates in `mathics-229.asy`: 

In creating an axis it uses:

```
draw(, rgb(0.2472, 0.24, 0.6)+linewidth(0.66667));
```

The empty list of points is mysteriously coming from the default TickStyle `{}`.

We are seeing this  Asymptote only

Also TicksStyle handling seems weird since we shouldn't try to create LineBox in the first place.

### DIFF
--- a/mathics/builtin/box/graphics.py
+++ b/mathics/builtin/box/graphics.py
@@ -1034,7 +1034,9 @@ class InsetBox(_GraphicsElementBox):
 
 
 class LineBox(_Polyline):
-    # Boxing methods for a list of Line.
+    """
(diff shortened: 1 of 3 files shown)


rocky:
I will let this sit for a day or so and if there are no comments or objections, I will merge.

mmatera (approved):
LGTM

PR #719, opened 2023-01-04
More doc updates

rocky:
More of the same. But apparently we weren't showing Units and Quantities. And there were bugs as a result of our more stringent Expressions which do not allow passing in Python strings or numeric values.


mathics/builtin/quantities.py
```diff
@@ -12,24 +14,42 @@
...
+# This tells documentation how to sort this module
+sort_order = "mathics.builtin.units-and-quantites"
 
 ureg = UnitRegistry()
 Q_ = ureg.Quantity
 
 
+def get_converted_magnitude(magnitude_expr, evaluation: Evaluation) -> float:
+    """
+    The Pythion "pint" library mixes in a Python numeric value as a multiplier inside
```
mmatera, line 33: Python instead of Pythion?

rocky: Good - thanks. fixed.

PR #720, opened 2023-01-04
fix wp link

grondilu:


rocky:
Thanks for the correction! (Will merge after CI finishes)

PR #721, opened 2023-01-04
Move more functions out of misc list...

rocky:
More eval-like code moved into eval.

More code should be moved out of the miscellaneous `mathics.builtin.list` into something more specific. But this is all I have time  for for today. And this kind of large already.


mathics/format/asy.py
```diff
@@ -750,4 +750,21 @@ def uniform_polyhedron_3d_box(self: RectangleBox, **options) -> str:
     )
 
 
+# def tetrahedron_polyhedron_3d_box(self: RectangleBox, **options) -> str:
```
mmatera, line 753: How is this related to the other changes?

rocky: It isn't. I had been working on that and this was on the disk when I committed. Rather than create another branch, may as well commit what I had so that later I can come back to this (after release probably).

rocky: Ok - I now have all this code (and other code that was on my local disk) put in a local branch. So this is now removed.

After CI clears I will merge this. There a still a bit more on the miscellaneous list functions.

PR #722, opened 2023-01-07
builtin.list reduction mathics.eval expansion..

rocky:
* Move mathics.builtin.list._IterationFunction to mathics.base.IterationFunction
* Move eval-like routines out of mathics.builtin.arithmetic into mathics.eval.numbers
* Move scoping routine out of builtin.base into mathics.eval.scoping


mmatera (approved):
LGTM

PR #739, opened 2023-01-10
Require newer PIL; handle Image failures

rocky:
Pillow 9.1.0 and greater support big TIF bigendian. We use that in hedy.tif.

Go over image loading failure handling and more generally import failure handling.


rocky:
Discussion: 

I opted for removing Python 3.6 from testing since 3.6 does not support Pillow 9.10, the minimum version that supports reading Big TIF bigendian byte order used in `lena.tif`.  Another possibility would be to use a TIF in another format. 

More generally `Import` will returned `$Failed` when it fails to import. 

Tests should be written for this aspect, but I leave that for others or some other time. 

Python 3.6 will still build, it just won't pass the image tests and should come back with a `$Failed` when trying to load such an image.  We could put effort into having the tests get skipped on Python 3.6 so there is no doctest failure, but I don't think it worth the effort.

Note that 3.6 is no longer supported by the Python. And after the next release we will remove the existing code that has Python 3.6 specific workarounds. (3. 7 is slated to be removed soon too. And we should try to get 3.11 working and 3.10 into the normal testing.)

Your thoughts?

mathics/builtin/image/misc.py
```diff
@@ -69,11 +69,24 @@ class ImageImport(Builtin):
      = -Image-
     """
 
+    messages = {
+        "imgtype": "Can't identify image file type for `1`.",
```
rocky, line 73: @mmatera I have no idea what Mathematica does in this situation. Please feel free to suggest or adjust the error messages.

.github/workflows/osx.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches: [ master ]
   pull_request:
-    branches: [ master ]
+    branches: '**'
```
TiagoCavalcante, line 7: Why two stars instead of one?

TiagoCavalcante (approved):
LGTM

mathics/builtin/image/misc.py
mmatera: In WMA, there is not a symbol `System`ImageImport`. If I fill a file `a.tif` with random bytes, using `Import`, in WMA I get


```
In[1]:= Import["a.tif"]                                                         

Import::infer: Cannot infer format of file a.tif.

Out[1]= $Failed
```

rocky: @mmatera  thanks - 7a2567ce0 has this message adjusted. When CI clear I will merge that the `replace-lena` branch.

PR #745, opened 2023-01-12
Add Random[]. We also assume some sort of NumPy

rocky:
so we can simplify code here.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -71,10 +71,21 @@ Bugs
 ++++
 
 # ``0`` with a given precision (like in ```0`3```) is now parsed as ``0``, an integer number.
+#. ``Random[]`` works now.
 #. ``RandomSample`` with one list argument now returns a random ordering of the list items. Previously it would return just one item.
 #. Origin placement corrected on ``ListPlot`` and ``LinePlot``.
 #. Fix long-standing bugs in Image handling
 #. Units and Quantities were sometimes failing. Also they were omitted from documentation.
+
+PyPI Package requirements
++++++++++++++++++++++++++
+
+Mathics3 now aims at a more richer set of functionality.
+
+Therefore NumPy and Pillow are required Python packages where they had
+been optional before.  In truth, probably running Mathics without one
+or both probably did not work well if it worked at all; we had not
+been testing setups that did not have NumPy.
 
 
 Enhancements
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -71,11 +71,22 @@ Bugs
...
 
+PyPI Package requirements
++++++++++++++++++++++++++
+
+Mathics3 now aims at a more richer set of functionality.
+
+Therefore NumPy and Pillow are required Python packages where they had
+been optional before.  In truth, probably running Mathics without one
+or both probably did not work well if it worked at all; we had not
+been testing setups that did not have NumPy.
```
mmatera, line 88: I guess NumPy is a requirement for SymPy

mmatera (approved):
LGTM

CHANGES.rst
rocky: https://libraries.io/pypi/sympy/1.10.1/tree says that only mpmath is required.  However looking at the code I see lots of numpy imports for a lot of functions. I wonder if this too is not tested that well. (Or if so not very usable without NumPy). 

scikit-image requires numpy. 
pillow  like sympy doesn't per se, but probably is not very usable without numpy. 

If Mathics3 was working well  without NumPy, NumPy detection was done cleanly, and the code was working fast without it then this wouldn't be an issue. But currently we need the speed and functionality of Numpy and what it provides to Pillow, SymPy, and Scikit-image.  In the future I am hoping that our large numeric data and multi-dimensional arrays will get stored there as well since numpy is performant while just about anything else isn't.

And NumPy is a well-established package in the Python ecosystem.

rocky:
There is also one other thing I'd like to say about this . For my part, people who are looking to run an open-source WMA on a reduced memory and CPU system such as on Android, raspberry Pi or a mobile phone, should consider Symja. It is Java based and doesn't need NumPy or SymPy, or Pillow. When our resources are spread so thin, I think it a perfectly fine thing that if there is an open-source platform that has a base covered, suggest that and not try to duplicated effort. 

Again, our code has gotten a lot of complexity, especially in the graphics area, and possibly other places, when it tries write homegrown code (often not as good or as carefully) than what can be had if we just use some other package. 

The recent breakage with handling TIFF images was an example of that.

mmatera:
From my point of view, reducing dependencies was mainly about making the CI simpler. At some point, I would like to split builtins into separate external (pymathics) modules (like image, statistics, linear algebra and calculus) in a way to make them easier to maintain.
Then, `mathics-core` would contain just the basic API and infrastructure for these modules, which would depend on other python libraries. However, to reach that point, we would need to have a more stable API. In any case, numpy probably should be inside `mathics-core`.

PR #747, opened 2023-01-12
Improving grid

mmatera:
This PR improves a little bit the compatibility of `Grid`  (and `GridBox`) with WMA.   Now it supports as input not just rectangular lists (`MatrixQ`) but general list. Also, it follows the semantics of WL. See the examples in the doctest.


rocky:
LGTM

PR #748, opened 2023-01-13
Last of the apply->eval conversions

rocky:
Some type annoations for options as a dict were done here.


rocky:
Unless there are concerns otherwise, I'll merge this in a day or so.

mmatera:
I could not see this in detail right now, but it seems to be straightforward. From my side, merge it when you consider this is ready.

PR #749, opened 2023-01-13
Remove using "apply" in signature match ...

rocky:
"eval" *must* be used now.

I will be making a pass over pymathics modules to adjust those that aren't done already


rocky:
Unless there are concerns, I'll merge this in a day or so after merging apply-reduction-7

mmatera:
Ok, but also check the contribute method in `mathics.builtin.base.Builtin`. I am pretty sure that there is a block that looks for apply* methods

rocky:
Note that there are a number of functions that start with "apply". I am not against the use of the work "apply" in a method name, just when it is used _improperly_.  
 
A lot has been mentioned (rightly) about the inefficient code. It is impossible to start to fix this when we can't get our concepts straight.

mmatera:
> Note that there are a number of functions that start with "apply". I am not against the use of the word "apply" in a method name, just when it is used _improperly_.
> 
> A lot has been mentioned (rightly) about the inefficient code. It is impossible to start to fix this when we can't get our concepts straight.

No, I was talking about the lines that load the apply_ methods, but I guess it were already removed.

PR #751, opened 2023-01-13
Image2latex

mmatera:
Work in progress. The aim of this PRis to be able to show images in the PDF documentation.


setup.py
```diff
@@ -43,7 +43,7 @@
     "Mathics-Scanner >= 1.3.0.dev0",
     # Pillow 9.1.0 supports BigTIFF with big-endian byte order.
     # ExampleData image hedy.tif is in this format.
-    "pillow >= 9.1.0",
+    "pillow == 9.2.0",
```
rocky, line 46: Why is this not ">= 9.2.0" ?

rocky:
There may be a conceptual problem here in any approach where any formatting routine writes in the name of a file: a the time that LaTeX the files in the filesystem may not exist.  In fact LaTeX may be run on a completely separate machine.

As a result, it is more reliable to have something closer to the LaTeX run create the files. In particular we use `doc2latex`  just before running LaTeX, and internally it uses  the `mathics.doc.latex_doc`. module. 

The second thing I'd like to mention regarding formatting is something I saw in producing SVGs. We wrap in 

```
<svg width= ... svg="http://www.w3.org/2000/svg"
                xmlns="http://www.w3.org/2000/svg"
                version="1.1"
                viewBox="%s">
```
and 

```
</svg>
```

in the output while Mathematica does not. And not surprisingly I think Mathematica is right and has thought about this possibly more than we have. 

The front-end should be adding that kind of text. We have the same `<asy>` wrapper for asymptote output. 

So corresponding thing here would be about adding `\includegraphics[width=3cm]{`. I am not totally sure though.

mathics/builtin/box/image.py
```diff
@@ -27,4 +30,12 @@ def boxes_to_mathml(self, elements=None, **options):
         )
 
     def boxes_to_tex(self, elements=None, **options):
-        return "-Image-"
+        fp = tempfile.NamedTemporaryFile(delete=True, suffix=".png")
```
rocky, line 33: Consider using pillow routines for writing PNG.

setup.py
mmatera: This change do not belong here. It stay here by accident.

mmatera:
> There may be a conceptual problem here in any 

I am aware of that. The problem is that , as far as I know, LaTeX does not have a native way to embed raster images. Then, we have the option of inventing a new homebrew syntax, but I guess that you would agree about  that we want to avoid it.

mathics/builtin/box/image.py
mmatera: The conversion was done when Image is converted in ImageBox. imageBox stores a B64 encoded version of the PNG file. Here we just decode it and store it in a file.

rocky: If ImageBox can keep the pillow structure, that may be a win.  A more general problem we have is that in digesting things for M-Expressions we lose the efficient and sometimes more flexible properties of whatever the object was before.  And we spend a lot of time in conversion.

rocky:
> > There may be a conceptual problem here in any
> 
> I am aware of that. The problem is that , as far as I know, LaTeX does not have a native way to embed raster images. Then, we have the option of inventing a new homebrew syntax, but I guess that you would agree about that we want to avoid it.

What I have been finding so far, is the that later can delay stuff like rasterization or "materializing" an object the better of or more flexibly we can handle results.  I think there was a mistake made early on and propagated that had the idea that mathics-core will do everything. A front-end just gives it an Mathics expression and suggests some preferences and then  off mathics core goes and takes it from there.  No more intervention by the front end is needed.

This downsides we have seen is that mathics core becomes much larger - a huge monolith in fact, there is a lot of code that duplicates work better done in specialized libraries like Pillow or SymPy, but isn't as good, and isn't as well maintained. And then on top of that, everything is slow too, partly because of the conversions going on or because it is inefficiently represented and isn't as cleverly implemented. 

Instead, what seems like a good alternative, is to make sure the evaluation and formatting are very well separate phases. And then have the front-end involved in dictating the formatting based on the evaluation results looking at it from a high-level.

Right now the front-ends are a little messy at doing this, but I expect after we have better Boxing and Form processing, exposing Symbols better (which we've started), and going over the APIs between the front-end and mathics core, (e.g. this Result  class), we can do much better. 

It is okay to come up with a protocol like Graphics3D that supports the kind 3D Graphics that Mathics should support.  It would be nice if there is an existing 3D Graphics protocol that is powerful enough to express the kinds of things WMA supports. There might be. Ditto for 2D, and for images. I would imagine that something like Pillow can handle images. (But we would want some sort of JSON-friendly way to represent things when the front end is not Python-based).  But using our own solution wrapped around a base-64 encoded PNG as the intermediate image language is a bad idea. It  has already been tried in Mathics, and at least for me, doesn't work.

rocky:
It occurs to me that WMA probably as the same problem of how does it communicate an image over a wire protocol. So that should be considered too.

mmatera:
> It occurs to me that WMA probably as the same problem of how does it communicate an image over a wire protocol. So that should be considered too.

Actually, it does not seem to do it well. Just try ```Import["Hedy.tif"]//TeXForm``` in the WMA REPL

rocky:
> > It occurs to me that WMA probably as the same problem of how does it communicate an image over a wire protocol. So that should be considered too.
> 
> Actually, it does not seem to do it well. Just try `Import["Hedy.tif"]//TeXForm` in the WMA REPL

Actually, the box has been down the last couple of days. But I will try it when it is available. Ok, if there isn't good precedent to follow we will use our own.

mmatera:
I am going to be away from the computer until the next Tuesday. After that, I will try to reformulate this.

mmatera:
Now this works well enough to produce the documentation. I removed the code in `mathics.format.latex` in favor of the `boxes_to_*` methods. Also, I deferred the base64 encoding to these methods instead of the constructor of ImageBox, which now stores the Pillow encoded image.

The part of embedding the images in the latex code would come for another round because it requires some extra discussion. Some possibilities would be 

* https://ctan.org/tex-archive/macros/latex/contrib/inline-images
* https://tex.stackexchange.com/questions/208819/embedding-images-in-tex-file-as-base64-strings
* https://github.com/zerotoc/pdfinlimg

mathics/doc/latex_doc.py
```diff
@@ -79,8 +79,8 @@
     r"(?P<content>.*?)\\end\{(?P=tag)\}"
 )
 
-LATEX_TESTOUT_DELIM_RE = re.compile(r",")
-NUMBER_RE = re.compile(r"(\d*(?<!\.)\.\d+|\d+\.(?!\.)\d*|\d+)")
+LATEX_TESTOUT_DELIM_RE = re.compile(r", ")
+NUMBER_RE = re.compile(r"([ -])(\d*(?<!\.)\.\d+|\d+\.(?!\.)\d*|\d+)")
```
mmatera, line 83: This deserves a small comment: without this change,
```
\includegraphics[width=5cm,height=3cm]{/tmp/img.png}
```
is converted into
```
\includegraphics[width=\allowbreak{}5\allowbreak{}cm,\allowbreak{}height=\allowbreak{}3\allowbreak{}cm]{/tmp/img.png}
```
because the parameters are treated as numbers to be displayed.

Here I exploded the fact that positive numbers in the parameters do not have spaces in front of them, different from the real output text. In any case, this conversion code should be removed at some point, in favor of a more standard way to parse LaTeX/markdown.

mathics/builtin/box/image.py
mmatera: This is done now.

mathics/builtin/box/image.py
```diff
@@ -1,6 +1,25 @@
...
+import base64
+import tempfile
+from copy import deepcopy
+from io import BytesIO
+from typing import Tuple
+
 from mathics.builtin.box.expression import BoxExpression
+from mathics.eval.image import pixels_as_ubyte
+
+try:
```
rocky, line 12: No try please. PIL is now _required_ .

mathics/builtin/box/image.py
```diff
@@ -13,18 +32,80 @@ class ImageBox(BoxExpression):
     an Image object.
     """
 
+    def boxes_to_b64text(self, elements=None, **options):
```
rocky, line 35: Type annotation and a docstring would be nice here.

mathics/builtin/box/image.py
```diff
@@ -1,6 +1,25 @@
 # -*- coding: utf-8 -*-
 
+import base64
```
rocky, line 3: Without a docstring at the top, this doesn't appear in any printed documentation or in Django docs. If that is intended, then the homegrown tagging such as `ImageBox` (on line 26) is not used. 

We should decide way we want to go and either add a docstring at the top or remove the homegrown tagging.

mathics/builtin/box/image.py
```diff
@@ -13,18 +32,80 @@ class ImageBox(BoxExpression):
     an Image object.
     """
 
+    def boxes_to_b64text(self, elements=None, **options):
+        contents, size = self.boxes_to_png(elements, **options)
+        encoded = base64.b64encode(contents)
+        encoded = b"data:image/png;base64," + encoded
+        return encoded, size
+
+    def boxes_to_png(self, elements=None, **options) -> Tuple:
```
rocky, line 41: The return type is more than just a generic tuple it is something like `Tuple[<type of contents>, Tuple[int, int]]`

Please add and fill in "type of contents".

mathics/builtin/box/image.py
```diff
@@ -13,18 +32,80 @@ class ImageBox(BoxExpression):
...
-        if elements is None:
-            elements = self._elements
+        encoded, size = self.boxes_to_b64text(elements, **options)
         # see https://tools.ietf.org/html/rfc2397
-        return '<mglyph src="%s" width="%dpx" height="%dpx" />' % (
-            elements[0].get_string_value(),
-            elements[1].get_int_value(),
-            elements[2].get_int_value(),
-        )
+        return '<mglyph src="%s" width="%dpx" height="%dpx" />' % (encoded, *size)
```
rocky, line 93: There is a little bit of magic here going on knowing that size has two items that exactly fills the two slots. (Automated checkers are probably not going to be able to figure this out, especially when the return type was that Tuple and one of the components of that was _also_ a tuple with two items).

Better would be to use an f-string and be explicit about the use of size: 

```python
return f'<mglyph src="{encoded}" width="{size[0]}px" height="{size[1]}px" />' 
```

mathics/doc/latex_doc.py
rocky: Thanks for the information. If it deserves a comment in a PR commit, it deserves a comment in the code.  Thanks.

rocky:
> Now this works well enough to produce the documentation. 

A big thanks!  I just tried this on the LaTeX documentation and it looks great! 

I am concerned about the proliferation of not-well-thought-out code do we do have to follow up on this after release and address this in a less hacky and more robust way.

mmatera:
> > Now this works well enough to produce the documentation.
> 
> A big thanks! I just tried this on the LaTeX documentation and it looks great!
> 
> I am concerned about the proliferation of not-well-thought-out code do we do have to follow up on this after release and address this in a less hacky and more robust way.

I guess that the next step would be to use markdown instead of LaTeX for building the documentation. This would allow to avoid most of the regexp hacks. Then, we could use pandoc to build the PDF.

rocky:
> The part of embedding the images in the latex code would come for another round because it requires some extra discussion. Some possibilities would be
> 
> * https://ctan.org/tex-archive/macros/latex/contrib/inline-images
> * https://tex.stackexchange.com/questions/208819/embedding-images-in-tex-file-as-base64-strings
> * https://github.com/zerotoc/pdfinlimg

Of these, first option https://ctan.org/tex-archive/macros/latex/contrib/inline-images seems like the most established and is likely to be the most robust. 

The front end should have a way to indicate whether the result should be put in a file or put inline. Mathicsscript for example decide in certain situations that having this in the filesystem is better than an inline PDF.

rocky:
> > > Now this works well enough to produce the documentation.
> > 
> > 
> > A big thanks! I just tried this on the LaTeX documentation and it looks great!
> > I am concerned about the proliferation of not-well-thought-out code do we do have to follow up on this after release and address this in a less hacky and more robust way.
> 
> I guess that the next step would be to use markdown instead of LaTeX for building the documentation. This would allow to avoid most of the regexp hacks. Then, we could use pandoc to build the PDF.

Sphinx has a markdown plugin. But it also supports RsT as well. I don't think we'll need to use pandoc directly. But this stuff is so far out on the horizon that it doesn't make sense to me to speculate. Who know, by the time we get around to it, things will have improved or new ways added. 

This I think has happened with PIL, some of the cool features that fix some of the problems we have such as Big-Endian Big TIFFs  (and whatever it was that broke MadTeaParty.jpg) fall into this category.

mathics/doc/latex_doc.py
rocky: Looks much better - thanks!

rocky (approved):
LGTM - see small comment though.

PR #755, opened 2023-01-15
Little things for improving document pdf  generation and appearance

rocky:
* latex *.pdf: add logos that aren't autobuild and require inkscape
* .gitignore: classes.pdf does get built, so ignore that when for when it automatically appears
* doc2latex.py: correct the version-info.tex location
* mathics.tex: Add TOC entries for Index and Colophon


TiagoCavalcante (approved):
LGTM

mmatera (approved):
LGTM

PR #757, opened 2023-01-18
Remove Doc Boxing (for now) and replace some autogenerated summaries

rocky:
In #751 I raised the question about about the homegrown docstring tagging for boxing that we have. 

There <strike>seems</strike> seemed to be a tug to go in the direction for including the built-in Box routines  like `GraphicsBox` in the documentation, and I think that is reasonable and in some ways preferrable. It is honest: a user _can_ see and call these so the user may as well know what they are.

The problem though is in the code because we hide this at the subsection level and expose it at the section level.

And the information in  the docstrings are lacking too.  For example for a GraphicsBox, we should spell out what the parameters are and give some indication of what options are available, and some indication about what to expect on the output. 

This starts to get this aspect under control or at least tries to regularize it more.

However, for now, we will remove Box documentation until it is gone over more thoroughly, is more consistent, and useful.

Please feel free to commit on top of this.

There is a corresponding branch in Django for the corresponding doc processing changes.


mathics/builtin/box/compilation.py
```diff
@@ -1,4 +1,7 @@
 # -*- coding: utf-8 -*-
+"""
+Boxing Routines for compiled code
+"""
```
rocky, line 4: The final period is left off since that seems what is done elsewhere.

mathics/builtin/box/graphics.py
```diff
@@ -361,24 +361,24 @@ class CircleBox(_ArcBox):
...
     <dl>
       <dt>'CircleBox[...]'
-      <dd>box structure for a 'Circle' element.
+      <dd>box a 'Circle' element.
     </dl>
     """
 
     face_element = False
-    summary_text = "internal box representation for 'Circle' elements"
+    summary_text = "box a 'Circle' element"
```
rocky, line 369: summary text start out with an active verb.

rocky:
I have been looking at this aspect in more detail and apparently something like `ArrowBox[]` is not invoked or evaluated like `Arrow`. In other words while it is a Symbol, it is not a built-in function.
 
So it is wrong then to have help that suggests it can be called like this which is what we seem to indicate in our help.

mathics/builtin/box/compilation.py
```diff
@@ -1,21 +1,20 @@
...
 
 from mathics.builtin.box.expression import BoxExpression
 
 
 class CompiledCodeBox(BoxExpression):
     """
     <dl>
-      <dt>'CompiledCodeBox[...]'
-      <dd> holds the compiled code generated by 'Compile'.
+      <dt>'CompiledCodeBox'
```
rocky, line 12: We are documenting a symbol here. This is not a built-in function.

mathics/builtin/drawing/graphics_internals.py
```diff
@@ -15,30 +15,6 @@
 
 
 class _GraphicsDirective(BuiltinElement):
-    def __new__(cls, *args, **kwargs):
-        # This ensures that all the graphics directive have a well formatted docstring
```
rocky, line 19: Since we don't cache Graphics object this also ensures overhead on every object creation. 

To paraphrase our former President Obama: I am not against automation, I am against _dumb_ automation. 

The default information here is lacking. And once we get to the point where we are trying to adjust articles in text and that sort of thing in a homegrown manner we've probably gone a little bit too far.

mathics/builtin/box/compilation.py
rocky: The disappearing period in formatting is a separate issue that needs addressing.

mathics/builtin/box/graphics.py
```diff
@@ -1121,6 +1123,9 @@ def extent(self):
 
 
 class PolygonBox(_Polyline):
+    # We have no documentation for this (yet).
+    no_doc = True
```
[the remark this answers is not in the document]
mmatera: Then ```nodoc``` must be set explicitly to ```False``` in the subclasses if we want that them contribute to the documentation, isn't it?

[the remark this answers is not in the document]
rocky: Right now `no_doc` when omitted is taken to mean `True` on those places where documentation is expected (builtin modules and builtin functions). 

If we were to specify this on everything, I probably would have dropped the `no` at the front and called it `add_doc` or something like that.  I am open to how you want to indicate this.  

The part I have a revulsion to is the behavior in the document generation where the name ends in "Box".  It is going to be hard for someone who doesn't understand the code to know to look there. I don't know why but that kind of thing feels like bad design. Whereas adding a `no_doc = False` in a base class for Built-ins as way to ensure every class has a `no_doc` class variable doesn't feel hacky.

mathics/builtin/box/graphics.py
rocky: That is for builtin _functions_.  for Symbols, this doesn't work.

mathics/doc/common_doc.py
```diff
@@ -823,6 +817,9 @@ def __init__(self, want_sorting=False):
                             if not builtin.__class__.__name__.endswith("Box")
                         ]
                         for instance in subsections:
+                            if hasattr(instance, "no_doc") and instance.no_doc:
```
rocky, line 820: This code can be simplified if we ensure a "no_doc" or better, "add_doc"  (or suggest something without "no_") , is added to a base Builtin class somewhere. 

Let me know your thoughts.

mathics/builtin/drawing/graphics_internals.py
mmatera: The goal of producing automatic documentation text was to establish certain standard form of the documentation, that we coluld check automatically. Eventually, if we converge to a criteria, then we could copy and paste the documentation as "static" doctrings.

rocky: The question I have right now (which was the same as before): do we want to do the work to more properly document the Boxing symbols before release? Or remove all mention boxes in docs?  For me, doing a pretty sloppy and not-very-useful job which is the current state in master, is not an acceptable option.

mmatera: Regarding the problem of building the documentation several times, we could avoid it adding the condition

```
if kwargs.get("expression", False) is False:
```
and then the documentation would be build just when the Definitions object is populated.

Regarding if it worth to write the documentation, I do not know. I guess that for the release, we could comment out the autogenerated documentation and set "no_doc=True" for these classes.

rocky: > Regarding the problem of building the documentation several times, we could avoid it adding the condition
> 
> ```
> if kwargs.get("expression", False) is False:
> ```
> 
> and then the documentation would be build just when the Definitions object is populated.

It feels like another hack on top of a bunch of existing hacks. Please no. 

> 
> Regarding if it worth to write the documentation, I do not know. I guess that for the release, we could comment out the autogenerated documentation and set "no_doc=True" for these classes.

This weekend I will remove mentions of Box symbols and put that on top if this PR.

rocky:
@mmatera @TiagoCavalcante  Unless dissent otherwise, I will merge this in about a day.

My plans for the weekend.  The top-level goal is to get yet another docker release out by the end of the week. This one will have the embedded images for the first time - thank you mmatera. 

Currently Django display of images is broken, so that will get fixed beforehand. I will also reduces the builtin functions in the grab-bag miscellaneous List function section. (Some of these have nothing to do with lists). 

The docker image will also have some of plots from the loaded workspaces fixed because we fix axis placement a while back (the plots though in workspaces just have the old less-good rendering). 

After the docker update (if I get all of the misc section removed), I think the only thing that remains to do again go over docs for everything and then mathics-core and scanner and rudimentary front-ends can go on. The front-ends however all could use work to make better use of the API changes. 

In sum we are then maybe a week or two (let's say two) weeks from finally getting to a Mathics3 version 6.0.0 release.

PR #758, opened 2023-01-21
Moving builtins to message

mmatera:
Just moving builtins in https://reference.wolfram.com/language/guide/Messages.html to mathics.builtin.messages and sorting the classes


rocky:
LGTM - thanks!

PR #761, opened 2023-01-22
Add setuptools to install requirements

davide125:
mathics/settings.py imports and uses `pkg_resources`, which comes from setuptools. While this is normally installed, it might not be present in very minimal deployments, so it's good to list it explicitly.


rocky (approved):
LGTM

PR #763, opened 2023-01-22
Move Rectangular fns into statistics...

rocky:
* Also add no_docs for base classes.
* Remove leading underscore where it isn't right
* More image evaluation function into 'mathics.eval.image'
* EdgeDetection does not need matplotlib and networkx.


TiagoCavalcante (approved):
LGTM

PR #764, opened 2023-01-22
Set requires in classes, rather than in Helper functions; adjust `requirements-full.txt`

rocky:
Set requires in builtins; adjust `requirements-full.txt`.
    
 `_SkimageRequires` and `_ImageTest` set only some requires and that wasn't working.
    
But more fundamentally we shouldn't use class inheritance for just setting a class variable.
    
Instead, set the class variable inside the class. This is more direct and clear.
    
Add additional PyPI packages that are used in Manipulate and Transliterate Set requires in builtins; adjust requirements-full.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,8 +3,12 @@
 CHANGES
 =======
 
-5.0.3dev0
+6.0.0dev0
 ---------
+
+A fair bit of code refactoring has gone on so that we might be able to scale the code, get it to be more performant, and more in line with other interpreters.
+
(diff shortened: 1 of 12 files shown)


rocky:
@mmatera @TiagoCavalcante because I'd like to get another docker release out today, I am working an a somewhat abbreviated PR fashion. 

This last commit on this PR which avoids infinite recursion Expression.SameQ (and also simplifies it, and probably speeds it up too), should have been in separate PR. 

As always feel free to review the PR and make comments and suggestions and they will be addressed in another PR.

PR #765, opened 2023-01-23
Changes needed to get Pymathics working again

rocky:
@mmatera These are changes I needed to make to get "pymathics.graph" and "pymathics.hello" to work.

Please review for reasonableness. 

To check these use the mathics-6.0.0 branches of each of those repositories.

### DIFF
--- a/mathics/builtin/__init__.py
+++ b/mathics/builtin/__init__.py
@@ -148,15 +148,18 @@ def name_is_builtin_symbol(module, name: str) -> Optional[type]:
     # via the inspect module which is not implementation specific.
 
     # Skip those builtins defined in or imported from another module.
-    if module_object.__module__ != module.__name__:
-        return None
+    # if module_object.__module__ != module.__name__:
+    #     return None
 
     # Skip objects in module mathics.builtin.base.
     if module_object.__module__ == "mathics.builtin.base":
         return None
 
     # Skip those builtins that are not submodules of mathics.builtin.
-    if not module_object.__module__.startswith("mathics.builtin."):
+    if not (
+        module_object.__module__.startswith("mathics.builtin.")
+        or module_object.__module__.startswith("pymathics.")
+    ):
         return None
 
     # If it is not a subclass of Builtin, skip it.
--- a/mathics/core/pymathics.py
+++ b/mathics/core/pymathics.py
@@ -4,12 +4,40 @@ Pymathics module handling
 """
 
 import importlib
+import inspect
 import sys
 
+from mathics.builtin import name_is_builtin_symbol
+from mathics.builtin.base import Builtin
 from mathics.core.evaluation import Evaluation
 
-# This dict probably does not belong here.
-pymathics = {}
+pymathics_modules = set()
+pymathics_builtins_by_module = {}
+
+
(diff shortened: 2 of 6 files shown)


rocky:
I should mention that I was able to get out a docker update. Things are ever closer, although as we see with PyMathics there is still more things to do.

mathics/builtin/__init__.py
```diff
@@ -148,15 +148,18 @@ def name_is_builtin_symbol(module, name: str) -> Optional[type]:
     # via the inspect module which is not implementation specific.
 
     # Skip those builtins defined in or imported from another module.
-    if module_object.__module__ != module.__name__:
-        return None
+    # if module_object.__module__ != module.__name__:
+    #     return None
```
rocky, line 152: It seems that the check was added for a consistency check, and it is now preventing pymathics routines from loading.

I don't think we should be checking names. What exactly does this prevent?

rocky:
@mmatera Even though this isn't finished, I would like to get this into master relatively soon. It blocks other work from easily being worked on. 

What will remain for another PR is to adjust `mathics.doc.latex_doc` along the lines of https://github.com/Mathics3/mathics-django/pull/181/files#diff-0dac66e07d151e9c5a5d8c77277b0091fdec920550348aa08131b55f088a5592

Right now with the the 6.0.0 branch of pythmathics.graph and https://github.com/Mathics3/mathics-django/pull/181 and this branch, Pymathics documentation renders in Django for those PyMathics3 modules that have been loaded.

However pymathics.graph works fine in the current master for mathicsscript. 

There still is a lot of work everywhere, but most of it is straightforward work. 

Some of the remaining things elsewhere: 

* Reorganize code into sections like we do for builtins. 
* Go over docs in pymathics.graph and pymathics.natlang
* Add doctests for pymathics modules - sort of along the lines of what we do for Mathics-Django. However testing should only be done of the Pymathics3 module that is getting added
* Add pytests. I have started some work here in pymathics.graph.
* Add

rocky:
Lastly, let me explain why I am holding my nose from the stench here and working on this. 

1. In going over pymathics.graph having docs has had a couple of benefits. First it suggests how to organize the modules better. We could do this anyway without the doc system, but having it working does make it easy to check things over for style and consistency and, as I said, organization which include module organization. 
2. It occurs to be that PyMathics modules make a perfect testbed for how to migrate docs. Compared to core there are small and isolated.
3. It makes interactive testing via Django easy (even though the tests involving graphs don't work.)
4. For pymathics.graph there is a massive amount of code there and it makes it much easier for people to know what is there
5. I don't think we've ever had a release with PyMathics3 modules documented.

It may be that we release with somewhat incomplete docs for pymathics.graph and pymathics.natlang, but I'd like the changes to core and django to be in place so that after release we can improve these.

mmatera:
@rocky, the changes look good. I just didn't have time to check if it works with the existing modules. So, if you think it is ready, let's merge and iterate.

rocky:
@mmatera - thanks. Everything is not fully done. However it _is_ kind of cool or helpful. 

There is still a lot of simple straightforward stuff to do like we've seen in the past say for the sorting out the "miscellaneous list" functions.

PR #766, opened 2023-01-27
Arithmetic refactor

mmatera:
Well, it took a lot of time, but I think this is now ready for revision. The idea is to move the code associate to the evaluation of arithmetic expressions, especially those expressions composed mainly of numbers and arithmetic operations, to a dedicated module in `mathics.core.eval`. 

There are several motivations to do this. In the first place, the code in master is quite involved, and not very well documented. Also, implies conversions back and forward to SymPy. Moreover, the way in which sympy evals arithmetic expressions is not fully compatible with the expected behavior in WMA. Then, more conversions and hacking over the returned code were needed. 
With the code in this branch, I hope the logic is a little bit clearer, and several useless conversions are avoided. Also, the issue with ```DirectedInfinity``` that prevents the application of some rules associated with `Gudermannian` and other complex functions should be fixed now.


mathics/eval/arithmetic.py
```diff
@@ -0,0 +1,812 @@
...
+
+NUMERICAL_CONSTANTS = {
+    SymbolE: 2.718281828,
+    SymbolEulerGamma: 0.5772156649,
+    SymbolKhinchin: 2.685452001,
+    SymbolPi: 3.141592654,
+}
+
+
+# Here I used the convention of calling eval_* to functions that can produce a new expression, or None
```
mmatera, line 61: @rocky, I leave this long comment here about the eval<-> apply semantics. Maybe we want to go over this again before merging this.

mathics/builtin/numbers/hyperbolic.py
```diff
@@ -325,11 +325,11 @@ class Gudermannian(Builtin):
         "Gudermannian[0]": "0",
         "Gudermannian[2*Pi*I]": "0",
         # FIXME: handling DirectedInfinity[-I] leads to problems
-        # "Gudermannian[6/4 Pi I]": "DirectedInfinity[-I]",
+        "Gudermannian[6/4 Pi I]": "DirectedInfinity[-I]",
         # Handled already
-        # "Gudermannian[Infinity]": "Pi/2",
+        "Gudermannian[DirectedInfinity[1]]": "Pi/2",
```
mmatera, line 330: Notice that the rule requires to use `DirectedInfinity[1]` instead of `Infinity`. This is due to `Infinity` is evaluated to `DirectedInfinity[1]` before looking at the `Gudermannian`  set of rules.

rocky:
Thanks for looking untangling this.  With regards to: 

>  Moreover, the way in which sympy evals arithmetic expressions is not fully compatible with the expected behavior in WMA. 

I would like to understand this in more detail.  Is this documented somewhere? (I am currently looking myself). 

What exactly are the differences?

rocky:
With respect to

> conversions back and forward to SymPy. 

I would like more details here too. Obviously we want to reduce this, but generally the ideal flow is convert into SymPy as much as possible and convert back to Mathics as late or as little as possible. In fact this idea is true for other things as well, like numpy arrays, constant literals that map into Python, etc.

mathics/eval/arithmetic.py
```diff
@@ -0,0 +1,812 @@
...
+    evaluates Sign[expr]
+    """
+    if isinstance(expr, Atom):
+        return eval_sign_number(expr)
+    if expr.has_form("Power", 2):
+        base, exp = expr.elements
+        if exp.is_zero:
+            return Integer1
+        if isinstance(exp, (Integer, Real, Rational)):
+            sign = eval_sign(base) or Expression(SymbolSign, base)
```
rocky, line 495: My Editor tells and flake8 tell me SymbolSign is not defined.

mathics/eval/arithmetic.py
```diff
@@ -0,0 +1,812 @@
...
+        if test_arithmetic_expr(exp):
+            sign = eval_sign(base) or Expression(SymbolSign, base)
+            return Expression(SymbolPower, sign, exp)
+        return None
+    if expr.has_form("Exp", 1):
+        exp = expr.elements[0]
+        if isinstance(exp, (Integer, Real, Rational)):
+            return Integer1
+        if isinstance(exp, Complex):
+            return Expression(SymbolExp, exp.imag)
```
rocky, line 509: My editor and flake8 tell me that SymbolExp is not defined.

mathics/eval/arithmetic.py
```diff
@@ -0,0 +1,812 @@
...
+    SymbolTimes,
+)
+from mathics.eval.numbers import eval_accuracy
+
+RationalMOneHalf = Rational(-1, 2)
+RealM0p5 = Real(-0.5)
+RealOne = Real(1.0)
+
+
+# Here we store numerical constants with their approximate
```
rocky, line 50: This smells like a wrong thing to do. I see it is used in `test_arithmetic_expr` so that is probably not advisable either. What is the history behind these?

mmatera:
> Thanks for looking untangling this. With regards to:
> 
> > Moreover, the way in which sympy evals arithmetic expressions is not fully compatible with the expected behavior in WMA.
> 
> I would like to understand this in more detail. Is this documented somewhere? (I am currently looking myself).
> 
> What exactly are the differences?

The first difference that I noticed is that SymPy rationalizes fractions. For example, `(1/3)^(1/2)` goes to `Sqrt[3]/3`, but in WMA, the result is `1/Sqrt[3]`.  Also, there are other conversions, like what happens with Complex numbers, which in SymPy are expressions, not atoms, and how Infinity quantities are handled. In WMA, `Infinity` and `ComplexInfinity` are evaluated as `DirectedInfinity[1]` and `DirectedInfinity[]`.

mathics/eval/arithmetic.py
mmatera: In a couple of places in the code, I need to multiply by that constant. The same with `RationalMOneHalf`. It is just to about avoiding calling the constructor to get the instance.

mathics/eval/arithmetic.py
rocky: I looked at this and there is information here that I a do not believe is elsewhere. I think it should be elsewhere but I don't know right now _where_ that elsewhere is.

mathics/eval/arithmetic.py
mmatera: > With respect to
> 
> > conversions back and forward to SymPy.
> 
> I would like more details here too. Obviously we want to reduce this, but generally the ideal flow is convert into SymPy as much as possible and convert back to Mathics as late or as little as possible. In fact this idea is true for other things as well, like numpy arrays, constant literals that map into Python, etc.

Suppose you are going to evaluate

```
3^(1/2) +  5 * 3^(-1/2) + 3 * Pi^2 
```
`1/2` is converted to sympy twice back and forward, to see that the expression didn't changed, so `1/2` is kept. Then  we have the evaluation of `3^(-1/2)` and `3^(-1/2)` where `3` and `1/2`, and `3^(1/2)` are converted again back and forward, to again get the same result. And again in the multiplication by 5, and again in the addition, to get at the end the same expression.

mathics/eval/arithmetic.py
mmatera: Thanks for catching this.

mmatera: I have to add more tests for these routines, but at this point, I need a little bit of feedback (not necessarily right now, but before continuing with this).

mathics/eval/arithmetic.py
rocky: All of this can be addressed by having a special form of Expression that is known to have a SymPy equivalent and that is saved.  We will probably do the same for literals that have exact Python correspondences, and numerous other things. 

If this is the only thing that being addressed, please don't do it this way.

mathics/eval/arithmetic.py
rocky: I didn't mean RealOne, I was referring to SymolE, SymbolEulerGamma, and SymbolPi.

rocky: BTW Computing RealOne once and using it many times is fine. And if we find that other modules tend to need RealOne, we can put it in a more global place.

mathics/eval/arithmetic.py
mmatera: The problem is that for certain expressions, the replacement rules in WMA and SymPy are different, which makes that new  rules written in WL fails to be applied, like in the case of DirectedInfinity. Actually, I did all of this because it was not possible to fix these rules without disentangling all the arithmetic module.

mathics/eval/arithmetic.py
```diff
@@ -0,0 +1,812 @@
...
+
+    if len(powers) == 0:
+        return base
+    elif len(powers) == 1:
+        return Expression(SymbolPower, base, powers[0])
+    result = Expression(SymbolPower, base, Expression(SymbolTimes, *powers))
+    return result
+
+
+def distribute_factor(expr: BaseElement, factor: BaseElement) -> BaseElement:
```
rocky, line 141: I am not seeing `distribute_factor` or `distribute_powers` used anywhere. Where is this used?

mathics/eval/arithmetic.py
rocky: I think here we need more knowledgeable conversion routines between the Mathics and SymPy representations.

rocky:
> > Thanks for looking untangling this. With regards to:
> > > Moreover, the way in which sympy evals arithmetic expressions is not fully compatible with the expected behavior in WMA.
> > 
> > 
> > I would like to understand this in more detail. Is this documented somewhere? (I am currently looking myself).
> > What exactly are the differences?
> 
> The first difference that I noticed is that SymPy rationalizes fractions. For example, `(1/3)^(1/2)` goes to `Sqrt[3]/3`, but in WMA, the result is `1/Sqrt[3]`. Also, there are other conversions, like what happens with Complex numbers, which in SymPy are expressions, not atoms, and how Infinity quantities are handled. In WMA, `Infinity` and `ComplexInfinity` are evaluated as `DirectedInfinity[1]` and `DirectedInfinity[]`.

This was said above but I think it applies here as well.  The real issue is about whether there is loss of information in either representation. If not, then I would _first_ to get this by doing a better conversion.  In a sense this is also related to the the Form stuff as well. 

It may turn out that this is too tricky, but I strongly encourage starting out with that first (and don't ask me for the reasons why unless you are prepared for the consequences of the answer).

rocky:
> With respect to
> 
> > conversions back and forward to SymPy.
> 
> I would like more details here too. Obviously we want to reduce this, but generally the ideal flow is convert into SymPy as much as possible and convert back to Mathics as late or as little as possible. In fact this idea is true for other things as well, like numpy arrays, constant literals that map into Python, etc.

So based on what has followed, I have to back of here, but only a little. You know how we have this overly complex caching mechanism for Expressions? Well, we might have something like that for Expressions with sympy representations. However the invalidation is done only when a rewrite rule kicks in.

mathics/builtin/numbers/hyperbolic.py
rocky: Ok - Thanks for the explanation.  I'd like to understand this more though. What is it that turns `Infinity` into `DirectedInfinity[1]` ? Is this SymPy doing this?

mmatera:
> 

Let me put this in a different way. I am pretty sure that your idea of implementing specialized expressions that stores the SymPy representation would provide a better implementation over what I have write here. However, even with that implementation, I think that it would be better to split the evaluation of arithmetic expressions in a part that could depend on the context (provided by the Evaluation object) and a part that is independent of the context. This is what this or is about. Then you could replace part of the implementation with proper calls to SymPy, and use a smarter  class hierarchy, but still I think we want to have a way to do arithmetic without passing throw the standard evaluation process.

mathics/builtin/numbers/hyperbolic.py
mmatera: We should if we want to get the same result if we apply the same rules as written for WMA.

rocky:
> I think that it would be better to split the evaluation of arithmetic expressions in a part that could depend on the context (provided by the Evaluation object) and a part that is independent of the context. 

I totally agree. And then this can be much much smaller and simpler. 

The other change for now that makes sense is changing the rule with `Gudermannian[Infinity]` to `Gudermannian[DirectedInfinity[1]]`. 

Doing both of these better is something we should keep in mind and do when the opportunity presents itself.

mathics/builtin/numbers/hyperbolic.py
rocky: Ok. If this is simple for now to do, then let's do. Or we can write `DirectedInfinity[1]` in the rules and come back to this some other time. Your choice - assuming this is a simple fix.

mathics/eval/arithmetic.py
mmatera: > Guder

That part of the PR are in this branch: https://github.com/Mathics3/mathics-core/tree/tests_from_arithmetic_refactor

However, the new tests do not pass without the other changes.

rocky: A number of them do, so those could co in. The ones that don't can be commented out to remind us to address various issues.

rocky: @mmatera if you would like me to make a stab and extracting the non-controversial parts, I'll do.

mmatera:
@rocky, I was trying to merge this with the current state of master, but I found a lot of conflicts. My plan is to take the tests added here, and move the changes to another new branch, and then, take the part of the code needed to pass the tests. It will take a while.

mmatera:
I completed the rebase, but for some reason, now the performance is lower than in master. Also, I found some extra bugs in non directly related code.  I am going to mark them, then propose fixing for them, and then continue over this.

rocky:
> I completed the rebase, but for some reason, now the performance is lower than in master. Also, I found some extra bugs in non directly related code. I am going to mark them, then propose fixing for them, and then continue over this.

Small PRs are always appreciated over large ones like this.

mmatera:
@rocky, yes, I am slowly splitting this  up.

mathics/builtin/arithfns/basic.py
```diff
@@ -528,15 +538,15 @@ class Power(BinaryOperator, _MPMathFunction):
     # Remember to up sympy doc link when this is corrected
     sympy_name = "Pow"
 
+    def eval_exp(self, x, evaluation):
+        "Power[E, x]"
+        return eval_exponential(x)
```
rocky, line 543: Shouldn't this be eval_Exponential() ? (And same for eval_power_*.

mmatera: yes, it should. However, what I did here was just to rebase this PR, in order to split this into smaller parts. As I will split the small pieces, I plan to fix these name conventions.

PR #767, opened 2023-01-28
Improve infinity and Precision

mmatera:
This would be the simplest part of #766, including tests and changes in `Infinity` and `Precision`. Some of the new tests are commented / marked as xfail because require the other changes.

### DIFF
--- a/mathics/builtin/numbers/hyperbolic.py
+++ b/mathics/builtin/numbers/hyperbolic.py
@@ -58,9 +58,9 @@ class ArcCosh(_MPMathFunction):
 
     rules = {
         "ArcCosh[Undefined]": "Undefined",
-        "ArcCosh[I Infinity]": "Infinity",
-        "ArcCosh[-I Infinity]": "Infinity",
-        "ArcCosh[ComplexInfinity]": "Infinity",
+        "ArcCosh[DirectedInfinity[I]]": "Infinity",
+        "ArcCosh[DirectedInfinity[-I]]": "Infinity",
+        "ArcCosh[DirectedInfinity[]]": "Infinity",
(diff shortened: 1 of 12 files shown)


mathics/builtin/numbers/hyperbolic.py
```diff
@@ -325,11 +325,11 @@ class Gudermannian(Builtin):
         "Gudermannian[0]": "0",
         "Gudermannian[2*Pi*I]": "0",
         # FIXME: handling DirectedInfinity[-I] leads to problems
-        # "Gudermannian[6/4 Pi I]": "DirectedInfinity[-I]",
+        "Gudermannian[3 I / 2 Pi]": "DirectedInfinity[-I]",
         # Handled already
```
rocky, line 329: This comment is not correct, right?

rocky:
LGTM - thanks for doing this. There is one small comment regarding possibly crud from before.

mathics/builtin/numbers/hyperbolic.py
mmatera: BTW, this is another reason to have a way to do arithmetic in an context-independent way: When rules are loaded, it makes a difference if we write ```Gudermannian[6 / 4 Pi I]-> repl``` and ```Gudermannian[3 I / 2 Pi]-> repl```: 
the first rule is never applied. Ideally, at loading time, the pattern should be brought to a canonical form. 

This could be achieved by evaluating the rule. However, in our current framework, it is not possible, because the evaluation requires definitions that might not already be loaded. But most of the needed changes are just arithmetic evaluations, which could be perfectly done without an Evaluation object.

mmatera: I removed the commented lines now. I am going to merge this when CI finishes.

PR #768, opened 2023-01-29
evaluation.message now returns a value - adjust

rocky:
This commit deserves a description of what was up.

In a previous PR, `evaluation.message()` returned `None` and that was changed to return the message it creates. So first, why this change?

`evaluation.message()` constructs a `Message()` object and then prints it.

But whatever care was made into deciding the structure of the `Message` is lost after printing.

A front end would like to be able to use this information too. The structure has for example:

* tagging message classes e.g. "syntxb", or "argmu", and
* Builtin/Variable type, e.g. "Syntax" or "Check"

A front end can use the Message object to know whether a result return represents an error or a valid result. And here in Django, the indirect method chosen was to see if the evaluation result is `None`. Note that comments also have an evaluation result that is `None`. (And there may be other things of that kind that return `None` for good reason).

We can and should do better than this.

Ok. So yes, we return the Message object now, but then a new misfeature kicks in. There were some places (but not all) in evaluation code where, *it was expected* that `evaluation.message()` return `None`!

Returning `None` in evaluation code signals not to do further evaluation.

So now one asks, well if `evaluation.message()` had always returned `None` unconditionally, why is `return evaluation.message()` used instead of `evaluation.message(); return`?

There are two reasons I can think of. It is slightly shorter (although in a number of cases, `evaluation.message()` falls to the end of the method, so here the `return` can be omitted and then *that* is in fact shorter).

The second possibility this was used (assuming this was a conscious decision) is that what was communicated was to say that whatever `evaluation.message()` returns, well, that's what I want to return. In other words, It might be that the idea was I don't want to understand what the right thing to do is, so I will "delegate" the decision to `evaluation.message()`.

There are a couple of problems with this.

First it locks in the wrong-minded behavior of `evaluation.message()`. Is it that evaluation message's return value is to _only_ be used to indicated what evaluation should do next? This a screwy thing for this thing to want to return. 

And it locks in the wrong-minded behavior that somehow mathics-core or parts of it know all and will do all, and the front-end then just pass in initially some form options it promises to handle and then after that for each evaluation afterwards evaluations are supposed to come back beautiful and perfect. But we have seen that this is impossible to do, it doesn't allow users to interactively decide what _they_ want and it creates convoluted behavior just to trying to get close to this misinformed goal. 

Second, returning `evaluation.message()` obscures in the eval code what is really going on. When you have an explicit `return` statement or even an implicit fall to the end, you know what is getting returned.

You don't have to know return the behavior of `evaluation.message()` to know what gets returned in an evaluation method.


rocky:
@mmatera and @TiagoCavalcante this really should have been two PRs instead of one. But in my desire to try to speed things along this is one PR. 

The comment above describes the bulk of the largish but pattern-oriented change of changing one statement into two. 

However the are _also_ some changes to `format/svg.py`. Some where clarifying docstrings and some Mathics->Mathics3 changes.

But there I made some formatting changes to the SVG output (not functional but just in terms of spacing to make it look nicer before base64 encoding). 

But then I ran into the fragility of `test/format/test_format.py` for MathML strings.

I made some small code changes to simplify this a little. But I also commented out the MathML base64-encoded strings for SVG that were failing. 

I don't want to get in the business of having the change base64-encoded strings. This kind of testing should be done some other way such as unit testing on the strings before base64-encoding. And even there, there should be flexibility in terms of ignoring white space.

rocky:
As before, as soon as this clears, I will be merging to master, but I'll take changes and address concerns on the other side of the merge.

mmatera (commented):
LGTM. Regarding the tests, if you add the comment "fragile!"  the test is automatically is tagged as xfail.

rocky:
Ok - but down the line we would like non-fragile tests. I was time-limited earlier. If you want to improve this according to the conventions set up, please go ahead.

PR #769, opened 2023-01-30
More compatibility improvements in Accuracy and Precision. Adding tests.

mmatera:
This PR is another fix coming from issues found when I was working on #766. In particular, the way in which Mathics handles Precision and Accuracy for real numbers with 0. nominal value. Also, a bug that prevents parsing Real numbers with fixed accuracy was fixed.


rocky:
The thrust or idea behind this is good. However I would like to spend some time understanding things in detail  and that may take a little while.

mathics/eval/numbers.py
```diff
@@ -1,10 +1,109 @@
...
 
+from mathics.core.atoms import Complex, MachineReal, PrecisionReal, Real
 from mathics.core.convert.sympy import from_sympy
+from mathics.core.element import BaseElement
 from mathics.core.expression import Expression
+from mathics.core.number import MACHINE_PRECISION_VALUE, ZERO_MACHINE_ACCURACY, dps
 from mathics.core.symbols import SymbolPlus
 
 
+def eval_accuracy(z: BaseElement) -> Optional[float]:
```
rocky, line 14: We seem to have started to have two conventions now for the naming of the `eval_` methods here. There are some that keep lower case like this here. And some that use the uppercase of the builtin name like `eval_N`, `eval_Plot`. 

We should decide on a convention one way or another and stick to it. I have a slight preference for using the uppercase letter because it makes the connection between the Mathics3 built-in name and the particular function. If it is a little unconventional from the normal Python naming, that might be a good thing in order to make the developer reading the code that this has a special function.

But more important is to be consistent. Your thoughts?

CHANGES.rst
```diff
@@ -92,9 +94,10 @@ Bugs
 
 #. Units and Quantities were sometimes failing. Also they were omitted from documentation.
 #. Better handling of ``Infinite`` quantities.
-#. Fix ``Precision`` compatibility with WMA.
+#. Fix ``Precision`` and ``Accuracy``compatibility with WMA. In particular, ``Precision[0.]`` and ``Accuracy[0.]``
+#. Accuracy in numbers usint the notation ``` n.nnn``acc ```  now is properly handled.
```
rocky, line 98: usint -> using

rocky: A small thing about using the word "Fix". Does it mean that all the problems with Precision have really been corrected or is it that the previous problems have been reduced, that is, that Precision has been _improved_? 

The section on "Precision" says

> This is rather a proof-of-concept than a full implementation

Is that still the case after this PR? 

The section on "Accuracy" says: 

> Notice that the value is not exactly equal to the [value] obtained in WMA

Is this still accurate?

mathics/builtin/atomic/numbers.py
```diff
@@ -183,11 +186,17 @@ class Accuracy(Builtin):
...
-     = 14.8861
+     = 15.8406
 
     'Accuracy' for the value 0 is a fixed-precision Real number:
      >> 0``2
       = 0.00
+     >> Accuracy[0.``2]
+      = 2.
+
+    For 0.`, the accuracy is the
```
rocky, line 197: is the -> is:

mathics/builtin/atomic/numbers.py
```diff
@@ -948,6 +934,23 @@ class Precision(Builtin):
...
+
+    >> (Accuracy[z] == Precision[z] + Log[z])/.z-> 37.`
+     = True
+
+    The case of `0.` values is special. Following WMA, in a Machine Real representation, the precision is
+    set to $MachinePrecision:
+    >> Precision[0.]
+     = MachinePrecision
+
+    On the other hand, for a Precision Real with fixed accuracy,
```
rocky, line 949: Add a backslash after "fixed accuracy,"

mathics/builtin/atomic/numbers.py
```diff
@@ -948,6 +934,23 @@ class Precision(Builtin):
...
 
+    For non-zero Real values, it holds in general
+
+    $$Accuracy[z] == Precision[z] + Log[z]$$
+
+    >> (Accuracy[z] == Precision[z] + Log[z])/.z-> 37.`
+     = True
+
+    The case of `0.` values is special. Following WMA, in a Machine Real representation, the precision is
+    set to $MachinePrecision:
```
rocky, line 945: $MachinePrecision -> '$MachinePrecision'

mathics/builtin/atomic/numbers.py
```diff
@@ -948,6 +934,23 @@ class Precision(Builtin):
     >> Precision[{{1, 1.`},{1.`5, 1.`10}}]
      = 5.
 
+    For non-zero Real values, it holds in general
```
rocky, line 937: general -> general:

mathics/builtin/numbers/constants.py
```diff
@@ -575,6 +581,59 @@ class Overflow(Builtin):
     summary_text = "overflow in numeric evaluation"
 
 
+class MaxMachineNumber(Predefined):
+    """
+    Largest normalizable machine number (<url>
```
rocky, line 586: I really like the summary title here - thanks!

mathics/core/number.py
```diff
@@ -2,28 +2,33 @@
...
+machine_digits = float_info.dig
 
-machine_epsilon = 2 ** (1 - machine_precision)
+MACHINE_EPSILON = float_info.epsilon
+MACHINE_PRECISION_VALUE = float_info.mant_dig / C
+MIN_MACHINE_NUMBER = float_info.min
+MAX_MACHINE_NUMBER = float_info.max
+ZERO_MACHINE_ACCURACY = -mpmath.log(MIN_MACHINE_NUMBER, 10.0) + MACHINE_PRECISION_VALUE
+
+# backward compatibility
```
rocky, line 30: Is this necessary? This release we are bumping the major number. If so, suggest possibly in the comment when we can remove this. (I would find it hard to believe that anyone other than ourselves is making use of this.)

mathics/eval/numbers.py
```diff
@@ -1,10 +1,109 @@
...
+    """
+    Determine the accuracy of an expression expr.
+    If z is a Real value, returns the difference between
```
rocky, line 17: add "\" at the end of lines mid-sentence. I would probably add one following the "."  on line 16 as well.

mathics/builtin/atomic/numbers.py
rocky: Please add "\" to the end of the line ("the precision is").

mathics/builtin/numbers/constants.py
```diff
@@ -575,6 +581,59 @@ class Overflow(Builtin):
...
+    https://reference.wolfram.com/language/ref/$MinMachineNumber.html
+    </url>)
+
+    <dl>
+      <dt>'$MinMachineNumber'
+      <dd>Represents the smallest positive number that can be represented \
+          as a normalized machine number in the system.
+    </dl>
+
+    MachinePrecision minus the Log base 10 of this number is the
```
rocky, line 623: MachinePrecision  -> 'MachinePrecision'

mathics/eval/numbers.py
```diff
@@ -1,10 +1,109 @@
...
+    the number of significant decimal figures (Precision) and
+    log_10(z).
+
+    For example,
+    ```
+    12.345`2
+    ```
+    which is equivalent to 12.`2  has an accuracy of
```
rocky, line 25: of -> of:

mathics/eval/numbers.py
```diff
@@ -1,10 +1,109 @@
...
+    ```
+    0.908509 == 2. -  log(10, 12.345)
+    ```
+
+    If the expression contains Real values, returns
+    the minimal accuracy of all the numbers in the expression.
+
+    Otherwise returns None, representing infinite accuracy.
+    """
+    if isinstance(z, MachineReal):
```
rocky, line 35: Add type annotation for return type?

test/builtin/atomic/test_numbers.py
```diff
@@ -177,43 +178,87 @@ def test_n():
...
-    for str_expr, str_expected in (
-        ("0`4", "0"),
-        ("Accuracy[0.0]", "15."),
-        ("Accuracy[0.000000000000000000000000000000000000]", "36."),
-        ("Accuracy[-0.0]", "15."),
-        # In WMA, this gives 36. Seems to be a rounding  issue
-        # ("Accuracy[-0.000000000000000000000000000000000000]", "36."),
-        ("1.0000000000000000 // Accuracy", "15."),
-        ("1.00000000000000000 // Accuracy", "17."),
+@pytest.mark.parametrize(
```
rocky, line 181: I think Accuracy is based on computer hardware characteristics, right? If so we should add a `pytest.skipif`  when those hardware characteristics are not met.

mathics/builtin/atomic/numbers.py
```diff
@@ -948,6 +934,23 @@ class Precision(Builtin):
...
+
+    $$Accuracy[z] == Precision[z] + Log[z]$$
```
rocky, line 939: This does not render right in Django. I don't think we handle `$$`.  It would be the more awkward: 

```
'Accuracy'[$z$] == Precision[$z$] + 'Log[$z$]'
```

Sigh. When we move to a sphinx or a real documentation system, it is going to be a major pain to go back and simplify all of our home-grown tagging.

mathics/builtin/numbers/constants.py
```diff
@@ -575,6 +581,59 @@ class Overflow(Builtin):
...
+    accuracy of 0`:
+    >> MachinePrecision -Log[10., $MinMachineNumber]==Accuracy[0`]
+     = True
+
+    """
+
+    name = "$MinMachineNumber"
+    summary_text = "smallest normalized positive machine number"
+
+    def evaluate(self, evaluation) -> MachineReal:
```
rocky, line 633: In new code I have been adding a type annotation for `evaluation`:  `evaluation: Evaluation`

CHANGES.rst
mmatera: In all the cases I have tested, I get the same result as in WMA (up to small numeric differences due to the mpmath implementation).

mathics/core/number.py
mmatera: I leave this commentary because that variable appears in several modules. I can make the change, but then this PR would become larger. Also, I wondered if it was worth making the change. Thoughts?

mathics/eval/numbers.py
mmatera: where?

test/builtin/atomic/test_numbers.py
mmatera: I used the native `sys.float_info`, so it would adjust to each architecture. If we discover one architecture where it fails, then we can handle it.

mathics/core/number.py
rocky: I think it worth making the change. If you want to do in another PR, that's okay.

mathics/eval/numbers.py
rocky: My mistake - I don't know why I thought this  a function definition.

test/builtin/atomic/test_numbers.py
rocky: Doesn't accuracy depend on MACHINE_PRECISION_VALUE and MIN_MACHINE_NUMBER? and those depend of mpmath values and you say that mpmath varies? 

I am okay with leaving out `pytest.mark.skipif`  _provided_ there is an assertion  on these underlying values. 

And then that way we will know that things have changed.

Recall we had a recent situation with reading bytes from a file that changed due to the endian-ness of the machine. 

I would feel differently if we were testing regularly using different CPU architectures and 32-bit machines versus 64-bit OS setups and so on. But we are not.

rocky: One other thing here. There should be unit test precisely for the underlying assumptions that are in effect.

One purpose of _unit_ tests is exactly this. (There are other test like functional and integration tests). Our tests tend to be on the heavier side - integration and functional rather than on the smaller lighter, more precise unit tests.

mathics/builtin/atomic/numbers.py
rocky: There are a number of changes that are marked "resolved" that I am not seeing. This is one.

mathics/builtin/numbers/constants.py
rocky: I am not seeing quotes around 'MachinePrecision'  on line 623. Maybe a commit that wasn't pushed?

mathics/eval/numbers.py
rocky: Ok - it looks like we are going with `eval_Accuracy`.  Thanks!  I think this is the right thing. 

I will resolve this but note there are other `eval_` functions that need adjusting: `evall_fulform_makeboxes`, `eval__makeboxes`.  and `eval_p_norm`. That can be done in another PR.

mathics/builtin/atomic/numbers.py
rocky: Colon at the end after "is" please.

rocky:
@mmatera Branch merged with master because I can't run Django using this and the current master.

mathics/builtin/atomic/numbers.py
```diff
@@ -934,19 +934,19 @@ class Precision(Builtin):
     >> Precision[{{1, 1.`},{1.`5, 1.`10}}]
      = 5.
 
-    For non-zero Real values, it holds in general
+    For non-zero Real values, it holds in general:
 
-    $$Accuracy[z] == Precision[z] + Log[z]$$
+    Accuracy[$z$] == Precision[$z$] + Log[$z$]
```
rocky, line 946: Single quotes around 'Accuracy', 'Precision', and 'Log' please.

mathics/builtin/atomic/numbers.py
```diff
@@ -183,11 +187,17 @@ class Accuracy(Builtin):
      = Infinity
 
     >> Accuracy[F[1.3, Pi, A]]
-     = 14.8861
+     = 15.8406
```
rocky, line 190: If this can vary, the please let's use '...' and make precise tests as pytests where we also have assertions or checks on the underlying factors that make this so.

rocky: As you know with the MLMath stuff, I am not a fan of fragile tests because they fail unexpectedly at a time when you really don't want to start dealing with the failure.

mathics/builtin/atomic/numbers.py
mmatera: Sorry, I tried to go one by one, correct, and push here before marking them. I am going over it again.

mathics/builtin/atomic/numbers.py
```diff
@@ -927,8 +914,8 @@ class Precision(Builtin):
       <dt>'Precision[$expr$]'
       <dd>examines the number of significant digits of $expr$.
     </dl>
-
-    <i>This is rather a proof-of-concept than a full implementation.</i>
+    <i>Notice that the result could be slighly different than the obtained
```
rocky, line 917: backslash after "obtained" please.  For myself I go into Django and _look_ at the rendered output. it can help catching things like this.

CHANGES.rst
mmatera: The differences I found are in this order:

```'Accuracy[0.32`30*^30]``` 

WMA:      0.4948500216800894
Mathics: 0.49485002168009373

For Complex, there could be more differences:

```Precision[12`20*^30+13`15*^24]```
WMA:       19.9553
Mathics:   20.

```Accuracy[12`20*^30+13`15*^24]```
WMA:       -11.1239
Mathics:  -11.0792

Probably in this case the normalization is different. In any case, differences are of at most ~O(1/10). So, all the currently detected issues are fixed. But OK, I can rephrase it.

File: mathics/builtin/atomic/numbers.py (hunk 17 above)
Comment (mmatera) [reply mathics/builtin/atomic/numbers.py:190]: In this case, the "fragility" on the test allows for checking the behavior. But maybe I could surround it with `Round`.

File: mathics/core/number.py (hunk 8 above)
Comment (mmatera) [reply mathics/core/number.py:30]: Then, I will do that in the following PR.

File: mathics/builtin/atomic/numbers.py (hunk 17 above)
Comment (rocky) [reply mathics/builtin/atomic/numbers.py:190]: You could, but this is the wrong philosophy: the _primary_ purpose of these tests is to enlighten users as to what this Builtin does; testing is secondary. So complicating an example to make testing is not the right philosophy.

If you want a _test_, the add this as a pytest where you can get as elaborate as you want, like within so many digits. But better here is to _also_ add checks for the fundamental properties that the this calculation is based on. That way when that assumption breaks for whatever reason, we have pinpointed at a low level what the problem is rather than have to recall or look around in the code what Accuracy assumes that has been violated.

File: test/builtin/atomic/test_numbers.py (hunk 13 above)
Comment (mmatera) [reply test/builtin/atomic/test_numbers.py:181]: > Doesn't accuracy depend on MACHINE_PRECISION_VALUE and MIN_MACHINE_NUMBER? and those depend of mpmath values and you say that mpmath varies?

`MACHINE_PRECISION_VALUE` and `MIN_MACHINE_NUMBER` depends on numbers coming from the `sys` module, so it would work in any platform that runs Python. What should we catch?



> 
> I am okay with leaving out `pytest.mark.skipif` _provided_ there is an assertion on these underlying values.
> 
> And then that way we will know that things have changed.
> 
> Recall we had a recent situation with reading bytes from a file that changed due to the endian-ness of the machine.
> 
> I would feel differently if we were testing regularly using different CPU architectures and 32-bit machines versus 64-bit OS setups and so on. But we are not.

Comment (rocky) [reply test/builtin/atomic/test_numbers.py:181]: I looked at the Python code, and most definitely these things  change per system - that is why it is in Python's `sys` module. In particular this seems to come from  the compiler's `<float.h>` values. 

I think we should have a function that is called by pytest.mark.skipif that returns False if any of MACHINE_EPSILON, MACHINE_PRECISION_VALUE, MIN_MACHINE_NUMBER, MAX_MACHINE_NUMBER are not the values that are assumed by the test.

Comment (mmatera) [reply test/builtin/atomic/test_numbers.py:181]: Ah! Now I understand. What I think would be better is to change the tests in a way that the connection with these numbers be reinforced.

Comment (rocky) [reply test/builtin/atomic/test_numbers.py:181]: Yes, that is even better. And if you are up for it,  go for it! 

Maybe you can find a 32-bit OS around and see what values you get or try a Python made from a 32-bit C compiler. Or if there are 128-bit architectures or C compiler and see what those do. 

There might be github workflows CI images that have these properties of 32-bitness or 32-bitness compilers (or other than 64-bitness) and we could run a test using that image.

Comment (rocky) [reply test/builtin/atomic/test_numbers.py:181]: And the same thing along the lines of different mpc versions.

File: CHANGES.rst (hunk 2 above)
Comment (rocky) [reply CHANGES.rst]: Thanks for performing the experiments. It is nice to know this. I think we should document this somewhere so we don't lose this useful information.

File: test/builtin/atomic/test_numbers.py (hunk 13 above)
Comment (rocky) [reply test/builtin/atomic/test_numbers.py:186]: It occurs to me that is you feel confident in knowing how what MACHINE_EPSILON, etc change things, then in a test you could artificially set these to different values and make the Precision and Accuracy tests. 

It would however be nice to know what alternate values these possibly can take on.

Comment (mmatera) [reply test/builtin/atomic/test_numbers.py:186]: I think that with the new changes, the tests are robust across different platforms. The only thing that could go wrong is that certain platform does not have support for floating point numbers, but I guess that in such a case everything gets broken.

File: mathics/builtin/atomic/numbers.py [hunk 19]
```diff
@@ -195,7 +195,7 @@ class Accuracy(Builtin):
      >> Accuracy[0.``2]
       = 2.
 
-    For 0.`, the accuracy is
+    For 0.`, the accuracy satisfies
```
rocky, line 198: ok, but please add a colon after "satisfies" this what we do everywhere (and it happens to be the correct punctuation).

test/builtin/atomic/test_numbers.py
```diff
@@ -183,38 +188,40 @@ def test_n():
     [
         # Accuracy for 0
         ("0", "Infinity"),
-        ("0.", "323.607"),
-        ("0.00", "323.607"),
-        ("0.00`", "323.607"),
-        ("0.00`2", "323.607"),
-        ("0.00`20", "323.607"),
+        ("0.", ZERO_MACHINE_ACCURACY_STR),
```
rocky, line 191: This is actually, more readable. Thanks.

test/builtin/atomic/test_numbers.py
```diff
@@ -183,38 +188,40 @@ def test_n():
...
         ("10.00`2", "1."),
         ("10.00`20", "19."),
         ("10.00000000000000000000", "20."),
         ("10.``2", "2."),
         ("10.``20", "20."),
-        # Returns the accuracy of ```2.4```
-        (" 0.4 + 2.4 I", "15.5744"),
+        # For some reason, the following test
+        # would fail in WMA
+        ("1. I", "Accuracy[1.]"),
```
rocky, line 215: Again this and the below are more understandable. Thanks. (And again this is shows how pytests are very different from tests that happen to get run in documentation.)

rocky:
LGTM and seems useful. Thanks. There may be some small lingering punctuation things, but those can be addressed any time.

PR #770, opened 2023-02-01
Place options 'All' and 'None' in correct section

rocky:
And go over docs and summary text for these. Add examples.

### DIFF
--- a/mathics/builtin/lists.py
+++ b/mathics/builtin/lists.py
@@ -5,28 +5,13 @@ List Functions - Miscellaneous
 Functions here will eventually get moved to more suitable subsections.
 """
 
-from mathics.builtin.base import Builtin, Predefined, Test
+from mathics.builtin.base import Builtin, Test
 from mathics.builtin.box.layout import RowBox
 from mathics.core.attributes import A_LOCKED, A_PROTECTED
 from mathics.core.exceptions import InvalidLevelspecError
 from mathics.core.list import ListExpression
 from mathics.eval.lists import list_boxes
 from mathics.eval.parts import python_levelspec
-
-
-class All(Predefined):
-    """
-    <url>
-    :WMA link:
-    https://reference.wolfram.com/language/ref/All.html</url>
-
-    <dl>
-      <dt>'All'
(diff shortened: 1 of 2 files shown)


mmatera (approved):
LGTM

rocky:
@mmatera I am not seeing some of the remaining functions `LevelQ` and `NotListQ`  listed on the Wolfram site. Do they still work?  (Server I use to check this has been unavailable).

mmatera:
> @mmatera I am not seeing some of the remaining functions `LevelQ` and `NotListQ` listed on the Wolfram site. Do they still work? (Server I use to check this has been unavailable).

nop, these symbols seem to be removed from WMA

rocky:
If we know when they were removed I will create a Pymathics module for compatibility with XXX, and add that there.

mmatera:
> If we know when they were removed I will create a Pymathics module for compatibility with XXX, and move this out of core and into this..

I never have used that symbol. Legacy versions of the WMA documentation are online. For example,

https://reference.wolfram.com/legacy/v4

has the documentation of version 4. However, I could not find a reference to these symbols.

mmatera:
On the other hand, we use `LevelQ` for defining rules in the module `mathics.builtin.functional.apply_fns_to_lists`

rocky:
This is weird. Something similar happens for 'NotListQ' - it is used only in signature testing. 

And it is used a _lot_ in SymJa rules in signatures as well which leads me to believe this _is_ a part of Mathematica, just one not documented. 

We have the option of keeping this but removing it from user-printed documentation. 

Your thoughts?

rocky:
Or we could _admit_ that it exists for use in function signature applicability and ask people not to use. I don't know.

But personally, I don't like the idea of saying nothing and not informing what's up.

mmatera:
I think we should keep it and document it, but maybe move the symbols to another section.

mmatera:
A good place for these symbols would be `mathics.builtin.testing_expressions.list_oriented`. Also probably we should remove the (unexisting) link to Wolfram Reference.

PR #771, opened 2023-02-01
renaming constants according to Python convension.

mmatera:
This is another round of changes in tiding up the arithmetic/precision code.


rocky:
LGTM  - thanks!

PR #772, opened 2023-02-02
mathics.eval eval_Capital fns

rocky:
Go over mathics.eval putting eval functions that might go into an instruction interpreter in the form eval_[Mathics-builtin-name] when there is only one evaluation function and some variation of that when there is more than one.

Some docstrings and summaries have been revised.

### DIFF
--- a/mathics/eval/nevaluator.py
+++ b/mathics/eval/nevaluator.py
@@ -46,9 +46,9 @@ def eval_NValues(
     expr: BaseElement, prec: BaseElement, evaluation: Evaluation
 ) -> Optional[BaseElement]:
     """
-    Looks for the numeric value of ```expr`` with precision ``prec`` by appling NValues rules
+    Looks for the numeric value of ```expr`` with precision ``prec`` by applying NValues rules
     stored in ``evaluation.definitions``.
-    If `prec` can not be evaluated as a number, returns None, otherwise, returns an expression.
+    If ``prec`` can not be evaluated as a number, returns None, otherwise, returns an expression.
     """
 
     # The first step is to determine the precision goal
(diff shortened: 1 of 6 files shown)


mmatera (approved):
LGTM

mathics/eval/nevaluator.py
```diff
@@ -34,15 +34,15 @@ def eval_N(
...
+    result = eval_NValues(evaluated_expression, prec, evaluation)
     if result is None:
         return expression
     if isinstance(result, Number):
         return result
     return result.evaluate(evaluation)
 
 
-def eval_nvalues(
+def eval_NValues(
```
mmatera, line 45: Maybe my only comment is that   `eval_NValues`  do not produce the same result as evaluating `NValues[x]` (which is return a list of `NValues`), but to apply the NValues rules associated with the elements in `x` over `x`. Maybe, adding a docstring would be enough to clarify this.

rocky: Good point - thanks for the information. Will think about and rename.

rocky: @mmatera - I just had a chance to look at.

Here are the only two  calls of `eval_NValues`  both in `mathics.builtin.numeric` that I see: 

```python
def eval_with_prec(self, expr, prec, evaluation, options=None):
    "N[expr_, prec_, OptionsPattern[%(name)s]]"
...
    return eval_NValues(expr, prec, evaluation)
```

and later in that file

```python
def eval_N(self, expr, evaluation: Evaluation):
    """N[expr_]"""
    # TODO: Specialize for atoms
    return eval_NValues(expr, SymbolMachinePrecision, evaluation)
```

Since eval routines return exactly the result that `NValues` returns, I think calling this `eval_NValues` is exactly the right thing to do.

PR #773, opened 2023-02-02
Comment and optimize eval for plus and times

mmatera:
This is another round of tidying up the arithmetic evaluations. Here
* `Accuracy[z_Complex]` and `Precision[z_Complex]` were adjusted to match better the WMA behaviour.
* `Plus.eval` and `Times.eval` were slightly rewritten and commented. Also, I change the order of some parts of the code to check the more frequent conditions to be checked first and to avoid waste computations.

### DIFF
--- a/mathics/builtin/arithfns/basic.py
+++ b/mathics/builtin/arithfns/basic.py
@@ -6,18 +6,12 @@ The functions here are the basic arithmetic operations that you might find on a 
 
 """
 
-
-import mpmath
-import sympy
-
 from mathics.builtin.arithmetic import _MPMathFunction, create_infix
 from mathics.builtin.base import BinaryOperator, Builtin, PrefixOperator, SympyFunction
 from mathics.core.atoms import (
     Complex,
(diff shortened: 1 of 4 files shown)


rocky:
If I have it correct, there are two separate things in this PR. 

First, the calculation of `Accuracy[]` has become more accurate. And there are a couple of pytests that show this.

Second, the order of performing Plus and TImes have been changed. And here I am not seeing anything to show the improvement. 

The first change I don't have a problem with, I think I understand, and I think that should go in as is. 

The second kind of change to Plus and TImes, would like to understand better. So it would be great to split this into two PRs.

mmatera:
As many other things it happens that handling accuracy and arithmetic operations are intrinsically entangled. So, in order to fix one in a comprensible way, I had to  comment and reorganize a little bit the code. The change in the order of the conditionals is just to avoid computing quantities that at the end are not used. In any case, the effect of avoiding to find the minimal precision would be appreciable just for very large expressions.

PR #774, opened 2023-02-04
fix accuracy and precision for complex numbers

mmatera:
This is the part of #773 related to the accuracy and precision of complex numbers that I could split without touching arithmetic. Then, I have to keep some funny behaviors like that  
```
Accuracy[Complex[1.``2, 2.``2 ]]
```
is different from
```
Accuracy[1.``2+2.``2 I]
```


rocky:
LGTM

PR #775, opened 2023-02-04
from_mpmath precision twaks

mmatera:
`PrecisionReal` numbers use `sympy.Float` numbers as their internal representation. At the time, `sympy.Float` uses `mpmath.mpf` numbers as their internal representation. Even if `sympy.Float` and `mpmath.mpf` accepts as a parameter a floating point number as decimal precision (dps), internally the precision is stored as an integer number bits. 

In our current master, we have an inconsistent use of the precision attribute: when we create a  `PrecisionReal` number, precision is measured in bits, but in `from_mpmath`, precision is passed as binary.  Apart from requiring several unneeded conversions, each time we pass from binary to decimal and backward, we lost accuracy in the estimation of the error, which is a problem if we want a consistent implementation for `Accuracy` and  `Precision`.

This PR amends that by adding this information in the docstrings, and normalizing the conversion between PrecisionReal and Mathics numbers by using always bit precision.

### DIFF
--- a/mathics/builtin/arithfns/basic.py
+++ b/mathics/builtin/arithfns/basic.py
@@ -446,24 +446,31 @@ class Plus(BinaryOperator, SympyFunction):
                     last_item = rest
                     last_count = count
         append_last()
-
         if numbers:
+            # TODO: reorganize de conditions to avoid compute unnecesary
+            # quantities. In particular, is we check mathine_precision,
+            # we do not need to evaluate prec.
             if prec is not None:
                 if is_machine_precision:
                     numbers = [item.to_mpmath() for item in numbers]
                     number = mpmath.fsum(numbers)
                     number = from_mpmath(number)
                 else:
-                    # For a sum, what is relevant is the minimum accuracy of the terms
-                    acc = (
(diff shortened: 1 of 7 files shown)


mmatera:
BTW: maybe we should consider using directly  `mpmath.mpf` as the internal representation for floating point numbers.

rocky:
> BTW: maybe we should consider using directly `mpmath.mpf` as the internal representation for floating point numbers.

While there may be special cases where float will do, I think in general this is a good thing to do.

mmatera:
> > BTW: maybe we should consider using directly `mpmath.mpf` as the internal representation for floating point numbers.
> 
> While there may be special cases where float will do, I think in general this is a good thing to do.
> 
> But let's please do this after a release. All of this work is just delaying getting the current release done.

Sure, it was just a comment. What I would like to have before the release is to bring the  Accuracy/Precision behavior as close as possible to the WMA one. This is kind of easy, and would help to drive the next step, which would be improving the arithmetic evaluation.

rocky:
LGTM - (I had intended to let you merge).  Please though let us focus on the things that will make a release happen sooner.

PR #776, opened 2023-02-05
mathics.core.pymathics -> mathics.eval.pymathics

rocky:
What is in mathics.eval is primarily for builtin LoadModule

mathics.core are for atoms, symbols, parsing, conversion - things at the very lowest level that aren't focused directly around evaluation of a particular Mathics3 builtin function.

### DIFF
--- a/mathics/builtin/assignments/assignment.py
+++ b/mathics/builtin/assignments/assignment.py
@@ -17,9 +17,9 @@ from mathics.core.attributes import (
     A_PROTECTED,
     A_SEQUENCE_HOLD,
 )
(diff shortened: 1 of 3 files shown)


rocky:
@mmatera and @TiagoCavalcante as soon as this clears it will be merged in, but feel free to comment and it will be handled later.  There is a corresponding change on the Mathics3 Django side as well (possibly mathicsscript too) which depend on this and will be merged in. 

There will also be a largish refactoring of the Doc stuff which removes _a_ lot of duplicated code in doing documentation and testing.  In that pymathics should be addressed on the LaTeX side.

PR #778, opened 2023-02-05
Reinstate pymathics doc

rocky:
This allows Mathics3 Modules loaded via `-l` or `--load-module` on the `docpipetest.py` command like to get loaded 
using  `eval_LoadModule` the underlying eval method for `LoadModule[]`.

In other words, from Mathics3 core you can test Mathics3 Modules or select individual sections Mathics3 Module tests. But that also means, we are collecting XML-like  documentation and test data which can then be put into the LaTeX-generated PDF.

 I believe the information will go into the PCL (Python Pickle file), but I haven't tested that yet.

 And I  haven't looked at the piece that puts creates `documentation.tex`  (`mathics/doc/latex/doc2latex.py`)

 I think that will be another (small) PR.  This is large enough as it is.

For expediency in being able to get this accomplished over the weekend in a limited time, there are/were several separable commits here.  I am sure you know the pattern - you want to get X done but bug Y gets in the way. So you fix Y, and then bug Z pops up. 

Over the week I will try to extract these so that they can be reviewed/discussed independently.

Here are two that I can think of though.

I would like to change the name Pymathics or Pymathics module to the simpler "Mathics3 Module" which has always been loaded via `LoadModule`. 

The messages when a Mathics3 Module can't be loaded have been changed.

### DIFF
--- a/mathics/builtin/assignments/assignment.py
+++ b/mathics/builtin/assignments/assignment.py
@@ -66,17 +66,17 @@ class LoadModule(Builtin):
       <dd>'Load Mathics definitions from the python module $module$
     </dl>
     >> LoadModule["nomodule"]
-     : Python module nomodule does not exist.
+     : Problem importing Python module: nomodule.
      = $Failed
     >> LoadModule["sys"]
-     : Python module sys is not a pymathics module.
+     : Python module "sys" is not a Mathics3 module.
      = $Failed
     """
 
     name = "LoadModule"
     messages = {
-        "notfound": "Python module `1` does not import.",
-        "notmathicslib": "Python module `1` is not a Mathics3 module.",
+        "notfound": """Problem importing Python module: `1`.""",
+        "notmathicslib": """Python module "`1`" is not a Mathics3 module.""",
     }
     summary_text = "load a pymathics module"
 
--- a/mathics/builtin/assignments/assignment.py
+++ b/mathics/builtin/assignments/assignment.py
@@ -75,8 +75,8 @@ class LoadModule(Builtin):
 
     name = "LoadModule"
     messages = {
-        "notfound": "Python module `1` does not exist.",
-        "notmathicslib": "Python module `1` is not a pymathics module.",
+        "notfound": "Python module `1` does not import.",
+        "notmathicslib": "Python module `1` is not a Mathics3 module.",
     }
     summary_text = "load a pymathics module"
 
--- a/mathics/builtin/assignments/assignment.py
+++ b/mathics/builtin/assignments/assignment.py
@@ -11,6 +11,7 @@ from mathics.core.assignment import (
     assign_store_rules_by_tag,
     normalize_lhs,
 )
+from mathics.core.atoms import String
 from mathics.core.attributes import (
     A_HOLD_ALL,
     A_HOLD_FIRST,
@@ -66,7 +67,7 @@ class LoadModule(Builtin):
       <dd>'Load Mathics definitions from the python module $module$
     </dl>
     >> LoadModule["nomodule"]
-     : Problem importing Python module: nomodule.
+     : Python import errors with: No module named 'nomodule'.
      = $Failed
     >> LoadModule["sys"]
      : Python module "sys" is not a Mathics3 module.
@@ -75,7 +76,7 @@ class LoadModule(Builtin):
 
     name = "LoadModule"
     messages = {
-        "notfound": """Problem importing Python module: `1`.""",
+        "loaderror": """Python import errors with: `1`.""",
         "notmathicslib": """Python module "`1`" is not a Mathics3 module.""",
     }
     summary_text = "load a pymathics module"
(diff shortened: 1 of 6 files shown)


mathics/doc/latex_doc.py
```diff
@@ -666,258 +655,22 @@ def latex(self, doc_data: dict):
...
+        self.doc_part_fn = LaTeXDocPart
+        self.doc_section_fn = LaTeXDocSection
+        self.doc_subsection_fn = LaTeXDocSubsection
         self.latex_pcl_path = settings.DOC_LATEX_DATA_PCL
         self.parts = []
         self.parts_by_slug = {}
-        self.pymathics_doc_loaded = False
-        self.doc_data_file = settings.get_doc_latex_data_path(should_be_readable=True)
         self.title = "Overview"
-        files = listdir(self.doc_dir)
```
rocky, line 678: A programmer who write such a long piece of code like this should be graduated out of programming.

mathics/builtin/assignments/assignment.py
```diff
@@ -66,17 +66,17 @@ class LoadModule(Builtin):
...
-     : Python module sys is not a pymathics module.
+     : Python module "sys" is not a Mathics3 module.
      = $Failed
     """
 
     name = "LoadModule"
     messages = {
-        "notfound": "Python module `1` does not exist.",
-        "notmathicslib": "Python module `1` is not a pymathics module.",
+        "notfound": """Problem importing Python module: `1`.""",
```
mmatera, line 78: If `LoadModule` produces the error `notfound`, it is because the module can not be found, not because there was a generic error in importing it. So, why the new message is better?

mathics/doc/common_doc.py
```diff
@@ -689,251 +1017,32 @@ def __str__(self):
         return self.test
 
 
+# FIXME: think about - do we need this? Or can we use DjangoMathicsDocumentation and
+# LatTeXMathicsDocumentation only?
```
mmatera, line 1021: Are we using this? Maybe we could replace it with a function that produces the right class of object.

mmatera:
Apart from my comments, and the need of merging first #781 to avoid the hanging, LGTM.

mathics/builtin/assignments/assignment.py
rocky: There is a difference between something existing and something importing. In fact another check could be made to see if the _file_ exists and give that message.  The file may exist, but there is invalid Python or it may import modules that are not importable.

mathics/doc/common_doc.py
rocky: Yes, we are using this. Merging the two is another mess that I don't want to deal with right now. 

If you want to try to do it go head.  I will note that there are still lots of opportunities for DRYing the code and improving it 

The motivation here was simply that we had code in Django to handle Mathics3 Modules and I wanted that used on the LaTeX/PDF side. So moving this code to common code seemed like the way to do this which both accomplishes what I want to do while improving the code. 

However in doing so there have been small breakages here and there precisely because there were two sets of code and an improvement or necessary change was made in one copy but not the other. And the one with the improvement did not make it to the common code. 

Last weekend I had my fill of trying to find and fix these kinds of things. Doing more in order to DRY the code more is left for the future for me.

mmatera:
There is another thing: for some reason, the pymathics module is not loaded when the module doctests are run

mathics/builtin/assignments/assignment.py
mmatera: OK, but the error we handle in importing the python module is about the existence of the library.

rocky: "Exist" is a vague or misleading word. The library can exist in some sense but not work. When I see an error message or warning I generally want to know what was tried. What was tried is that we ran a Python kind of "import": and that did not work,.

mmatera: In Python, what we are capturing is:
``` 
>>> import amadeupmodulename
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'amadeupmodulename'
```
So, maybe "Python module `1` was not found in the system"
could be a more accurate message, isn't it?

rocky:
> There is another thing: for some reason, the pymathics module is not loaded when the module doctests are run

Works for me: 

```
pyston mathics/docpipeline.py -x -l pymathics.graph -s BinomialTree
Mathics3 Module pymathics.graph loaded
Testing section(s): BinomialTree
Testing section: Graphs - Vertices and Edges / BinomialTree
b'   1 ( 0): TEST BinomialTree[3]'
```

What do you get when you run `pyston mathics/docpipeline.py -x -l pymathics.graph -s BinomialTree`

mathics/builtin/assignments/assignment.py
rocky: There is a differences here. You issued the `import` and not "not found" so it is clear that this is in response to Python "import".  In Mathics3 though, the user a priori doesn't know that LoadModule runs a Python `import`. 

So using the words Python and import suggest what kind of thing is failing. In fact it suggest trying to run a Python "import".

mmatera: What about ```Pymathics module `1` was not found in the system```? Otherwise,  it was just a suggestion. Please just keep it as you prefer.

rocky: I have been thinking about this some more. I do listen to your suggestions...

I think we can get the best of both worlds by catching _any_  Python exception and passing back the message in the exception. In those cases where Python reports "not found" the message has that (indicating this is from an import). 

In fact, it was probably a mistake using `except ImportError` rather then `except Exception`.

PR #779, opened 2023-02-05
Small bugs...

rocky:
* Fix InterpretedBox[] bug
* Turn flaky UniqueSym[] doctests into pytests. More could be done here
* typo: enlapsed -> elapsed
* Use `N[Pi, 100]` -> `N[Pi, 30]` until we get the mp.mpath precision reset bug fixed

### DIFF
--- a/mathics/doc/documentation/1-Manual.mdoc
+++ b/mathics/doc/documentation/1-Manual.mdoc
@@ -168,8 +168,8 @@ Of course, \Mathics has complex numbers:
 ('!' denotes the factorial function.)
 The precision of numerical evaluation can be set:
 
-  >> N[Pi, 100]
-  = 3.141592653589793238462643383279502884197169399375105820974944592307816406286208998628034825342117068
+  >> N[Pi, 30]
+  = 3.14159265358979323846264338328
 
 Division by zero is forbidden:
   >> 1 / 0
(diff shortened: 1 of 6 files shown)


rocky:
Currently most of this is in #778. However after this lands, that branch will be rebased to have this stuff removed.

mathics/doc/documentation/1-Manual.mdoc
```diff
@@ -168,8 +168,8 @@ Of course, \Mathics has complex numbers:
 ('!' denotes the factorial function.)
 The precision of numerical evaluation can be set:
 
-  >> N[Pi, 100]
-  = 3.141592653589793238462643383279502884197169399375105820974944592307816406286208998628034825342117068
```
TiagoCavalcante, line 172: Maybe a test that is expected to fail would be nice here, to detect when the regression is fixed

TiagoCavalcante (approved):
LGTM

mathics/doc/documentation/1-Manual.mdoc
rocky: @TiagoCavalcante Just so you know...

The problem right now that @mmatera isolated is that the function `mp_constant`  does not reset the precision after doing something like this. And then subsequent functions that use mpmath get very very slow.

We see this in testing where a 2 minute test jumps to 10 minutes, after that mpmath precision has gotten set to 100 or so. 

@mmatera offers: 

```
mathics -e 'p=N[Pi,100];' -e 'Timing[Plot[BesselK[1,x],{x,0,2}]]'
```

as an example to try to show the problem. 

He is currently working on a fix to this.  

I think a test to show that the precision is reset is fine. However this would _not_ be a doctest but rather a pytest.  

And I imagine that will appear along with the code to fix. The code above is just a temporary workaround.

PR #781, opened 2023-02-06
fix mpmath precision handling

mmatera:
This PR fixes another issue with precision. In particular, this issue was the reason of the hangings in doctest in #778.

### DIFF
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -52,7 +52,12 @@ from mathics.core.convert.sympy import SympyExpression, from_sympy, sympy_symbol
 from mathics.core.element import ElementsProperties
 from mathics.core.expression import Expression
 from mathics.core.list import ListExpression
-from mathics.core.number import SpecialValueError, dps, min_prec
+from mathics.core.number import (
+    FP_MANTISA_BINARY_DIGITS,
+    SpecialValueError,
+    dps,
+    min_prec,
+)
 from mathics.core.symbols import (
     Atom,
     Symbol,
@@ -79,23 +84,26 @@ from mathics.core.systemsymbols import (
 from mathics.eval.nevaluator import eval_N
 
 
-# Caching this function produce problems tracking
-# precision.
-#
-# @lru_cache(maxsize=4096)
-def call_mpmath(mpmath_function, mpmath_args):
-    try:
-        return mpmath_function(*mpmath_args)
-    except ValueError as exc:
-        text = str(exc)
-        if text == "gamma function pole":
-            return SymbolComplexInfinity
-        else:
-            raise
-    except ZeroDivisionError:
-        return
-    except SpecialValueError as exc:
-        return Symbol(exc.name)
+@lru_cache(maxsize=4096)
+def call_mpmath(mpmath_function, mpmath_args, prec=None):
+    if prec is None:
+        prec = FP_MANTISA_BINARY_DIGITS
+    with mpmath.workprec(prec):
+        try:
+            result_mp = mpmath_function(*mpmath_args)
+            if prec != FP_MANTISA_BINARY_DIGITS:
+                return from_mpmath(result_mp, prec)
+            return from_mpmath(result_mp)
+        except ValueError as exc:
+            text = str(exc)
+            if text == "gamma function pole":
+                return SymbolComplexInfinity
+            else:
+                raise
+        except ZeroDivisionError:
+            return
+        except SpecialValueError as exc:
+            return Symbol(exc.name)
 
 
 class _MPMathFunction(SympyFunction):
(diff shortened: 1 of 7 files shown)


mmatera:
Also, (in my machine) the doctest time was reduced in 40%

mathics/builtin/arithmetic.py
```diff
@@ -79,7 +79,10 @@
 from mathics.eval.nevaluator import eval_N
 
 
-@lru_cache(maxsize=4096)
+# Caching this function produce problems tracking
+# precision.
+#
+# @lru_cache(maxsize=4096)
```
rocky, line 85: Interesting - caching considered bad. Well, actually it is _improper_ caching.   A fix here is to see if one can add in the current precision to the cache.  This would be done inside the code rather than as a decorator on the code. However the lru_cache package would be used (rather than some homegrown code like we have for Expression).

rocky:
LGTM - many thanks. The coding style has improved _a lot_.  40% speedup on the doctests is _awesome_!

There is one suggestion for a future PR.

rocky:
@TiagoCavalcante suggested in another PR tha a unit test be created to check that the precision gets reset properly.

mathics/builtin/arithmetic.py
mmatera: Yes, it could be implemented in that way. This is why I added a comment instead of erasing the line. By now, I think it is enough to fix the issue and improve the performance. Then, we can write another patch to restore the cache having into account the precision. If you want, I can do that in this same branch.

rocky: Ok  that makes sense. 

But I would ask that we add a unit test for the bug here. 

That in general is good practice for any bugs that come up and get fixed. Here, it would be a test that precision is not altered across any Mathics operation that changes the mpmath precisiono.

mmatera: I added a bunch of basic tests now. If you have more in mind, please add over them.

mmatera: Also, I could reinstall the lru_cache. Doing that, I think I got a cleaner code, which (even without the cache) is even faster.

mathics/builtin/arithmetic.py
```diff
@@ -86,6 +86,13 @@
 
 @lru_cache(maxsize=4096)
 def call_mpmath(mpmath_function, mpmath_args, prec=None):
+    """
+    calls the mpmath_function with mpmath_args parms
+    if prec=None, use floatting point arithmetic.
```
rocky, line 91: floatting -> floating

mathics/builtin/arithmetic.py
```diff
@@ -86,6 +86,13 @@
...
+    Otherwise, work with prec bits of precision.
+    """
+    # TODO: rocky, please help me with the annotations
```
rocky, line 94: * `mpmath_function`  is a `Callable` 
* `mpmath_args` is a `tuple` 
* `prec` is `Optional[int]`

mathics/builtin/arithmetic.py
```diff
@@ -150,38 +158,18 @@ def eval(self, z, evaluation):
...
-                    # FIXME: replace try/except as a context manager
-                    # like "with evaluation.from_mpmath()...
-                    # which can be instrumented for
-                    # or mpmath tracing and benchmarking on demand.
-                    # Then use it on other places where mpmath appears.
-                    try:
-                        result = from_mpmath(result)
-                    except OverflowError:
-                        evaluation.message("General", "ovfl")
-                        result = Expression(SymbolOverflow)
```
rocky, line 170: SymbolOverlow can also be removed from the systemsymbol import.

mathics/builtin/arithmetic.py
```diff
@@ -150,38 +158,18 @@ def eval(self, z, evaluation):
...
-
-            if isinstance(result, (mpmath.mpc, mpmath.mpf)):
-                if mpmath.isinf(result) and isinstance(result, mpmath.mpc):
-                    result = SymbolComplexInfinity
-                elif mpmath.isinf(result) and result > 0:
-                    result = Expression(SymbolDirectedInfinity, Integer1)
-                elif mpmath.isinf(result) and result < 0:
-                    result = Expression(SymbolDirectedInfinity, IntegerM1)
-                elif mpmath.isnan(result):
-                    result = SymbolIndeterminate
```
rocky, line 159: Symbol Indeterminate can be removed from the import from systemsymbol

rocky:
Excellent! - thanks for adding the tests and restoring the cache using the precision parameter. I wonder if there was any speedup noticed with that. 

All good. One thing I notice and nothing to change but just to keep in mind. We tend to use top-level Mathics tests at the top -level to check to make sure that precision is saved/restored. This is good from the standpoint of making sure that the end-goal is correct. 

These are called end-to-end test or "black-box" tests. 

However it is also possible to do a lighter-weight  and  faster test which when it fails narrows the problem specifically to `call_mpmath`.  These are called "white box" (but "transparent box" is a more accurate name) and somethings this is what people mean when they talk about "unit tests" - testing a small unit by virtue of being able to look inside the black box and tailor the test to the working that is seen inside the black box. 

These tend to be a little more fragile when there is a big refactor.

rocky:
merge when satisfied.

rocky:
Great! ship when ready.

PR #783, opened 2023-02-12
Include Mathics3 modules in docs...

rocky:
Add the ability to include Mathics3 Modules into the LaTeX/PDf documentionat 

Make it clear that Python Pick PCL info has just doctests - test input and output result, keyed by part/section/subsection/test/number)

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -23,7 +23,8 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         python -m pip install wheel
-        choco install llvm tesseract
+        choco install --force llvm
+        choco install tesseract
         set LLVM_DIR="C:\Program Files\LLVM"
     - name: Install Mathics3 with Python dependencies
       run: |
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -23,6 +23,9 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         python -m pip install wheel
+        # use --force because llvm may already exist, but it also may not exist.
+        # so we will be safe here. Another possibility would be check and install
+        # conditionally.
         choco install --force llvm
         choco install tesseract
         set LLVM_DIR="C:\Program Files\LLVM"
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ MATHICS3_MODULE_OPTION ?= --load-module pymathics.graph,pymathics.natlang
    dist \
    doc \
    doctest \
-   doc-data \
+   doctest-data \
    djangotest \
    gstest \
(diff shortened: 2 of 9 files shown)


rocky:
A guide to this PR with my usuallong  commentary...  

In the first commit (which has been squashed down with later commits,  I did basically the minimum get this working. That was basically adding a `--load-module` option to `doc2latex.py`.

This is needed because as I _now_ realize, the Python PCL file has just is a dictionary of `mathics.core.evaluation.Result` converted to a dictionary.  Somehow I had thought that the docsting/XML-like data would be in there too.  Nope, that is loaded via via importing mathics for Builtins, and via `eval_LoadModule` for Mathics3 modules.

While this may seem cool from Django docs because when combined with the automatic reload mechanism that Python provides, from a design and robustness standpoint it is _horrible_. 

(And on the Django side at some point this as documentation grows it may be become too slow and unusable.)
 
I shouldn't be amazed, but I am constantly amazed by the bone headedness of the code that we haven't replaced. A good way to hide the bone headedness seems to be omit comments, docstrings, and type annotations, and that then limits discussion and you can hope no one will notice because all of is more hidden from scrutiny. 

At any rate, back to the code. Right now, the docstrings and the test result data written out to the PCL file are assumed to match, but there is no guarantee or check that in fact they do. (Please let's not write additional code for this kind of checking ...) 

And Mathics3 load-module options on `docpipeline.py` (which creates the PCL file) should match options on `doc2latex` (which reads the PCL file). If not, you get the intersection of the Mathics3 module options. 

Of course a system like sphinx or _any_ documentation system just doesn't work like this - of _course_ the doc is internally consistent and there is no stitching together data from different parts. _And_ we get back our docstring of the classes back for writing about Python class information.

Having documentation data all extracted rather spread between  Python docstrings read at runtimes and some from a pre-generated PCL file is more complex than just having all the documentation extracted. And I suppose that contributed to my confusion. I assumed the simple straight-forward approach that documentation systems use was in used here, even if the code was homegrown rather than using an existing, better written, more maintained package. 

But whatever....

After doing the minimum, I want to make sure the code has comments and variable names that to make nonstandard behavior and intent clearer.  When something is nonstandard we need even _more_ documentation, not less.

That is the motivation for changing "doc_data" to doctest_data" and things like that. So this PR is a little more pervasive, than a minimum change.

I will be making the corresponding change to Django as well.

Lastly, let me say that what remains now is to get TeXForm working for pymathics.graph so that it produces asymptote output. Right now it produces SVGs only for graphs.

rocky:
Merging  and iterating. This PR gets the Mathics3 module documentation into the LaTeX/PDF. 

The next one will deal with getting NetworkX graph images included.

PR #784, opened 2023-02-13
Format to LaTeX when we are writing to LaTeX

rocky:
I noticed this when looking into what it would take to get pymathics.graph graphs to format properly. 

<strike>I don't see a beneficial effect when doing this, but I imagine that</strike> as we get more proper formatting it <strike>may be</strike>  is an issue, so may as well fix it here now. 

(Short story there - all of this should be doable on the pymathics.graph side, but it will take writing some MakeBox rules, and adding a asymptote formatter for GraphBox on the pymathics.graph side. This is probably not something we should undertake before release.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,8 +124,9 @@ gstest:
 
 
 #: Create doctest test data and test results that is used to build LaTeX PDF
+# For LaTeX docs we assume Unicode
 doctest-data: mathics/builtin/*.py mathics/doc/documentation/*.mdoc mathics/doc/documentation/images/*
-	MATHICS_CHARACTER_ENCODING="ASCII" $(PYTHON) mathics/docpipeline.py --output --keep-going $(MATHICS3_MODULE_OPTION)
+	MATHICS_CHARACTER_ENCODING="UTF-8" $(PYTHON) mathics/docpipeline.py --output --keep-going $(MATHICS3_MODULE_OPTION)
 
 #: Run tests that appear in docstring in the code.
 doctest:
--- a/mathics/builtin/exp_structure/general.py
+++ b/mathics/builtin/exp_structure/general.py
@@ -48,32 +48,49 @@ class ApplyLevel(BinaryOperator):
 class BinarySearch(Builtin):
     """
     <url>
-    :WMA link:
-    https://reference.wolfram.com/language/ref/BinarySearch.html</url>
+    :Binary search algorithm:
+    https://en.wikipedia.org/wiki/Binary_search_algorithm</url> (<url>
+    :WMA:
+    https://reference.wolfram.com/language/ref/BinarySearch.html</url>)
 
     <dl>
       <dt>'CombinatoricaOld`BinarySearch[$l$, $k$]'
-      <dd>searches the list $l$, which has to be sorted, for key $k$ and returns its index in $l$. If $k$ does not
-        exist in $l$, 'BinarySearch' returns (a + b) / 2, where a and b are the indices between which $k$ would have
-        to be inserted in order to maintain the sorting order in $l$. Please note that $k$ and the elements in $l$
-        need to be comparable under a strict total order (see https://en.wikipedia.org/wiki/Total_order).
+      <dd>searches the list $l$, which has to be sorted, for key $k$ and \
+          returns its index in $l$.
+
+          If $k$ does not exist in $l$, 'BinarySearch' returns ($a$ + $b$) / 2, \
+          where $a$ and $b$ are the indices between which $k$ would have  \
+          to be inserted in order to maintain the sorting order in $l$.
+
+          Please note that $k$ and the elements in $l$ need to be comparable \
+          under a <url>
+          :strict total order:
+          https://en.wikipedia.org/wiki/Total_order</url>.
 
       <dt>'CombinatoricaOld`BinarySearch[$l$, $k$, $f$]'
-      <dd>the index of $k in the elements of $l$ if $f$ is applied to the latter prior to comparison. Note that $f$
-        needs to yield a sorted sequence if applied to the elements of $l.
+      <dd>gives the index of $k$ in the elements of $l$ if $f$ is applied to the \
+          latter prior to comparison. Note that $f$ \
+          needs to yield a sorted sequence if applied to the elements of $l$.
     </dl>
+
+    Number 100 is found at exactly in the fourth place of the given list:
 
     >> CombinatoricaOld`BinarySearch[{3, 4, 10, 100, 123}, 100]
      = 4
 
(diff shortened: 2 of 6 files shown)


TiagoCavalcante (approved):
LGTM

mmatera:
Is there any difference when the documentation is built?

rocky:
> Is there any difference when the documentation is built?

Actually, I have been looking at the behavior of 

```
filename = $TemporaryDirectory <> "/example_file"
Put[x + y, 2x^2 + 4z!, Cos[x] + I Sin[x], filename]
FilePrint[filename]
```

Starting this release is breaking for LaTeX, and it may be related to this. 

With this change, it doesn't _fix_ that, but it seems to break things in Django which I think assumes the format is `text`.

mmatera:
> 

Both in  WMA and Mathics,
```
In[1]:= filename = $TemporaryDirectory <> "/example_file"                       

Out[1]= /tmp/example_file

In[2]:= Put[x + y, 2x^2 + 4z!, Cos[x] + I Sin[x], filename]                     

In[3]:= FilePrint[filename]                                                     
x + y
2*x^2 + 4*z!
Cos[x] + I*Sin[x]
```

rocky:
> > 
> 
> Both in WMA and Mathics,
> 
> ```
> In[1]:= filename = $TemporaryDirectory <> "/example_file"                       
> 
> Out[1]= /tmp/example_file
> 
> In[2]:= Put[x + y, 2x^2 + 4z!, Cos[x] + I Sin[x], filename]                     
> 
> In[3]:= FilePrint[filename]                                                     
> x + y
> 2*x^2 + 4*z!
> Cos[x] + I*Sin[x]
> ```

But the real question is - what is the contents of `/tmp/example_file` in Both Mathics-Django and in the command-line version. Do they match? And if not is that correct? (Probably not)

mmatera:
> desired

Pytest would also pass in mathics-core. The issue comes from the mathics-django hack. It should be fixed now with #786.

rocky:
With the last batch of changes, and @mmatera's change to create PNG's  on the pymathics'graph side when TeXForm is requested, we now have all of the graphs appearing in the PDF. 

_And_ the problem I mentioned before with Put is goine. 

_And_ generally there is an improvement since the output is formatted more properly for LaTeX. 

That said, I think I want to revise this a little more....

The synergy around flawed thinking around the design of the doctest system and how formatting works is making document generation look worse. 

I have silenced that here, but it is still wrong, and I think we can reduce this wrongness a little bit more. 

One conceptual flaw in design is that the output that you expect to see in testing using StandardForm is not the same output as you expect for TeXform.  And the homegrown testing assumes and tests on, I guess, StandardForm, or something other than TeXForm. But TeXForm is what is preferable for the docs. 

We saw a little of the before with the need to set MATHICS_CHARACTER_ENCODING.

(And so rather than deal with this,  in the past the solutions was to add more text substitutions to basically duplicate what TeXForm is supposed to do. Needless to say, even though our TeXForm isn't ideal, it is beats search and replace text.) 

The second conceptual flaw which synergizes with the first flaw is that evaluation by default has been eval + format. And what we want here are two format phases: to StandardForm for the purpose of testing, and then Format for writing test output. 

I will try adding this.

rocky:
I should also mention that I see numerous LaTeX errors. Some around hrefs.  I think some we could fix by improving TeXForm.

rocky:
> > desired
> 
> Pytest would also pass in mathics-core. The issue comes from the mathics-django hack. It should be fixed now with #786.

A pytest in mathics django for the contents would verify this.

mmatera:
we want here are two format phases: to StandardForm for the purpose of testing, and then format to TeXForm for writing test output.
> 
> I will try adding this.

Just a couple of comments. doctests compares against  our `OutputForm`  not `StandardForm`. I say 'our' `OutputForm` because in WMA `OutputForm` means 2d  text,  something like the prettyprint output in SymPy.

On the other hand, `TeXForm` (`MathMLForm`) produces, by default, a TeX (MathML) version of the `StandardForm`.

rocky:
> > we want here are two format phases: to StandardForm for the purpose of testing, and then format to TeXForm for writing test output.
> > I will try adding this.
> 
> Just a couple of comments. doctests compares against our `OutputForm` not `StandardForm`. I say 'our' `OutputForm` because in WMA `OutputForm` means 2d text, something like the prettyprint output in SymPy.
> 
> On the other hand, `TeXForm` (`MathMLForm`) produces, by default, a TeX (MathML) version of the `StandardForm`.

Sure, you say so  - whatever. 

What is called a "StandardForm",  "OutputForm",  "XML" and what is called "TeX" in the code base like too many other things that was inherited in this code base, is vague or confused if not plain wrong in some cases.  (WMA helps with the confusion though for  what "TeX" means.)

So all I meant to say is there is the kind of output format that we want to use in the LaTeX PDF in test cases, and there is a kind of output is supposed to match home-grown expected test output - this is the textf that follows `= ` in class docstrings for builtins.

mmatera:
BTW, `make doc` works without complains!  Merge this when you feel it is ready.

rocky:
> BTW, `make doc` works without complains! Merge this when you feel it is ready.

It is not complete in that it should do the stuff I mentioned - run tests twice, once for test expectation and once once for LaTeX formatting. And do that in a way that doesn't use the eval+format call but instead eval, format for test, format for LeTeX. 

However, I am not going to be able to get to this in the near term. So I'll merge, and we can continue later.

rocky:
A couple of slightly random things when doing this. I tried adding a rule like

```
    rules =  {"MakeBoxes[Pi,(StandardForm|TraditionalForm)]": '"\\[Pi]"'", "MakeBoxes[Pi,TeXForm]": r'\Pi'}
```

To try to do Pi formatting correctly without the search and replace string. This doesn't work. Do you know why? 

The second thing I noticed is that I haven't see Mathics3 module natlang get pulled in. I suspect it is a problem on the natlang side. But I haven't had a chance to investigate.

mmatera:
Actually, it seems to  work: look at 
https://github.com/Mathics3/mathics-core/tree/correct-docpipeline-form-on-latex-generation

This is how the test look in the pdf documentation:

imagen

mmatera:
Update: the problem is how `mathics.eval.makeboxes.format_element` is processing the format. Inside the CLI, it always forces `OutputForm`. This is something that would be fixed with the refactor in which I was working on.

rocky:
> Actually, it seems to work: look at https://github.com/Mathics3/mathics-core/tree/correct-docpipeline-form-on-latex-generation
> 
> This is how the test look in the pdf documentation:
> 
> imagen

It works because of a hack. And we want to remove the hack.

mmatera:
#792 would allow to remove the hack.

PR #786, opened 2023-02-13
removing the use of Evaluation.format_output in builtins

mmatera:
This PR fixes an issue reported by @rocky, where the behavior or `Put` was modified when is was called from mathics-server.  Now, in `mathics.builtin` all the references to `format_output` which is modified by mathics-server  were rempaced by calls to `format_element` and `boxes_to_text`.

### DIFF
--- a/mathics/core/evaluation.py
+++ b/mathics/core/evaluation.py
@@ -12,7 +12,7 @@ from mathics_scanner import TranslateError
 from mathics import settings
 from mathics.core.atoms import Integer, String
 from mathics.core.convert.python import from_python
-from mathics.core.element import KeyComparable, ensure_context
+from mathics.core.element import BaseElement, KeyComparable, ensure_context
 from mathics.core.interrupt import (
     AbortInterrupt,
     BreakInterrupt,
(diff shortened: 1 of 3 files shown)


mathics/core/evaluation.py
```diff
@@ -379,7 +379,17 @@ def get_stored_result(self, eval_result, output_forms):
     def stop(self) -> None:
         self.stopped = True
 
-    def format_output(self, expr, format=None):
+    def format_output(
```
mmatera, line 382: probably, `format_output` should be a parameter of `Evaluation.evaluate`, or in the `Evaluation.__init__` method.

mmatera: Not sure about that, but now this is used in all the backends, so let's keep for a while

mmatera:
> This is all provisional until it is all replaced by more coherent Boxing. So LGTM.
> 
> After this PR is merged, I guess it would be nice to remove the code on the mathics-django side.

I think you wouldn't like to do that before the release since this would require going deeper into the formatting problem.

rocky:
> > This is all provisional until it is all replaced by more coherent Boxing. So LGTM.
> > After this PR is merged, I guess it would be nice to remove the code on the mathics-django side.
> 
> I think you wouldn't like to do that before the release since this would require going deeper into the formatting problem.

Ahh - I thought this was a simple matter of removing corresponding code. If not, that's fine we can keep Django the way it currently is. 

After release there will be plenty of time to do further Django releases and pymathics.graph releases.

PR #792, opened 2023-02-17
fix do_format_element

mmatera:
This fix the issue mentioned in https://github.com/Mathics3/mathics-core/issues/784#issuecomment-1434767955:

With this PR, in the CLI,
```
In[1]:=Pi
Out[1]=Pi

In[2]:=OutputForm[Pi]
Out[2]=Pi

In[3]:=StandardForm[Pi]
Out[3]= π
```

### DIFF
--- a/mathics/builtin/forms/output.py
+++ b/mathics/builtin/forms/output.py
@@ -754,8 +754,12 @@ class OutputForm(FormBaseClass):
      = f'[x]
     >> OutputForm[Derivative[1, 0][f][x]]
      = Derivative[1, 0][f][x]
-    >> OutputForm["A string"]
-     = A string
+
+    'OutputForm' is used by default:
+    >> OutputForm[{"A string", a + b}]
+     = {A string, a + b}
+    >> {"A string", a + b}
+     = {A string, a + b}
     >> OutputForm[Graphics[Rectangle[]]]
      = -Graphics-
     """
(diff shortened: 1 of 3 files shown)


mathics/builtin/forms/output.py
```diff
@@ -850,12 +854,9 @@ class StandardForm(FormBaseClass):
     </dl>
 
     >> StandardForm[a + b * c]
-     = a + b c
+     = a+b c
```
mmatera, line 857: With our current implementation, `StandardForm` does not add spaces around the operators. If we change `MakeBoxes[Infix[...],...]`   to allow the spaces, then TeXForm and MathMLForm get broken.
Spaces should be added at the level of `boxes_to_format` routines, probably controlled by a certain parameter in the `RowBox` object. But this is something for another round.

mathics/eval/makeboxes.py
```diff
@@ -142,7 +142,7 @@ def do_format_element(
         # removes the format from the expression.
         if head in OutputForms and len(expr.elements) == 1:
             expr = elements[0]
-            if not (form is SymbolOutputForm and head is SymbolStandardForm):
+            if not form.sameQ(head):
```
mmatera, line 145: This is the root of the problem that I try to solve here. What we want to remove are duplicated formatting heads, not the special combination of `form=SymbolOutputForm` and  `head=SymbolStandardForm`.

rocky:
Ok - thanks for the explanation. I have no way to independently verify this, so I'll take your word for it. 

LGTM.

PR #794, opened 2023-02-19
SymbolStringExpression is used in natlang

rocky:
Small change will be merged in when it passes CI. 

Used in natlang. Still trying to get dockerhub image to build and notice this in the interim.


mmatera (approved):
LGTM

PR #796, opened 2023-02-20
Revert get_section_test() addition

rocky:
@mmatera  - Does this fix the problem that ~ 2/3 of the doctests do not get loaded? 

For reference, 1c35f0feeeaab9e26da1c1dd9c69db2af6c565b9 is the SHA before this merge and f0e04d647 I think is the commit after that. 

If this works, just merge it in. Or feel free to adjust this. 

Sorry for the breakage.


mmatera (approved):
It does. Thanks!

LGTM

PR #798, opened 2023-02-24
Intro revision

rocky:
I'd like to do a revision of the roadmap `FUTURE.rst` but that will be in a separate PR. 

Some of this is not a neat as I would like, I think there is still a bit of duplication to remove. But I wanted definitely to remove the homegrown doctest stuff as a "feature" (in the positive sense).

### DIFF
--- a/mathics/doc/documentation/1-Manual.mdoc
+++ b/mathics/doc/documentation/1-Manual.mdoc
@@ -36,7 +36,7 @@ Performance of \Mathics is not, right now, practical in large-scale projects and
 
 <section title="What does \Mathics offer?">
 
-Some of the features of \Mathics tries to be compatible with Wolfram Alpha within the confines of the Python ecosystem.
+Some of the features of \Mathics tries to be compatible with Wolfram-Language kernel within the confines of the Python ecosystem.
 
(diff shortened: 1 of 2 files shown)


mathics/doc/documentation/1-Manual.mdoc
```diff
@@ -36,30 +36,36 @@ Performance of \Mathics is not, right now, practical in large-scale projects and
 
 <section title="What does \Mathics offer?">
 
-Some of the features of \Mathics are:
+Some of the features of \Mathics tries to be compatible with Wolfram Alpha within the confines of the Python ecosystem.
```
mmatera, line 39: Why Wolfram alpha? I guess that Wolfram Language or if you want, Wolfram kernel would be a  more accurate description. wolfram alpha points more to a natural language web interface, I guess

mmatera (approved):
Apart from the comment, LGTM

mathics/doc/documentation/1-Manual.mdoc
rocky: > Why Wolfram alpha?

Sorry, since I don't use this stuff, I often get this wrong. 

Is "Wolfram-Language kernel" okay?

PR #799, opened 2023-02-24
Start noting what's up and what goals have been met

rocky:
I should have started this earlier and we should have started a longer discussion of this. 

But hey this is a start. I want to have something in place for the 6.0.0 release.

Your comments and ideas are welcome. 

I seem to recall something around `\!`  that involved precedence that was deferred that might go on the list.

### DIFF
--- a/FUTURE.rst
+++ b/FUTURE.rst
@@ -2,6 +2,91 @@
 
 .. contents::
 
+The following 2023 roadmap that appears the 6.0.0 hasn't gone through enough discussion. This provisional.
+Check the github repository for updates.
+
+
+2023 Roadmap
+============
+
+
+When the release settles, "Forms, Boxing, And "Formatting" is the next
+large refactor slated.  Having this will allow us to supporting Jupyter or other front
+ends. And it is something that is most visibly wrong in Mathics3 output.
+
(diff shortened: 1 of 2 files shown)


mmatera (approved):
LGTM. What we could do later would be to create discussions for each of the topics in "FUTURE.rst", and then link them in the file.

Regarding the issue of pattern matching, @rljackobson talked about certain possibilities here:https://github.com/Mathics3/mathics-omnibus/issues/14 

If he is up for it, I guess it could be possible to start by trying to decouple a little bit more the pattern matching code from the rest of the mathics-core, and then see if it is possible to replace some part with the existing python libraries without breaking the tests in the CI.

rocky:
> LGTM. What we could do later would be to create discussions for each of the topics in "FUTURE.rst", and then link them in the file.

Good idea - I'll start doing that and adjust this to point to that.

rocky:
@mmatera Ok - I have made basically the most minimum stab at this. Feel free to add more discussion and commit on top of this with the links. Sometime before release, I merge whatever has been accumulated. I mostly wanted to give people  a heads up and get _something_ start here, on  this aspect I should have been documenting better.

PR #803, opened 2023-02-26
Use recently-found V0.91 version

rocky:

PR #804, opened 2023-02-27
storing attributes when a pattern is created

mmatera:
Inspired in discussion #800, this changes a little bit the implementation of `Pattern`, in a way that attributes are stored in the object instead of looking at it in the Definitions object. Just with this, doctest cuts other 7s (~7% faster than master in Pyston).  
Also, I added a pytest with the tests proposed in #800.

### DIFF
--- a/mathics/builtin/patterns.py
+++ b/mathics/builtin/patterns.py
@@ -4,7 +4,7 @@
 Rules and Patterns
 
 The concept of transformation rules for arbitrary symbolic patterns is key \
-in \Mathics.
+in \\Mathics.
 
 Also, functions can get applied or transformed depending on whether or not \
 functions arguments match.
@@ -40,6 +40,8 @@ The attributes 'Flat', 'Orderless', and 'OneIdentity' affect pattern matching.
 
 # This tells documentation how to sort this module
 sort_order = "mathics.builtin.rules-and-patterns"
+
+from typing import Optional as OptionalType
 
 from mathics.builtin.base import (
     AtomBuiltin,
@@ -525,8 +527,8 @@ class PatternTest(BinaryOperator, PatternObject):
     precedence = 680
     summary_text = "match to a pattern conditioned to a test result"
 
-    def init(self, expr):
-        super(PatternTest, self).init(expr)
+    def init(self, expr, evaluation: OptionalType[Evaluation] = None):
+        super(PatternTest, self).init(expr, evaluation=evaluation)
         # This class has an important effect in the general performance,
         # since all the rules that requires specify the type of patterns
         # call it. Then, for simple checks like `NumberQ` or `NumericQ`
@@ -549,7 +551,7 @@ class PatternTest(BinaryOperator, PatternObject):
             "System`NonNegative": self.match_nonnegative,
         }
 
-        self.pattern = Pattern.create(expr.elements[0])
+        self.pattern = Pattern.create(expr.elements[0], evaluation=evaluation)
         self.test = expr.elements[1]
         testname = self.test.get_name()
         self.test_name = testname
@@ -745,9 +747,11 @@ class Alternatives(BinaryOperator, PatternObject):
     precedence = 160
     summary_text = "match to any of several patterns"
 
-    def init(self, expr):
-        super(Alternatives, self).init(expr)
-        self.alternatives = [Pattern.create(element) for element in expr.elements]
+    def init(self, expr, evaluation: OptionalType[Evaluation] = None):
+        super(Alternatives, self).init(expr, evaluation=evaluation)
+        self.alternatives = [
+            Pattern.create(element, evaluation=evaluation) for element in expr.elements
+        ]
 
     def match(self, yield_func, expression, vars, evaluation, **kwargs):
         for alternative in self.alternatives:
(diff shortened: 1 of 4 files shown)


mathics/builtin/patterns.py
```diff
@@ -4,7 +4,7 @@
 Rules and Patterns
 
 The concept of transformation rules for arbitrary symbolic patterns is key \
-in \Mathics.
+in \\Mathics.
```
mmatera, line 7: This is unrelated, but came up from flake...

mathics/builtin/options.py
```diff
@@ -54,8 +54,8 @@ class All(Predefined):
 
     In <url>
     :Plot:
-    /doc/reference-of-built-in-symbols/graphics-and-drawing/plotting-data/plot</url> \
-    </url>, setting the <url>
```
mmatera, line 58: Another unrelated fix, but this prevents to have a clean CI.

mmatera: (This also affected #803)

rocky: Thanks for finding and fixing, but please do this outside of this PR please and just go ahead and get this in.

mmatera: I just fixed it in master.

rocky:
> Inspired in discussion #800, this changes a little bit the implementation of `Pattern`, in a way that attributes are stored in the object instead of looking at it in the Definitions object. 

The _what_ I think is right, but the _how_ or implementation I need to think about.

> Just with this, doctest cuts other 7s (~7% faster than master in Pyston). 

Pretty amazing.  This may also be an indicator that there other problems in how the definitions object  is structured and implemented. 

> Also, I added a pytest with the tests proposed in #800.

I am not seeing that in this PR yet. Perhaps you forgot to `git add`.

mathics/builtin/patterns.py
rocky: This too can be separated from this PR. But not a big deal to get in though.

mmatera:
> 
> > Also, I added a pytest with the tests proposed in #800.
> 
> I am not seeing that in this PR yet. Perhaps you forgot to `git add`?

Indeed. Now it is in.

test/core/test_rules.py
```diff
@@ -0,0 +1,93 @@
+from test.helper import check_evaluation, evaluate_value
```
rocky, line 1: This is definitely more of the kind of thing we need - thanks!

In fact we actually could use even lower-level tests. These test at the user level, but we we need a lot more especially for pattern matching is tests of the various routines in more isolation of the rest of the system.

mmatera: The reason of proposing these high level tests is that I am pretty sure that we need to change the low level API. What I am trying to do with this is to check that the changes do not move us away from our current degree of compatibility with WL.

rocky: Yes, this is why end-to-end tests or integration tests are typically written. 

We _also_  (not _instead_) should be considering smaller unit tests., I agree that things here are likely to change and when that happens unit tests are the most fragile and the first to break.

So I guess here I am thinking that in doing larger changes, then it would be great to start adding unit (transparent-box) unit tests.

rocky: > What I am trying to do with this is to check that the changes do not move us away from our current degree of compatibility with WL.

Thinking about this more. I think this should be in a separate PR and that can go in now.

rocky:
LGTM - thanks

I just had a chance to go over this in detail. This is a good step in the right direction. 

The one question I have is whether there is a possibility of what we are doing now being wrong by _not_ accumulating attributes from the head, but instead getting that from later on from an evaluation object. 

And the one suggestion I have would be to give a _simple_ example of a nested pattern rule where information is needed from elements.

mmatera:
> LGTM - thanks
> 
> I just had a chance to go over this in detail. This is a good step in the right direction.
> 

Great!

> The one question I have is whether there is a possibility of what we are doing now being wrong by _not_ accumulating attributes from the head at the time of creation, but instead getting that from later on from an evaluation object.
> 

In principle, it would be possible but would require much more rewriting, and I don't know if I am able to get something more clear or more efficient. If you or something else knows how to do that, go ahead!


> For example, is it possible that head attributes might have changed in the interim?
> 

That is one of the points: the rules use the attributes as they were when the rules were defined, not the attributes that the expressions have at the moment of the application.

> The word "cache" in
> 
> https://github.com/Mathics3/mathics-core/blob/9510283dd60523949ce4f593a28ac2a4fa0f715c/mathics/core/pattern.py#L91
> 
> is misleading or wrong. A cache is something that you know is optional and will be the same as some computed value. Here, I don't know if the value is _guaranteed_ to be the same as the value in the evaluation object.
> 

The value used is the same of the attribute according to the evaluation object at the moment of the rule creation (for the rules created during the session), or, when it is not available (when the rule is created to populate the initial `evaluation.definitions`), at the first use of the rule.

If you are up and have a clearer way to express this, please improve my explanation.


> And the one suggestion I have would be to give a _simple_ example of a nested pattern rule where information is needed from elements.

This is the simplest example that comes to my mind:

```
rule=Times[c__, Plus[Q[a_],Q[b_]]]->Q[c*(a+b)]
```
Notice that `Times` has its attributes and default values, `Plus` has other attributes, and potentially, `Q` too.
Then, when the pattern `Times[c__, Plus[Q[a_],Q[b_]]]` is created, the subpatterns
`c__`, `Plus[Q[a_],Q[b_]]`, `Q[a_]`, `Q[b_]`, `a_` , and `b_` are created too. The constructors of all of these patterns need to know what the attributes assigned to the head of these expressions are at the build time.

mmatera:
I added now an example for a nested pattern, and some discussion of the corner case `Alternatives` as the head of a pattern, which works differently than the general case. I will merge this when the CI finished. Then we can improve this in another iteration.

rocky:
> > The one question I have is whether there is a possibility of what we are doing now being wrong by _not_ accumulating attributes from the head at the time of creation, but instead getting that from later on from an evaluation object.
> 
> In principle, it would be possible but would require much more rewriting, and I don't know if I am able to get something more clear or more efficient.



As this is new to you please do not try right now, but leave it for others. And that might be a good principle in general. When we get to the phase where focus on pattern-matching and rule rewriting specifically, we can come back to this.

Right now, all that matters is the observation and that doing things this way may lead to small subtle semantic differences.

mmatera:
> pattern

Yep, that's the idea. My goal here was to put as much information as possible for someone who wants to tackle this in the future.

PR #809, opened 2023-03-06
Improve directed infinity

mmatera:
This is based on one part of #766  related to how DirectedInfinity is handled. To simplify this, I remove part of the `eval_Times` code handling this kind of expression and replace it with upvalues rules. Also, some of the new pytests in #766 were included here.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,25 @@
 CHANGES
 =======
 
+
+API
+---
+
+* ``eval_sign`` and ``eval_abs`` are now in ``mathics.eval.arithmetic``.
+
+Compatibility
+-------------
+
+* ``*Plot`` does not show messages during the evaluation.
+
+
+
+Bugs
+----
+
+* Improved support for ``DirectedInfinity`` and ``Indeterminate``.
+
+
 6.0.1
 -----
 
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -7,7 +7,7 @@ Mathematical Functions
 Basic arithmetic functions, including complex number arithmetic.
 """
 
-from mathics.eval.numerify import numerify
+import sympy
 
 # This tells documentation how to sort this module
 sort_order = "mathics.builtin.mathematical-functions"
@@ -16,7 +16,6 @@ sort_order = "mathics.builtin.mathematical-functions"
 from functools import lru_cache
 
 import mpmath
-import sympy
 
 from mathics.builtin.base import (
     Builtin,
@@ -55,11 +54,9 @@ from mathics.core.number import dps, min_prec
 from mathics.core.symbols import (
     Atom,
     Symbol,
-    SymbolAbs,
     SymbolFalse,
     SymbolList,
     SymbolPlus,
-    SymbolPower,
     SymbolTimes,
     SymbolTrue,
 )
@@ -72,8 +69,11 @@ from mathics.core.systemsymbols import (
     SymbolTable,
     SymbolUndefined,
 )
-from mathics.eval.arithmetic import eval_mpmath_function
+from mathics.eval.arithmetic import eval_abs, eval_mpmath_function, eval_sign
 from mathics.eval.nevaluator import eval_N
+from mathics.eval.numerify import numerify
+
+ExpressionComplexInfinity = Expression(SymbolDirectedInfinity)
 
 
 class _MPMathFunction(SympyFunction):
@@ -207,6 +207,13 @@ class Abs(_MPMathFunction):
     }
     summary_text = "absolute value of a number"
     sympy_name = "Abs"
+
+    def eval(self, x, evaluation):
+        "%(name)s[x_]"
+        result = eval_abs(x)
+        if result is not None:
+            return result
+        return super(Abs, self).eval(x, evaluation)
 
 
 class Arg(_MPMathFunction):
@@ -590,8 +597,7 @@ class DirectedInfinity(SympyFunction):
      = Indeterminate
 
     >> DirectedInfinity[0]
-     : Indeterminate expression 0 Infinity encountered.
-     = Indeterminate
+     = ComplexInfinity
 
     #> DirectedInfinity[1+I]+DirectedInfinity[2+I]
      = (2 / 5 + I / 5) Sqrt[5] Infinity + (1 / 2 + I / 2) Sqrt[2] Infinity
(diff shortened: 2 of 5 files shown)


mathics/eval/plot.py
```diff
@@ -102,15 +100,10 @@ def quiet_f(*args):
 
             return quiet_f
     expr: Optional[Type[BaseElement]] = Expression(SymbolN, expr).evaluate(evaluation)
-    quiet_expr = Expression(
```
mmatera, line 105: This came up when I run the doctests with the other changes: In WMA, when the points are evaluated, `*Plot` hides all the warnings, not just "Power::infy". This could go in a different PR, but the natural tests needs of the changes introduced here. (I also tested with functions like `F[x_]:=(Open["nonexistingfile"];x^2)` but it is just not so nice.

CHANGES.rst
```diff
@@ -3,6 +3,25 @@
 CHANGES
 =======
 
+
+API
```
rocky, line 7: Usually API refers the external API - things that a person using core would have to worry about. 

I suspect this is just about internal organization and under the general category of we are moving evaluation functions into its own module.

mathics/builtin/arithmetic.py
```diff
@@ -622,39 +625,59 @@ class DirectedInfinity(SympyFunction):
...
-        "DirectedInfinity[DirectedInfinity[x___]]": "DirectedInfinity[x]",
+        "a_?NumericQ * DirectedInfinity[b_]": "DirectedInfinity[a * b]",
+        "a_ DirectedInfinity[]": "DirectedInfinity[]",
+        "DirectedInfinity[a_] * DirectedInfinity[b_]": "DirectedInfinity[a * b]",
     }
 
     formats = {
         "DirectedInfinity[1]": "HoldForm[Infinity]",
-        "DirectedInfinity[-1]": "HoldForm[-Infinity]",
+        "DirectedInfinity[-1]": "PrecedenceForm[-HoldForm[Infinity], 390]",
```
rocky, line 641: We should be getting numbers like 390 from the operator tables of Mathics scanner.

mathics/builtin/arithmetic.py
```diff
@@ -622,39 +625,59 @@ class DirectedInfinity(SympyFunction):
...
         "DirectedInfinity[]": "HoldForm[ComplexInfinity]",
-        "DirectedInfinity[DirectedInfinity[z_]]": "DirectedInfinity[z]",
-        "DirectedInfinity[z_?NumericQ]": "HoldForm[z Infinity]",
+        "DirectedInfinity[z_]": "PrecedenceForm[z HoldForm[Infinity], 390]",
```
rocky, line 643: 390 - see above.

test/builtin/arithmetic/test_basic.py
```diff
@@ -164,7 +177,7 @@ def test_multiply(str_expr, str_expected, msg):
         ("I^(2/3)", "(-1) ^ (1 / 3)", None),
         # In WMA, the next test would return ``-(-I)^(2/3)``
         # which is less compact and elegant...
-        ("(-I)^(2/3)", "(-1) ^ (-1 / 3)", None),
+        #        ("(-I)^(2/3)", "(-1) ^ (-1 / 3)", None),
```
rocky, line 180: Do I have it correct that this was working before, and not it is not?  _If_ that it the case, what's the insight here. (And if not, what is the insight here.)

mmatera: This works in #766 but not here. The issue comes from `Power.eval_check`.

mathics/builtin/arithmetic.py
mmatera: Times has Precedence 400, this needs to have a slightly lower value to ensure the parenthesis.

rocky: There is something then incomplete here. What comes in contact with what? Is there a value in [Robert's table](https://wltools.github.io/LanguageSpec/Specification/Syntax/OperatorTableHTML/) that is correct here?

mmatera: Maybe what is missing is better documentation of what `PrecedenceForm` means.
https://reference.wolfram.com/language/ref/PrecedenceForm.html

Consider this:
```
a DirectedInfinity[-1]  + b
```
In WMA is formatted as

```
c + b (-Infinity)
```

`PrecedenceForm` modifies the precedence of the outer `Times` in  `Times[b, PrecedenceForm[Times[-1, HoldForm[Infinity]], 390] ]`  to have a lower precedence value than the inner `Times` (which has the default precedence value of 400). Then, `-1 Infinity` is surrounded by a parenthesis, according to that precedence.
`390` and `399` have the same effect, as far as the expression is not a subexpression of another operator with (for example) precedence `395` .
For example, `Plus` has precedence `310`. So, let's consider now
```
 DirectedInfinity[3+4 I] + DirectedInfinity[-1]
```

This expression is formatted as
``` 
-Infinity + (3 / 5 + 4 / 5 I) Infinity
```
Notice that now there are not any parenthesis sorrounding `DirectedInfinity`. If instead of `390` we had used `290`,
then it would be formatted as

``` 
(-Infinity) + ((3 / 5 + 4 / 5 I) Infinity)
```


Now, in WL there is an operator with precedence `390`: `Star`. If we try with it (in WMA):

```
In[1]:= Star[b, DirectedInfinity[-1]]                                                                                                                                        

Out[1]= b ⋆ -Infinity
```
we see that it does not add parenthesis, so the precendece value for `Star` is not higher that `390`. Notice that if `Star` is understood as the convolution product, this is not OK, but is what WL does.

rocky: This is good and important information. Give me some time to study this and compare with the existing tables and Robert's tables.

mmatera: Notice also that Robert's tables have several values depending on the context. I guess that  this is related to the use of PrecedenceForm in some internal  the format rules in WMA.

mathics/builtin/arithmetic.py
```diff
@@ -208,6 +208,13 @@ class Abs(_MPMathFunction):
     summary_text = "absolute value of a number"
     sympy_name = "Abs"
 
+    def eval(self, x, evaluation):
+        "%(name)s[x_]"
+        result = eval_abs(x)
```
rocky, line 213: Should follow convention using capital letter. Also it would be nice to start removing `%(name)s`. 

I will be putting in a PR to merge into this to correct such things.

mathics/builtin/arithmetic.py
rocky: Given that this is a much larger issue, and I think it is not going to be solved in the way proposed above, let's remove this change.  I created https://github.com/Mathics3/mathics-core/issues/811 to capture the discussion above, and track this problem.

rocky:
After CI clears - LGTM

PR #813, opened 2023-03-18
Try using autoload to define BesselI and BesselJ

rocky:
Not working - it Looks like autoload rules wipe out builtin rules


mmatera:
The problem is here:
https://github.com/Mathics3/mathics-core/blob/730a4885adb793754d5dcf70a01b1785e9b984eb/mathics/core/definitions.py#L167


Instead of update, what we should do is to look at all the keys in `this.user`, look for the merged definition (`self.get_definition(key)`) and replace the built-in definition by the merged one:


```
for key in self.user:
    self.builtin[key] = self.get_definition(key)

self.user = {}
```

mmatera:
Another possibility would be to temporarily replace self.user by a reference to self.builtin before loading the module:
```
user, self.user = self.user, self.builtin

... Autoload code ...
self.user = self.builtin
```

The effect would be the same, but is more hacky and less safe, because if the module fails to load, the definition is still overwritten.

mmatera:
> The problem is here:
> 
> https://github.com/Mathics3/mathics-core/blob/730a4885adb793754d5dcf70a01b1785e9b984eb/mathics/core/definitions.py#L167
> 
> Instead of update, what we should do is to look at all the keys in `this.user`, look for the merged definition (`self.get_definition(key)`) and replace the built-in definition by the merged one:
> 
> ```
> for key in self.user:
>     self.builtin[key] = self.get_definition(key)
> 
> self.user = {}
> ```

And talking about safety, this code should be inside the autoload function, to be run for each module separately.

rocky:
> 

I agree with you here about the potential problems.

rocky:
@mmatera If you are okay with this and the PR this is based on, then so am I.

test/builtin/specialfns/test_bessel.py
```diff
@@ -0,0 +1,32 @@
...
+import pytest
+
+
+@pytest.mark.parametrize(
+    ("str_expr", "str_expected", "assert_failure_msg"),
+    # Basically special rules from autoload/rules/Bessel.m that are not covered
+    # by SymPy.
+    [
+        (
+            "BesselI[1/2,z]",
```
rocky, line 16: I should note that the rule we have here for BesselI and BesselJ (from SymJa) apparently is not the most general rule in WMA. See for example: https://wolfram.com/xid/0e7q3bup-rrrwu

mmatera: Do you mean the spherical Bessel case?

rocky: > Do you mean the spherical Bessel case?

I found what I was looking at before: https://wolfram.com/xid/0cq0os16-dlr12e

From BesselI Possible Issues:
> For symbolic arguments they [the BesselI arguments] are: 
> In[2]: Bessel[11/2, X] 

Note that we only handle +/- 1/2.

rocky: And also under Specific values: 

> For half-integer orders, [BesselI](https://reference.wolfram.com/language/ref/BesselI.html) evaluates to elementary functions ...

mmatera: Yep, for half-integer indices, Bessel functions reduce to Spherical Bessel functions, which can be written in terms of  powers of e^{i r} and r. We can implement this later. By now, let's merge this as it is and then we can improve it.

PR #814, opened 2023-03-18
Improving autoload

mmatera:
This PR 

* set system as the default context in autoload files.
* merge the definitions in autoload files instead of overwriting the definition.
* as an example, implement some rules for Bessel functions.


rocky:
LGTM

PR #815, opened 2023-03-19
Parsing annotations and modernization

rocky:
I will start aadding RowBox, SuperscriptBox, SubscriptBox, UnderscriptBox, FractionBox, RadicalBox, FormBox parsing shortcuts. But before that there is this stuff to modernize the code according to current practice, type annotations, adding some literals, etc.


mmatera (approved):
LGTM

PR #818, opened 2023-03-20
adding Elements. Local symbols in rules/Bessel.m  inside a private context.

mmatera:
This PR adds the built-in `Element`, which is used in defining certain integrals. In particular, it would be useful to extend more the support for integral representations of special functions.

Also, I noticed that when `Set`/`SetDelayed` expressions with a LHS involving named patterns appear in autoload files, the temporal names can spoil the `System` namespace, making some test fail. A way to prevent that is to enclose the code of the modules in a `Begin`/`End` statement. Public symbols defined in the module appear then before the private block, by assigning a usage string.

### DIFF
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -705,20 +705,23 @@ class Element(Builtin):
       <dt>'Element[$expr$, $domain$]'
       <dd>returns $True$ if $expr$ is an element of $domain$
       <dt>'Element[$expr_1$|$expr_2$|..., $domain$]'
-      <dd>returns $True$ if all the $expr_i$ belongs to $domain$, and $False$ if \
-      one of the items doesn't.
+      <dd>returns $True$ if all the $expr_i$ belongs to $domain$, and \
+    $False$ if one of the items doesn't.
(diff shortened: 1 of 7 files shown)


mathics/builtin/arithmetic.py
```diff
@@ -696,6 +696,71 @@ def to_sympy(self, expr, **kwargs):
...
+    <dl>
+      <dt>'Element[$expr$, $domain$]'
+      <dd>returns $True$ if $expr$ is an element of $domain$
+      <dt>'Element[$expr_1$|$expr_2$|..., $domain$]'
+      <dd>returns $True$ if all the $expr_i$ belongs to $domain$, and $False$ if \
+      one of the items doesn't.
+    </dl>
+
+
+    Check if $3$ and $a$ are both integers. If $a$ is not defined, then 'Element' reduces the condition:
```
rocky, line 713: It would be nice not to have long lines by using `\`

mmatera:
What actually happens is that if the package fails at some point, then the `End[]` expression is not evaluated, and the context handler gets compromised. I can check that later.

PR #820, opened 2023-03-20
Unprotect RealNumberQ 

rocky:
This is not a WMA builtin so it should not be protected. Possibly we should remove this altogether and add ExactNumberQ

Rubi 5 defines RealNumberQ so this is a problem.

### DIFF
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -1128,6 +1128,8 @@ class RealNumberQ(Test):
      = True
     """
 
+    attributes = A_NO_ATTRIBUTES
+
     summary_text = "test whether an expression is a real number"
 
     def test(self, expr):


mmatera:
> I am now thinking that indeed implementing WMA ExactNumberQ and removing this altogether is a better way to go.
> 
> Your thoughts?

I agree: let's remove this symbol and implement the one which is in WL.

rocky:
> Just one comment: we use this symbol in a couple of places to check that some parameter is a Real number, using the syntax `_?RealNumberQ`. This should be transformed into something like `x_?(Not[ExactNumberQ[#1]] && Element[#1, Reals])&`
> 
> Another possibility would be to put `RealNumberQ` inside an internal context, like `` internals`tests` ``, in a way that keep the current form of the rule. Another possibility would be to make the check inside the `eval_` method.

Right now I am thinking one of these depending on the context. In #821 for the case of `Range[]` the better thing to do was to make a check inside the function since that is what WMA seems to do.  So we have been conformance there. However  I note that our `Range[]` conformance is still not exact - closer, but not exact.

rocky:
Superceded by #821

PR #823, opened 2023-03-23
Reduce RealNumberQ in builtin.image.basic and ...

rocky:
Disallow Python 3.6
move is_integer_rational_or_real to mathics.core.atoms

### DIFF
--- a/mathics/builtin/image/basic.py
+++ b/mathics/builtin/image/basic.py
@@ -46,13 +46,14 @@ class Blur(Builtin):
      = -Image-
     """
 
-    summary_text = "blur an image"
     rules = {
         "Blur[image_Image]": "Blur[image, 2]",
         "Blur[image_Image, r_?RealNumberQ]": (
             "ImageConvolve[image, BoxMatrix[r] / Total[Flatten[BoxMatrix[r]]]]"
         ),
     }
+
+    summary_text = "blur an image"
 
 
 class ImageAdjust(Builtin):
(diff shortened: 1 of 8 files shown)


mathics/builtin/image/basic.py
```diff
@@ -40,12 +46,15 @@ class Blur(Builtin):
      = -Image-
     """
 
-    summary_text = "blur an image"
     rules = {
         "Blur[image_Image]": "Blur[image, 2]",
-        "Blur[image_Image, r_?RealNumberQ]": "ImageConvolve[image, BoxMatrix[r] / Total[Flatten[BoxMatrix[r]]]]",
+        "Blur[image_Image, r_?RealNumberQ]": (
```
rocky, line 51: I haven't been able to figure out the replacement pattern here.

mmatera: One possibility would be    `r:(_Integer|_Real|_Rational)`. On the other hand, we can keep  `RealNumberQ` but put it in a private context.

mmatera: Here is another one: `r_/;(NumberQ[r] && (Im[r]==0))`. Also we can just ask for NumberQ, and then in the "body" check `r` has a nonzero imaginary part and send a message. Then pass `BoxMatrix` the real part.

mmatera: Still, there is another one: `r_?NumberQ`  and then define another rule specific for Complex numbers, that just show the warning.

In any case, I wonder why there is no keyword in WL that allows having the behavior we obtain when we implement the rule as a method,  and the method returns (Python's) `Null`.

rocky: One thing I am coming to understand is that this rewrite-rule approach generally leads to different behavior than WMA. 

First, when there is no match, Mathics3 silently fails whereas in WMA a type error message is given.

Second, when there is an error detected a rewrite may have already occurred leading to a result that was different from the original input given. For example If I write; 

```
ImageAdjust[x, "a"]
```

By the time the message occurs that "a" is invalid, This has already been rewritten to `ImageAdjust[x, {"a", 0, 1}]` and this is confusing. 

While on the one hand it is nice that there was an attempt to use more Mathics3 as opposed to Python, on the other hand it leads to less WMA conformance, and more user confusion. 

So again, to me this is another example of "too clever by a half".

rocky:
Merging and iterating...

PR #825, opened 2023-03-25
fix name_is_builtin_symbol

mmatera:
This function is used to check if `name` is the name of a builtin symbol defined inside a module. This PR reinstate a check that was previously removed for being fragile, using the `inspect` module.


rocky:
LGTM - and thanks!

PR #826, opened 2023-03-25
Arithmetic refactor 0

mmatera:
Another smaller chunk of #766.


mathics/eval/arithmetic.py
```diff
@@ -145,176 +155,213 @@ def append_last():
...
+        # For a sum, what is relevant is the minimum accuracy of the terms
+        with mpmath.workprec(prec):
+            numbers = (item.to_mpmath() for item in numbers)
+            number = mpmath.fsum(numbers)
+            return from_mpmath(number, precision=prec)
+    else:
+        return from_sympy(sum(item.to_sympy() for item in numbers))
+
+
+def eval_multiply_numbers(*numbers: Tuple[Number]) -> Number:
```
rocky, line 315: For now we should leave off a type annotation here. `Tuple[Number]` is not quite right.

mathics/eval/arithmetic.py
```diff
@@ -145,176 +155,213 @@ def append_last():
...
+        return None
+    return Expression(
+        SymbolTimes,
+        *elements,
+        elements_properties=ElementsProperties(False, False, True),
+    )
+
+
+def eval_add_numbers(
+    *numbers: Tuple[Number],
```
rocky, line 288: See comment below about the use of  use of `Tuple[Number]`

mathics/eval/arithmetic.py
rocky: `-> Number` is not statically correct either, so for now let's leave it off.  We have no guarantee that the return of `from_sympy()` is indeed a Number. Statically it could be a Symbol.

I think we would need something like `assert instance(return_value, Number)` for static typing to accept this.

rocky:
When adding type annotations, we also need a way verify what is getting added is correct. 

What have you been using to very the annotations? 

Some possibilities I have tried are pyright which hooks into GNU Emacs with lsp-mode. mypy is another possibility.

mmatera:
When I put the annotations, I remember I was looking in the internet and at the end, I just used the output of  `type(arg)`. Now I change this in a way that pyright does not complain.  In any case, I am pretty sure that the result of an arithmetic operation performed over `Number` should return a `Number`. For example, `eval_Power_number(IntegerMOne, RationalOneHalf)` should return `Complex(0,1)` and not `SymbolI`

mmatera:
I realized that `Log[-2]` indeed could produce an expression from SymPy (`Log[2] + I Pi`).

rocky:
> I realized that `Log[-2]` indeed could produce an expression from SymPy (`Log[2] + I Pi`).

Ok - I but I believe you are thinking about this the wrong way.: Type annotations is not about your ability to come pu with counterexamples to disprove or whatever theorems you can come up with to prove that types you declare are correct. Instead type annotation correctness has to do with what the computer system is easily able to deduce. 

Presumably over time Python will get better at this. But for now we have to live or adjust to its level.

As a specific example: 

```
def eval_multiply_numbers(*numbers: Number) -> Number:
       ...
        return from_mpmath(number)
```

Is not not correct from Python's standpoint.  `from_mpmath()` returns an `Atom` and this function says that a `Number` type is returned. And there a variable of type Atom does not necessarily have to be a Number.

So either `Number` should be changed to `Atom` for the above to be correct, or you would need something like: 

```
     result = from_mpmath(number)
     assert isinstance(result, Number)
     return Number
```

Again, we have to go with not the kinds of things we can prove, but rather what the level at which the Python type checkers work at.

mathics/eval/arithmetic.py
```diff
@@ -145,176 +155,213 @@ def append_last():
...
+    """
+    From a list of elements, produce two lists, one with the numeric items
+    and the other with the remaining. Differently than `segregate_numbers`,
+    this function assumes that elements are sorted with the numbers at
+    the begining.
+    """
+    for pos, element in enumerate(elements):
+        if not isinstance(element, Number):
+            return list(elements[:pos]), list(elements[pos:])
+    return list(elements), []
```
rocky, line 367: See https://github.com/Mathics3/mathics-core/pull/826#issuecomment-1483986319

mathics/eval/arithmetic.py
```diff
@@ -145,176 +155,213 @@ def append_last():
...
+        factors = (item.to_mpmath() for item in numbers)
+        number = mpmath.fprod(factors)
+        return from_mpmath(number)
+
+    prec = min_prec(*numbers)
+    if prec is not None:
+        with mpmath.workprec(prec):
+            factors = (item.to_mpmath() for item in numbers)
+            number = mpmath.fprod(factors)
+            return from_mpmath(number, prec)
```
rocky, line 335: See https://github.com/Mathics3/mathics-core/pull/826#issuecomment-1483986319

mmatera:
> > I realized that `Log[-2]` indeed could produce an expression from SymPy (`Log[2] + I Pi`).
> 
> Ok - I but I believe you are thinking about this the wrong way.: Type annotations is not about your ability to come up with counterexamples to disprove or whatever theorems you can come up with to prove that types you declare are correct. Instead type annotation correctness has to do with what the computer system is easily able to deduce.
> 
> Presumably over time Python will get better at this. But for now we have to live or adjust to its level.
> 
> As a specific example:
> 
> ```
> def eval_multiply_numbers(*numbers: Number) -> Number:
>        ...
>         return from_mpmath(number)
> ```
> 
> Is not not correct from Python's standpoint. `from_mpmath()` returns an `Atom` and this function says that a `Number` type is returned. And a variable of type Atom does not necessarily have to be a Number.
> 
> So either the return type `Number` should be changed to `Atom` for the above to be correct, or you would need something like:
> 
> ```
>      result = from_mpmath(number)
>      assert isinstance(result, Number)
>      return Number
> ```
> 
> Again, we have to go with not the kinds of things we can prove, but rather what the level at which the Python type checkers work at.

OK, now I think I understand your point better: `from_mpmath` in certain cases can return `SymbolInfinity` or `SymbolComplexInfinity`. This is why the return type is  `Atom`. `eval_add_numbers` calls that function and, independently that is not possible that by adding a finite list of numbers we get an infinity quantity (theorem), the return type for `eval_add_numbers` should be  `Atom` (type consistency), isn't it?

My question is now, would be fine to use `BaseElement` as the output type on these `eval_` functions?

mmatera:
BTW, `from_mpmath` sometimes returns `DirectedInfinity[1]`, so the return type annotation of this function is also wrong.

mathics/eval/arithmetic.py
```diff
@@ -1,4 +1,12 @@
-from typing import Callable, Optional, Tuple
+# -*- coding: utf-8 -*-
+
+"""
+helper functions for arithmetic evaluation, which do not
+depends on the evaluation context. Conversions to Sympy are
```
rocky, line 5: depends -> depend

mathics/eval/arithmetic.py
```diff
@@ -145,176 +155,213 @@ def append_last():
...
+    return items[True], items[False]
+
+
+# TODO: Annotate me
+def segregate_numbers_from_sorted_list(
+    *elements: BaseElement,
+) -> Tuple[List[Number], List[BaseElement]]:
+    """
+    From a list of elements, produce two lists, one with the numeric items
+    and the other with the remaining. Differently than `segregate_numbers`,
```
rocky, line 360: Differently than -> Different from

rocky:
LGTM - see small grammar corrections.    Some of the underlying intent that has been there from before I wonder about. But at some point in the future maybe we'll come back to this.

mmatera:
> LGTM - see small grammar corrections. Some of the underlying intent that has been there from before I wonder about. But at some point in the future maybe we'll come back to this.

OK. Let's merge this when the CI finishes. I hope that this helps to simplify and make the code more maintainable.

PR #827, opened 2023-03-26
improve from_mpmath. adding pytests

mmatera:
This PR improves the annotations and the implementation of `from_mpmath`. Also, some pytests are included.

### DIFF
--- a/mathics/core/convert/mpmath.py
+++ b/mathics/core/convert/mpmath.py
@@ -8,26 +8,30 @@ import sympy
 
 from mathics.core.atoms import (
     Complex,
+    Integer0,
     Integer1,
     IntegerM1,
     MachineReal,
     MachineReal0,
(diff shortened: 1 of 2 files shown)


rocky:
LGTM  - I like the tests! thanks for doing this. 

I wonder if I could interest you in doing the same for sympy conversions, and improving the conversions that you mentioned for square route or cubed root.

mmatera:
> LGTM - I like the tests! thanks for doing this.
> 
> I wonder if I could interest you in doing the same for sympy conversions, 

This would be an small detour from my plans, but sure. 

> and improving the conversions that you mentioned for square route or cubed root.

once #826 converges and lands, then the next step is to proceed with `eval_Power_*`. At that point, maybe I ask for your help to see if it is possible to handle that at the level of the conversions.

PR #828, opened 2023-03-28
Adding pytests for from_sympy / to_sympy. 

mmatera:
This PR adds some basic tests to check the consistency in the behavior of `from_sympy`. When I was writing the tests, I found some issues in the behavior of these functions, and also some failures in the modularity. So, this PR

* moves `mathics_to_sympy` and `sympy_to_mathics` dicts from `mathics.builtin` to `mathics.core.convert.sympy`. This allows for avoiding some local imports.
* `to_numeric_sympy_args` was moved from `mathics.core.convert.expression` to `mathics.core.convert.sympy`, which again helps to reduce local imports.
* moves the implementation of `Expression.to_sympy` and `Symbol.to_sympy` to `mathics.core.convert.sympy`.
* Constant expressions like `MATHICS_*INFINITY` and numbers `MATHICS_COMPLEX_I` were included in `mathics.core.expression` to avoid local duplication.
* Improves the implementation of `Symbol.to_sympy` by using a dictionary to translate some predefined symbols in SymPy, like `pi`, `E`, `GoldenRatio` or `I`.
* adds pytests for `from_sympy` and `to_sympy`.

In the code, it is also a proposal for a rewriting of `from_sympy` using dictionaries, which could help to get a clearer code.


rocky:
At first glance, looks like a welcome improvement. Would like to try this out and go over in detail later.

mathics/core/atoms.py
```diff
@@ -942,6 +942,8 @@ def is_zero(self) -> bool:
 
 
 RationalOneHalf = Rational(1, 2)
+MATHICS_COMPLEX_I = Complex(Integer0, Integer1)
```
rocky, line 945: We have been moving from Mathics to Mathics3, and this is a Mathics3-only change. Consider whether we want to use MATHICS or MATHICS3.

mathics/core/expression.py
```diff
@@ -1937,3 +1924,14 @@ def convert_expression_elements(
 
 def string_list(head, elements, evaluation):
     return atom_list_constructor(evaluation, head, "String")(elements)
+
+
+# These are constant expressions used in several places.
+# TODO: See if the is a better place to declare them.
```
rocky, line 1930: There is no reason this can't be put in its own separate module as was done between the symbols and systemsymbols module .

More smaller focused modules is better than one sprawling module

mmatera: OK, any suggestions for the name of such a module?

rocky: These all seem to be symbolic constants. 

So `mathics.core.symboli_constants`.

rocky: At one time I had thought that we might set elements_properties, but that is not needed - once set on the first use will no longer need to get set again. 

(And computing the properties for any of these is very quick.)

We might consider if it is worth it to turn this create an ExpressionConstant class which has a special and simple evaluate method (and other methods like that).

mmatera: OK, maybe I can make a separate PR with these objects and module, and then reformulate over it.

mmatera: I think this was solved in #831

rocky:
LGTM

PR #833, opened 2023-04-06
Doc part as separate files2

mmatera:
Just a rebase


mmatera:
Maybe the main issue here is that `BoxExpression`s dump and load from pickle is not working.

PR #834, opened 2023-04-08
Cache via lru_cache mpmath calls

rocky:
Also, a number of other small style changes, and grammar corrections were done as well.

### DIFF
--- a/mathics/eval/arithmetic.py
+++ b/mathics/eval/arithmetic.py
@@ -6,6 +6,7 @@ depend on the evaluation context. Conversions to Sympy are
 used just as a last resource.
 """
 
+from functools import lru_cache
 from typing import Callable, List, Optional, Tuple
 
 import mpmath
@@ -34,22 +35,39 @@ from mathics.core.systemsymbols import SymbolComplexInfinity, SymbolIndeterminat
 RationalMOneHalf = Rational(-1, 2)
 
 
-# @lru_cache(maxsize=4096)
 def call_mpmath(
-    mpmath_function: Callable, mpmath_args: tuple, prec: Optional[int] = None
-):
+    mpmath_function: Callable, mpmath_args: tuple, precision: Optional[int] = None
+) -> Optional[BaseElement]:
+    """A wrapper around ``call_mpmath_with_precision()`` which is defined below
+    below.
+
+    The parameter ``precision`` specifies the number bit of precsion
+    of the result.
+
+    If the parameter ``precision`` is not specified, IEEE-754 "double
+    precision" or (about 53 bits) is used.
     """
-    calls the mpmath_function with mpmath_args parms
-    if prec=None, use floating point arithmetic.
-    Otherwise, work with prec bits of precision.
+    if precision is None:
+        precision = FP_MANTISA_BINARY_DIGITS
+    return call_mpmath_with_precision(mpmath_function, mpmath_args, precision)
+
+
+@lru_cache(maxsize=4096)
+def call_mpmath_with_precision(
+    mpmath_function: Callable, mpmath_args: tuple, precision: int
+) -> Optional[BaseElement]:
     """
-    if prec is None:
-        prec = FP_MANTISA_BINARY_DIGITS
-    with mpmath.workprec(prec):
+    A wrapper that calls
+       mpmath_function(mpmath_args *mpmathargs)
+    setting precision to the parameter ``precision``.
+
+    The result is cached.
+    """
+    with mpmath.workprec(precision):
         try:
             result_mp = mpmath_function(*mpmath_args)
-            if prec != FP_MANTISA_BINARY_DIGITS:
-                return from_mpmath(result_mp, prec)
+            if precision != FP_MANTISA_BINARY_DIGITS:
+                return from_mpmath(result_mp, precision)
             return from_mpmath(result_mp)
         except ValueError as exc:
             text = str(exc)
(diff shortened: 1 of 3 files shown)


mathics/eval/arithmetic.py
```diff
@@ -34,22 +35,39 @@
...
+
+    The parameter ``precision`` specifies the number bit of precsion
+    of the result.
+
+    If the parameter ``precision`` is not specified, IEEE-754 "double
+    precision" or (about 53 bits) is used.
+    """
+    if precision is None:
+        precision = FP_MANTISA_BINARY_DIGITS
+    return call_mpmath_with_precision(mpmath_function, mpmath_args, precision)
```
TiagoCavalcante, line 52: Maybe this is silly, but can't this function call slow down the code as it is used in a lot of places?

mathics/eval/arithmetic.py
```diff
@@ -354,7 +372,7 @@ def segregate_numbers(
 # TODO: Annotate me
 def segregate_numbers_from_sorted_list(
     *elements: BaseElement,
-) -> Tuple[List[Number], List[BaseElement]]:
+) -> Tuple[List[BaseElement], List[BaseElement]]:
```
mmatera, line 375: The point of this function is that the first element of the output tuple should be `Number`s. But of course, they are also `BaseElement`s.

mathics/eval/arithmetic.py
mmatera: I remember I tried this, and I didn't see any improvement in performance, but neither a noticeable overhead. The problem is that in general, we do not call this function over the same values many times.

rocky: > Maybe this is silly, but can't this function call slow down the code as it is used in a lot of places?

In fact right now we can remove it since the single caller doesn't call this at all any more. 

> I remember I tried this, and I didn't see any improvement in performance, but neither a noticeable overhead. 

When this happens, it would be helpful to _communicate_ this, and then I can focus on things other than this.

> The problem is that in general, we do not call this function over the same values many times.

The situation right now is that we only call this from one place, gamma. And that's because mpmath seems to provide gamma while sympy doesn't  Will that always be the case? I don't know.

Unless I ran this over a large corpus of WMA code designed for benchmarking, I would be hesitant to offer generalizations based on insufficient data. 

At any rate, what to do?  The PR has a number of other small fixes. So I can: 

* So I can pull the changes to call_mpmath and leave the other changes. 
* Another possibility is to require `call_mpmath()` to pass the precision and not make it optional.   So basically `call_mpmath_with_precision` basically renamed back to  `call_mpmath` but there is no longer a default value. 
* Something else

Let me know which way to go.

mathics/eval/arithmetic.py
rocky: Again, there is a confusion about how type annotation works. It is not about what you can prove prove what the type system sees and its limitations. 

In the return on line 385 the type system is not smart enough reason that if there is an element that is not a number an early return is taken. 

From its less sophisticated level of understanding, the type of `*elements` has to match the return type of the first argument. The code can't guarantee that the type of *elements is a Number which is why it has to check in the loop.

mmatera: > In the return on line 385 the type system is not smart enough reason that if there is an element that is not a number an early return is taken.

If the program reaches the line 385, is because all the items of `elements` are `Number`, so indeed, the return type is correct, but OK. In that case, what about the line 361 of the `segregate_numbers` function?

mathics/eval/arithmetic.py
mmatera: > > I remember I tried this, and I didn't see any improvement in performance, but neither a noticeable overhead.
> 
> When this happens, it would be helpful to _communicate_ this, and then I can focus on things other than this.
> 

Yes, I planned to do that, but I didn't finish all the tests.

> > The problem is that in general, we do not call this function over the same values many times.
> 
> The situation right now is that we only call this from one place, gamma. And that's because mpmath seems to provide gamma while sympy doesn't Will there always only be a single caller? I don't know.
> 

I guess we can keep both callers, and at some point see where it is useful to specialize it.

> Unless I ran this over a large corpus of WMA code designed for benchmarking, I would be hesitant to offer generalizations based on insufficient data.
> 

This is the reason why I didn't report this. The argument about the frequency when the cache is used is just about my intuition on how I use WMA. Maybe in some context, it would be useful to call BesselJ over an array of data with many repeated values, and then it could make more sense. Also, I think that this depends on the particular function.


> At any rate, what to do? The PR has a number of other small fixes. So I can:
> 
>     * Pull the changes to call_mpmath and leave the other changes.
> 
>     * Another possibility is to require `call_mpmath()` to pass the precision and not make it optional.   So basically `call_mpmath_with_precision` basically is renamed back to  `call_mpmath` but there is no longer a default value.
> 
>     * Something else
> 
> 
> Let me know which way to go.

I think that there is not an appreciable overhead, this is fine as it is, and it is easy to remove. I just mentioned this because of the Thiago's comment.

mmatera:
Apart from the comment on the annotations, LGTM.

mathics/eval/arithmetic.py
rocky: > > In the return on line 385 the type system is not smart enough reason that if there is an element that is not a number an early return is taken.
> 
> If the program reaches the line 385, is because all the items of `elements` are `Number`, so indeed, the return type is correct, but OK. In that case, what about the line 361 of the `segregate_numbers` function?

That one mypy and other type checkers think is okay because there _that list never had a type_. It was implicitly defined `[]` so it matches anything. 

Here is what mypy thinks of arithmetic.py master: 

[mypy.log](https://github.com/Mathics3/mathics-core/files/11183533/mypy.log)

mmatera: OK, the linter rules. Merge then when you feel this is ready.

rocky:
In the last commit, I have remove `call_mpmath_with_precision` and opted for a more `call_mpmath()` that require precision. 

However in thinking about this, the original problem was not having an optional value per se, but that we weren't saving and restoring the precision when it wasn't the default.

PR #838, opened 2023-04-22
Correct pythmathics module loading...

rocky:
when built-in classes are imported properly.

Apparently pymathics importing broken when we started imported builtin-class functions properly in the pymathics modules.

### DIFF
--- a/mathics/builtin/__init__.py
+++ b/mathics/builtin/__init__.py
@@ -144,7 +144,20 @@ def name_is_builtin_symbol(module, name: str) -> Optional[type]:
         return None
 
     # Skip those builtins defined in or imported from another module.
-    if inspect.getmodule(module_object) is not module:
+
+    # rocky: I think this is a code smell. It doesn't feel like
+    # we should have to do this if things are organized and modularized
+    # well - builtins should have mostly top-level builtin function information,
+    # like doc definition, function attributes, rules, etc, and not
+    # lower-level things like Symbols, or helper funtions that would
+    # be imported. If the implementation of another builtins needs
+    # to call an eval method that would be found in mathics.eval;
+    # if a symbol is needed, that should be in mathics.core.systemsymbols.
+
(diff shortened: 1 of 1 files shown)


mmatera:
@rocky, this comes from the way in which we use the nested module structure to load Builtin symbols in the Definitions module, in the documentation, and in the consistency tests. 

In the `contribute` method, we want to load each Builtin just once. The way to do that is to walk over all the Builtin sub-modules, and look for the classes that belong to the module. Then, if the module `B` imports the class `C` from the module `A`, when the method walks on `B` avoid to load `A`.

In the documentation, on the other hand, we want to know where the Builtin class belongs, in order to put generate the documentation in the right section.  Something similar happens with the consistency tests.

If we want to put the exposed Builtins through the `__init__.py` file, then we should reformulate how the walk and load mechanism is implemented.

rocky:
> @rocky, this comes from the way in which we use the nested module structure to load Builtin symbols in the Definitions module, in the documentation, and in the consistency tests.
> 
> In the `contribute` method, we want to load each Builtin just once. The way to do that is to walk over all the Builtin sub-modules, and look for the classes that belong to the module. Then, if the module `B` imports the class `C` from the module `A`, when the method walks on `B` avoid to load `A`.
> 
> In the documentation, on the other hand, we want to know where the Builtin class belongs, in order to put generate the documentation in the right section. Something similar happens with the consistency tests.
> 
> If we want to put the exposed Builtins through the `__init__.py` file, then we should reformulate how the walk and load mechanism is implemented.

This description is a bit complicated, and in general, Python handles imports just fine by itself. My gut feeling is that the walk and load mechanism is probably too complicated and can be done simpler. But it feels like it would come at a bit of effort.  

But on the other hand there are just so many other things that are more important that in the grand scheme of things, this is not that important to address now. 

I suspect over time this kind of thing will become easier.

rocky:
@mmatera  is the comment revision in 5d97d18 accurate.

PR #839, opened 2023-04-23
Start adding Limit[] rules

rocky:
These were culled from symja_android_library/rules/LimitRules.m. More to come. 

I have noticed that as we add add more autoloaded rules, initial loading slows down. At some point I hope to rectify this by some sort of image loading mechanism that Lisps and Smalltalk have.

But also, we need to speed up readling loading overall.

### DIFF
--- a/mathics/autoload/rules/Limit.m
+++ b/mathics/autoload/rules/Limit.m
@@ -4,14 +4,14 @@
    These were culled from symja_android_library/rules/LimitRules.m.
  *)
 
-BeginPackage["Limit"]
+(* BeginPackage["Limit"] *)
 
-Begin["`private`"]
+(* Begin["`private`"] *)
 Unprotect[Limit];
 Limit[Tan[System`Limit`x_], System`Limit`x_Symbol->Pi/2] = Indeterminate;
 Limit[Cot[System`Limit`x_], System`Limit`x_Symbol->0] = Indeterminate;
-Limit[System`Limit`x_ Sqrt[2 Pi]^(System`Limit`x_^-1) (Sin[System`Limit`x_]/(System`Limit`x_!))^(System`Limitx_^-1), System`Limit`x_Symbol->Infinity] = E;
+Limit[System`Limit`x_ Sqrt[2 Pi]^(System`Limit`x_^-1) (Sin[System`Limit`x_]/(System`Limit`x_!))^(System`Limit`x_^-1), System`Limit`x_Symbol->Infinity] = E;
 Protect[Limit];
-End[]
+(* End[] *)
 
-EndPackage[]
+(* EndPackage[] *)
--- a/mathics/autoload/rules/Limit.m
+++ b/mathics/autoload/rules/Limit.m
@@ -14,9 +14,7 @@ To be false when it should be true.
  Huh?
  ****)
 
-(****
-Limit[Tan[x_], x_Symbol->Pi/2] = Indeterminate;
-Limit[Cot[x_], x_Symbol->0] = Indeterminate;
-Limit[x_*Sqrt[2*Pi]^(x_^-1)*(Sin[x_]/(x_!))^(x^-1), x_Symbol->Infinity] = E;
- ****)
+Limit[Tan[Sysetm`Limit`x_], System`Limit`x_Symbol->Pi/2] = Indeterminate;
+Limit[Cot[System`Limit`x_], System`Limit`x_Symbol->0] = Indeterminate;
+Limit[System`Limit`x_ Sqrt[2 Pi]^(System`Limit`x_^-1) (Sin[System`Limit`x_]/(System`Limit`x_!))^(System`Limitx_^-1), System`Limit`x_Symbol->Infinity] = E;
 Protect[Limit];
--- a/mathics/autoload/rules/Limit.m
+++ b/mathics/autoload/rules/Limit.m
@@ -5,7 +5,18 @@
  *)
 
 Unprotect[Limit];
-Limit[Tan[x_], x_Symbol->Pi/2] := Indeterminate;
-Limit[Cot[x_], x_Symbol->0] := Indeterminate;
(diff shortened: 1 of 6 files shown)


mmatera:
@rocky, I don't get why you comment out the rules. Is it because it slows down the loading?

mathics/autoload/rules/Limit.m
```diff
@@ -0,0 +1,22 @@
...
+   or not covered by SymPy.
+
+   These were culled from symja_android_library/rules/LimitRules.m.
+ *)
+
+Unprotect[Limit];
+(*****
+ Fixme  Uncommenting any of the below causes:
+
+  G[x_Real]=x^2; a={G[x]}; {x=1.; a, x=.; a} == {{1.}, {G[x]}}
```
mmatera, line 11: I guess the reason is because of the context of the patterns. What about using 
```
Limit[Tan[System`limits`x_], System`limits`x_Symbol->Pi/2]=Indeterminate;
```
?

rocky: I assume you meant `System`Limit`x_` ?  That seems to work. 

Does WMA require this kind of thing too? I would find it hard to believe because I see many packages out there that use short names without context qualification. 

And exactly why does changing one cause something that seems totally unrelated break?

mmatera: There are two different things. On the one hand, in WMA packages, the definitions are embraced on a ```Begin["context`"]/End[] ```
This prevents `x` and other symbols associated with patterns from being stored in the ```Global` ``` context.

The other thing is maybe related to this:

https://github.com/Mathics3/mathics-core/blob/49ed7c9f8ba08722836bc415dcd584ac96c91dba/mathics/core/definitions.py#L59

We set the context in autoload packages to ```System` ``` by default. Then, ```x``` is interpreted as ```System`x ``` instead of  ``` Global`x  ```, which hides this other definition.

rocky: Again, my question is - is this how WMA works? 

In particular - 

Does WMA have a mechanism whereby the user can load definitions that change the behavior like add rules like this to Built-in functions? 

If it does have a mechanism, does it also requery specifying  the "System" (and other nested) contexts? 

If it doesn't then we have a bug in Mathics3 somewhere.

mmatera: In Mathematica, by default, the Context used when Autoload packages are loaded is ``` Global` ```. Also, by looking at what happens if I define a variable in an autoload module and I clear it afterward, I would say that these definitions are not moved into the "builtin" definitions, as we does in Mathics.

For example, if I add the following file to the autoload folder:
``` (*  ~/Autoload/TestAutoload/init.m  *)

Print[$Context]

x="Hola"
```


each time I start a WMA REPL, ```  Global`   ```  is printed before the prompt 
```
 In[1]:=
```
Now, if I evaluate `x`,  I get
```
 In[1]:=x
Out[1] Hola
```
and 
```
 In[2]:=? x

Out[2]= Global`x
        x = "Hola"

```
Then,
```
 In[3]:= Clear[x]; x

Out[3]= x
```

mmatera:
Regarding the speed on loading these definitions, maybe we can also consider implementing something like `DumpSave`, and use it to load complex/ long set of  rules.

https://reference.wolfram.com/language/ref/DumpSave.html

rocky:
> Regarding the speed on loading these definitions, maybe we can also consider implementing something like `DumpSave`, and use it to load complex/ long set of rules.
> 
> https://reference.wolfram.com/language/ref/DumpSave.html



> Regarding the speed on loading these definitions, maybe we can also consider implementing something like `DumpSave`, and use it to load complex/ long set of rules.
> 
> https://reference.wolfram.com/language/ref/DumpSave.html

Yes, this is similar to what previous functional languages like Lisp and Smalltalk call image dumping and loading.

mathics/autoload/rules/Limit.m
rocky: > In Mathematica, by default, the Context used when Autoload packages are loaded is `` Global` ``. Also, by looking at what happens if I define a variable in an autoload module and I clear it afterward, I would say that these definitions are not moved into the "builtin" definitions, as we does in Mathics.
> 
> For example, if I add the following file to the autoload folder:
> 
> ```
>  (*  ~/Autoload/TestAutoload/init.m  *)
> 
> Print[$Context]
> 
> x="Hola"
> ```
> 
> each time I start a WMA REPL, `` Global` `` is printed before the prompt
> 
> ```
>  In[1]:=
> ```
> 
> Now, if I evaluate `x`, I get
> 
> ```
>  In[1]:=x
> Out[1] Hola
> ```
> 
> and
> 
> ```
>  In[2]:=? x
> 
> Out[2]= Global`x
>         x = "Hola"
> ```
> 
> Then,
> 
> ```
>  In[3]:= Clear[x]; x
> 
> Out[3]= x
> ```

Thanks for the explanation. So as I understand it, the "limits" in ``System`limits`x_`` was correct in that we just needed a new name that is not likely to be used elsewhere. "Limit" matches the builtin name, but all that matters is that it is not likely to be used elsewhere.

I guess putting things in System is a better or more protected than Global, and this awkwardness is limited just to predefined builtin functions. So I guess the way things are is good. For all we know Mathematics _does_ do a copy for these autoloaded rules that you do not see.

mathics/autoload/rules/Limit.m
```diff
@@ -0,0 +1,11 @@
+(* Additions to mathics.builtin.numbers.trig that are either wrong
+   or not covered by SymPy.
+
+   These were culled from symja_android_library/rules/LimitRules.m.
+ *)
+
+Unprotect[Limit];
```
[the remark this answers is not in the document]
rocky: Good idea. - I think this is now done, but I am not totally sure this is what was intended. 

If there is something missing, please just commit on top.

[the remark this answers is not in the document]
mmatera: I mean, instead of using the explicit context, using the begin/end structure. But I can do this in another round.
LGTM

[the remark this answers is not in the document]
rocky: Please do!

PR #841, opened 2023-04-24
Fix serialization for `Definitions`, `Builtin` and `MathicsSession` objects

mmatera:
This PR fixes a long-standing issue related to the compatibility of `Rule` and `Definitions` objects with serialization. With these changes, it is possible to store a `Definitions` object in a pickle file, and then load it afterward.

For example, now it is possible to store a MathicsSession and open it afterward:

```
import pickle
from mathics.session import MathicsSession()
session = MathicsSession()
session.evaluate("x=4")

with open("/tmp/session.pkl", "wb") as f:
    pickle.dump(session, f)
```

stores the `session` object in the file "/tmp/session.pkl". Then, we can load
the Mathics session in another Python session:

```
import pickle

with open("/tmp/session.pkl", "rb") as f:
    session = pickle.load(f)

session.evaluate("x=4")
```

This also could be useful to implement something like `DumpSave` https://reference.wolfram.com/language/ref/DumpSave.html  or to implement a faster start in Mathics clients.

### DIFF
--- a/mathics/core/rules.py
+++ b/mathics/core/rules.py
@@ -259,14 +259,25 @@ class BuiltinRule(BaseRule):
 
     def __getstate__(self):
         odict = self.__dict__.copy()
-        del odict["function"]
-        odict["function_"] = (self.function.__self__.get_name(), self.function.__name__)
+        # If function is a method of a builtin class
+        # store the name of the class and the name
+        # of the method instead of the callable.
+        if hasattr(self.function, "__self__"):
+            odict["function_"] = (
+                self.function.__self__.get_name(),
+                self.function.__name__,
+            )
+            del odict["function"]
         return odict
 
     def __setstate__(self, dict):
         from mathics.builtin import _builtins
 
         self.__dict__.update(dict)  # update attributes
-        class_name, name = dict["function_"]
-
-        self.function = getattr(_builtins[class_name], name)
+        # If "function_=(class_name, name)" is provided,
+        # then store the method, then store the method
+        # name of the builtin class class_name as the
+        # replacement function.
+        if "function_" in dict:
+            class_name, name = dict["function_"]
+            self.function = getattr(_builtins[class_name], name)
--- a/mathics/core/rules.py
+++ b/mathics/core/rules.py
@@ -259,31 +259,9 @@ class BuiltinRule(BaseRule):
 
     def __getstate__(self):
         odict = self.__dict__.copy()
-        # The attribute self.function stores the eval_ method
-        # associated to the rule, or an `UnavalableFunction` callable if
-        # the method requires a library that is not available.
-        # In order to store a method of a Builtin class,
-        # the "function" key is removed, and instead, a tuple
-        # ("class_name", "method_name") is stored in the key
-        # `"function_"`.
-        # Then, `__setstate__` can reconstruct the "function"
-        # value from these names.
-        if hasattr(self.function, "__self__"):
-            odict["function_"] = (
-                self.function.__self__.get_name(),
-                self.function.__name__,
-            )
-            del odict["function"]
         return odict
 
     def __setstate__(self, dict):
         from mathics.builtin import _builtins
 
         self.__dict__.update(dict)  # update attributes
-        # If "function_=(class_name, name)" is provided,
-        # then store the method, then store the method
-        # name of the builtin class class_name as the
-        # replacement function.
-        if "function_" in dict:
-            class_name, name = dict["function_"]
-            self.function = getattr(_builtins[class_name], name)
--- a/mathics/core/rules.py
+++ b/mathics/core/rules.py
@@ -265,7 +265,7 @@ class BuiltinRule(BaseRule):
         # In order to store a method of a Builtin class,
         # the "function" key is removed, and instead, a tuple
         # ("class_name", "method_name") is stored in the key
-        # `"function_"`. 
+        # `"function_"`.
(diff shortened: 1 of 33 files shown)


mmatera:
On the top of this, https://github.com/Mathics3/mathics-core/tree/cache_builtin_defs  is a proof of concept for testing the advantage of caching the builtin definitions in a file between sessions.

mathics/core/rules.py
```diff
@@ -259,14 +259,25 @@ def __repr__(self) -> str:
 
     def __getstate__(self):
         odict = self.__dict__.copy()
-        del odict["function"]
-        odict["function_"] = (self.function.__self__.get_name(), self.function.__name__)
+        # If function is a method of a builtin class
+        # store the name of the class and the name
+        # of the method instead of the callable.
```
rocky, line 264: If this is being done to make the definition pickle-able,  please state that here.

rocky:
When we are using a name as a proxy for specific code, we also need a way to ensure that the name written matches the code. Also this potentially opens up a security hole when loading a dumped file, because some malicious person may  definitions that are different from the standard definitions. 

I assume WMA does not allow DumpSave of its internals or has a way of checking tampering?

rocky:
@mmatera I closed this accidentally, will be reopening as soon as I figure out how.

mathics/core/rules.py
mmatera: Check if this makes more sense.

mmatera:
> When we are using a name as a proxy for specific code, we also need a way to ensure that the name written matches the code. Also this potentially opens up a security hole when loading a dumped file, because some malicious person may definitions that are different from the standard definitions.
> 

OK. Any idea to do that? In any case, here we are not storing code, but the name of a function and a class that should be present in a `mathics.builtin.` module /submodule. Then, this should not introduce any risk of code injection. But if there is a better way, please tell me.




> I assume WMA does not allow DumpSave of its internals or has a way of checking tampering?

I try to save ```` System`Plus ``` in WMA, and I guess that the builtin rules are not saved. But I can not be sure because it is a binary file.

mathics/core/rules.py
rocky: The comment is repeating what is done. But the _why_ is missing.  

There is much bravado over the benefit and what it provides, but the place where the crux of the change happens, there was no mention for the reason behind the change.

When one changes code for a very specific or important reason, we should state what the reason is at the place of change. 

Otherwise, someone may just revert the change because it seems complicated and the reason is not given.

rocky:
> > When we are using a name as a proxy for specific code, we also need a way to ensure that the name written matches the code. Also this potentially opens up a security hole when loading a dumped file, because some malicious person may definitions that are different from the standard definitions.
> 
> OK. Any idea to do that? In any case, here we are not storing code, but the name of a function and a class that should be present in a `mathics.builtin.` module /submodule. Then, this should not introduce any risk of code injection. 

We still have a problem that _our_ code changes over time and possibly in an incompatible way. And so someone can take advantage of that. 

> But if there is a better way, please tell me.

This part is not _wrong_, just _incomplete_. And the fear is that it can be a stepping stone for something that is potentially dangerous is further security measures are not added. 

Of course I have ideas on how to fix. They are not my ideas or observations. However the second I mention something you will put 2 and 2 together and get 22 and then thus I always find myself in "catch-up" mode where you've rushed ahead to implement something that was incompletely described or understood. 

If you want to merge this, okay. But please do not go further. 

> 
> > I assume WMA does not allow DumpSave of its internals or has a way of checking tampering?
> 
> I try to save ```` System`Plus ``` in WMA, and I guess that the builtin rules are not saved. But I can not be sure because it is a binary file.

mathics/core/rules.py
mmatera: Interesting. 
Maybe you can help me to improve the comment. The reason for the change is that at some point, we allow to the `Builtin.contribute` method to create rules which instead of calling an eval method, called an internal `unavailable_function` defined in the line 456 of the `mathics.builtin.base` module. The goal of this function is to show a message telling the user that the method is not available because a certain library is not available in the system.

This breaks the pickle mechanism, so here I proposed to replace it with a callable object, that essentially does the same but accepts to be pickled. 

Then we have two possibilities for the content of the `function` attribute: either it is a method of a `Builtin` class, or is a callable object, which can be pickled. If it is a method, then the code makes the same that it did before the PR, but if it another kind of callable, it just stores it.

In any case, if someone change this to a simpler form, and the new serialization test passes, everything would be OK.

rocky: Give me some time and I will happily add/revise the comment and you can review. 

There is a lot of information in these  PR  comments/discussion (in a not compact or even rambling form). 

The goal I think is basically to say what has been said here inside the code somehow.

mmatera:
I just tried and if I removed the special case and the `function_` hack, it seems to work too. 
Again, there is no rush here. I see that there was something broken, and I tried to fix it with the smaller possible changes. If this is useful, then merge this. Otherwise, just discard it.

rocky:
> I just tried and if I removed the special case and the `function_` hack, it seems to work too. Again, there is no rush here. I see that there was something broken, and I tried to fix it with the smaller possible changes. If this is useful, then merge this. Otherwise, just discard it.

This _is_ useful, much needed, and appreciated - thanks! It just needs a little touching up to make the intent clearer. 

I will try to get to this over the weekend. Again, thanks.

rocky:
LGTM - sorry for the delay.

PR #842, opened 2023-04-25
Enclosing autoload Limit rules between Begin/End context

mmatera:
This just adds the Begin/End block to the `autoload/rules/Limit.m`  module. The `BeginPackage`  is not used here because we are not trying to define a package, but just to define the default context.


rocky (approved):
Thanks for cleaning this up! 

I am not sure I would have been able to come up with this without a lot of work.

PR #843, opened 2023-04-29
Make explicit the need for serialization ...

rocky:
in Definition and Rule.

Some small linting was doine to shorten long lines, remove an uneeded import, correct a type annotation here and there.

test_serialization was expanded to include a rule test.

test_calculus was split and moved into test/builtin/calculus.

### DIFF
--- a/mathics/builtin/base.py
+++ b/mathics/builtin/base.py
@@ -60,10 +60,16 @@ no_doc = True
 
 class DefaultOptionChecker:
     """
-    Callable used to check options
+    Callable class that is used in checking that options are valid.
+
+    If initialized with ``strict`` set to True,
+    then a instantance calls will return True only if all
+    options listed in ``options_to_check`` are in the constructor's
+    list of options. In either case, when an option is not in the
+    constructor list, give an "optx" message.
     """
 
-    def __init__(self, builtin, options, strict):
+    def __init__(self, builtin, options, strict: bool):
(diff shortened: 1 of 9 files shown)


rocky:
Googling for pickle replacements turns up [dill](https://pypi.org/project/dill/) and JSON which seems less general.

We should check that dill works.

mmatera (approved):
LGTM

rocky:
Sorry for the delay  - this work you did is good and I think will allow us to separate module loading from definitions loading.

mmatera:
@rocky, thanks for the review and the improvements.

mmatera:
> Googling for pickle replacements turns up [dill](https://pypi.org/project/dill/) and JSON which seems less general.
> 
> We should check that dill works.

I remember that some years ago I tried dill, and for some reason pickle still was the best option. But I can look again.

rocky:
> > Googling for pickle replacements turns up [dill](https://pypi.org/project/dill/) and JSON which seems less general.
> > We should check that dill works.
> 
> I remember that some years ago I tried dill, and for some reason pickle still was the best option. But I can look again.

It might have been because pickle can store Python objects that have pointers in them. But this might not be a good thing. 

As for the security issues see https://medium.com/ochrona/python-pickle-is-notoriously-insecure-d6651f1974c9 and https://nedbatchelder.com/blog/202006/pickles_nine_flaws.html

PR #851, opened 2023-05-24
3.11 support preparation

rocky:
Things needed to run under Python 3.11. 

Start redoing conversion from Mathics3 regexps to Python regexps. 

Unit tests for this are needed. We might also consider filling out missing regexp patterns.


test/builtin/calculus/test_solve.py
```diff
@@ -8,6 +8,10 @@
 
 def test_solve():
     for str_expr, str_expected, message in (
+        # Clean the definitions, because
```
mmatera, line 11: @rocky, I pushed this directly here by mistake. In any case, this seems to make all the tests pass now.

mathics/builtin/string/regexp.py
```diff
@@ -24,9 +24,9 @@ class RegularExpression(Builtin):
      = RegularExpression[[abc]]
 
     ## Mathematica doesn't seem to verify the correctness of regex
-    #> StringSplit["ab23c", RegularExpression["[0-9]++"]]
-     : Element RegularExpression[[0-9]++] is not a valid string or pattern element in RegularExpression[[0-9]++].
-     = StringSplit[ab23c, RegularExpression[[0-9]++]]
+    #> StringSplit["ab23c", RegularExpression["[0-9]++)"]]
+     : Element RegularExpression[[0-9]++)] is not a valid string or pattern element in RegularExpression[[0-9]++)].
```
mmatera, line 28: Still, maybe the best would be to move this to a pytest.

mathics/eval/strings.py
```diff
@@ -1,13 +1,241 @@
...
+
+# A better thing to do would be to write a pymathics module that
 def eval_ToString(
     expr: BaseElement, form: Symbol, encoding: String, evaluation: Evaluation
 ) -> String:
     boxes = format_element(expr, evaluation, form, encoding=encoding)
     text = boxes.boxes_to_text(evaluation=evaluation)
     return String(text)
+
+
```
mmatera, line 71: Maybe this stuff belongs to a module `mathics.core.convert.regex` instead of `mathics.eval.string`

test/builtin/calculus/test_solve.py
rocky: Thanks! You should not feel that doing this is a mistake! In the future if you want to push directly to the branch, please do.

mathics/eval/strings.py
rocky: Yes, I think that is better. I was thinking that too.

mathics/builtin/string/regexp.py
rocky: I totally agree. The long-term plan is to remove all of the `#>` tests and move these to pytests. 

And the other aspect here, is that we should be testing `to_regex` more thoroughly there as well. 

When I say that the code is monolithic, part of what this means is that we have this module which contains the class definition for `StringSplit` as well as has regex conversion routine (that should be in `mathics.core.convert.regex`) and we test this as part of StringSplit instead of testing `to_regex()` directly.  

Testing via StringSplit makes it harder to pinpoint to_regex problems because we are going through layers of code  in order to get to the parts that we want (and want to strengthen) and when you find these the component you want to strengthen it is mingled with lots of other layers that are of different concerns. 

And then we also have a great deal code like this. This is just one of _several_ places where we have this. 

Other unrelated thing, there are many callers of `to_regex()` that are not quite right because they do not test to see if the result is None.  Those callers should be fixed.

rocky:
Although there is still quite a bit of work that could be done here, now that this is passing thanks to @mmatera we could consider merging and iterating. I leave it up to you @mmatera as to which way you'd prefer to go - continue here or merge and iterate.

mmatera:
I think that the only thing is missing is to add a workflow to test against  3.11
Then I think it would be ready to merge.

mmatera:
BTW: https://www.phoronix.com/review/python311-pyston-pypy

rocky:
> BTW: https://www.phoronix.com/review/python311-pyston-pypy

I too have noticed in Mathics3 that Pyston is faster than 3.11. We have had problems with PyPy though. We have never gotten the speed improvements that other allege we should get, and we don't get as complete a code base either due to various incompatibilities with PyPy.

mmatera:
> > BTW: https://www.phoronix.com/review/python311-pyston-pypy
> 
> I too have noticed in Mathics3 that Pyston is faster than 3.11. We have had problems with PyPy though. We have never gotten the speed improvements that other allege we should get, and we don't get as complete a code base either due to various incompatibilities with PyPy.

In the current state, is very hard to have a full compatibility with Pyston and PyPy because we have many dependencies, and many builtins. 

For example, each time I want configure  Mathics to use Pyston, I find issues with the Pillow version. Pillow should not be a dependency of the core...


My hope is that if at some point we can make the pymathics module interface stable, then we could split this huge mathics-core in a true "core" with few dependencies and fully compatible,  and many smaller modules with different levels of compatibility.

rocky:
> My hope is that if at some point we can make the pymathics module interface stable, 

This is a lot more work than you think.  

And even after that, with PyPy we'd have a less functional  Mathic3 than we get than either Pyston or Python 3.11.

PR #855, opened 2023-05-24
mathics.eval.string.to_regex -> mathics.core.convert.regex.to_regex

mmatera:
This PR just moves the `to_regex` function to the place that I think it belongs, with the other converters.


mmatera:
Something that came up to my mind is that certain functions that do not involve an evaluation require an  `Evaluation` parameter to show messages. We could simplify that kind of dependency by passing instead a callable, that shows the message. For example, the signature of `to_regex` could be
```
def to_regex(
    expr: Expression,
    messages: Callable=None,
    q=_regex_longest,
    groups=None,
    abbreviated_patterns=False,
)
```

and then, the caller could call it as
```
regex_str =to_regex(regex_expr, evaluation.message)
```

rocky:
> Something that came up to my mind is that certain functions that do not involve an evaluation require an `Evaluation` parameter to show messages. We could simplify that kind of dependency by passing instead a callable, that shows the message. For example, the signature of `to_regex` could be
> 
> ```
> def to_regex(
>     expr: Expression,
>     messages: Callable=None,
>     q=_regex_longest,
>     groups=None,
>     abbreviated_patterns=False,
> )
> ```
> 
> and then, the caller could call it as
> 
> ```
> regex_str =to_regex(regex_expr, evaluation.message)
> ```

Yes, I had that thought too. It would be better.

PR #856, opened 2023-05-24
Evaluation message to callable

mmatera:
This implements the idea of passing a `Callable` object instead of an `Evaluation` object when the object is just used to show a message. 
Also, a default `Callable` was introduced which does the same than `Evaluation.message` when an evaluation object is not available.

### DIFF
--- a/mathics/builtin/__init__.py
+++ b/mathics/builtin/__init__.py
@@ -26,6 +26,7 @@ import inspect
 import os.path as osp
 import pkgutil
 import re
+from sys import stderr
 from typing import List, Optional
 
 from mathics.builtin.base import (
@@ -35,6 +36,8 @@ from mathics.builtin.base import (
     SympyObject,
     mathics_to_python,
 )
+from mathics.core.convert.python import from_python
+from mathics.core.element import ensure_context
 from mathics.core.pattern import pattern_objects
 from mathics.core.symbols import Symbol
 from mathics.eval.makeboxes import builtins_precedence
@@ -184,6 +187,26 @@ def name_is_builtin_symbol(module, name: str) -> Optional[type]:
     if module_object in getattr(module, "DOES_NOT_ADD_BUILTIN_DEFINITION", []):
         return None
     return module_object
+
+
+def print_message(symbol_name: str, tag, *msgs, channel=stderr) -> "Message":
+    """
+    Print to the standard output a message using the default Builtin messages.
+    """
+    builtins = builtins_dict()
+    full_symbol_name = ensure_context(symbol_name)
+    builtin = builtins.get(full_symbol_name, None)
+    text = builtin.messages.get(tag) if builtin else None
+
+    if text is None:
+        text = builtins["System`General"].messages.get(tag)
+
+    if text is None:
+        text = "Message %s::%s not found." % (symbol_name, tag)
+
+    for i, arg in enumerate((from_python(arg) for arg in msgs)):
+        text = text.replace(f"`{i+1}`", str(arg))
+    print(text, file=channel)
 
 
 # FIXME: redo using importlib since that is probably less fragile.
--- a/mathics/builtin/__init__.py
+++ b/mathics/builtin/__init__.py
@@ -189,9 +189,11 @@ def name_is_builtin_symbol(module, name: str) -> Optional[type]:
     return module_object
 
 
-def print_message(symbol_name: str, tag, *msgs, channel=stderr) -> "Message":
+def print_message(symbol_name: str, tag: str, *msgs, channel=stderr):
     """
-    Print to the standard output a message using the default Builtin messages.
+    Print to the standard error a message using the default Builtin messages.
+    Using this instead `evaluation.message` instead of `evaluation.message`
+    when an `Evaluation` object is not available.
     """
     builtins = builtins_dict()
     full_symbol_name = ensure_context(symbol_name)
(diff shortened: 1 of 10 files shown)


mathics/builtin/files_io/filesystem.py
```diff
@@ -1350,7 +1350,10 @@ def eval_3(self, forms, paths, n, evaluation, **options):
         if options.get("System`IgnoreCase", None) is SymbolTrue:
             patterns = [
                 re.compile(
-                    "^" + to_regex(p, evaluation, abbreviated_patterns=True),
+                    "^"
+                    + to_regex(
```
rocky, line 1354: Note that there was a bug in both the original and here in that `to_regex()` can return `None` and then the string concatenation will raise an exception.

mathics/builtin/files_io/filesystem.py
```diff
@@ -1359,7 +1362,14 @@ def eval_3(self, forms, paths, n, evaluation, **options):
         else:
             patterns = [
                 re.compile(
-                    "^" + to_regex(p, evaluation, abbreviated_patterns=True) + "$"
+                    "^"
+                    + to_regex(
```
rocky, line 1366: See comment above regarding `to_regex()`. 

Since this code is exactly the same as the "then" branch and the only difference is whether `re.IGNORECASE`  is used, I think it would be clearer if this the list comprehension were turned into an internal function which takes a boolean "ignore_case" parameter. 

Again, I realize this is a flaw of the original code.

mathics/builtin/__init__.py
```diff
@@ -186,6 +189,28 @@ def name_is_builtin_symbol(module, name: str) -> Optional[type]:
     return module_object
 
 
+def print_message(symbol_name: str, tag: str, *msgs, channel=stderr):
```
rocky, line 192: We should not be adding more code to `builtin/__init__.py`  but instead should be removing it. 

It doesn't look like this is used anywhere. I would imagine it could be used in testing. Is it tested at all?  Inside a test routine is where it would belong.

rocky (commented):
LGTM - but the main thing I would like to see is `builtin.__init__.py`  not get larger, especially with code that is not used anywhere. 

It would be nice if some of the not-quite-right places of `to_regex()`  that do not check the return type were fixed, even though that is not the focus of this PR.

mathics/builtin/__init__.py
mmatera: Regarding this, I initially wrote it as a default message function for the functions that require one as a parameter, but in the end, I used it just for doing tests.  I can remove it from this and put it in another branch in case we want to use it for low-level tests.

mmatera: Done. The function is now in https://github.com/Mathics3/mathics-core/tree/context_free_print_message, in `mathics.builtin.base`

mathics/builtin/files_io/filesystem.py
mmatera: OK, if you have an idea about how it should be, and you are up for it, just commit the change over here.

rocky: Ok - will do later.

mathics/builtin/files_io/filesystem.py
mmatera: OK, but what would be the best? return `None`, an empty string or to raise an exception?

mmatera:
[7613968](https://github.com/Mathics3/mathics-core/pull/856/commits/76139681b4ee92bbb0cb0c593cbe512fab608078) is a possible way to handle the issue, which is compatible with the current use of `to_regex` in the code.

rocky:
> [7613968](https://github.com/Mathics3/mathics-core/pull/856/commits/76139681b4ee92bbb0cb0c593cbe512fab608078) is a possible way to handle the issue, which is compatible with the current use of `to_regex` in the code.

This has some advantages. There are places in the code where regexps are used like in listing files in a directory where it looks like we want to hide messages when something is invalid. Although this kind of thing should not happen, I am not sure we prevent that from happening.

mmatera:
> > [7613968](https://github.com/Mathics3/mathics-core/pull/856/commits/76139681b4ee92bbb0cb0c593cbe512fab608078) is a possible way to handle the issue, which is compatible with the current use of `to_regex` in the code.
> 
> This has some advantages. There are places in the code where regexps are used like in listing files in a directory where it looks like we want to hide messages when something is invalid. Although this kind of thing should not happen, I am not sure we prevent that from happening.

Ok, I leave this open so you can make the last adjustments.  Then merge it when you think it is ready.

mathics/builtin/files_io/filesystem.py
rocky: > OK, but what would be the best? return `None`, an empty string or to raise an exception?

This kind of thing has been debated a bit in programming languages. In Go and Rust they have gone the direction of returning `None` and the caller has to check. This is the more type-checked solution, because the compiler statically can detect whether this condition has been handled. 

In Python, C++ and many other languages the answer is looser. Python sometimes offers two methods, one that returns a special value and one that raises an error.  For example the `index` vesuss `find` methods on a string.

rocky:
I started making the change, but I see a bit of bitrot in `mathics.builtin.files_io.filesystem`. So after CI clears, I will merge this in and start a new PR with the more extensive changes.

There is more than one outright bug in the code. Another bug is trying `+ "$"` to compiled regexp.

PR #857, opened 2023-05-26
Arithmetic test zero

mmatera:
This is the part of #766 related to the tests of positivity / zero equality. Also, provides a better way to populate the `NUMERICAL_CONSTANTS`  dict.


rocky (approved):
LGTM - thanks for adding the tests.

PR #858, opened 2023-05-26
`session.parse`

mmatera:
Adding this method which is  handy for low-level tests.  
Also,
* Removing unused imports. 
* Adding symbols to mathics.core.systemsymbols


rocky (approved):
LGTM.

 The new Symbols in systemsymbols do not seem to be used anywhere yet. This is in anticipation for possible use in the future?  Maybe #859.  If so, #859 should be been based off of this and merged to this rather than master. 

Not a big deal, just something to keep in mind. 

Line 15 of session.py: 

```python
import mathics.settings
```

Looks like it is not used anywhere in the module. It probably could be removed.

rocky (approved):
LGTM.

 The new Symbols in systemsymbols do not seem to be used anywhere yet. This is in anticipation for possible use in the future? 

Line 15 of session.py: 

```python
import mathics.settings
```

Looks like it is not used anywhere in the module. It probably could be removed.

mmatera:
> LGTM.
> 
> The new Symbols in systemsymbols do not seem to be used anywhere yet. This is in anticipation for possible use in the future? Maybe #859. If so, #859 should be been based off of this and merged to this rather than master.
> 
> Not a big deal, just something to keep in mind..
> 
> Line 15 of session.py:
> 
> ```python
> import mathics.settings
> ```
> 
> Looks like it is not used anywhere in the module. It probably could be removed.

OK, let me merge this and then I can remove the unused import.

PR #862, opened 2023-05-28
Start to out mathics.builtin.files.io.filesystem

rocky:
Split out section previously under: Input/Output, Files, and Filesystem

There are a couple more sections: "User File Directories " and "System File Directories"  that should be split out as well. 

However this PR is already a bit large.


mmatera (approved):
LGTM

PR #864, opened 2023-06-03
Add --cprofile option to main and

rocky:
small styling stuff of misc files looked a in looking for preformance bottlnecks

### DIFF
--- a/mathics/main.py
+++ b/mathics/main.py
@@ -1,8 +1,16 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
+"""
+Simpler Command-line interface to Mathics3.
+
+See also mathicsscript for a more sophisticated, and full
+featured CLI, which uses in more add-on Python packages and modules
+to assist in command-line behavior.
+"""
 
 import argparse
 import atexit
+import cProfile
 import locale
 import os
 import os.path as osp
@@ -44,7 +52,8 @@ def show_echo(query, evaluation):
             stream = stream_manager.lookup_stream(strm.elements[1].value)
             if stream is None or stream.io is None or stream.io.closed:
                 continue
-        stream.io.write(query + "\n")
+        if stream is not None:
+            stream.io.write(query + "\n")
 
 
 class TerminalShell(MathicsLineFeeder):
@@ -305,10 +314,21 @@ Please contribute to Mathics!""",
         "multiple times)",
     )
 
+    # Python 3.7 does not support cProfile as a context manager
+    if sys.version_info >= (3, 8):
+        argparser.add_argument(
+            "--cprofile",
+            help="run cProfile on --execute argument",
+            action="store_true",
+        )
+
     argparser.add_argument(
         "--colors",
         nargs="?",
-        help="interactive shell colors. Use value 'NoColor' or 'None' to disable ANSI color decoration",
+        help=(
+            "interactive shell colors. Use value 'NoColor' or 'None' to disable "
+            "ANSI color decoration"
+        ),
     )
 
     argparser.add_argument(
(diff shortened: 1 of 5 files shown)


mathics/main.py
```diff
@@ -1,8 +1,16 @@
...
+Simpler Command-line interface to Mathics3.
+
+See also mathicsscript for a more sophisticated, and full
+featured CLI, which uses in more add-on Python packages and modules
+to assist in command-line behavior.
+"""
 
 import argparse
 import atexit
+import cProfile
```
mmatera, line 13: I just wonder, is this package available by default?

rocky: It is part of the standard library.

mmatera:
Just one small comment: this that was put inside the CLI, could be implemented as a symbol, in the same way that  `Timing` works. Then, it would be available for the different front-ends

rocky:
> Just one small comment: this that was put inside the CLI, could be implemented as a symbol, in the same way that `Timing` works. Then, it would be available for the different front-ends

Feel free to do or add on.

mmatera:
> > Just one small comment: this that was put inside the CLI, could be implemented as a symbol, in the same way that `Timing` works. Then, it would be available for the different front-ends
> 
> Feel free to do or add on.

I mean, some thing like that:
```
class PythonProfileEvaluation(Builtin):
    """
    <dl>
      <dt>'PythonProfileEvaluation[$expr$]'
      <dd>profile $expr$
    </dl>
    """
    summary_text = "profile the internal evaluation of an expression"
    
    def eval(self, expr:Expression, evaluation:Evaluation):
        "PythonProfileEvaluation[expr_]"
        import cProfile
        profile_dict = SymbolNull
        if sys.version_info >= (3, 8):
            with cProfile.Profile() as pr:
                result = expr.evaluate(evaluation)
                # pr.create_stats()
                # profile_dict = from_python(pr.stats)
                # comment mm: It is a pitty that Profile.print_stats
                # does not have an interface to pass a stream, or
                # any other way to get a string instead of a direct
                # print to stdout...
                pr.print_stats()
        else:
            result = expr.evaluate(evaluation)
        return ListExpression(result, profile_dict)
       
```

mmatera:
> > > Just one small comment: this that was put inside the CLI, could be implemented as a symbol, in the same way that `Timing` works. Then, it would be available for the different front-ends
> > 
> > 
> > Feel free to do or add on.
> 
> I mean, some thing like that:
> 
> ```
> class PythonProfileEvaluation(Builtin):
>     """
>     <dl>
>       <dt>'PythonProfileEvaluation[$expr$]'
>       <dd>profile $expr$
>     </dl>
>     """
>     summary_text = "profile the internal evaluation of an expression"
>     
>     def eval(self, expr:Expression, evaluation:Evaluation):
>         "PythonProfileEvaluation[expr_]"
>         import cProfile
>         profile_dict = SymbolNull
>         if sys.version_info >= (3, 8):
>             with cProfile.Profile() as pr:
>                 result = expr.evaluate(evaluation)
>                 # pr.create_stats()
>                 # profile_dict = from_python(pr.stats)
>                 # comment mm: It is a pitty that Profile.print_stats
>                 # does not have an interface to pass a stream, or
>                 # any other way to get a string instead of a direct
>                 # print to stdout...
>                 pr.print_stats()
>         else:
>             result = expr.evaluate(evaluation)
>         return ListExpression(result, profile_dict)
>        
> ```

The only problem with this is that it is not a direct way to get the text printed by `Profile.print_stats()` in the form of a `str`, in a way to be able to return it, or to show it as a message instead sending it to the stdout.

rocky:
> The only problem with this is that it is not a direct way to get the text printed by `Profile.print_stats()` in the form of a `str`, in a way to be able to return it, or to show it as a message instead sending it to the stdout.

I am sure there is a way to dig into the API to do these kinds of things. But let's leave that for some other time and some other PR.

PR #865, opened 2023-06-03
Move get_rules_list from element to expression...

rocky:
Also, the return type of get_sequence should be a tuple, not a Union.

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -748,6 +748,28 @@ class Expression(BaseElement, NumericOperators, EvalMixin):
                 option_values[name] = option.elements[1]
         return option_values
 
+    def get_rules_list(self) -> list:
+        """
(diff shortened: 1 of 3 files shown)


mathics/core/expression.py
```diff
@@ -748,6 +748,28 @@ def get_option_values(
                 option_values[name] = option.elements[1]
         return option_values
 
+    def get_rules_list(self) -> list:
```
mmatera, line 751: Since this happens just for Lists, which are always converted to `ListExpression`, why do not put this in that subclass?

rocky: I tried that and I got an error in test_assignment.py. I think for expressions we need to do something. 

And was also not 100% certain we never use Expression when we mean ListExpression.

mmatera (approved):
LGTM

PR #866, opened 2023-06-04
Another look over expression.py

rocky:
The motivation for this was originally just to note as a TODO that `get_rules_list()` should be eventually moved inside the `ListExpression` class. 

In doing that though LSP was showing a number of errors, some have been corrected. 

And then in breaking up some (but not all) of the long comment lines, I had a hard time understanding some of the  comments. I have tried to rephrase, but @mmatera please make sure I haven't mangled the meaning in the process.

The comment at the end of `rewrite_apply_eval_step`  I don't quite understand.

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -10,8 +10,6 @@ from typing import Any, Callable, Iterable, List, Optional, Tuple, Type
 import sympy
 
 from mathics.core.atoms import Integer, String
-
-# FIXME: adjust mathics.core.attributes to uppercase attribute names
 from mathics.core.attributes import (
     A_FLAT,
     A_HOLD_ALL,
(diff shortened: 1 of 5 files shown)


mathics/core/expression.py
```diff
@@ -1327,12 +1339,14 @@ def rules():
...
-        # Step 8: Update the cache. Return the new compound Expression and indicate that no further evaluation is needed.
+        # Step 8: Update the cache. Return the new compound Expression and
+        #        indicate that no further evaluation is needed.
         new._timestamp_cache(evaluation)
         return new, False
 
     #  Now, let's see how much take each step for certain typical expressions:
-    #  (assuming that "F" and "a1", ... "a100" are undefined symbols, and n0->0, n1->1,..., n99->99)
+    #  (assuming that "F" and "a1", ... "a100" are undefined symbols, and
+    #  n0->0, n1->1,..., n99->99)
```
mmatera, line 1349: The idea here was to add relative benchmarks to these cases. I thought I have done it, but they are not there... Maybe we can just move this as a pytest and directly benchmark it.

mmatera (approved):
LGTM

mathics/core/expression.py
rocky: Agreed - it would be better.

PR #870, opened 2023-06-25
Expose more settings...

rocky:
* Set `sys.set_int_max_str_digits() `on systems that need it - this is everything except pyston
* expose this setting and `MATHICS_CHARACTER_ENCODING` to front-ends via mathics.system_info

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,8 @@ Internals
 ---
 
 * ``eval_abs`` and ``eval_sign`` extracted from ``Abs`` and ``Sign`` and added to ``mathics.eval.arithmetic``.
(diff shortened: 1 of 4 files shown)


rocky:
I was looking at the [35 year celebration of Mathematica](https://www.youtube.com/watch?v=HxWg8exJxNY) and trying the examples given there. (This may be a Django workspace in the future.) 

In that I was seeing that `1989 ^ 1989` failed due to a CPython default max. This corrects that, and exposes a couple more settings.  

There will be a change in Django afterwards to pick up the settings.

rocky:
Feel free to comment on this at any time in the future.

PR #871, opened 2023-06-25
Move more code out of mathics.builtin.__init__

rocky:
Moved to mathics.eval.builtin. Possibly not the best place, but it is better and is a start.


mmatera:
LGTM. 

My only comment is that I think it `mathics.core.builtins` would be a better place than `mathics.eval.builtin` for the code used to load builtins. 

My guess is that at some point `mathics.core.builtin` namespace should contain only subclasses of `Builtin` that defines built in symbols, and the loading mechanisms should be in the "core" of the library.  But this is something we can discuss later: the split proposed here seems a good starting point.

rocky:
> LGTM.
> 
> My only comment is that I think it `mathics.core.builtins` would be a better place than `mathics.eval.builtin` for the code used to load builtins.
> 
> My guess is that at some point `mathics.core.builtin` namespace should contain only subclasses of `Builtin` that define built in symbols, and the loading mechanisms should be in the "core" of the library. But this is something we can discuss later: the split proposed here seems a good starting point.

I had been thinking about this too, and I think `mathics.core` is better. 

The bigger picture here is that I'd like to be able speed up initial startup time by loading only a minimal set of built in modules, and have the others load as they get called. In GNU Emacs this is called autoloading.

mmatera:
@rocky, thanks. Merge this when you feel it is ready.

PR #873, opened 2023-07-05
Add complex expand

mmatera:
@rocky, I open this PR to avoid losing the last changes. I think it is almost ready, but I put this as a draft in case you think there are pending things.


rocky:
One very small thing that was missing from ComplexExpand is to mention it in CHANGES.rst. 

And there was the duplication of the same expression as a doctest and a pytest. There are problem a lot more examples that could be tried - if they are of interest to users then a statement of what aspect this is showing interesting would be in the doctest. (I think that is missing in the current doctest that I started). Or if it is just an edge case that we want to test then it would go in pytest. 

I will try to look at it more detail late today.

PR #875, opened 2023-07-09
Move more builtin module loading code out of `mathics/builtin/__init__.py`

rocky:
Next iteration in sequence of refactorings. 

Here, the remaining code that should have been put into a function has been moved out of `mathics/builtin/__init__.py` and into `mathics.core.builtin_load` and imports have been adjusted which reduces circular dependencies. 

Remaining refactoring: 

Initiate the module loading from  the `mathics.core.builtin_load` module. 

Loading currently needs to be rewritten. It currently gets _all_ builtin module names. then from that returned collection imports _all_ module names, then from that collection ... _all_ ... _all_ ...

Instead,  of many "all" loops, we need to do follow through everything a specific module name.  To do everything, that is in one loop (over builtin module names) and no further loops would follow.

### DIFF
--- a/mathics/builtin/__init__.py
+++ b/mathics/builtin/__init__.py
@@ -16,81 +16,40 @@ builtin class such as the Builtin's Attributes, its Information text,
 among other things.
 """
 
-import glob
-import importlib
+import os
 import os.path as osp
-import pkgutil
-import re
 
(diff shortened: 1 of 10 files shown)


mmatera (approved):
LGTM

PR #877, opened 2023-07-13
Yet another round of documentation-inspired fixes

rocky:
Move graphics-drawing options under drawing. (The creation of this predated when we had guide sections. This not only makes more sense, but currently we can't reference chapters in LaTeX

Makefile: build we build documentation in include the non-test/tutorial Pymathics modules.

sed-hack.sh work around doctest bug that strips the "s" in a Section name.


mmatera (approved):
LGTM

PR #879, opened 2023-07-15
Move builtin loading code entirely out of __init__

rocky:
... YAY!

This paves the way for further improvements.

### DIFF
--- a/mathics/core/load_builtin.py
+++ b/mathics/core/load_builtin.py
@@ -12,6 +12,7 @@ import os
 import os.path as osp
 import pkgutil
 from glob import glob
+from types import ModuleType
 from typing import List, Optional
 
 from mathics.core.pattern import pattern_objects
@@ -19,13 +20,16 @@ from mathics.core.symbols import Symbol
 from mathics.eval.makeboxes import builtins_precedence
 from mathics.settings import ENABLE_FILES_MODULE
 
-# List of Mathics3 Builtin modules.
-# This is initialized via below import_builtins modules
-modules = []
+# List of Python modules contain Mathics3 Builtins.
+# This list used outside to gather documentation,
+# and test module consistency. It is
+# is initialized via below import_builtins modules
+mathics3_builtins_modules: List[ModuleType] = []
 
 _builtins = {}
 builtins_by_module = {}
 display_operators_set = set()
+
 
 # The fact that are importing inside here, suggests add_builtins
 # should get moved elsewhere.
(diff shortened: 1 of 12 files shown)


mathics/version.py
```diff
@@ -5,4 +5,4 @@
 # well as importing into Python. That's why there is no
 # space around "=" below.
 # fmt: off
-__version__="6.0.2dev0"  # noqa
+__version__="6.1.0dev0"  # noqa
```
rocky, line 8: This probably should be 7.0.0. Even before this change there were a number of API breaking changes. You'll see them in the recent Mathics3 Modules (natlang, and graph). And we'll have to change these more for this as well.

mathics/core/load_builtin.py
```diff
@@ -16,6 +17,11 @@
 from mathics.core.pattern import pattern_objects
 from mathics.core.symbols import Symbol
 from mathics.eval.makeboxes import builtins_precedence
+from mathics.settings import ENABLE_FILES_MODULE
+
+# List of Mathics3 Builtin modules.
+# This is initialized via below import_builtins modules
+modules = []
```
rocky, line 24: modules is a too vague a name. Probably mathics3_builtin_modules is better.

mathics/main.py
```diff
@@ -24,13 +24,16 @@
...
 from mathics.core.expression import Expression
+from mathics.core.load_builtin import import_and_load_builtins
 from mathics.core.parser import MathicsFileLineFeeder, MathicsLineFeeder
 from mathics.core.read import channel_to_stream
 from mathics.core.rules import BuiltinRule
 from mathics.core.streams import stream_manager
 from mathics.core.symbols import SymbolNull, strip_context
 from mathics.timing import show_lru_cache_statistics
 
+import_and_load_builtins()
```
rocky, line 35: Now that this section is isolated we could and should be able to time how long running this subroutine takes.

mathics/core/load_builtin.py
mmatera: agree

mmatera (approved):
All the changes look good. Merge this when you feel it is ready.

rocky:
running `mathics -e '1+1' ` with the timing decorator around loading builtin modules show that 2.2 seconds out of 7 elapsed are spent in loading. 

Next changes will be to be able to load builtin modules one at a time rather than as a set of sequential which do phases of module loading.

Issue #881, opened 2023-07-19
(sympy 1.12) test_calculus fails with AssertionError: Issue 208


badshah400:
## Description

On openSUSE Tumbleweed, our sympy packages have moved on to version 1.12, and now we see one failure when running pytest for Mathics 6.0.1 in test_calculus with the message: "AssertionError: Issue 208". I understand that the requirements specify sympy < 1.12, but this is not an option for us when building openSUSE packages for Mathics any more, which is why I just tried running the tests and reporting any issues here anyway.

## How to Reproduce

Using Mathics 6.0.1, python 3.9 and 3.10, numpy 1.24.2, and sympy 1.12, the error may be reproduced by running tests using pytest from the command line terminal.

## Output Given

```
FAILED test/test_calculus.py::test_calculus - AssertionError: Issue 208
```

Full build fail log attached: 
[log.zip](https://github.com/Mathics3/mathics-core/files/12095031/log.zip)

## Expected behavior

Tests should pass.

**All: Mathics version, Installed Python Python packages.

```
[   21s] Mathics 6.0.1
[   21s] on CPython 3.10.12 (main, Jun 28 2023, 17:56:55) [GCC]
[   21s] using SymPy 1.12, mpmath 1.2.1, numpy 1.24.2, cython 0.29.36
```

In addition

**Desktop (please complete the following information):**
 - OS: [e.g. iOS]: Linux/openSUSE Tumbleweed
 - Browser [e.g. chrome, safari]: NA

## Additional context

Found when trying to build RPM packages for openSUSE Tumbleweed

rocky:
Thanks for the information. I'll look at and take care of when I get a chance. 

As for the process though, since our API has become incompatible with 6.x, a branch off of the 6.0.1 release tag is needed and a change should be applied there. 

It may be that a parallel change may be applied to the master branch as well.

rocky:
@badshah400 Let us know if commit https://github.com/Mathics3/mathics-core/commit/4dccfb81a1c45d8a7068b62931dcb776110b94a2 or branch `601_basebranch` addresses things.

@mmatera thanks for addressing.

badshah400:
Absolutely perfect. Big thanks to both of you @mmatera and @rocky. Please feel free to close.

rocky:
@badshah400 Thanks for packaging on SuSE (and letting us know). 

[...]

PR #882, opened 2023-07-20
support for sympy 1.12

mmatera:
This PR allows to use sympy 1.12, without breaking the tests, fixing #881.


rocky:
This is good. Make a new branch from tag 6.0.1 (I think the SHA1 is 62b8b7f1 and open an additional PR applying this commit to that branch. Thanks.

mmatera:
For some reason, when I try to make a PR over these branches, github just shows me the differences, but no a PR button.

rocky:
The branch current

> For some reason, when I try to make a PR over these branches, github just shows me the differences, but no a PR button.

That is because the branch currently has the first commit as part of it.

rocky:
LGTM

PR #886, opened 2023-07-21
refactor arithmetic power

mmatera:
This is another chuck of #766. This part is related to the way in which powers are evaluated. 

Part of this code is dealing with the fact that the output from Sympy is not compatible with the way in which WL handles expressions involving powers. Tests were fixed here according to the behavior in WMA.

### DIFF
--- a/mathics/builtin/arithfns/basic.py
+++ b/mathics/builtin/arithfns/basic.py
@@ -7,6 +7,9 @@ on a calculator.
 
 """
 
+
+import sympy
+
 from mathics.builtin.arithmetic import create_infix
 from mathics.builtin.base import (
     BinaryOperator,
@@ -45,7 +48,6 @@ from mathics.core.symbols import (
     Symbol,
     SymbolDivide,
     SymbolHoldForm,
-    SymbolNull,
     SymbolPower,
     SymbolTimes,
 )
@@ -56,10 +58,17 @@ from mathics.core.systemsymbols import (
     SymbolInfix,
     SymbolLeft,
     SymbolMinus,
+    SymbolOverflow,
     SymbolPattern,
-    SymbolSequence,
 )
-from mathics.eval.arithmetic import eval_Plus, eval_Times
+from mathics.eval.arithmetic import (
+    associate_powers,
+    eval_Exponential,
+    eval_Plus,
+    eval_Power_inexact,
+    eval_Power_number,
+    eval_Times,
+)
 from mathics.eval.nevaluator import eval_N
 from mathics.eval.numerify import numerify
 
@@ -535,15 +544,15 @@ class Power(BinaryOperator, MPMathFunction):
     # Remember to up sympy doc link when this is corrected
     sympy_name = "Pow"
 
+    def eval_exp(self, x, evaluation):
+        "Power[E, x]"
+        return eval_Exponential(x)
+
     def eval_check(self, x, y, evaluation):
         "Power[x_, y_]"
-
-        # Power uses MPMathFunction but does some error checking first
-        if isinstance(x, Number) and x.is_zero:
-            if isinstance(y, Number):
-                y_err = y
-            else:
-                y_err = eval_N(y, evaluation)
+        # if x is zero
+        if x.is_zero:
+            y_err = y if isinstance(y, Number) else eval_N(y, evaluation)
             if isinstance(y_err, Number):
                 py_y = y_err.round_to_float(permit_complex=True).real
                 if py_y > 0:
@@ -557,17 +566,47 @@ class Power(BinaryOperator, MPMathFunction):
                     evaluation.message(
                         "Power", "infy", Expression(SymbolPower, x, y_err)
                     )
-                    return SymbolComplexInfinity
-        if isinstance(x, Complex) and x.real.is_zero:
-            yhalf = Expression(SymbolTimes, y, RationalOneHalf)
-            factor = self.eval(Expression(SymbolSequence, x.imag, y), evaluation)
-            return Expression(
-                SymbolTimes, factor, Expression(SymbolPower, IntegerM1, yhalf)
-            )
+                return SymbolComplexInfinity
 
-        result = self.eval(Expression(SymbolSequence, x, y), evaluation)
-        if result is None or result != SymbolNull:
-            return result
+        # If x and y are inexact numbers, use the numerical function
+
+        if x.is_inexact() and y.is_inexact():
+            try:
+                return eval_Power_inexact(x, y)
+            except OverflowError:
+                evaluation.message("General", "ovfl")
+                return Expression(SymbolOverflow)
+
+        # Tries to associate powers a^b^c-> a^(b*c)
+        assoc = associate_powers(x, y)
+        if not assoc.has_form("Power", 2):
+            return assoc
+
+        assoc = numerify(assoc, evaluation)
+        x, y = assoc.elements
+        # If x and y are numbers
+        if isinstance(x, Number) and isinstance(y, Number):
+            try:
+                return eval_Power_number(x, y)
+            except OverflowError:
+                evaluation.message("General", "ovfl")
+                return Expression(SymbolOverflow)
+
+        # if x or y are inexact, leave the expression
+        # as it is:
+        if x.is_inexact() or y.is_inexact():
+            return assoc
+
+        # Finally, try to convert to sympy
+        base_sp, exp_sp = x.to_sympy(), y.to_sympy()
+        if base_sp is None or exp_sp is None:
+            # If base or exp can not be converted to sympy,
+            # returns the result of applying the associative
+            # rule.
+            return assoc
+
+        result = from_sympy(sympy.Pow(base_sp, exp_sp))
+        return result.evaluate_elements(evaluation)
 
 
 class Sqrt(SympyFunction):
--- a/mathics/eval/arithmetic.py
+++ b/mathics/eval/arithmetic.py
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
 
 """
-arithmetic-related evaluation functions.
+helper functions for arithmetic evaluation, which do not
+depends on the evaluation context. Conversions to Sympy are
+used just as a last resource.
 
 Many of these do do depend on the evaluation context. Conversions to Sympy are
 used just as a last resource.
@@ -320,6 +322,28 @@ def eval_Sign(expr: BaseElement) -> Optional[BaseElement]:
     return sign or eval_complex_sign(expr)
 
 
+def eval_Sign_number(n: Number) -> Number:
+    """
+    Evals the absolute value of a number.
+    """
+    if n.is_zero:
+        return Integer0
+    if isinstance(n, (Integer, Rational, Real)):
+        return Integer1 if n.value > 0 else IntegerM1
+    if isinstance(n, Complex):
+        abs_sq = eval_add_numbers(
+            *(eval_multiply_numbers(x, x) for x in (n.real, n.imag))
+        )
+        criteria = eval_add_numbers(abs_sq, IntegerM1)
+        if test_zero_arithmetic_expr(criteria):
+            return n
+        if n.is_inexact():
+            return eval_multiply_numbers(n, eval_Power_number(abs_sq, RealM0p5))
+        if test_zero_arithmetic_expr(criteria, numeric=True):
+            return n
+        return eval_multiply_numbers(n, eval_Power_number(abs_sq, RationalMOneHalf))
+
+
 def eval_mpmath_function(
     mpmath_function: Callable, *args: Number, prec: Optional[int] = None
 ) -> Optional[Number]:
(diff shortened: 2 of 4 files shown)


mmatera:
@rocky, notice that this part was one of the most *polemical* in the discussion of #766. The reasons I have for insisting in this are:
* Provides a closer behavior to the WMA, so I can build tests that are compatible with its outputs.
* Despite the implementation of `eval_Power_`can be improved, I think that splitting it from the symbol definition would help to improve it in the future.
* In particular, this could help to think about how to improve the sympy back and forward conversions, if we still want to use it for these arithmetic evaluations.

rocky:
> @rocky, notice that this part was one of the most _polemical_ in the discussion of #766. The reasons I have for insisting in this are:
> 
> * Provides a closer behavior to the WMA, so I can build tests that are compatible with its outputs.
> * Despite the implementation of `eval_Power_`can be improved, I think that splitting it from the symbol definition would help to improve it in the future.
> * In particular, this could help to think about how to improve the sympy back and forward conversions, if we still want to use it for these arithmetic evaluations.

Ok - noted. I note this and will try to take a look at over the weekend.

mathics/builtin/arithfns/basic.py
```diff
@@ -520,6 +528,8 @@ class Power(BinaryOperator, _MPMathFunction):
     rules = {
         "Power[]": "1",
         "Power[x_]": "x",
+        "Power[I,-1]": "-I",
```
rocky, line 531: In master, this rule is not needed.: `Power[I, -1] ` returns `-I`. 

In this branch though I am seeing `(-I) ^ (-1 / 2)`.  Which one matches WMA. And why the discrepancy?

mathics/builtin/arithfns/basic.py
```diff
@@ -520,6 +528,8 @@ class Power(BinaryOperator, _MPMathFunction):
...
+        "Power[-1, 1/2]": "I",
```
rocky, line 532: As before, in master, this is not needed. Why is it needed here?

mathics/eval/arithmetic.py
```diff
@@ -319,48 +321,27 @@ def eval_complex_sign(n: BaseElement) -> Optional[BaseElement]:
     sign = eval_RealSign(expr)
```
rocky, line 321: I now see that SymPy has a "Sign" function that works over reals and complexes. 

I missed this fact before.  We should be using that.

test/builtin/arithmetic/test_basic.py
```diff
@@ -153,8 +153,8 @@ def test_multiply(str_expr, str_expected, msg):
         ("a  b  DirectedInfinity[q]", "a b (q Infinity)", ""),
         # Failing tests
         # Problem with formatting. Parenthezise are missing...
-        #        ("a  b  DirectedInfinity[-I]", "a b (-I Infinity)",  ""),
-        #        ("a  b  DirectedInfinity[-3]", "a b (-Infinity)",  ""),
+        ("a  b  DirectedInfinity[-I]", "a b (-I Infinity)", ""),
```
rocky, line 156: In this branch I am not seeing parenthesis around the result. I assume WMA adds parenthesis? 

I wonder if this is a formatting problem or a precedence problem

test/builtin/arithmetic/test_basic.py
```diff
@@ -153,8 +153,8 @@ def test_multiply(str_expr, str_expected, msg):
...
+        ("a  b  DirectedInfinity[-3]", "a b (-Infinity)", ""),
```
rocky, line 157: Same thing here.

test/builtin/arithmetic/test_basic.py
```diff
@@ -197,7 +197,7 @@ def test_directed_infinity_precedence(str_expr, str_expected, msg):
         ("I^(2/3)", "(-1) ^ (1 / 3)", None),
         # In WMA, the next test would return ``-(-I)^(2/3)``
         # which is less compact and elegant...
-        #        ("(-I)^(2/3)", "(-1) ^ (-1 / 3)", None),
+        ("(-I)^(2/3)", "(-1) ^ (-1 / 3)", None),
```
rocky, line 200: Hmm. master returns -1 right now, so what we have is wrong. 

Do you know what part of the code fixes this?

test/format/test_format.py
```diff
@@ -456,34 +456,53 @@
     "Sqrt[1/(1+1/(1+1/a))]": {
         "msg": "SqrtBox",
         "text": {
-            "System`StandardForm": "Sqrt[1 / (1+1 / (1+1 / a))]",
-            "System`TraditionalForm": "Sqrt[1 / (1+1 / (1+1 / a))]",
-            "System`InputForm": "Sqrt[1 / (1 + 1 / (1 + 1 / a))]",
-            "System`OutputForm": "Sqrt[1 / (1 + 1 / (1 + 1 / a))]",
+            "System`StandardForm": "1 / Sqrt[1+1 / (1+1 / a)]",
```
rocky, line 459: Why the change in test?

mathics/eval/arithmetic.py
```diff
@@ -688,8 +694,88 @@ def eval_Times(*items: BaseElement) -> BaseElement:
     )
 
 
+# Here I used the convention of calling eval_* to functions that can produce a new expression, or None
+# if the result can not be evaluated, or is trivial. For example, if we call eval_Power_number(Integer2, RationalOneHalf)
+# it returns ``None`` instead of ``Expression(SymbolPower, Integer2, RationalOneHalf)``.
+# The reason is that these functions are written to be part of replacement rules, to be applied during the evaluation process.
+# In that process, a rule is considered applied if produces an expression that is different from the original one, or
```
rocky, line 701: Something is weird here in terminology. 

In normal terminology, a rule is applied if it matches. period. It has nothing to do with whether things change. The WMA evaluation process has decided to look for fixed points for its own reasons.

mathics/eval/arithmetic.py
```diff
@@ -688,8 +694,88 @@ def eval_Times(*items: BaseElement) -> BaseElement:
     )
 
 
+# Here I used the convention of calling eval_* to functions that can produce a new expression, or None
```
rocky, line 697: I don't understand this comment. Based on past experience, I have a sinking feeling this is going off into some misguided direction. (I think of the days when text substitutions on MathML boxes was proposed as a way of addressing formatting?)

mathics/eval/arithmetic.py
```diff
@@ -688,8 +694,88 @@ def eval_Times(*items: BaseElement) -> BaseElement:
...
+# From my point of view, it would make more sense to use the following convention:
+#  * if the method has signature ``eval_method(...)->BaseElement:`` then use the prefix ``eval_``
+#  * if the method has the siguature ``apply_method(...)->Optional[BaseElement]`` use the prefix ``apply_`` or maybe ``repl_``.
+#
+# In any case, let's keep the current convention.
+#
+#
+
+
+def associate_powers(expr: BaseElement, power: BaseElement = Integer1) -> BaseElement:
```
rocky, line 727: Are we sure this kind of thing can't be done in Sympy. Have we tried this? 

For another thing brought up and in a previous PR, with Sign, I see in that a problem there was just about not tagging a variable as being in the right type (complex vs real).

rocky:
It feels to me there is more here than just 

> > @rocky, notice that this part was one of the most _polemical_ in the discussion of #766. The reasons I have for insisting in this are:
> > 
> > * Provides a closer behavior to the WMA, so I can build tests that are compatible with its outputs.

I would like more detail here. Be specific about what behavior is not close. Elsewhere I see specific differences in output even with this PR.  I think we need to understand in better detail what is is that is making the behavior different. 

> > * Despite the implementation of `eval_Power_`can be improved, I think that splitting it from the symbol definition would help to improve it in the future.

I am not opposed in theory to having something that can let us control how power works. But again, I'd appreciate understanding exactly _what_.  Here I see the associativity of turning repeated powers into multiplications. But I am not convinced Sympy can't do this as well. Or that we couldn't do this via a rule. 

> > * In particular, this could help to think about how to improve the sympy back and forward conversions, if we still want to use it for these arithmetic evaluations.

Until a good cases are made for not using sympy for arithmetic evaluation let's use sympy for arithmetic evaluation. In this PR I see that  for `I^(2/3)` where we get the wrong result, but it is not clear to me that this is the fault of Sympy versus how we feed stuff to Sympy. 

So let's take some specific examples like this see what's up here for specific examples.

rocky:
This code still seems to have bits of code that aren't associated with arithmetic power. Please let's split off anything that is not related to the association stuff and put that in a different PR so we can disentangle whatever issues those have from the association part.

test/builtin/arithmetic/test_basic.py
mmatera: In WMA
```
In[1]:= a b DirectedInfinity[-I]                                                

Out[1]= a b (-I Infinity)

In[2]:= %//FullForm                                                             

Out[2]//FullForm= Times[a, b, DirectedInfinity[Complex[0, -1]]]

```

On the other hand, in master,

```
In[1]:= a b DirectedInfinity[-I]
Out[1]= a b -I Infinity

In[2]:= %//FullForm
Out[2]//FullForm= Times[a, b, DirectedInfinity[Complex[0, -1]]]

```

Notice that the first output is wrong, and indeed, is a formatting problem.

rocky: Thanks for the information.  It looks like this is an operator precedence problem.

mathics/builtin/arithfns/basic.py
mmatera: Here the previous rule seems to be repeated. I have to check.

test/builtin/arithmetic/test_basic.py
mmatera: I guess this is related to the new rules you mention before...

mathics/eval/arithmetic.py
mmatera: `eval_` methods are used as the `function` attribute of a `BuiltinRule`. When the rule is applied, the `eval_` method is filled with the parameters from the pattern, and produce an output. If the output is `None`, the rule is assumed to "fail" and the original expression is kept. Otherwise, if the `eval_` method returns an `Element`, then this element is used as the result of the replacement, and the rule is considered  "applied".

Here we have certain `eval_` functions that are not `eval_` methods but can be used as a part of them. If you are up, please help me to rephrase the comment.

mathics/eval/arithmetic.py
mmatera: The point is not if we can find a way to use Sympy, but if we can do it "natively", without the cost of back-and-forth conversions. Even if we can make these conversions more efficient, avoiding them at all would always be a win, mostly if the native implementation is not too complicated.

mathics/eval/arithmetic.py
mmatera: `sympy.Sign` works like WL `Sign` and, even if it is true to force `sympy.Sign` as WL `RealSign`, it would mean doing an extra work in conversion, by checking that the expression is a Real expression. This can be done by hacking `to_sympy` to handle `RealSign` in a special way, but that work is more or less equivalent to performing the actual computation in pure Python, with the overhead of the back-and-forth conversion.

test/builtin/arithmetic/test_basic.py
mmatera: OK, I see the problem now: in this branch, the test is marked as "skip". I solved this precedence issue in another branch, and for some reason, the change was lost. I will try to put this fix and tests in another PR.

test/builtin/arithmetic/test_basic.py
mmatera: In any case, the test seems to be right. The problem is that if we activate it, it does not pass it. I can revert the change, but it will not have any effect until the underlying problem gets fixed, and the tests get activated.

test/builtin/arithmetic/test_basic.py
mmatera: Nop. The rule at the end seems to be useless. The computation is done by `mathics.eval.arithmetic.eval_Power_number`, which now is called from `Power.eval_check` ( `mathics.builtin.arithfns.basic`,  line 591).

test/format/test_format.py
mmatera: This is the effect of implementing the associativity of powers. In master, 

```
Sqrt[1 / (1 + 1 / (1 + 1 / a))]//FullForm
```

produces
```
Power[Power[Plus[1, Power[Plus[1, Power[a, -1]], -1]], -1], Rational[1, 2]]
```

while here,
```
Power[Plus[1, Power[Plus[1, Power[a, -1]], -1]], Rational[-1, 2]]
```

On the other hand, looking at WMA, it seems that it does not implement the associativity in this case either.

mathics/eval/arithmetic.py
mmatera: An alternative way to solve the same problem would be to add a parameter `expression` into the `eval_` methods. Then, we can just check if the result of the replacement is the same as the original expression without any extra cost. However, it will take to change the code in many many places...

mathics/eval/arithmetic.py
mmatera: BTW, `Sign[3 - 4 I]` can not be evaluated by using only sympy. The problems of using `sympy.sign` are shown in the tests by replacing the implementation of `mathics.eval.arithmetic.eval_Sign` by

```
def eval_Sign(expr: BaseElement) -> Optional[BaseElement]:
    """
    if expr is a number, return its sign.
    """
    sympy_expr = expr.to_sympy()
    if sympy_expr is not None:
        print("   sympy_expr", sympy_expr)
        result = from_sympy(sympy.sign(sympy_expr))
        if result is not None:
            return result
    return None

```


In a similar way, if we have expressions with products of  negative rational powers, the work of convert back the product to Mathics is more or less the same than evaluating the product without sympy.

mathics/eval/arithmetic.py
mmatera: To simplify this PR, I am going to remove these comments, that still are available in #766.

mathics/eval/arithmetic.py
mmatera: See comment above.

mathics/eval/arithmetic.py
rocky: Here is what I see when I evaluate the Python/Sympy equivalent of `Sign[3 - 4I]`: 

```
>>> from sympy import *
>>> sign(3 - 4j)
0.6 - 0.8*I
>>> 
````

It looks like Complex.to_sympy() is not doing the right thing.

mathics/builtin/arithfns/basic.py
rocky: I do not consider your different implementation as a good thing but a bad thing. Given the choice of relying on Sympy code or mmatera code we should go with Sympy. It is more used, more maintained ,and generally more well thought out and more complete. 

Our job should be to figure out how to get WMA thinks done in Sympy when it can.

PR #889, opened 2023-07-24
Start to reduce/refactor arithmetic ...

rocky:
Remove stuff from `mathics.builtin.arithmetic` that does not belong there:

* `_MPMathFunction` -> `MPMathFunction` and move to `mathics.buitin.base`
* `_MPMathMultiFunction` -> `MPMathFunction` and move to `mathics.builtin.base`
* `Abs`, `Piecewise`, `RealAbs`, `RealSign`, `Sign` moved to `mathics.builtin.numeric` to follow WMA organization better

The corresponding `eval_` routines will be gone over in another PR. This one is already large.

Url's gone over to make this not exceed standard line limit. Note that the formatting has been gone over to follow the existing pattern that we have been using.

cythonization in mathics.builtin class files removed. It is not clear this has benefit in modern Pythons, especially in this kind of builtin function and all of this needs to be retested if not rethought.

### DIFF
--- a/mathics/builtin/arithfns/basic.py
+++ b/mathics/builtin/arithfns/basic.py
@@ -2,12 +2,19 @@
 """
 Basic Arithmetic
 
-The functions here are the basic arithmetic operations that you might find on a calculator.
+The functions here are the basic arithmetic operations that you might find \
+on a calculator.
 
 """
 
-from mathics.builtin.arithmetic import _MPMathFunction, create_infix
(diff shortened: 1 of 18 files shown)


mmatera:
@rocky,
I am OK in general with these changes, but once merged, they will produce many merge conflicts with #886. I admit that #886 still requires a little work,  but that work is going to be easier if we avoid touching the involved files.

mmatera:
@rocky, I check it out and it seems that this just produces a moderated number of conflicts with  #886, so let's merge this if you feel it is ready.

PR #890, opened 2023-07-24
More on sympy

mmatera:
@rocky, this PR completes #888  by rewriting `from_sympy` in the form of a dispatch table instead of a very long  `if/elif` routine. Also, I tried to implement a way to avoid the conversions back-and-forth.
Finally, I included some code to implement `eval_Plus` and `eval_Times` fully based on Sympy.

### DIFF
--- a/mathics/core/symbols.py
+++ b/mathics/core/symbols.py
@@ -372,7 +372,7 @@ class Symbol(Atom, NumericOperators, EvalMixin):
 
     # __new__ instead of __init__ is used here because we want
     # to return the same object for a given "name" value.
-    def __new__(cls, name: str, sympy_dummy=None):
+    def __new__(cls, name: str):
         """
         Allocate an object ensuring that for a given ``name`` and ``cls`` we get back the same object,
         id(object) is the same and its object.__hash__() is the same.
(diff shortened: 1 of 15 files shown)


mathics/eval/arithmetic.py
```diff
@@ -49,6 +49,8 @@
 RealOne = Real(1.0)
 
 
+use_sympy_for_arithmetic = False
```
mmatera, line 52: Setting this variable to `True` we use a `eval_Times` and `eval_Plus` implementation based on sympy. Notice that it fails to pass several tests.

test/core/test_serialization.py
```diff
@@ -1,8 +1,11 @@
 import io
 import pickle
 
+from mathics.core.load_builtin import import_and_load_builtins
 from mathics.session import MathicsSession
 
+import_and_load_builtins()
```
mmatera, line 7: I had to add this line on each test that I want to run individually. Maybe it should be added in `mathics.session`.

mathics/core/symbols.py
```diff
@@ -402,18 +401,6 @@ def __new__(cls, name: str, sympy_dummy=None):
             # For example, this can happen with String constants.
 
             self.hash = hash((cls, name))
-
```
mmatera, line 405: Dummy symbols are something related to sympy and not with the basic Mathics data structures. It is not needed anymore.

mmatera:
Most of this is deprecated

PR #891, opened 2023-07-28
Times should have always at least 2 elements

mmatera:
This PR fixes a tiny inconsistency. In WMA, p`HoldForm[Times[expr]]` produces as a formatted output `Times[expr]`. Our current `format_times` function hides this, but the issue is revealed if we rewrite it to make it closer to the WMA routine.


rocky:
I tried in the master branch and in this branch and in both cases I see: 

```
In[1]:= HoldForm[Times[expr]]
Out[1]= expr
```

Is there a test that shows the difference?

mmatera:
> 

This is what you get in WMA:
```
In[1]:= HoldForm[Times[expr]]                                                   

Out[1]= Times[expr]

```

rocky:
LGTM. Still seems valid. Sorry this got overlooked.

PR #895, opened 2023-08-01
Moving private doctests to pytests. Fix the compatibility issue 0. I-> 0.

mmatera:
In this PR, I start to move some private doctests to Pytest. I also propose some extra related tests and check that the expected behavior matches with obtained in WMA.
From this, it came up that we are handling wrong how ```  0. I``` is evaluated: In master, it is evaluated to `0.`, but in WMA, it is evaluated to ```  Complex[0.`, 0.`] ```.
To match this behavior, I had to modify `from_mpmath`, and deactivate its lru_cache.  The problem now with lru_cache is that  `mpmath.mpf(0.)` produces the same hash than `mpmath.mpc(0., 0.)`, which looks wrong.

### DIFF
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -279,45 +279,6 @@ class Complex_(Builtin):
      = 1 + 2 I / 3
     >> Abs[Complex[3, 4]]
      = 5
-
-    #> OutputForm[Complex[2.0 ^ 40, 3]]
-     = 1.09951×10^12 + 3. I
-    #> InputForm[Complex[2.0 ^ 40, 3]]
-     = 1.099511627776*^12 + 3.*I
-
-    #> -2 / 3 - I
-     = -2 / 3 - I
-
-    #> Complex[10, 0]
-     = 10
-
-    #> 0. + I
-     = 0. + 1. I
-
-    #> 1 + 0 I
-     = 1
-    #> Head[%]
-     = Integer
-
-    #> Complex[0.0, 0.0]
-     = 0. + 0. I
-    #> 0. I
-     = 0.
-    #> 0. + 0. I
-     = 0.
-
-    #> 1. + 0. I
-     = 1.
-    #> 0. + 1. I
-     = 0. + 1. I
-
-    ## Check Nesting Complex
-    #> Complex[1, Complex[0, 1]]
-     = 0
-    #> Complex[1, Complex[1, 0]]
-     = 1 + I
-    #> Complex[1, Complex[1, 1]]
-     = I
     """
 
     summary_text = "head for complex numbers"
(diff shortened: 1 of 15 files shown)


mathics/builtin/arithmetic.py
```diff
@@ -999,7 +875,7 @@ class RealNumberQ(Test):
     >> RealNumberQ[0 * I]
      = True
     >> RealNumberQ[0.0 * I]
-     = True
+     = False
```
rocky, line 878: Recall there is no `RealNumberQ` in WMA. 

Eventually I'd like to remove RealNumberQ from Mathics3. It would be okay to have an `eval_RealNumberQ` though. Can we revise this revised to use that?

mathics/core/convert/mpmath.py
```diff
@@ -42,8 +56,8 @@ def from_mpmath(
         # HACK: use str here to prevent loss of precision
         return PrecisionReal(sympy.Float(str(value), precision=precision - 1))
     elif isinstance(value, mpmath.mpc):
-        if value.imag == 0.0:
-            return from_mpmath(value.real, precision=precision)
+        # if value.imag == 0.0:
```
rocky, line 59: If this is going to stay as a comment, then it would be helpful to explain why this does not work. For example, list an example where the evaluation does the wrong thing.

mmatera: I asked in the mpmath project https://github.com/mpmath/mpmath/issues/714. In any case, the problem with the lru_cache is fixed by passing the parameter `typed=True`.

rocky:
> To match this behavior, I had to modify from_mpmath, and deactivate its lru_cache. The problem now with lru_cache is that mpmath.mpf(0.) produces the same hash than mpmath.mpc(0., 0.), which looks wrong.

See comments in code for how this can probably be addressed keeping lru_cache.

Overall, this looks good. Thanks for undertaking this! 

LGTM - but see the other small comments.

mathics/builtin/arithmetic.py
rocky: Ok - but please consider doing this _before_ contemplating anything that includes big picture design work.

mmatera: Sure!

mathics/core/convert/mpmath.py
rocky: In another PR then let's reinstate the lru_cache adding the `typed=True` parameter. This is a better solution than the one proposed before.  But there a other places in the code where lru_cache was used, and the split function technique was used. That then should be adjusted not to split the function and use `typed=True` instead.

mathics/builtin/arithmetic.py
mmatera: I was investigating a little bit about this, and I found this discussion:https://mathematica.stackexchange.com/questions/18647/how-to-check-if-an-expression-is-a-real-valued-number

So, in WMA, there is an internal symbol that makes exactly what RealNumberQ does, but with a slightly different name.

mmatera: And actually, in the last version, `RealValuedNumberQ` is a system symbol.

PR #899, opened 2023-08-06
A second pass at redoing load_builtin.py

rocky:
Second pass at making it possible to load a single module inside `mathics.builti`  rather than all modules at once.
 
* the loop updating ``display_operators_set`` is now done inside module import rather than its own independent loop.    
*  `add_builtins_from_builtin_module()` isolated from inside `add_builtins_from_builtin_modules()` so that we can eventually remove  the latter.
*  More type annotations, docstrings and comments.


mmatera (approved):
LGTM

PR #903, opened 2023-08-13
Load single builtin module

rocky:
Inch forward towards being able to reduce startup time by loading only essential module.

The change here is that we can now import a single module by import name, e.g. `mathics.builtin.quantum_mechanics`. 

Note that frontends may need to reload sessions, which is not desirable.


test/core/test_load_builtin.py
```diff
@@ -0,0 +1,42 @@
...
+    Test that add_builtins_from_module() loads a single Builtin module
+    and updates definitions.
+    """
+    # Set up a session with all but one module.
+    # Then evaluate a builtin in that module and see that we
+    # now have the function defined.
+
+    # First, load in many modules except quantum_mechanics.
+    _builtins = {}
+    import_and_load_builtins(exclude_files={"quantum_mechanics"}, clean_all=True)
```
mmatera, line 22: Pytest loads all the tests before running them. So, as other unit tests load  `test.helper`, and it calls `import_and_load_builtins`, when we reach this test, all the modules are already loaded. So, if we want to do a test over a "fresh" instance of the module, we need this second parameter.

mmatera:
In the back of my mind, I have the feeling that it does not make too much sense to store these lists and dictionaries outside of the Definitions object. In any case, I guess this PR is a good step forward.

rocky:
> In the back of my mind, I have the feeling that it does not make too much sense to store these lists and dictionaries 
outside of the Definitions object. In any case, I guess this PR is a good step forward.

I agree they should not exist and are signs of things that are still misfeatures in the code. Some information is used by the document system. The builtin operator precedence tables may be legitimate somehow but probably not in this form.

It has taken something like 3 or 4 iterations just to get to this stage, which is annoying and has been maddening and reflects ill-designed things have been, make worse by "incremental" changes.  It reinforces why going forward we should be taking a more high-level view inspired more by existing patterns for interpreters and Python coding.

PR #904, opened 2023-08-14
moving private doctests from mathics.builtin.list

mmatera:
Another round of moving private doctests to pytests.


rocky:
I am seeing many tests removed but not added to test/builtin. Maybe these need to be `git add`'ed?

mmatera:
> I am seeing many tests removed but not added to test/builtin. Maybe these need to be `git add`'ed?

yep, I had forgotten that silly step...

rocky:
Thanks - LGTM.

PR #907, opened 2023-08-14
Move private doctests to pytest4

mmatera:
And also another round.


rocky:
LGTM after running `black test/builtin/test_strings.py` 

I look forward to the day when we can _remove_ `#>` from developer documentation and code. One step closer then to making it possible to ditch the home-grown doc system.

PR #908, opened 2023-08-19
files_io  doctest-> pytest

mmatera:
Here I started to take some nontrivial decisions:
* "..." does not work in pytests. I just checked the structure from the output by checking the match against a pattern, or by looking at the head and certain elements expected in the output.
* In some places of true Doctests, there are private lines (#>) for erasing variables, closing a file or deleting a file created by the test. I am not sure what to do with them.


mathics/builtin/files_io/files.py
```diff
@@ -528,29 +502,9 @@ class OpenRead(_OpenAction):
 
     >> OpenRead["ExampleData/EinsteinSzilLetter.txt", CharacterEncoding->"UTF8"]
      = InputStream[...]
-    #> Close[%];
+    >> Close[%];
 
     S> Close[OpenRead["https://raw.githubusercontent.com/Mathics3/mathics-core/master/README.rst"]];
```
mmatera, line 507: I keep these sandbox tests for another round, because I need to figure out how to mark them in the pytests. Any idea?

rocky: I don't have a thought right now about sandboxed tests. 

As for `#> Close[%];` vs. `>> Close[%];`,  if it is instructive that is okay. For example: 

```mathematica
>> OpenRead["ExampleData/EinsteinSzilLetter.txt", CharacterEncoding->"UTF8"]
     = InputStream[...]

Streams should be closed when done with them: 
 >> Close[%];
```

But if it is not, but just done to make testing more robust and for bookkeeping keep the `#>` so we don't pollute the docs.

mmatera: Good: I leave then `S>`for another round. Also, I have reverted  `#> Close` ->`>> Close` when it is not instructive.

mmatera: In Sphinx, the cleanup is implemented in a special code block directive `testcleanup`:
https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html

rocky:
LGTM - thanks. Sorry I didn't get to this sooner.

test/builtin/files_io/test_files.py
```diff
@@ -80,6 +80,262 @@ def test_close():
...
+            None,
+        ),
+        (
+            'OpenRead[""]',
+            ("File specification  is not a string of one or more characters.",),
+            "OpenRead[]",
+            None,
+        ),
+        (
+            'OpenRead["MathicsNonExampleFile"]',
```
rocky, line 143: @mmatera There are a problems with this test.

It assumes this file does not exist. Another test in this parameterize loop creates a file with exactly this name.

Instead, the test should ensure that a file does not exist is used. One way to do this is to create a _new_ named temporary directory which is empty and then look for some name there. https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory can help here.

So probably this should be removed from this parameterize loop and be put in a separate test that has the additional code

test/builtin/files_io/test_files.py
```diff
@@ -80,6 +80,262 @@ def test_close():
...
+            ("Cannot open MathicsNonExampleFile.",),
+            "OpenRead[MathicsNonExampleFile]",
+            None,
+        ),
+        (
+            'fd=OpenRead["ExampleData/EinsteinSzilLetter.txt", BinaryFormat -> True, CharacterEncoding->"UTF8"]//Head',
```
rocky, line 149: @mmatera The sequence of tests from here to line 172 should be put in a separate test. Subsequent tests depend on the actions of earlier tests.  If an early test fails, there is no cleanup. I am seeing for example "MathicsNonExampleFIle" appearing in the file system on failure and equally worse, in the mathics-core directory which should not be touched by testing. 

In that new separate test we should be doing cleanup when something goes wrong or a test fails. Context managers can help, or possibly the use of a temporary new temporary file or directory. 

For tests that are writing files, I do not think we are ensuring that we have access to a writable directory.

test/builtin/files_io/test_importexport.py
```diff
@@ -82,6 +81,185 @@ def test_export():
                 assert data.endswith("</svg>")
 
 
+"""
```
rocky, line 84: Double quotes _and_ comments? Comments with a TODO is probably good enough.

PR #923, opened 2023-09-22
fix some pylinter warnings

mmatera:
adding doctrings, using formatted strings, and simplifying conditionals.


rocky:
Thanks - let's just get this in.

PR #925, opened 2023-09-23
Move private doctests to pytest11

mmatera:
Another bunch of private doctests converted in pytests. Maybe the most drastic change here regarding the tests is that with these changes, the content of the syntax error messages is not tested anymore.


rocky:
LGTM - Thanks for moving forward on this!

PR #926, opened 2023-10-16
moving private doctests to pytests for physchemdata, tensor, statistics and vector

mmatera:
Another round


rocky:
LGTM - thanks for doing/continuing this.

PR #927, opened 2023-10-16
move private doctests to pytests for exp_structure and fuctional

mmatera:
and another


rocky:
Looks great.  As before, many thanks!

PR #928, opened 2023-10-16
moving private doctests to pytest for testing_expressions

mmatera:
and the last one for this Sunday...


rocky:
Great again - thanks!.

PR #931, opened 2023-10-17
MathicsSession.evaluate_as_in_cli

mmatera:
This PR implements a method in the `MathicsSession` class that uses the `Evaluation.evaluate` method. This allows to handle exceptions and special symbols like % or Line references. This method is used in certain pytests.

TODO: define a better name for the method and improve docstrings...


rocky:
I can't think of a better name. And this is an improvement on what we currently have: adds some docstrings and and removes some . So +1 for adding.

PR #935, opened 2023-11-16
Update tensors.py

Li-Xiang-Ideal:
Add LeviCivitaTensor


rocky:
Pretty good adherence to myriad of little details we have in adding a new built-in function.  Looks like you should run [isort](https://pypi.org/project/isort/3.8.1/) over this.  If you install [`pre-commit`](https://pypi.org/project/pre-commit/) then these checks will be run at `git commit` time.   

I suppose that too is missing from the developer's guide. 

A couple of other small things. Please move those Python `import` statements from inside functions to the top. Yeah, I find this counter-intuitive. In fact much of the Python style stuff I find counter-intuitive or wrong. But I have been beaten down by so many Python fascists who insist there is "only one way to write things in Python" that I have given up here.

Also, please consider adding an entry to `CHANGES.rst`.

But again, all of this is very small stuff. Thanks for doing this.

Li-Xiang-Ideal:
First time to experience isort and black 😅 Hope it passes the checks this time.

mmatera:
Very nice! Thank you very much! LGTM

rocky:
> Very nice! Thank you very much! LGTM

I agree! Thanks for sticking with this.

PR #945, opened 2023-12-04
Combine all ``rec()`` related to ``Outer``

Li-Xiang-Ideal:
``rec()`` and ``rec_sparse()`` in ``Outer`` is combined into ``unpack_outer``. Now anyone can write a new tensor-product-like function with it.

I found it a tough task to name the variables/functions clearly while avoiding too many line breaks caused by isort & black. So any suggestions on variable/function naming is welcomed.


rocky:
I am sorry I haven't been able to look a this. 

I hope to be able to look at this seriously next weekend. I have been short of time recently. Starting January 7, 2024 I should have more time though.

rocky:
I just started looking at this and overall I like it. Thanks for doing this. 

I will make some comments in the code in case you are interested in them. They are more as a "by the way" or "if you are interested", or "in case you didn't know" kind of thing rather than something that needs to change.

I okay though with merging the code as is. @mmatera your thoughts?

mathics/eval/tensors.py
```diff
@@ -76,6 +74,85 @@ def get_dimensions(expr, head=None):
...
+def to_std_sparse_array(sparse_array, evaluation: Evaluation):
+    if sparse_array.elements[2] == Integer0:
+        return sparse_array
+    else:
+        return Expression(
+            SymbolSparseArray, Expression(SymbolNormal, sparse_array)
+        ).evaluate(evaluation)
+
+
+def unpack_outer(item, rest_lists, current, level: int, const_etc: tuple):
```
rocky, line 86: As best as I can tell, this returns an BaseElement (a node of our M-Expression structure).

It would be helpful to indicate that as `-> BaseElement:`.  To do this though, you'll have to add at the top:

```python
from mathics.core import BaseElement
```

mathics/eval/tensors.py
```diff
@@ -76,6 +74,85 @@ def get_dimensions(expr, head=None):
...
+    to current (if at lowest level)
+
+    ``rest_lists``: the rest of lists to be unpacked
+
+    ``current``: the current lowest level elements
+
+    ``level``: the current level (unused yet, will be used in
+    ``Outer[f_, lists__, n_]`` in the future)
+
+    ``const_etc``: a tuple of functions used in unpacking, remains constant
```
rocky, line 107: One thing that is done for this kind of thing, is to have an inner function that is recursive and the other function then just serves to make the these variables accessible within the inner recursive function. 

```python
def unpack_outer(item, rest_lists, current, level: int, const_etc: tuple) -> BaseElement:
    (
        cond_next_list,  # return True when the next list should be unpacked
        get_elements,  # get elements of list, tuple, ListExpression, etc.
        apply_head,  # e.g. lambda elements: Expression(head, *elements)
        apply_f,  # e.g. lambda current: Expression(f, *current)
        join_elem,  # join current lowest level elements (i.e. current) with a new one
        if_nested,  # True for result as nested list ({{a,b},{c,d}}), False for result as flattened list ({a,b,c,d}})
        evaluation,  # evaluation: Evaluation
    ) = const_etc

     def _unpack_outer(item, reset_lists, current, level: int) -> BaseElement:
             # use con_next_list, get_elements, apply_head and so on.
	     # For recursive calls, use _unpack_outer()


     _unpack_outer(item, rest_lists, current, leven)

File: mathics/eval/tensors.py [hunk 3]
```diff
@@ -156,74 +233,51 @@ def eval_Outer(f, lists, evaluation: Evaluation):
...
-                        ListExpression(*(current[0] + _pos.elements)),
-                        current[1] * _val,
-                    ),
-                )
-
     # head != SparseArray
     if not head.sameQ(SymbolSparseArray):
-        return rec(lists[0], lists[1:], [])
+        etc = (
+            (lambda item, level: isinstance(item, Atom) or not item.head.sameQ(head)),
```
rocky, line 239: Many linters frown on using lambda and prefer `def` functions instead.

Here this would look like;

```python
            def sparse_get_item_fn(item, level):
	    	return: isinstance(item, Atom) or not item.head.sameQ(head)
	    def sparse_get_elements_fn(head, *elements)
```

As before, there could be a whole topic on the wisdom of deprecating lambda's which personally I find questionable other than the crappy irregular syntax of `lambda` that was a mistake made from a long time ago and not corrected.

I am just reporting what Python linters report and what my understanding of Python convention is.

rocky:
We might consider writing an independent unit test for "unpack_outer".  This would be in a file under `test/eval/test_tensor.py` but I see that right now we don't have _any_ tests for any of the functions under `mathics.eval`. This is our bad.

mathics/eval/tensors.py
Li-Xiang-Ideal: As a frequent mathematica user I'm used to pure functions. Given that there have been hundreds of ``lambda`` in Mathic3, I don't feel it necessary to rewrite pure functions like ``lambda elements: elements``, which is clear and can be written in a single line, with ``def``. However, I agree that rewriting longer pure functions with ``def`` is helpful. Now all pure functions longer than one line is replaced with ``def``.

mathics/eval/tensors.py
```diff
@@ -76,6 +74,88 @@ def get_dimensions(expr, head=None):
         return [len(expr.elements)] + sub
 
 
+def to_std_sparse_array(sparse_array, evaluation: Evaluation):
+    if sparse_array.elements[2] == Integer0:
```
mmatera, line 78: A docstring here would be welcome.

mathics/eval/tensors.py
```diff
@@ -76,6 +74,88 @@ def get_dimensions(expr, head=None):
...
+        get_elements,  # get elements of list, tuple, ListExpression, etc.
+        apply_head,  # e.g. lambda elements: Expression(head, *elements)
+        apply_f,  # e.g. lambda current: Expression(f, *current)
+        join_elem,  # join current lowest level elements (i.e. current) with a new one
+        if_nested,  # True for result as nested list, False for result as flattened list
+        evaluation,  # evaluation: Evaluation
+    )
+    ```
+    """
+    (
```
mmatera, line 124: I would think that passing a dictionary would be more "Pythonic" (explicit) than passing a magic tuple as an argument. One way to do this is

```
def unpack_outer(item, rest_lists, current, level: int, const_etc: dict):
      locals().update(const_etc)
```

mmatera, line 124: @Li-Xiang-Ideal, is there any advantage in using a magic tuple instead of a dictionary to pass these parameters?

mmatera:
@Li-Xiang-Ideal, thank you very much for this work.  Apart from the comments, which are optional, LGTM. Let's merge when you feel it is ready.

mathics/eval/tensors.py
rocky: Actually, the better thing to use here is what is called [NamedTuple](https://docs.python.org/3.11/library/typing.html?highlight=namedtuple#typing.NamedTuple)

Like a tuple this is immutable, which is what we want for clarity. And like a dictionary access is via a meaningful name. However this is accessed as an attribute of the object rather than as a key, which is also considered by those who are into the style kind of thing.

mmatera: @rocky, a NamedTuple would make sense to me if the bunch of functions were built in some place and then passed to other functions. Here it is just a bunch of parameters packed to be passed to a function, whose first instruction is to unpack it in local variables. For some reason, the standard structure used to do this in Python is the `dict`. Am I missing something?

mmatera: @Li-Xiang-Ideal, this was more a question for @rocky. Again, for me it is ok to merge this as it is now.

rocky: > Here it is just a bunch of parameters packed to be passed to a function, whose first instruction is to unpack it in local variables

IMHO then a tuple is fine and it happens to be efficient too.  

Use `*args`,  and `**kwargs` is frowned upon more nowadays except in very rare and custom circumstances.

One reason for this is that nowadays type annotations are desirable and used more by linters if not other kinds of things. 

I would not be surprised if, in the future, instead of being ignored in running the code, that they are used to speed up code as you find in mypy. This seems the direction that 3.12 is going towards. Because Guido is no longer in the center of things as before, attitudes have been shifting (in my opinion for the better). 

The problem with `*args`, and `**kwargs` is that these can't be type checked.

mathics/eval/tensors.py
rocky: You are correct that clarity over efficiency, especially micro-efficiency  is important. Micro-efficiency: I timed the current Python compiler and it _currently_ handles  this construct better than it handles that construct.

There is of course "order-of-magnitude efficiency", as in Big O notation efficiency, that is most important, but that rarely interferes with clarity. And this comes up so rarely, that we sometimes even forget to mention.

In 2024, I hope to deal with the order-of-magnitude efficiency. I don't see anything here impacting this that much. Or if it does, I am appreciative that, in contrast to before, this code is now to the stage where I think I and others would have no trouble understanding what is currently done in order to decide whether it should be changed as part of some sort of larger reorganization. 

I don't know what "self-containment" is so I can't comment on that. 

If this sometimes returns a list and other times a BaseElement, the way to annotate that is `Union[list, BaseElement]`  (This is the 3.7 compatible way to notate this. Newer Pythons have slicker notation for this using `|`.)

mathics/eval/tensors.py
rocky: > Here apply_head is identity function, so return apply_head(elements) will give a list instead of our M-Expression like ListExpression. This is because rules in SparseArray is a flattened list of pos -> val. So I set if_nested=False to automatically get a flattened list with elements.extend() and apply ListExpression manually in the end. 

Thanks for this clarification. Putting this is as a comment in the code would be helpful.

Li-Xiang-Ideal:
Apart from the suggestions above, I replace ``if_nested`` with ``if_flatten`` to make it clearer for Mathematica users.

Now there are two things left to do:
1. Decide what to use to pass the functions ``const_etc``, ``tuple``, ``NamedTuple`` or ``dict``? @rocky 
2. Write some unit tests for ``unpack_outer``.

p.s. In my vision, ``unpack_outer`` will not be called by other functions for at least several months, since the functions that might call it are basically more fundamental functions like ``Product`` and ``Table``. ``TensorProduct`` (which I intend to write in a few weeks) will definitely call it, but I think rules like ``TensorProduct[a__, b_, c_, d__]:=Tensor[a, Outer[Times, b, c], d]`` are enough.

rocky:
> Apart from the suggestions above, I replace `if_nested` with `if_flatten` to make it clearer for Mathematica users.

Excellent idea! 

> 
> Now there are two things left to do:
> 
> 1. Decide what to use to pass the functions `const_etc`, `tuple`, `NamedTuple` or `dict`? @rocky

tuple as I believe you had are simple and fine. 

> 2. Write some unit tests for `unpack_outer`.
> 

Much appreciated!

> p.s. In my vision, `unpack_outer` will not be called by other functions for at least several months, since the functions that might call it are basically more fundamental functions like `Product` and `Table`. `TensorProduct` (which I intend to write in a few weeks) will definitely call it, but I think rules like `TensorProduct[a__, b_, c_, d__]:=Tensor[a, Outer[Times, b, c], d]` are enough.

Ok. Thanks for the information!

mathics/eval/tensors.py
```diff
@@ -76,6 +74,88 @@ def get_dimensions(expr, head=None):
...
+        if cond_next_list(item, level):  # unpack next list
+            if rest_lists:
+                return _unpack_outer(
+                    rest_lists[0], rest_lists[1:], join_elem(current, item), 1
+                )
+            else:
+                return apply_f(join_elem(current, item))
+        else:  # unpack this list at next level
+            elements = []
+            for element in get_elements(item):
```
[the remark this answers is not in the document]
mmatera: BTW, I checked that this form of building  the `elements` array is 10% - 30% faster (if we do not count the cost of the `_unpack_outer` call)

Li-Xiang-Ideal:
Since the inital ``item, rest_lists, level`` of ``unpack_outer`` is always ``lists[0], lists[1:], 1``, I put them inside ``unpack_outer`` as the initial values of ``_unpack_outer``, leaving only ``list, current, const_etc`` as input arguments. Then ``unpack_outer`` is renamed to ``construct_outer`` to make its effect clearer.

In the next few days I won't have time to write more unit tests for ``construct_outer()``. The tests to be written are basically functions that can be implemented by ``construct_outer()``, here is a list of them:

- ``Sum``
- ``Product``
- ``Table``
- ``TensorProduct``

Li-Xiang-Ideal:
Now everything is done. If you don't have any comments it's ready to be merged.

rocky:
Very thorough - thank you for your attention to detail. I don't see anything. 

And if there happens to be more, like we need to generalize the kinds of things that `di` will allow or have several implementation for speed ups, that can go in another PR.

rocky:
Actually, I do see something that should eventually be addressed, but it is in `Range`: we do not describe the 3-argument form of this in the description of it, just in the examples. 

At some point, this should be addressed. 
Range

PR #946, opened 2023-12-12
Fix ArrayQ for SparseArray

Li-Xiang-Ideal:
At present ``ArrayQ`` returns ``False`` for ``SparseArray`` (which is inconsistent with Mathematica). This (draft) PR will fix this problem. 

It's also a preparation for ``TensorProduct`` that I plan to bring in in a few weeks, which will make rules of ``TensorProduct`` clearer.

For me there are two things of concern about this (draft) PR:
1. ``ArrayQ`` (``MatrixQ``) has been used throught out Mathics3. Though all tests have been passed, numerous problems may occur (e.g. ``Transpose[a_SparseArray]``, which returns the same expression as input now, but will cause an ``AttributeError`` in this PR). In a way this reminds us that there's still quite a bit to be done about ``SparseArrays``.
2. Rules like ``"ArrayQ[expr_]": "ArrayQ[expr, _, True&]"`` don't look efficient, all the evaluation on pattern matching and ``True&`` can be omitted entirely. Actually I've seen quite a few rules in Mathics3 like this: replacing a simple function with a more complicated but more general function. While these rules are unlikely to affect order-of-magnitude efficiencies, this issue may still worth attention.

### DIFF
--- a/mathics/builtin/testing_expressions/list_oriented.py
+++ b/mathics/builtin/testing_expressions/list_oriented.py
@@ -7,10 +7,10 @@ from mathics.core.builtin import Builtin, Test
 from mathics.core.evaluation import Evaluation
 from mathics.core.exceptions import InvalidLevelspecError
 from mathics.core.expression import Expression
-from mathics.core.rules import Pattern
 from mathics.core.symbols import Atom, SymbolFalse, SymbolTrue
-from mathics.core.systemsymbols import SymbolSubsetQ
+from mathics.core.systemsymbols import SymbolSparseArray, SymbolSubsetQ
 from mathics.eval.parts import python_levelspec
+from mathics.eval.testing_expressions import check_ArrayQ, check_SparseArrayQ
 
 
 class ArrayQ(Builtin):
@@ -39,6 +39,14 @@ class ArrayQ(Builtin):
      = False
     >> ArrayQ[{{a, b}, {c, d}}, 2, SymbolQ]
      = True
+    >> ArrayQ[SparseArray[{{1, 2} -> a, {2, 1} -> b}]]
+     = True
+    >> ArrayQ[SparseArray[{{1, 2} -> a, {2, 1} -> b}], 1]
+     = False
+    >> ArrayQ[SparseArray[{{1, 2} -> a, {2, 1} -> b}], 2, SymbolQ]
+     = False
+    >> ArrayQ[SparseArray[{{1, 1} -> a, {1, 2} -> b}], 2, SymbolQ]
+     = True
     """
 
     rules = {
@@ -51,37 +59,10 @@ class ArrayQ(Builtin):
     def eval(self, expr, pattern, test, evaluation: Evaluation):
         "ArrayQ[expr_, pattern_, test_]"
 
-        pattern = Pattern.create(pattern)
+        if not isinstance(expr, Atom) and expr.head.sameQ(SymbolSparseArray):
+            return check_SparseArrayQ(expr, pattern, test, evaluation)
 
-        dims = [len(expr.get_elements())]  # to ensure an atom is not an array
-
-        def check(level, expr):
-            if not expr.has_form("List", None):
-                test_expr = Expression(test, expr)
-                if test_expr.evaluate(evaluation) != SymbolTrue:
-                    return False
-                level_dim = None
-            else:
-                level_dim = len(expr.elements)
-
-            if len(dims) > level:
-                if dims[level] != level_dim:
-                    return False
-            else:
-                dims.append(level_dim)
-            if level_dim is not None:
-                for element in expr.elements:
-                    if not check(level + 1, element):
-                        return False
-            return True
-
-        if not check(0, expr):
-            return SymbolFalse
-
-        depth = len(dims) - 1  # None doesn't count
-        if not pattern.does_match(Integer(depth), evaluation):
-            return SymbolFalse
-        return SymbolTrue
+        return check_ArrayQ(expr, pattern, test, evaluation)
 
 
 class DisjointQ(Test):
--- a/mathics/builtin/testing_expressions/list_oriented.py
+++ b/mathics/builtin/testing_expressions/list_oriented.py
@@ -8,9 +8,9 @@ from mathics.core.evaluation import Evaluation
 from mathics.core.exceptions import InvalidLevelspecError
 from mathics.core.expression import Expression
 from mathics.core.symbols import Atom, SymbolFalse, SymbolTrue
-from mathics.core.systemsymbols import SymbolSparseArray, SymbolSubsetQ
+from mathics.core.systemsymbols import SymbolSubsetQ #, SymbolSparseArray
 from mathics.eval.parts import python_levelspec
-from mathics.eval.testing_expressions import check_ArrayQ, check_SparseArrayQ
+from mathics.eval.testing_expressions import check_ArrayQ #, check_SparseArrayQ
 
 
 class ArrayQ(Builtin):
@@ -39,14 +39,6 @@ class ArrayQ(Builtin):
      = False
     >> ArrayQ[{{a, b}, {c, d}}, 2, SymbolQ]
      = True
-    >> ArrayQ[SparseArray[{{1, 2} -> a, {2, 1} -> b}]]
-     = True
-    >> ArrayQ[SparseArray[{{1, 2} -> a, {2, 1} -> b}], 1]
-     = False
-    >> ArrayQ[SparseArray[{{1, 2} -> a, {2, 1} -> b}], 2, SymbolQ]
-     = False
-    >> ArrayQ[SparseArray[{{1, 1} -> a, {1, 2} -> b}], 2, SymbolQ]
-     = True
     """
 
     rules = {
@@ -59,8 +51,8 @@ class ArrayQ(Builtin):
     def eval(self, expr, pattern, test, evaluation: Evaluation):
         "ArrayQ[expr_, pattern_, test_]"
 
-        if not isinstance(expr, Atom) and expr.head.sameQ(SymbolSparseArray):
-            return check_SparseArrayQ(expr, pattern, test, evaluation)
+        # if not isinstance(expr, Atom) and expr.head.sameQ(SymbolSparseArray):
+        #    return check_SparseArrayQ(expr, pattern, test, evaluation)
 
         return check_ArrayQ(expr, pattern, test, evaluation)
 
--- a/mathics/builtin/testing_expressions/list_oriented.py
+++ b/mathics/builtin/testing_expressions/list_oriented.py
@@ -8,9 +8,9 @@ from mathics.core.evaluation import Evaluation
 from mathics.core.exceptions import InvalidLevelspecError
 from mathics.core.expression import Expression
 from mathics.core.symbols import Atom, SymbolFalse, SymbolTrue
-from mathics.core.systemsymbols import SymbolSubsetQ #, SymbolSparseArray
+from mathics.core.systemsymbols import SymbolSubsetQ  # , SymbolSparseArray
 from mathics.eval.parts import python_levelspec
-from mathics.eval.testing_expressions import check_ArrayQ #, check_SparseArrayQ
+from mathics.eval.testing_expressions import check_ArrayQ  # , check_SparseArrayQ
 
 
 class ArrayQ(Builtin):
--- a/mathics/eval/testing_expressions.py
+++ b/mathics/eval/testing_expressions.py
@@ -2,13 +2,15 @@ from typing import Optional
 
 import sympy
 
-from mathics.core.atoms import Complex, Integer0, Integer1, IntegerM1
+from mathics.core.atoms import Complex, Integer, Integer0, Integer1, IntegerM1
+from mathics.core.evaluation import Evaluation
 from mathics.core.expression import Expression
-from mathics.core.systemsymbols import SymbolDirectedInfinity
+from mathics.core.rules import Pattern
+from mathics.core.symbols import SymbolFalse, SymbolTimes, SymbolTrue
+from mathics.core.systemsymbols import SymbolDirectedInfinity, SymbolSparseArray
 
 
 def do_cmp(x1, x2) -> Optional[int]:
-
     # don't attempt to compare complex numbers
     for x in (x1, x2):
         # TODO: Send message General::nord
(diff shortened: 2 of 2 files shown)


Li-Xiang-Ideal:
I'm kind of tempted to delete this branch and create a new "Adaption_for_SparseArray" branch and put all the changes for SparseArray in it. What do you think about this idea? @rocky @mmatera 

Btw, would a more OOP solution like
```
class SparseArray(Builtin):
    ...
    def eval_transpose(self):  # or simply transpose() and put eval_transpose() in mathics/eval
        ...
```
be better?

mathics/builtin/testing_expressions/list_oriented.py
```diff
@@ -39,6 +39,14 @@ class ArrayQ(Builtin):
      = False
     >> ArrayQ[{{a, b}, {c, d}}, 2, SymbolQ]
      = True
+    >> ArrayQ[SparseArray[{{1, 2} -> a, {2, 1} -> b}]]
+     = True
+    >> ArrayQ[SparseArray[{{1, 2} -> a, {2, 1} -> b}], 1]
+     = False
```
rocky, line 45: Please keep in mind that this section is here to elucidate or educate users of some aspect of `ArrayQ`.  As such, usually we describe _why_ the example is given. For example:

```
'ArrayQ' works on sparse arrays:
>> ArrayQ[SparseArray[{{1, 2} -> a, {2, 1} -> b}]]
     = True
``` 

When one wants just wants to make sure all kinds of situations are covered, and this is not of that much interest to a user, then this goes should be added as part of a pytest.

mathics/eval/testing_expressions.py
```diff
@@ -99,3 +101,64 @@ def expr_min(elements):
 
 def is_number(sympy_value) -> bool:
     return hasattr(sympy_value, "is_number") or isinstance(sympy_value, sympy.Float)
+
+
+def check_ArrayQ(expr, pattern, test, evaluation: Evaluation):
```
rocky, line 106: Thanks for moving this out of mathics/builtin/testing_expressions/list_oriented.py

rocky:
> I'm kind of tempted to delete this branch and create a new "Adaption_for_SparseArray" branch and put all the changes for SparseArray in it. What do you think about this idea? @rocky @mmatera
> 
> Btw, would a more OOP solution like
> 
> ```
> class SparseArray(Builtin):
>     ...
>     def eval_transpose(self):  # or simply transpose() and put eval_transpose() in mathics/eval
>         ...
> ```
> 
> be better?

It is about your intent. If you intend to adapt Mathics3 so that SparseArray is more on equal footing as Array, then yes, naming the branch `adaption-for-SparseArray` is a better name. (The convention, although not a strong one, is to lowercase things unless something is clearer with capital letters like "SparseArray". Dashes rather than underscores are used, unless that is part of a name e.g. `correct-eval_transpose`.)

If however this is more about correcting ArrayQ for SparseArray and some slight other things are changed, then this branch name is fine.  Sometimes this kind of thing happens when one realizes that fixing everything in one go is too difficult and too many changes. Reviewing many smaller changes is easier on reviewers than one massive change. 

I hope with these guidelines you have enough information to decide which is best. If not, just ask.

Independently, it is not surprising to find that SparseArray has not been completely implemented. This kind of thing happens a bit here. SparseArray was added somewhat late in version 2.1.0

> I'm kind of tempted to delete this branch and create a new "Adaption_for_SparseArray" branch and put all the changes for SparseArray in it. What do you think about this idea? @rocky @mmatera
> 
> Btw, would a more OOP solution like
> 
> ```
> class SparseArray(Builtin):
>     ...
>     def eval_transpose(self):  # or simply transpose() and put eval_transpose() in mathics/eval
>         ...
> ```
> 
> be better?

Here are the reasons we have `mathics.eval` and `eval_xxx()` functions in that. We would eventually like to be in a position to create sequences of _instructions_ (not M-expressions) that can be run. For this, Object Oriented structuring hurts us, not helps. 

Similar to this,  when a function does not belong to a class, it is easier to call that function. For example, if I have some other builtin function that I am implementing and it needs to take the transpose of a matrix, it is faster for it to create lower-level calls to the underlying implementation functions than creating an M-expression and calling `.evaluate()` on that.  The Expression evaluation method is quite heavyweight and _has_ to be like that because general evaluation may have to sort arguments, look for and possibly run rewrite rules, look up and figure out which method to apply. 

For top-level evaluation of a user expression, this is powerful and expected and makes Wolfram Language what it is. However, it is also very slow when it is not needed. Most of the time when implementing a built-in function, all that generality is not needed or desired, but speed _is_ needed and desired.

mmatera:
@Li-Xiang-Ideal, thanks for looking at this. Indeed, the support for `SparseArray` expressions is quite limited right now.  

A first observation is that functions like `ArrayQ`, `Part`, or `Equal` do not work in SparseArray expressions like in other expressions. For example, `SparseArray[{{1,2}->3},{3,3}][[1]]` does not return the first element `{{1,2}->3}`  but another sparse array `SparseArray[<1>, {3}]`. This special behavior only applies if the sparse array expression is a valid sparse array. 
For instance, if an expression with a `SparseArray` head is not called with valid arguments, then WMA does not recognize it as an array. For example, in WMA,
```
In[1]:= SparseArray[3,Pi]                                                       

SparseArray::list: List expected at position 1 in SparseArray[3, Pi].

Out[1]= SparseArray[3, Pi]

In[2]:= ArrayQ[%]                                                               

Out[2]= False
```
But also, if it has the right form, but the data is invalid, the expression is not recognized as an array:
```
In[3]:= SparseArray[{{1,8}->3},{3,3}]                                           

SparseArray::posd: 
   The left-hand side of {1, 8} -> 3 in {{1, 8} -> 3}
     is not a position or a pattern that will match the position of an element
     in an array with dimensions {3, 3}.

Out[3]= SparseArray[{{1, 8} -> 3}, {3, 3}]

In[4]:= ArrayQ[%]                                                               

Out[4]= False
```

The idea that I had when I started implementing this part of the code is that valid `SparseArray` expressions should be stored as a special type of `Expression` (let's say, `SparseArrayExpression`), just like we do in  `mathics.core.list.ListExpression`. Then, the special behavior could be encoded in the interface of this specialized object.

mathics/eval/testing_expressions.py
mmatera: @Li-Xiang-Ideal,  I think that just this step is valuable to be put alone in a PR.

mmatera:
> I'm kind of tempted to delete this branch and create a new "Adaption_for_SparseArray" branch and put all the changes for SparseArray in it. What do you think about this idea? @rocky @mmatera
> 
> Btw, would a more OOP solution like
> 
> ```
> class SparseArray(Builtin):
>     ...
>     def eval_transpose(self):  # or simply transpose() and put eval_transpose() in mathics/eval
>         ...
> ```
> 
> be better?

Again, I think that the way to implement this is by following the model of `mathics.core.list.ListExpression`:  When a `SparseArray` expression is found, it is evaluated to a special kind of object (let's say, `SparseArrayExpression`). The corresponding class does not have all the functionality inside (like `Tr`, `Part` or `Transpose`) but helps to decide the functions that implement the evaluation of these expressions which function must use.

rocky:
> The idea that I had when I started implementing this part of the code is that valid SparseArray expressions should be stored as a special type of Expression (let's say, SparseArrayExpression), just like we do in mathics.core.list.ListExpression. Then, the special behavior could be encoded in the interface of this specialized object.

Yes, specializing down the line is probably a good idea.  Another thing in the back of my mind is to consider the ability to convert to uses scipy's representation of sparse arrays (when scipy is available).

In general, we should be using libraries from other packages than implementing our own versions which typically are lacking and less good. In the past there has been too much reinvention rather than using existing libraries. 

The challenge of course is to get the semantics right. But IMHO spend cleverness effort in that aspect rather than in writing a sparse-array package inside Mathics3.

mathics/eval/testing_expressions.py
rocky: I agree that this is probably enough of a first independent step if the goal is to fix up SparseArray. It is a reasonable-size PR.

mmatera:
Then, the special behavior could be encoded in the interface of this specialized object.
> 
> Yes, specializing down the line is probably a good idea. Another thing in the back of my mind is to consider the ability to convert to uses scipy's representation of sparse arrays (when scipy is available).
> 

Yes, I had that idea too. But I think that should be implemented in a module outside Mathics-core. 

> In general, we should be using libraries from other packages than implementing our own versions which typically are lacking and less good. In the past there has been too much reinvention rather than using existing libraries.
> 

The problem with a Scipy-based implementation is that Scipy is a big package to be included in the Mathics-core dependencies. Also, I do not think that Scipy.sparse is particularly designed to work with symbolic expressions. Also, what we need here is just some basic implementation for accessing elements, reshaping, and maybe matrix multiplication for one kind of internal storage. That part of the code could be borrowed from the scipy implementation.





> The challenge of course is to get the semantics right. But IMHO spend cleverness effort in that aspect rather than in writing a sparse-array package inside Mathics3.

rocky:
> The problem with a Scipy-based implementation is that Scipy ... 

We've gone over this before ad nauseum. There are solutions including the ones you've mentioned. For now the only thing to keep in mind is:

* we are not experts in how to implement sparse arrays, there are others who are
* we don't have the resources to do a good job, and in the past when this has been tried the results have been a bit lacking to the point of ripping out that code and replacing it with someone else's library; this has generally been a win in terms of functionality, speed, and reduced maintenance. 
* to a limited extent, we sometimes do have to provide fallback code, but we shouldn't go too deep down here without good reason and justification
* if other libraries and packages are lacking, we should be working with those folks to improve things.

Li-Xiang-Ideal:
> @Li-Xiang-Ideal, I think that just this step is valuable to be put alone in a PR.

From the above discussion I feel it might be better to simply move ``check_ArrayQ`` to ``mathics.eval`` in this PR/branch, and leave ``check_SparseArrayQ`` to be applied in future.

> If however this is more about correcting ArrayQ for SparseArray and some slight other things are changed, then this branch name is fine. Sometimes this kind of thing happens when one realizes that fixing everything in one go is too difficult and too many changes. Reviewing many smaller changes is easier on reviewers than one massive change.

Fixing ArrayQ for SparseArray is necessary, but it has too many side effects (that's why I submitted a draft PR). Of course, accumulating many small changes into one big one would be favorable for both developing and reviewing, but merging a PR with side effects into ``master`` seems a bit inappropriate to me. IMHO a potentially better solution would be to create a new branch called ``support-for-SparseArray`` and merge all these small changes with side effects into it. After everything is done in ``support-for-SparseArray``, we will merge it into ``master``; and before that, we can put in PRs to ``support-for-SparseArray``.

> The idea that I had when I started implementing this part of the code is that valid `SparseArray` expressions should be stored as a special type of `Expression` (let's say, `SparseArrayExpression`), just like we do in `mathics.core.list.ListExpression`. Then, the special behavior could be encoded in the interface of this specialized object.

> In general, we should be using libraries from other packages than implementing our own versions which typically are lacking and less good. In the past, there has been too much reinvention rather than using existing libraries.

Either writing a new ``SparseArrayExpression`` or utilizing existing packages could be a massive task. At present I'm completely unaware of things like Mathics3's 2024 annual plan. If it's not already fully scheduled, add a paragraph like
```
TODO: Support for SparseArray
1. functions that need to be adapted to sparse arrays (e.g. ArrayQ, MatrixQ)
    ...
    their side effects to be resolved (e.g. LinearModelFit[{m_?MatrixQ, v_}])
    ...
2. SparseArrayExpression or using libraries from other packages
    ...
```
might be helpful in my humble opinion.

mmatera:
> From the above discussion I feel it might be better to simply move `check_ArrayQ` to `mathics.eval` in this PR/branch, and leave `check_SparseArrayQ` to be applied in future.

Agree.

> 
> > If however this is more about correcting ArrayQ for SparseArray and some slight other things are changed, then this branch name is fine. Sometimes this kind of thing happens when one realizes that fixing everything in one go is too difficult and too many changes. Reviewing many smaller changes is easier on reviewers than one massive change.
> 
> Fixing ArrayQ for SparseArray is necessary, but it has too many side effects (that's why I submitted a draft PR). Of course, accumulating many small changes into one big one would be favorable for both developing and reviewing, but merging a PR with side effects into `master` seems a bit inappropriate to me. IMHO a potentially better solution would be to create a new branch called `support-for-SparseArray` and merge all these small changes with side effects into it. After everything is done in `support-for-SparseArray`, we will merge it into `master`; and before that, we can put in PRs to `support-for-SparseArray`.
> 


One possibility we have discussed before would be to implement the SparseArray support as an external PyMathics module, like NatLang or Graph. If this mechanism is still immature, is less immature than three years ago, and allows us to 
extend the implementation of Mathics symbols without compromising Mathics-core. However, I think that if it is possible to work out the `SparseArray` implementation without being too entangled with the rest of the code, a specialized branch would be a good starting point. In the process, some new issues in Mathics-core implementation probably come up, but they could be fixed iteratively.








> 
> Either writing a new `SparseArrayExpression` or utilizing existing packages could be a massive task. At present I'm completely unaware of things like Mathics3's 2024 annual plan. If it's not already fully scheduled, add a paragraph like
> 
> ```
> TODO: Support for SparseArray
> 1. functions that need to be adapted to sparse arrays (e.g. ArrayQ, MatrixQ)
>     ...
>     their side effects to be resolved (e.g. LinearModelFit[{m_?MatrixQ, v_}])
>     ...
> 2. SparseArrayExpression or using libraries from other packages
>     ...
> ```

Sure. @rocky, do we have such a thing as Mathics3's 2024 annual plan? 

> 
> might be helpful in my humble opinion.

mmatera:
Well, I started here https://github.com/mmatera/pymathics-sparse with an external module where we improve the implementation of SparseArray without touching Mathics-core.

rocky:
> From the above discussion I feel it might be better to simply move check_ArrayQ to mathics.eval in this PR/branch, and leave check_SparseArrayQ to be applied in future.

I agree  - thanks for postponing this for now. More below.

> Fixing ArrayQ for SparseArray is necessary, but it has too many side effects (that's why I submitted a draft PR).



> Of course, accumulating many small changes into one big one would be favorable for both developing and reviewing, but merging a PR with side effects into master seems a bit inappropriate to me. IMHO a potentially better solution would be to create a new branch called support-for-SparseArray and merge all these small changes with side effects into it. After everything is done in support-for-SparseArray, we will merge it into master; and before that, we can put in PRs to support-for-SparseArray.

Yes, that is a good option. In the past for major disruptive kinds of changes, there have been long-running branches. And branches can be made off of them and can be merged back into the long-running branch.  Also, in a long-running branch, it is okay if CI fails in ways that are expected to fail. This happens  as a result of incompleteness or partial implementation that hasn't been fully realized.

> Either writing a new SparseArrayExpression or utilizing existing packages could be a massive task. 

I  agree and greatly appreciate the caution taken. Too often people have not given enough forethought or have rushed ahead with an idea before fully understanding the problem, size, or consequence of the changes. For people who are trying to track things, this causes a bit of additional work, and we lose focus on those items that we have previously prioritized. And it explains why we have some of the gaps that we find in the code, and 2023 items haven't been finished.

> At present I'm completely unaware of things like Mathics3's 2024 annual plan. If it's not already fully scheduled, add a paragraph like ...

[FUTURE.rst](https://github.com/Mathics3/mathics-core/blob/master/FUTURE.rst) has the 2023 plan. I just modified it to have a section for 2024. If you are willing to commit to taking the in doing the work (with backup assistance as needed) then feel free to add this to the 2024 section. Same goes for you @mmatera - feel free to add items you are planning to work on.

Sometime in early 2024,  I will add mine items as well. Of course, a number of the items not finished in 2023 will be considered for 2024 and will probably appear there. 

In 2024, we can have a discussion of the items and give priority assign leadership of the items.

Li-Xiang-Ideal:
Now this PR can be safely merged into master and we may go to [Discussions](https://github.com/Mathics3/mathics-core/discussions) to continue our discussion about implement of ``SparseArray``?

rocky:
> Now this PR can be safely merged into master and we may go to [Discussions](https://github.com/Mathics3/mathics-core/discussions) to continue our discussion about implement of `SparseArray`?

Yes, Discussions is a good place to discuss. Thanks. 

Merging now...

Li-Xiang-Ideal:
[New Discussion on SparseArray](https://github.com/Mathics3/mathics-core/discussions/948) added.

PR #949, opened 2023-12-15
fix int-str conversion in Python 3.11

mmatera:
These are my 2 cents to solve the issue presented in  #941


mathics/builtin/system.py
```diff
@@ -29,6 +30,67 @@
     have_psutil = True
 
 
+class MaxLengthIntStringConversion(Predefined):
+    """
+    <url>:Python 3.11:https://docs.python.org/3.11/library/stdtypes.html#int-max-str-digits</url>
+    <dl>
+      <dt>'$MaxLengthIntStringConversion'
+      <dd>A system constant that fixes the largest size of the String resulting from converting
```
rocky, line 38: String -> 'String' here and below.   Integer ->'Integer' 
"converting" -> "converting \"  so there is no hard line break. 

Yes, our docstring formatting is weird. I hope to live to see the day when we can ditch it.

mathics/builtin/system.py
```diff
@@ -29,6 +30,67 @@
...
+          an Integer into a String.
+    </dl>
+
+    >> originalvalue = $MaxLengthIntStringConversion
```
rocky, line 42: The example is good, but I think there should be comments explain better what is going on. In particular, that internally things are okay, but when we get down to displaying the results or formatting, then we have a problem in converting to the output string because it may take a long time.

In a PR to get merged into this, I will work on some suggested expanded verbiage.

mathics/eval/makeboxes.py
```diff
@@ -67,6 +67,27 @@ def _boxed_string(string: str, **options):
...
+    # Estimate the number of decimal digits
+    num_digits = int(value.bit_length() * 0.3)
+    len_num_digits = len(str(num_digits))
+    len_parts = (max_digits - len_num_digits - 8) // 2
+    msd = str(value // 10 ** (num_digits - len_parts))
+    lsd = str(abs(value) % 10**len_parts)
+    value_str = f"{msd} <<{num_digits - len(lsd)-len(msd)}>> {lsd}"
+    return String(value_str)
+
+
```
rocky, line 90: We could write a pytest for this.

rocky:
> These are my 2 cents to solve the issue presented in #941

I'll see your 2 cents and double it 4 cents!  - Sold? @kejcao Would you please try this and see if this solves your concerns?

mathics/eval/makeboxes.py
rocky: Thinking about this more, I think the string should contain "..." at the end (or something like that) to indicate that we have truncated the true value.

mmatera: I just used the WL convention used when the same expression is formatted with `Short`

rocky: Ok - we'll go with this kind of elision then. But we haven't tested it. And this kind of thing is not clear in the docstring either.

mathics/builtin/system.py
mmatera: I tried to improve the description, but I do not know how much we want to go deeper into this.

mathics/eval/makeboxes.py
mmatera: Now I put a bunch of pytests. Also, I inaugurated the `/test/eval/` directory.

mathics/builtin/system.py
rocky: One technique is to give something succinct but then use the examples to elucidate further. Examples though need some additional commentary more motivation was to why there are used, and what to look for, rather than be pure code.

test/eval/test_makeboxes.py
```diff
@@ -0,0 +1,51 @@
+# -*- coding: utf-8 -*-
```
rocky, line 1: Thanks!

rocky:
Whew! We've beat this one to death (for the time being). In the back of my mind is to mention the environment variable `DEFAULT_MAX_STR_DIGITS` which is set to 7000, but I'll leave that for some other time. 

Merge and iterate.

PR #950, opened 2023-12-16
Go over mathics.builtin.system documentation

rocky:
This isn't quite right. Needs a pytest and the failing doctest needs going over. 

But this is closer, I think


mmatera:
@rocky, thanks for the rewriting. I tried to merge it with the part I rewrote.   
 The problem with the tests is that in Python <3.10 the exception is not raised, and the function used to set the value is not available. By now, I think the best is to leave the results of the tests with a wildcard.

rocky:
> @rocky, thanks for the rewriting. I tried to merge it with the part I rewrote. The problem with the tests is that in Python <3.10 the exception is not raised, and the function used to set the value is not available. By now, I think the best is to leave the results of the tests with a wildcard.

Thanks for the information. Okay tests have been weakened. See if I have this and the commentary right though. Feel free to just commit on top of this.

mmatera:
Excellent. Let's merge this into the other branch when the tests finish.

PR #951, opened 2023-12-17
Fix ``Range`` for negative ``di``

Li-Xiang-Ideal:
Now ``Range`` is still not applicable to general ``di`` like ``Pi``. I'm not sure if it's worth sacrificing efficiency to make ``Range`` work for general real-valued ``di`` as ``IterationFunction`` does.

### DIFF
--- a/mathics/builtin/list/constructing.py
+++ b/mathics/builtin/list/constructing.py
@@ -221,6 +221,9 @@ class Range(Builtin):
     >> Range[-3, 2]
      = {-3, -2, -1, 0, 1, 2}
 
+    >> Range[5, 1, -2]
+     = {5, 3, 1}
+
     >> Range[1.0, 2.3]
      = {1., 2.}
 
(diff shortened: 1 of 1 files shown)


rocky:
> I'm not sure if it's worth sacrificing efficiency to make Range work for general real-valued di as IterationFunction does.

The way to get efficiency here is to note that with respect to the implementation "di" does not change. So at the expense of some complexity a determination at the top of the code can be made to decide what category "di" is and customized the implementation into one of several implementations based on that. 

This though can be done in future PR's if/when desired.

There was a similar idea used in other pr where some code that does not change inside a loop is moved outside of the loop.  In programming languages that do not allow dynamically changing variable types, like C, C++, Java, Go, or Rust there are "optimizing" compilers that do this kind of thing automatically.  In Python technology for stuff like this does not exist, so these kinds of things are done by changing the program in roughy the same kind of way that the optimizing compiler detects and does.  (It is possible Python may advance to the stage where this kind of stuff is done automatically)

So that's why, personally, I am more interested in algorithmic and overall structural changes rather than the kinds of changes that are typically done by "optimizing" compilers. That said, if this is how others want to spend their time and focus on, sure, go for it as long at it does not complicate the code inordinately.

PR #952, opened 2023-12-18
Add description for Range

Li-Xiang-Ideal:


mathics/builtin/list/constructing.py
```diff
@@ -213,6 +213,12 @@ class Range(Builtin):
 
       <dt>'Range[$a$, $b$]'
       <dd>returns a list of integers from $a$ to $b$.
+
+      <dt>'Range[$a$, $b$, $di$]'
+      <dd>returns a list of integers from $a$ to $b$ using step $di$.
```
axkr, line 218: Not only integers but appropriate expressions.

axkr: Not only numbers if you see this example:
- https://wolfram.com/xid/0giumy-bictat

`Together` can help to simplify the expressions generated by the iterator to real values:
- https://github.com/axkr/symja_android_library/commit/f0b7d5e08682c183192fda7837e304827013ec0b

Li-Xiang-Ideal: However, ``Range`` in Mathics3 only accepts real arguments for now. For some historical reason it does not use the same implementation as ``Table``, which makes it more efficient while having more limited applicability.

Anyway, your suggestion sounds good and I think it's worth considering in our future reorganization of historical codes including ``Range``. Would you mind putting your suggestions to [Issues](https://github.com/Mathics3/mathics-core/issues)? That way we can remember it after this PR is closed (which could be soon since I don't have time to rewrite ``Range`` in this PR for now).

axkr: I think you can better add details in an issue, how to use the `Table` iterator implementation for `Range`.

rocky:
This change is a correction and improvement as is. Please feel free to open issues or features in the issue/feature  part of github.

PR #958, opened 2024-01-09
Improve switch doc

mmatera:
Just adding the explanation about how parameters in `Switch` are handled in the documentation.


rocky:
LGTM - thanks for the additional comment.

PR #963, opened 2024-01-16
fix doctest for MaxLengthStringConversion in Pyston

mmatera:
Just a tiny adjustment to avoid a failure of the doctest in Pyston 3.8


rocky:
LGTM - and thanks!

Issue #964, opened 2024-01-16
Raising a` BoxExpressionError` is not handled properly?


rocky:
## Description

A  `BoxExpressionError`  does  not seem to be handled properly.

## How to Reproduce

Evaluate in Django or mathicsscript 

```
Graphics3D[Cylinder[0]]
```

## Output Given

```
Traceback (most recent call last):
...
  File "Mathics3/mathics-core/mathics/builtin/box/expression.py", line 51, in __new__
    instance = super().__new__(cls, *elements, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "Mathics3/mathics-core/mathics/core/builtin.py", line 465, in __new__
    instance.init(*args, **kwargs)
  File 'Mathics3/mathics-core/mathics/builtin/box/graphics3d.py", line 602, in init
    raise BoxExpressionError
mathics.core.exceptions.BoxExpressionError
```

## Expected behavior

Actually, I do not _know_ what is expected. I know that an uncaught exception is _not_ expected. In  `mathicsscript` or `mathics`, this terminates the session.

## Your Environment

All environments.

## Additional context

Noticed in testing out things for a possible new release.

rocky:
I can make a stab at addressing, but I am at a loss for what the right behavior is.

For `Graphics3D[Sphere[1, 0, 1]]` I have turned this into a Syntax error giving: 

```
mathics.core.exceptions.BoxExpressionError: In Sphere[], expecting 2 elements: point(s) and radius, got 3.
```

but I don't know if this is right.

mmatera:
Ideally, the failing box expression should be replaced by a red rectangle in the picture, and showing an error message:
![imagen](https://github.com/Mathics3/mathics-core/assets/2862288/a153e45e-52d6-415d-aa0b-192771d12bd1)

This can be done by catching the exception in line 233 of mathics/builtin/box/graphics3d.py

rocky:
This issue handled, although the specific case `Graphics3D[Cylinder[0]]` is not. #970 is the follow on.

PR #966, opened 2024-01-17
handle box errors in graphics and graphics3d

mmatera:
Here we go with a proposal to address #964. The idea is that when an invalid graphics primitive is processed, a pink rectangle is drawn instead. Notice that in WMA no error message is shown in the client. Just the pink background is shown in the image, plus a tooltip in the notebook interface.

### DIFF
--- a/mathics/builtin/graphics.py
+++ b/mathics/builtin/graphics.py
@@ -72,8 +72,7 @@ GRAPHICS_OPTIONS = {
 DEFAULT_POINT_FACTOR = 0.005
 
 
-ERROR_BOX_FACE_COLOR = RGBColor(components=[1, 0.3, 0.3, 0.25])
-ERROR_BOX_EDGE_COLOR = RGBColor(components=[1, 0, 0, 1.0])
+ERROR_BACKGROUND_COLOR = RGBColor(components=[1, 0.3, 0.3, 0.25])
 
 
 class CoordinatesError(BoxExpressionError):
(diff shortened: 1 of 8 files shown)


rocky:
```
Graphics3D[{Red, Opacity[0.25], Cuboid[{{0, 0, 0}, {0.5, 0.5, 0.5}}]}]
``` 
is probably a little bit closer. Also it looks like the error message is some sort of legend or alt for the image. 

The closest approximation I can come up with in our limited set of graphics is: 

```
Row[{Graphics3D[{Red, Opacity[0.25], Cuboid[{{0, 0, 0}, {0.5, 0.5, 0.5}}]}], "Cilinder is not a graphics primitive"}]]
```

An canned image would also work.

mmatera:
I think now the behaviour is closer, but requires an slight change in Mathics-Django: https://github.com/Mathics3/mathics-django/pull/198

Also, this makes that the option `Background` works again in `Graphics` and `Graphics3D`

mathics/builtin/graphics.py
```diff
@@ -1087,6 +1094,75 @@ def stylebox_style(style, specs):
                     raise BoxExpressionError
             return new_style
 
+        failed = []
+
+        def build_error_box1(style):
```
mmatera, line 1099: If we agree that setting the background is the right thing, these two functions can be safely removed.

rocky: I agree that setting the background is the right thing.

rocky:
> I think now the behaviour is closer, but requires an slight change in Mathics-Django: [Mathics3/mathics-django#198](https://github.com/Mathics3/mathics-django/pull/198)
> 
> Also, this makes that the option `Background` works again in `Graphics` and `Graphics3D`

This is awesome. Thanks!

mathics/builtin/graphics.py
```diff
@@ -1087,6 +1093,75 @@ def stylebox_style(style, specs):
...
+            error_style = style.klass(
+                style.graphics,
+                edge=ERROR_BOX_EDGE_COLOR,
+                face=ERROR_BOX_FACE_COLOR,
+            )
+            if isinstance(self, GraphicsElements):
+                error_primitive_head = Symbol("PolygonBox")
```
rocky, line 1105: Let's  define PolygonBox and Polygon3DBox, in `mathics.core.systemsymbols` and move the assignment of the boxes underneath the definitions of  `ERROR_BOX*COLOR`. Thanks.

mathics/builtin/graphics.py
```diff
@@ -1087,6 +1093,75 @@ def stylebox_style(style, specs):
...
+                error_primitive_expression = Expression(
+                    error_primitive_head,
+                    from_python([(-1, -1), (1, -1), (1, 1), (-1, 1), (-1, -1)]),
+                )
+            else:
+                error_primitive_head = Symbol("Polygon3DBox")
+                error_primitive_expression = Expression(
```
rocky, line 1112: See above.

rocky:
With some small changes noted I think this is ready.  It is much improved from the original draft. Thanks. 

After this PR we are in a position now to do the relatively simple task of addressing the other kinds of place where we raise a graphics box error of some sort.

mathics/builtin/graphics.py
mmatera: Since the PR in Mathics-django was merged, this is no longer needed.

rocky:
LGTM. Now that we handle RGB opacity in #968 do we want to add 0.25 opacity to the background to match what you reported before?

rocky:
Background should be documented, but we can do that in a separate PR.

mmatera:
> LGTM. Now that we handle RGB opacity in #968 do we want to add 0.25 opacity to the background to match what you reported before?

This and the documentation for the `Background` property is going to be included in another PR.

PR #968, opened 2024-01-19
Go over `mathics.builtin.colors.color_directives`

rocky:
Color docs and examples. Add RGB opacity parameter. Move Opacity to `mathics.core.systemsymbols`
Add wikipedia links, and note that RGB and opacity values go between 0 and 1.

Order RGB examples from simple to more complex and introduce each example.


mmatera (approved):
LGTM

mathics/builtin/colors/color_directives.py
```diff
@@ -660,6 +690,7 @@ class RGBColor(_ColorObject):
     def to_rgba(self):
         return self.components
 
+    rules = {"RGBColor[r, g, b, a]": "{RGBColor[r, g, b], Opacity[a]}"}
```
mmatera, line 693: Actually, to make this work as it should in `Background`, we need to improve the support for lists in that option. I think that the best is to remove this rule, and improve the support for the alpha channel in the formatter.

rocky: Ok. Please put in another PR to make this work better. Thanks.

Issue #971, opened 2024-01-21
Symbol xxx is locked issues.


rocky:
## Description

The `$PrintForms` doctest is failing with error message: 

```
ClearAttributes::locked: Symbol True is locked.
```

I suspect this has been happening for a while, but we just haven't been testing it. (Testing may have had a bug that skipped this.)

Similarly when loading CombinatoricaV0.9 we see this: 

```
    SetDelayed: write

    Tag Element in Element[a_List, {index___}] is Protected.
```

## How to Reproduce

In a Mathics3 session

Example 1: 
```
In[1]:= $PrintForms
Out[1]= {TraditionalForm, StandardForm, PythonForm, MathMLForm, SympyForm, FullForm, TeXForm, InputForm}

In[2]:= MemberQ[$PrintForms, MyForm]
Out[2]= False

In[3]:= Format[MyForm[F[x_]]]:= "F<<" <> ToString[x] <> ">>"'
Out[3]= None

In[4]:= Format[F[x_], MyForm]:= MyForm[F[x]]'
Out[4]= None

In[5]:= MemberQ[$PrintForms, MyForm]'
ClearAttributes::locked: Symbol True is locked.
UpSet::write: Tag True in True' is Protected.
...
$RecursionLimit::reclim: Recursion depth of 200 exceeded.
...
```

Example2:
```
In[1]:=  Needs["DiscreteMath`CombinatoricaV0.9`"]

SetDelayed::write: Tag Element in Element[a_List, {index___}] is Protected.
Out[1]= None
```
## Output Given

See output in previous section

## Additional Context

These are small things noticed in going over everything for a release.

mmatera:
@rocky, in the example, the problem seems to be the single quote at the end of lines 3, 4, and 5. Without it, it works as expected. 
The issue seems to be the result of the interaction between `MemberQ` and `Derivative`:
```
In[1]:= MemberQ[$PrintForms, MyForm]'
ClearAttributes::locked: Symbol False is locked.
UpSet::write: Tag False in False' is Protected.
ClearAttributes::locked: Symbol False is locked.
UpSet::write: Tag False in False' is Protected.
ClearAttributes::locked: Symbol False is locked.
UpSet::write: Tag False in False' is Protected.
...
$RecursionLimit::reclim: Recursion depth of 200 exceeded.
Out[1]= $Aborted
```

In any case, this deserves some investigation.

rocky:
[...]

PR #973, opened 2024-01-22
Fix Derivative issue discussed in #971

mmatera:
This PR prevents ``True'``, ``False'`` and ``List'`` be evaluated producing the error reported in #971. Comments explaining how the rule for evaluating `Derivative` over anonymous functions works was added.

### DIFF
--- a/mathics/builtin/attributes.py
+++ b/mathics/builtin/attributes.py
@@ -349,6 +349,7 @@ class Locked(Predefined):
      = 3
     """
 
+    attributes = A_PROTECTED | A_LOCKED
     summary_text = "keep all attributes locked (settable but not clearable)"
 
 
--- a/mathics/builtin/messages.py
+++ b/mathics/builtin/messages.py
@@ -7,7 +7,7 @@ import typing
 from typing import Any
 
 from mathics.core.atoms import String
-from mathics.core.attributes import A_HOLD_ALL, A_HOLD_FIRST, A_PROTECTED
+from mathics.core.attributes import A_HOLD_ALL, A_HOLD_FIRST, A_LOCKED, A_PROTECTED
 from mathics.core.builtin import BinaryOperator, Builtin, Predefined
 from mathics.core.evaluation import Evaluation, Message as EvaluationMessage
 from mathics.core.expression import Expression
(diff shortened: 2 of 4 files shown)


mmatera:
Remember that in an eval method, return None means that the expression is left without changes and the evaluation stops. Another possibility would be to add another condition in the LHS of the WL rule, to make that the expression does not match if applied to a Locked symbol.

rocky:
> Remember that in an eval method, return None means that the expression is left without changes and the evaluation stops. Another possibility would be to add another condition in the LHS of the WL rule, to make that the expression does not match if applied to a Locked symbol.

This seems even hackier. So let's go with what you propose. And again, thanks for documenting some of the mess we inherited.

Is the handling of  for `List'` correct though with this change? (I realize it is better behavior, but is it _correct_?)

rocky:
Add some small tests to `test/builtin/numbers/test_calculus.py`

mmatera:
> Is the handling of  for `List'` correct though with this change? (I realize it is better behavior, but is it _correct_?) 

This is the behavior in WMA.

rocky:
> > Is the handling of  for `List'` correct though with this change? (I realize it is better behavior, but is it _correct_?)
> 
> This is the behavior in WMA.

Ok. Why is `{1} &` printed?

mmatera:
> > > Is the handling of  for `List'` correct though with this change? (I realize it is better behavior, but is it _correct_?)
> > 
> > 
> > This is the behavior in WMA.
> 
> Ok. Why is `{1} &` printed?

Funny... Tomorrow I will check again.

mmatera:
> > > Is the handling of  for `List'` correct though with this change? (I realize it is better behavior, but is it _correct_?)
> > 
> > 
> > This is the behavior in WMA.
> 
> Ok. Why is `{1} &` printed?

@rocky: indeed, you are right. This makes sense because   `Differential[1][List][x]`  ->     `D[List[x],x]->{1}`
But this is the result of  applying  `{1}&` to `x`. I am going to modify the code to produce this.

mmatera:
I checked again by hand all the special cases that could produce the recursion error:

```
s = Select[s, MemberQ[Attributes[Evaluate[#1]], Locked] &];
Table[q -> (Derivative[1][Evaluate[ToExpression[q]]] // FullForm), {q,
    s}] // TableForm
```

PR #978, opened 2024-01-25
Support for Python 3.12

mmatera:
This PR tries the compatibility with Python 3.12. It seems that the is a kind of issue in catching some recursion limit exceptions.


mmatera:
The failing test seems to be related to this comment in another project: https://discuss.python.org/t/has-sys-setrecursionlimit-behaviour-changed-in-python-3-12b/30205

PR #979, opened 2024-01-26
Moving eval mixin methods to the eval module

mmatera:
This draft starts to move part of the implementation of the `Expression.evaluation` method to the `mathics.eval` module. Also, the fat routine `rewrite_apply_eval_step()` is decomposed into different subroutines. @rocky, I hope this helps with the upcoming work.

### DIFF
--- a/mathics/core/element.py
+++ b/mathics/core/element.py
@@ -446,16 +446,6 @@ class EvalMixin:
         """
         return False
 
-    def rewrite_apply_eval_step(self, evaluation) -> Tuple["BaseElement", bool]:
-        """
-        Performs a since rewrite/apply/eval step used in
-        evaluation.
-
-        Here we specialize evaluation so that any results returned
-        do not need further evaluation.
-        """
-        return self.evaluate(evaluation), False
-
     def sameQ(self, other) -> bool:
         """Mathics SameQ
         Each class should decide what is right here.
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -1044,186 +1044,13 @@ class Expression(BaseElement, NumericOperators, EvalMixin):
         s = structure(head, deps, evaluation, structure_cache=structure_cache)
         return s(list(elements))
 
-    def rewrite_apply_eval_step(self, evaluation) -> Tuple["Expression", bool]:
-        """Perform a single rewrite/apply/eval step of the bigger
-        Expression.evaluate() process.
-
-        We return the Expression as well as a Boolean which indicates
-        whether the caller `evaluate()` should consider reevaluating
-        the expression.
-
-        Note that this is a recursive process: we may call something
-        that may call our parent: evaluate() which calls us again.
-
-        Also note that this step is time consuming, complicated, and involved.
-
-        Therefore, subclasses of the BaseEvaluation class may decide
-        to specialize this code so that it is simpler and faster. In
-        particular, a specialization for a particular kind of object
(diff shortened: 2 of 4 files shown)


rocky:
While I agree that we tend to have modules and large functions and that `mathics.core` needs work, I have grave misgivings about this PR.   

`mathics.eval`  and `mathics.core` have a very specific well-defined purposes and this muddies or muddles those purposes. So let me try to describe those purposes. If we need to adjust the Mathics3 developer guide to make it clearer, please point out what needs to be specified and revised, then please let's do that. 

`mathics.core`  has the lowest level atoms  (in the physics sense) that everything is built up from. So we have things like Symbols, Integers, Strings and so on. These are degenerate S-Expressions (and therefore also degenerate M-Expressions). At the next level up at the molecular level, we have "compound" M-Expressions: those things that are built up from the atomic symbols, strings, numbers, etc. 

`rewrite_apply_eval_step()` for (compound) `Expression` is fat largely _because this is the ugly complex semantics of general WMA Expression evaluation_.    This is the second part of Einstein's well known phrase "as simple as possible..":  "_but no simpler_". 

I do believe right now this code has a little baggage because it is improperly called from Boxing routines when instead Boxing evaluation would have a separate routine which may _sometimes_ (and probably rarely) need to call general (complex) Evaluation. But in those conditions, I do not expect that Evaluation has to change one bit over what it would do if Boxing did not exist. So if I have this right, this is currently a flaw of the current Expression evaluation. 

The second complexity inside `mathics.core.evaluation` is due to the homegrown cache, which is probably somewhat useful until things get rewritten more properly so it is not needed because it is covered by other kinds of caches and because general compound Expression evaluation is not called unnecessarily. 

_Evaluation is a fundamental or "core" aspect of all objects._  For now, I don't see a problem with explicitly specifying what evaluation is for the different kinds of atoms. Except for (compound) Expression evaluation, the code is not so large. And the complexity of (compound) Expression evaluation was alluded to before. 

Now we come to `mathics.eval`.  These are intended helping evaluating the builtin functions which lie  _above_ the level of the core objects.  A reason it is good to have these in a separate module is that they can be called outside of the context of a specific single builtin.  We saw for example that cross-product kind of combination is useful in several Bulltin-functions. Or a common setup routine for calling sympy or mpmath. 

Also, we can avoid having to call the slow general compound Expression evaluation for something like a simple arithmetic expression by calling just the parts that we need to call and skip stuff like ordering arguments, looking for rewrite rules and so on. 

Notice that this customization and reusability aspect has very little use for the evaluation routine for atoms or compound Expressions. It is hard to imagine situations where one want to call compound Expression  `rewrite_apply_eval_step()`  outside of Expression. Or the same thing is true for the other specialized evaluation routines.  (There might be, but it feels pathological and probably would be is simpler and clearer to go through normal mechanism). 

Having said this, here might be a case for splitting core evaluation into a separate module _in `mathics.core`_. But such kinds of things I think we should discuss at a higher level before coding. And please let's not contemplate this until after the next release. Thanks.

mmatera:
I close this because there were too many changes in the base code to reconcile it with master would require starting it all over again.

rocky:
> I close this because there were too many changes in the base code to reconcile it with master would require starting it all over again.

This is probably all for the best. 

Rereading what I wrote, I don't see that things have changed, and I don't understand the need or motivation for this. 

So, as you say, starting again from scratch would probably be best if there is still a need for something.

PR #980, opened 2024-01-29
fixing make doc-data

mmatera:
@rocky 
This is the first bunch of silly changes. Most of them does not have any effect, but could help to clarify the code:
* Use `logging` for warnings
* Avoid monkey patch objects of the `DocTests` class.  
* Check what is the class of the input parameter `tests` in `mathics.docpipeline.create_output` and `mathics.docpipeline.test_tests`. The mixture of types was responsible for a failure here, when a generator was passed instead of a `Tests` object.
* The parameter `format_output` in `mathics.docpipeline.test_case` should not be there: all the reference results are given in `text` form, so the evaluation must be always formatted as `text`.

With the last two changes, make doc-data is working now.

### DIFF
--- a/mathics/doc/common_doc.py
+++ b/mathics/doc/common_doc.py
@@ -447,9 +447,17 @@ class DocTest:
 
 
 # FIXME: Turn into a NamedTuple?
+
+# Looking at https://github.com/mathics/Mathics/blob/eba5c22990fae01cd2579d13bbeabcfdb5221426/mathics/doc/doc.py
+#
+# The class Test was used to yield tests from the Documentation.get_tests() function,
+# while DocTests was used as items in the Doc object (now DocText). It is possible that Tests should subclass DocTests.
+#
+#
+#
 class Tests:
     """
-    A group of tests in the same section or subsection.
+    A group of tests in the same section or subsection of the Documentation.
     """
 
     def __init__(
--- a/mathics/doc/common_doc.py
+++ b/mathics/doc/common_doc.py
@@ -30,6 +30,7 @@ In the shorter, we might we move code for extracting printing to a separate pack
 """
 
 import importlib
+import logging
 import os.path as osp
 import pkgutil
 import re
@@ -199,7 +200,7 @@ def get_results_by_test(test_expr: str, full_test_key: list, doc_data: dict) -> 
             if result_candidate["query"] == test_expr:
                 if result:
                     # Already found something
-                    print(f"Warning, multiple results appear under {search_key}.")
+                    logging.warning(f"multiple results appear under {search_key}.")
                     return {}
                 else:
                     result = result_candidate
@@ -470,6 +471,7 @@ class DocChapter:
     """An object for a Documented Chapter.
     A Chapter is part of a Part[dChapter. It can contain (Guide or plain) Sections.
     """
+
     def __init__(self, part, title, doc):
         self.doc = doc
         self.guide_sections = []
@@ -674,6 +676,7 @@ class DocTests:
     subsection.
     Access ``tests`` to get these.
     """
+
     def __init__(self):
         self._tests = []
         self.text = ""
@@ -690,7 +693,7 @@ class DocTests:
         """
         Older form of ``.tests`` attribute. Don't use.
         """
-        print("Warning: get_tests is deprecated")
+        logging.warn("get_tests is deprecated")
         return self._tests
 
     def is_private(self) -> bool:
@@ -769,27 +772,30 @@ class DocSection:
                     index = 1
                     for doctest in section.items:
                         if isinstance(doctest, DocTests):
-                            doctest.index = index
-                            doctest.chapter = self.chapter.title
-                            doctest.part = self.chapter.part.title
-                            doctest.section = self.title
-                            doctest.key = (
-                                doctest.part,
-                                doctest.chapter,
-                                doctest.section,
+                            # Why monkey patch DocTests?
+                            # These attributes corresponds to a `Tests`
+                            # object.
+                            doctest_index = index
+                            doctest_chapter = self.chapter.title
+                            doctest_part = self.chapter.part.title
+                            doctest_section = self.title
+                            doctest_key = (
+                                doctest_part,
+                                doctest_chapter,
+                                doctest_section,
                                 index,
                             )
                             index += 1
                             new_items.append(doctest)
-
                             if len(new_items) > 0:
                                 self.tests = Tests(
-                                    doctest.part,
-                                    doctest.chapter,
-                                    doctest.section,
+                                    doctest_part,
+                                    doctest_chapter,
+                                    doctest_section,
                                     new_items,
                                 )
                                 yield self.tests
+                                1 / 0
                     return
 
                 for subsection in section.subsections:
@@ -1317,8 +1323,15 @@ class XMLDoc:
     Mathics core also uses this in getting usage strings (`??`).
     """
 
-    def __init__(self, doc_str: str, title: str, section: [DocSection],
-                 doctests_class=DocTests, doctest_class=DocTest, doctext_class=DocText):
+    def __init__(
+        self,
+        doc_str: str,
+        title: str,
+        section: [DocSection],
+        doctests_class=DocTests,
+        doctest_class=DocTest,
+        doctext_class=DocText,
+    ):
         self.title = title
         chapter = section.chapter
         part = chapter.part
@@ -1326,7 +1339,9 @@ class XMLDoc:
         key_prefix = (part.title, chapter.title, title)
 
         self.rawdoc = doc_str
-        self.items = gather_tests(self.rawdoc, doctests_class, doctest_class, doctext_class, key_prefix)
+        self.items = gather_tests(
+            self.rawdoc, doctests_class, doctest_class, doctext_class, key_prefix
+        )
 
     def __str__(self) -> str:
         return "\n".join(str(item) for item in self.items)
--- a/mathics/doc/common_doc.py
+++ b/mathics/doc/common_doc.py
@@ -803,7 +803,6 @@ class DocSection:
                                     new_items,
                                 )
                                 yield self.tests
-                                1 / 0
                     return
 
                 for subsection in section.subsections:
--- a/mathics/doc/latex_doc.py
+++ b/mathics/doc/latex_doc.py
@@ -653,7 +653,9 @@ class LaTeXDoc(XMLDoc):
     """
 
     def __init__(self, doc_str: str, title: str, section: Optional[DocSection]):
-        super().__init__(doc_str, title, section, LaTeXDocTests, LaTeXDocTest, LaTeXDocText)
+        super().__init__(
+            doc_str, title, section, LaTeXDocTests, LaTeXDocTest, LaTeXDocText
+        )
 
     def latex(self, doc_data: dict) -> str:
         """
@@ -664,6 +666,7 @@ class LaTeXDoc(XMLDoc):
                 # We have text but no tests
                 return escape_latex(self.rawdoc)
 
+        print([type(i) for i in self.items])
         return "\n".join(
             item.latex(doc_data) for item in self.items if not item.is_private()
         )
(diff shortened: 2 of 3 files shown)


rocky:
> @rocky This is the first bunch of silly changes. Most of them does not have any effect, but could help to clarify the code:
> 

> * Avoid monkey patch objects of the `DocTests` class.

I am not sure this is correct. This initializes fields that may have not or were not used initialized previously. That documentation data structure may be used in other place than in just testing. So let's leave this as is for now.

BTW, this is not monkey patching. Monkey patching is when you want to _change_ the behavior, not initialize possibly uninitialized data. (I agree that initializing uninitialized data possibly in this way is not good though). 

> * The parameter `format_output` in `mathics.docpipeline.test_case` should not be there: all the reference results are given in `text` form, so the evaluation must be always formatted as `text`.

Again, I am not sure about this because it is not clear you have considered what happens when we want to use this code from Django which needs HTML, or format to LaTeX. 

> 
> With these changes, make doc-data is working now.

Thanks for getting `doc-data` working again. 

You have a tendency to overstate things. The initialization of data, the hard-coding of the format, assert checking and changing to use "logging" has nothing to do with getting `doc-data` working again, there is something else not specified here that changed that _did_ fix this.

rocky:
For the record, I am still seeing these problems. 

```
********** Graphs - Vertices and Edges / Core routines for working with Graphs. / AdjacencyList **********
2687 ( 1): TEST AdjacencyList[{1 -> 2, 2 -> 3}, 3]
result != wanted
Test failed: in Mathics3 Modules / Graphs - Vertices and Edges / Core routines for working with Graphs.
Mathics3 Modules
Result: AdjacencyList[{1 -> 2, 2 -> 3}, 3]
Wanted: {2}

2688 ( 2): TEST AdjacencyList[{1 -> 2, 2 -> 3}, _?EvenQ]
result != wanted
Test failed: in Mathics3 Modules / Graphs - Vertices and Edges / Core routines for working with Graphs.
Mathics3 Modules
Result: AdjacencyList[{1 -> 2, 2 -> 3}, _ ? EvenQ]
Wanted: {1, 3}

2689 ( 3): TEST AdjacencyList[{x -> 2, x -> 3, x -> 4, 2 -> 10, 2 -> 11, 4 -> 20, 4 -> 21, 10 -> 100}, 10, 2]
result != wanted
Test failed: in Mathics3 Modules / Graphs - Vertices and Edges / Core routines for working with Graphs.
Mathics3 Modules
Result: AdjacencyList[{x -> 2, x -> 3, x -> 4, 2 -> 10, 2 -> 11, 4 -> 20, 4 -> 21, 10 -> 100}, 10, 2]
Wanted: {2, 11, 100, x}
```

However this has nothing to do with the bug in writing out data. This, I  believe is part of another bug that I had mentioned before, where the behavior in `test_tests()`  is different than either `test_chapters()` or `test_sections()`. 

If you run tests on specific chapters or sections in a Mathics3 Module then these tests work properly. 

Perhaps you don't have the two Mathics3 Modules installed yet?

mmatera:
> different

Yes, this seems to be related to the tests of the external modules. In my system it does not appears, but maybe because in the environment where I run the tests I didn't have installed pymathics-graph.

mmatera:
Before continuing, I have a question: What is the reason for having parallel `MathicsDocumentation`, `DocPart`, etc classes for the common and the formatted versions? It should not be simpler to have just one common set of classes describing the documentation, and then have formatting functions (like in `mathics.format`) to convert the "abstract" documentation into LaTeX, HTML, or any other format?

rocky:
> The problem that I have with adding these attributes is that then, `DocTests` and `Tests` are essentially the same thing. On the other hand, is it really necessary to have these two classes? Both classes are collections of `DocTest`, but the second also stores the titles of the Part, Chapter, and Section. 

The classes `DocTests` and `Tests` go back at v.04 around 2011. I am sure there was a reason for the two. We just don't know what it was. Git log on the commit doesn't reveal much: 

```
$ git log 9ce1e337
commit 9ce1e3376ed166f74da35e4da24405c2fedc4853
Author: Jan <jan@poeschko.com>
Date:   Mon Mar 7 18:01:14 2011 +0100

    initial version
```

And the code back then did not have any comments or docstrings (which existed in 2011) nor type annotations which of course didn't exist in 2011. To make things worse, vague names were also prevalent in the code back then.

The docstrings and type annotations that have been added since then, are my best guesses, and I could be wrong. 
 
> After looking more in detail, it seems that each section have just one `Tests` object, which would be a collection of both `DocTest` (a single line test) and `DocTests` (several consecutive Doctest lines) Is this the idea?

That makes sense. However we need to be careful here. Did you look at all the cases? For example where the this appears in a subsection, a section, and a guide section? Do things change when the use context changes between running tests, producing documentation and accessing via Django? 

Probably adding an assert for whatever it is you think is the situation can be added to try to get more confidence in this hypothesis.

In the future though, we'll have some pytest tests of varying sorts.

However, until we understand this and have tests in place, caution suggests not changing what we might not fully understand. 

> 
> > That documentation data structure may be used in other place than in just testing. So let's leave this as is for now, for those portions that initialize data. And go over the code for when the data type changes.
> 
> But in that case, why do we not use here the structure that already has these parameters (which is `Tests`)?
> 

I am not sure I understand what you mean specifically.  If we can simplify, of course, I am for that. As simple as possible, _but no simpler_.  Which means we need to make sure we are covering everything and not simplifying necessary parts of the operation.  Again, adding tests will help with making sure that changes do not break things. (Did I mention that in addition to the other problems, testing wasn't a big thing in the code around 2011?)  

> > BTW, this is not monkey patching. Or at least not intentionally. Monkey patching is when you want to mock or _change_ the behavior, not initialize possibly uninitialized data, possibly incorrectly because the data was already some other type. I agree that initializing uninitialized data possibly in this way is not good though.
> 
> OK, I am not sure what is exactly the term, but what happens here is that we add attributes/properties that are not declared in the class.

I do not know of a name for this. I know though it is not [monkey patching](https://en.wikipedia.org/wiki/Monkey_patch) whose use case is to alter code that you generally do not have access to or cannot easily modify.

Only recently with the improvement of linters do we typically get warnings about adding fields in a class that were added to after object creation.  The linters suggest adding these fields, if for no other reason to make sure everything is at least initialized in terms of the attributes. It avoids subtle bugs in the code. 

However the linters do not call this monkey patching either. 

> 
> The goal of `test_case` is to evaluate an expression, format it, and compare the result against a reference value. The reference values are given assuming a fixed format ("text") so if you change the format of the result, it is almost certain that the test will fail. Notice also that the result of this evaluation is not stored here. The place where it makes sense to have the "format_output" parameter is `create_output` where the expression is evaluated, formatted, and then stored in the `doctest_data` dictionary.

Yes, but it may be that in the Django situation the format we expect is an HTML format. Django uses a different pickle file the documentation system which uses LaTeX.  And in running `docpipeline`  in testing, we of course do not need or care about what is in a Pickle file. 

> Sorry for that. I just try to provide a concise report of my founds and questions, which is also the goal of this draft. Strictly, what made that the command make doctest-data works are related to the last two changes: fix the "format_output" parameter in `test_case`, and to control the class received by `create_output`.
>  The others were just "visual aids" to help to understand what was going on.

Ok. Detailed information about what were or are the bug(s) were and what addresses them is I was looking for. 

Keep in mind PR diffs themselves can mix essential from non-essential information, the code here is (unnecessarily) subtle, and sometimes vague. We are working with code that we don't understand and probably at this point no one understands. 

In this kind of situation, caution and not overgeneralizing especially when there is a chance that assumptions may be wrong, do not help. The PR is _one_ possibility, but there may be other possibilities we may want to discuss.

rocky:
> Before continuing, I have a question: What is the reason for having parallel `MathicsDocumentation`, `DocPart`, etc classes for the common and the formatted versions? It should not be simpler to have just one common set of classes describing the documentation, and then have formatting functions (like in `mathics.format`) to convert the "abstract" documentation into LaTeX, HTML, or any other format?

I had on a todo list to start a discussion about the kinds of changes we might want to make to the documentation system in the future which includes some of the things you had suggested or that follows how Mathematica works more closely.  In that I was going to add some of the material below. 

However, as often is the case, before I get around to doing this, the questions start, so here goes the background to what I had intended would go into the background of a broader discussion. 

The current information below should already be documented in docstrings in the code. If there are way to make this clearer or improve it, let's do that...

First let me describe the idea, and then the implementation which does not live up to the idea.

There 3 distinct kinds of things that the "doc system" does:

1. Run documentation examples and check output with known expected output to make sure things haven't broken. 
2. Collect and format  text, test examples with its output in LaTeX
3. Access the documentation inside Django.  We provide this not only hierarchically, but we allow searching for a particular builtin name, or by part, chapter, ... 

There is a 4th consideration, not a requirement but a very nice to have that personally, I would call a requirement:

We need a developer-friendly way to work on the documentation and tests. Currently, in Django, when the something in the doc system changes those changes are picked up and displayed automatically without need to run a batch process.
 
Contrast this to building the LaTeX do which historically has been to produce this huge LaTeX file and run a 3 or 4 pass long running LaTeX command to build everything. As a developer, all we want to check are some small changes, or possibly bugs in LaTeX or figure processing.

Or recall that traditionally we had the same "feature" in running doctests: we had to run everything rather than run just the few tests for a built-in function that were being developed.

Okay now more to the implementation and its history. 

Recall that Django is in its own repository. And building the documentation in LaTeX may someday be in its own repository as well.

Displaying tests in HTML form in the way that Django currently needs it is quite a bit different than formatting in the way that LaTeX needs it. And _both_ of these are different from just checking to see if the examples in the document give correct results. 

Initially in v0.5 or so, HTML (hyped as XML), LaTeX (noted as TeX) and plain text for testing were in fact all in one class. That is a possibility but I personally find it complicates things and can cause scaling problems. Conceptually it is wrong, because one should be able to build and use mathics core without ever having to care about Django HTML - Django HTML code should not be in core.  And Django should not care about having to build LaTeX documentation. 

When Django was split into its own repository, what I wanted to do was split the previous code into a base class of code, and then have 3 subclasses for each kind of thing that one wants to do. Although Mathics-core could have kept the Django (HTML)-specific code, I definitely wanted to reduce the bulk in mathics-core. So definitely this was subclassed. 

LaTeX was also subclassed since ideally that would eventually find its way to a new repository. However the testing of ASCII text was kept in the base class. Not by desire, but because refactoring just this much was a heavy lift as it was. So that part was deferred and is still not done. 

As I've said before, typically before a release I make a pass over documentation. And in doing that invariably it is a struggle with the whole homegrown system because we have made changes to the code (split off Django into a new repository, add Guide Sections, or add Mathics3 Modules). In the past, because my focus has been getting a release out and my time was limited there, work on the documentation code has always felt rushed and minimal.

mathics/doc/common_doc.py
```diff
@@ -769,27 +780,30 @@ def get_tests(self) -> Iterator[Tests]:
...
-                                    doctest.part,
-                                    doctest.chapter,
-                                    doctest.section,
+                                    doctest_part,
+                                    doctest_chapter,
+                                    doctest_section,
                                     new_items,
                                 )
                                 yield self.tests
+                                1 / 0
```
rocky, line 806: What's this line for?

mmatera: This is just a mistake. I put it to see if the function was called in some place, and I forgot to remove it.

mathics/doc/common_doc.py
```diff
@@ -769,27 +780,30 @@ def get_tests(self) -> Iterator[Tests]:
...
                         if isinstance(doctest, DocTests):
-                            doctest.index = index
-                            doctest.chapter = self.chapter.title
-                            doctest.part = self.chapter.part.title
-                            doctest.section = self.title
-                            doctest.key = (
-                                doctest.part,
-                                doctest.chapter,
-                                doctest.section,
+                            # Why monkey patch DocTests?
```
rocky, line 783: In this branch I reverted these changes. And "make doctest" still works. 

I suspect that your suspicion that `Tests` and `DocTests` can be combined. I am not seeing `Tests` used in either `mathics/doc/latex_doc.py` or `mathics_django/docpipeline.py`. So I don't know right now why we need this.

However let's old off of making this change until we get some testing in place.

mmatera: @rocky, this change aimed to determine if these attributes were used in any other place. And I can report that they are not used. Now, I am studying how the class hierarchy works. When I finish with it, I will try to improve the documentation in the common_doc.py module, and start adding some pytests. In the meantime, feel free to pick the parts from this PR that you find useful.

rocky: Ok. If you are going to start writing pytests, I'll hold off. (I have been procrastinating on this). There are other things that I have backlogged (which don't have anything to do with Mathics3).

rocky:
@mmatera When I run tests, e.g. `make doctest` I am not seeing the Part 1 Tutorial tests getting run.

We start out: 

```
MATHICS_CHARACTER_ENCODING="ASCII" SANDBOX= python3 mathics/docpipeline.py
Testing Mathics 7.0.0dev0
on CPython 3.11.7 (main, Dec 17 2023, 06:05:08) [GCC 13.2.0]
using SymPy 1.12, mpmath 1.3.0, numpy 1.26.0, cython Not installed
********** Arithmetic Functions / Basic Arithmetic / CubeRoot **********
   1 ( 1): TEST CubeRoot[16]
********** Arithmetic Functions / Basic Arithmetic / Divide **********
   2 ( 1): TEST 30 / 5
   3 ( 2): TEST 1 / 8
   4 ( 3): TEST Pi / 4
```

When we should start out: 

```
********** Language Tutorial / 3D Graphics **********
   1 ( 1): TEST Graphics3D[Polygon[{{0,0,0}, {0,1,1}, {1,0,0}}]]
   2 ( 2): TEST Graphics3D[{Orange, Polygon[{{0,0,0}, {1,1,1}, {1,0,0}}]}, Axes->True]
   3 ( 3): TEST Head[ToBoxes[Graphics3D[{Polygon[]}]]]
********** Language Tutorial / Basic calculations **********
   4 ( 1): TEST 1 + 2
   5 ( 2): TEST 1 - 2 * (3 + 5) / 4
   6 ( 3): TEST 1 - 2 (3 + 5) / 4
   7 ( 4): TEST 2 4
   8 ( 5): TEST 3 ^ 4
   9 ( 6): TEST 6 / 4
  10 ( 7): TEST N[6 / 4]
  11 ( 8): TEST Log[E]
  12 ( 9): TEST Sin[Pi]
```

And the count at the end is 2.686K  when it should be in the high 3.975K (without Mathics3 Modules loaded).

rocky:
> With the last two changes, make doc-data is working now.

But appearly working by omitting over 1K tests. The lesson here, again, is to use caution when changing things we don't quite fully understand.

mmatera:
Well, it seems that a shortcut is not possible, and then, it is better to start from the beginning and understand how the documentation system works.

mmatera:
Relevant changes already merged in master

PR #981, opened 2024-01-30
Fix gather tests and start adding pytests

mmatera:
This PR fixes an indentation error in `gather_tests`, and starts adding some pytests for the documentation system.

### DIFF
--- a/mathics/doc/common_doc.py
+++ b/mathics/doc/common_doc.py
@@ -345,9 +345,11 @@ def gather_tests(
             if tests is None:
                 tests = test_collection_constructor()
             tests.tests.append(test)
-        if tests is not None:
-            items.append(tests)
-            tests = None
+
+    # After the loop finishes, add the last bunch of tests
+    if tests is not None:
+        items.append(tests)
+        tests = None
     return items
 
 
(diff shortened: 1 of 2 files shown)


mathics/doc/common_doc.py
```diff
@@ -862,6 +862,27 @@ def get_tests(self) -> Iterator[Tests]:
...
+    """
+    `Documentation` describes an object containing the whole documentation system.
+
+    Documentation
+        |
+        +--------0> Parts
+                      |
+                      +-----0> Chapters
+                                 |
+                                 +-----0>Sections
```
rocky, line 874: It is a little more complicated than this.  There can be a Guide Section in between Chapters an Section. When this happens, the "Chapter" name is the same as the "Guide Section name" so in a sense there is an elision here. But Chapter properties and attributes and Guide Sections attributes are not quite the same. 

I think that subsection is optional.

mathics/doc/common_doc.py
```diff
@@ -345,9 +345,11 @@ def gather_tests(
             if tests is None:
                 tests = test_collection_constructor()
             tests.tests.append(test)
-        if tests is not None:
-            items.append(tests)
-            tests = None
+
```
mmatera, line 348: BTW, the name of this function should not be something like `parse_docstring_to_xmldoc`?

rocky: I see and agree with your point that the name is a little weird. Let me think about it some more.


I definitely would like to get away from referring to XML-this or that since this isn't anything like XML.

mathics/doc/common_doc.py
rocky: > 
> OK, but a Guide Section is not just another kind of sections with some special features, that also belong to a Chapter? 

Something I think it useful to keep in mind with Guide Sections is that the level of nesting is increased by one.  From the bottom a section can have either a guide section parent or a chapter.

PR #984, opened 2024-01-31
Doc code revision

rocky:
@mmatera this branch now rebases everything on the current master. 

One test in `test/doc/test_common.py` fails, but I don't understand what the test is testing. Would you please look over this and correct what's wrong. 

`make doctest-data` fails as it did before. Also, testing the additional loading Mathics3 modules still fails, although it narrowing to a chapter or section test does work. (Again, this is probably a difference in docpipeline between `test_tests()` and `test_chapters()` or `test_secitons()`.


test/doc/test_common.py
```diff
@@ -76,7 +75,10 @@ def test_gather_tests():
             ),
         )
         assert isinstance(result, list)
-        assert len(list_expected_types) == len(result)
+
+        # @mmatera this test is failing, why?
+        # assert len(list_expected_types) == len(result)
```
rocky, line 80: This test fails. I don't understand the test. Would you please advise or fix?

mmatera: The idea of this test is to check if the number of elements parsed from the docstring coincided with the number of expected elements. This detects the issue with the wrong indentation in the former `gather_tests` function.

rocky: The assert just says that it there were 9 expected but 10 were returned.

mathics/doc/common_doc.py
```diff
@@ -459,36 +462,25 @@ def __str__(self) -> str:
         return self.test
 
 
-class DocTests:
+# Tests has to appear before Documentation which uses it.
```
mmatera, line 465: I want this here because at some point I would like to DRY Tests and DocTests

mathics/doc/common_doc.py
```diff
@@ -596,6 +731,220 @@ def __str__(self) -> str:
...
+    """
+    Class to gather Part data and its sub-parts (chapters) and format that.
+    """
+
+    def __init__(self, doc, title, is_reference=False):
+        self.doc = doc
+        self.title = title
+        self.chapters = []
+        self.chapters_by_slug = {}
+        self.chapter_class = DocChapter
```
mmatera, line 744: Shouldn't be better to make this a class attribute?

test/doc/test_common.py
mmatera: For which test case? For some reason, the pytests seems to pass in the CI

mathics/doc/common_doc.py
rocky: Please move it when you need it later then. This can be done after release.

mathics/doc/common_doc.py
rocky: Yes, there are a lot of opportunities here and other places. 

However this is going to cause a lot more reorganization of the code.  When made a class attribute, all of the corresponding classes will now have to appear  before classes that use them. (Performance gains are marginal)

If you want to do that feel free, but again let's do that after we get things working for Django and LaTeX.

mmatera: @rocky, it is not a matter of performance, but of consistency and clarity: it is weird the need of passing the class of the sub-elements which are known before hand. As I see this, the design pattern that we use in mathics3.format would be clearer, simpler, and less susceptible to bugs. But if the aim is to follow the present design, and if the idea is to improve the design and behavior of the documentation system before the release, I think we need to consider these alternatives.

rocky: You may misunderstand the code. As currently written, in building the main documentation object we need class constructors for the chapter, and section. These constructors change whether the context is  testing, LaTeX or HTML.  The static assignments in `common_doc.py` are just _default_ values. 

There is probably a way to reorganize everything to reduce this more. But IMHO it would be nice to make another stab at this after release.

mathics/doc/latex_doc.py
```diff
@@ -961,17 +936,11 @@ def __init__(
         mathics/builtin/colors/__init__.py . In mathics/builtin/colors/named-colors.py we have
         the "section" name for the class Read (the subsection) inside it.
         """
+        super().__init__(
+            chapter, section, text, operator, installed, in_guide, summary_text
```
mmatera, line 940: chapter, section, title, text, operator, installed, in_guide, summary_text

(title is missing)

mmatera:
@rocky 

What I am doing now is comparing this branch against master, and start adding step-by-step the changes, to see when it starts to fail. As I discover issues, I will mark the suspicious lines here.

mathics/doc/latex_doc.py
```diff
@@ -1048,6 +1017,6 @@ class LaTeXDocText(DocText):
     Class to hold some (non-test) LaTeX text.
     """
 
-    def latex(self, doc_data) -> str:
+    def latex(self) -> str:
```
mmatera, line 1020: the parameter doc_data is necessary here because LaTeXDocumentationEntry.latex expects that it accepts the parameter. Otherwise, we need to change the logic there.

test/doc/test_common.py
rocky: That's the problem - when there is a failure it is hard to figure out which of the 9 or 10 things is being referred to. 

Of course, to see  a failure, uncomment out the assert and run it.

mathics/doc/latex_doc.py
rocky: Thanks -  `title` should be in there now.

mathics/doc/latex_doc.py
```diff
@@ -864,31 +853,17 @@ class LaTeXDocGuideSection(DocSection):
...
-        self.doc = LaTeXDocumentationEntry(text, title, None)
-        self.in_guide = False
-        self.installed = installed
-        self.section = submodule
-        self.slug = slugify(title)
-        self.subsections = []
-        self.subsections_by_slug = {}
+        super().__init__(chapter, title, text, submodule, installed)
+        self.doc = LaTeXDocumentationEntry(text, title, self)
         self.title = title
```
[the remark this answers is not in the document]
rocky: The assignment to `self.title` is no longer needed and should be removed now. Thanks.

mathics/doc/common_doc.py
```diff
@@ -596,6 +734,196 @@ def __str__(self) -> str:
...
+        return all(test.private for test in self.tests)
+
+    def __str__(self) -> str:
+        return "\n".join(str(test) for test in self.tests)
+
+    def test_indices(self) -> List[int]:
+        return [test.index for test in self.tests]
+
+
+class DocSection:
```
mmatera, line 773: The class is duplicated here.

mathics/doc/common_doc.py
```diff
@@ -517,8 +509,151 @@ def __str__(self) -> str:
...
+
+    def get_tests(self) -> Iterator[Tests]:
+        """
+        Generator which returns all of the Mathics3 Doctest-type tests for a given chapter.
+        It is assumed that everything has been "gathered" or parsed and
+        extracted from all modules previously.
+        """
+
+        for section in self.all_sections:
+            if section.installed:
```
mmatera, line 610: To avoid going so deep in the indentation level,

``
if not section.installed:
    continue
```

File: mathics/doc/common_doc.py [hunk 9]
```diff
@@ -517,8 +509,151 @@ def __str__(self) -> str:
...
+                if section.in_guide:
```
mmatera, line 611: I do not understand this logic: For some reason, a GuideSection is populated with subsections
that are already loaded in the Chapter as sections. Should not be easier to handle a GuideSection as a (shallow) section without subsections?

mathics/doc/common_doc.py
rocky: Hmm. This is an older version of the code. It will be fixed in the next update.

mathics/doc/common_doc.py
rocky: This code had been fixed locally. It should be addressed in next push.

mathics/doc/common_doc.py
rocky: I don't recall what's up here. It may be wrong.

mathics/doc/common_doc.py
```diff
@@ -596,6 +734,196 @@ def __str__(self) -> str:
...
+        else:
+            assert False
+            yield self.tests
+
+
+# This stores a pointer to the Mathics3 Module portion of docs In interactive
+# sessions, Mathics3 Modules get loaded on user demand. We need to update that
+# portion of the documenation separately. Storing a pointer here allows us to
+# retrive this portion
+mathics3_module_part: Optional[DocPart] = None
```
mmatera, line 916: IMHO, this does not belong here. On the one hand, once you have the ``documentation`` object (of the  suitable subclass) can be accessed through
```documentation.parts_by_slug.get("mathics3-modules", None)```. 
Notice that the part could not exist if in the session no modules were loaded.

rocky:
Superceded by #999

PR #986, opened 2024-02-01
Doc code revision order1

mmatera:
Another round of incremental changes. This branch requires small tweaks in mathics-django to keep compatibility (just about the names of certain properties in the documentation classes)
* Start adding pytests for LaTeX documentation.
* Fix `mathics.doc.Documentation` class, which seems to be mixed and smashed with the `DocChapter` class in an old merge.
* Split the part of the code associated with `mathics.doc.Documentation` which does not depend on the `mathics.builtin` code, which was moved to `mathics.doc.MathicsMainDocumentation`. 
* Small tweaks and reorganization to make the code closer to the @rocky's branch `doc-code-revision`.
* Classes that are not used like `LaTeXDocumentation` were removed.
* Tweaks to make the LaTeX documentation to compile


mmatera:
https://github.com/Mathics3/mathics-django/pull/201 is needed in Mathics-Django to keep the compatibility

mathics/data/ExampleData/EinsteinSzilLetter.txt
```diff
@@ -67,4 +67,3 @@ is now being repeated.
 Yours very truly,
 A. Einstein
 (Albert Einstein)
```
mmatera, line 69: These files where modified automatically by the new pre-commit script...

mathics/doc/latex_doc.py
```diff
@@ -594,79 +596,15 @@ def latex(self, doc_data: dict) -> str:
         return text
 
 
-class LaTeXDocumentation(Documentation):
```
mmatera, line 597: This class was never used. The class that we use along the code is `LaTeXMathicsDocumentation`.

mathics/doc/latex/sed-hack.sh
```diff
@@ -48,3 +48,11 @@ sed -i -e "s/°/\\\\degree{}/g" documentation.tex
 # from Properties in a Section heading.
 # TODO: figure out how to fix that bug.
 sed -i -e "s/Propertie\\\\/Properties\\\\/g" documentation.tex
+
```
mmatera, line 51: This fixes almost all the errors that prevent that mathics.tex from compiling.

mathics/core/load_builtin.py
```diff
@@ -144,6 +145,12 @@ def import_and_load_builtins():
     """
     Imports Builtin modules in mathics.builtin and add rules, and definitions from that.
     """
+    # TODO: Check if this is the expected behavior, or it the structures
```
mmatera, line 148: @rocky, please check if this is the expected behavior.

mmatera:
With these changes, the documentation is generated without complaint.

mathics/core/load_builtin.py
rocky: I think think this is correct. 

(And if not we can adjust later).

rocky:
LGTM

mathics/doc/latex/sed-hack.sh
rocky: Note though that for release we need the reorganization started in #984 which is what motivated all of this refactoring.

mmatera: I know, and I am trying to go in that direction, trying to work over something that works.

PR #987, opened 2024-02-04
Another baby step in improving the documentation system.

mmatera:
* Adding docstrings to modules that otherwise would not be loaded in the documentation.

* Adding a private function in MathicsMainDocumentation.doc_part  to ensure that we walk over top-level modules to build the chapters.


mathics/doc/common_doc.py
```diff
@@ -304,10 +304,10 @@ def skip_doc(cls) -> bool:
     return cls.__name__.endswith("Box") or (hasattr(cls, "no_doc") and cls.no_doc)
 
 
-def skip_module_doc(module, modules_seen) -> bool:
+def skip_module_doc(module, must_be_skipped) -> bool:
```
mmatera, line 307: There are several reasons to skip a module in the documentation system. However, to skip it because we already visited should not be a good criterion. The way in which we walk the modules should be more predictable to allow testing the process.

mathics/doc/common_doc.py
```diff
@@ -1102,6 +1139,10 @@ def doc_part(self, title, modules, builtins_by_module, start):
                         continue
 
                     submodule_name = get_doc_name_from_module(submodule)
+                    # This has the side effect that Symbols with the same
+                    # short name but in different contexts be skipped.
+                    # This happens with ``PlaintextImport`` that appears in
+                    # the HTML and XML contexts.
```
rocky, line 1145: The best thing would be to use the part name, chapter name, section name, and submodule name as the key.

mathics/doc/common_doc.py
rocky: Yes, I agree. Things were given to us in kind of a mess, and we haven't dug our way out yet.

rocky:
LGTM - merge and iterate.

PR #989, opened 2024-02-04
More on complete documentation

mmatera:
Adding docstrings to modules that otherwise would not be loaded in the documentation. The part of #987 which is related to the builtin modules.

### DIFF
--- a/mathics/builtin/compress.py
+++ b/mathics/builtin/compress.py
@@ -1,7 +1,5 @@
 # -*- coding: utf-8 -*-
-"""
-Compress functions
-"""
+
 
 import base64
 import zlib
--- a/mathics/builtin/compress.py
+++ b/mathics/builtin/compress.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 """
-Compress functions
+Compress Functions
 """
 
 import base64
--- a/mathics/builtin/compress.py
+++ b/mathics/builtin/compress.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
-
+"""
+Compress functions
+"""
 
 import base64
 import zlib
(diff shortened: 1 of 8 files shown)


mathics/builtin/intfns/misc.py
```diff
@@ -19,8 +26,13 @@ class BernoulliB(MPMathFunction):
 
     First five Bernoulli numbers:
 
-    >> Table[BernoulliB[k], {k, 0, 5}]
-     = {1, -1 / 2, 1 / 6, 0, -1 / 30, 0}
+    ## This test is desactivated because it fails in the CI:
+    ## >> Table[BernoulliB[k], {k, 0, 5}]
+    ## = {1, -1 / 2, 1 / 6, 0, -1 / 30, 0}
```
rocky, line 31: Don't comment out the example. Instead, make the expectation more general, e.g. 

```
= ...

File: mathics/builtin/compress.py [hunk 2]
```diff
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
-
+"""
+Compress functions
```
rocky, line 3: The pattern for titles that is in use is capitals, e.g. "Compress Functions". See "Form Variables" somewhere else I think in this PR.

rocky:
LGTM

mathics/builtin/intfns/misc.py
```diff
@@ -19,8 +26,15 @@ class BernoulliB(MPMathFunction):
 
     First five Bernoulli numbers:
 
+    ## This test is desactivated because it fails in the CI:
```
rocky, line 29: Typo desativated -> deactivated.

rocky: And actually, it is not deactivated, we just don't check the results.

PR #990, opened 2024-02-04
Documentation generation for Builtin Symbols with the same name and different contexts.

mmatera:
This PR improves the way in which we walk the `mathics.builtin` package to build the documentation. With these changes

* some tests that are loaded twice in master are now loaded just once (this reduces in ~300 the number of doctests reported when `make doctest` runs ).
* fix the issue in the documentation generation that prevents symbols with the same name and different contexts be loaded.
* Makes the algorithm more deterministic, simpler, and allows to decompose `MathicsMainDocumentation.doc_parts` in simpler functions. (as proposed in #984)


mmatera:
Already included in previous merged PR

PR #991, opened 2024-02-04
Fix $InputFileName

adamantinum:
Currently $InputFileName returns always an empty string.
The commit fixes this wrong behaviour, I don't know if I did it in the optimal way.

### DIFF
--- a/mathics/core/definitions.py
+++ b/mathics/core/definitions.py
@@ -121,6 +121,7 @@ class Definitions:
             "System`",
             "Global`",
         )
+        self.inputfile = ""
 
         # Importing "mathics.format" populates the Symbol of the
         # PrintForms and OutputForms sets.
@@ -243,6 +244,9 @@ class Definitions:
     def get_context_path(self):
         return self.context_path
 
+    def get_inputfile(self):
+        return self.inputfile if hasattr(self, "inputfile") else ""
+
     def set_current_context(self, context) -> None:
         assert isinstance(context, str)
         self.set_ownvalue("System`$Context", String(context))
@@ -258,6 +262,9 @@ class Definitions:
         )
         self.context_path = context_path
         self.clear_cache()
+
+    def set_inputfile(self, dir) -> None:
+        self.inputfile = dir
 
     def get_builtin_names(self):
         return set(self.builtin)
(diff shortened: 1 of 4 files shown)


mathics/core/read.py
```diff
@@ -316,3 +316,8 @@ def read_from_stream(stream, word_separators, msgfn, accepted=None):
                 break
 
             word += tmp
+
+
+def set_inputfile(filename):
+    global INPUTFILE_VAR
+    INPUTFILE_VAR = filename
```
mmatera, line 323: @adamantinum. Thanks for catching this and providing a solution. As a comment, a better place to store the `INPUT_FILE_NAME` variable would be as an attribute in the `Definitions` object.

mmatera (approved):
LGTM

mathics/main.py
```diff
@@ -442,6 +443,8 @@ def dump_tracing_stats():
             definitions.set_line_no(0)
         elif not args.execute:
             return exit_rc
+    else:
+        definitions.set_inputfile("")
```
mmatera, line 447: An alternative would be to leave the attribute undefined here. Then, in `definitions.get_inputfile` just check if the attribute is already set with  `hasattr(self, "inputfile")`. and if not, just return "".

mathics/core/definitions.py
```diff
@@ -243,6 +244,9 @@ def get_current_context(self):
     def get_context_path(self):
         return self.context_path
 
+    def get_inputfile(self):
+        return self.inputfile
```
mmatera, line 248:
```
         return self.inputfile if hasattr(self, "inputfile") else ""
```

This works even if the attribute was not set in main.

mathics/main.py
adamantinum: Yes, it seems better. I have just added it and removed the env var.

mathics/core/definitions.py
adamantinum: Idk, maybe `return self.inputfile` is enough, since in `mathics/core/definitions.py` it is already initialized to an empty string inside `__init__`

mmatera: That's true. So the initialization in the trivial case is not needed.

rocky:
LGTM too.

PR #992, opened 2024-02-04
docstring in evaluation module

mmatera:
For some reason, this change produce an error in the windows workflow


mmatera:
This is a mystery to be solved in the future...

PR #993, opened 2024-02-05
Doc code another round of tiny changes 2

mmatera:
This PR continues #990 by moving the inner loop that loads chapters in `MathicsMainDocumentation.doc_part` to a new method `MathicsMainDocumentation.doc_chapter`. With this change, the "on the fly" loading of documentation for Pymathics modules loaded in a Django session now is possible (and is implemented in https://github.com/Mathics3/mathics-django/pull/201)


rocky:
LGTM

PR #995, opened 2024-02-07
Tiny changes to tiny changes 2

rocky:


mathics/doc/common_doc.py
```diff
@@ -1143,20 +1143,17 @@ def doc_part(self, title, modules, builtins_by_module, start):
         # packages inside ``mathics.builtin``.
         modules_seen = set([])
 
-        want_sorting = True
-        if want_sorting:
-            module_collection_fn = lambda x: sorted(
+        def module_collection_fn(_) -> list:
```
mmatera, line 1146: why does this function need to have an argument that in the end is not used?

mathics/docpipeline.py
```diff
@@ -608,21 +605,6 @@ def main():
         action="store_true",
         help="print cache statistics",
     )
-    # FIXME: historically was weird interacting going on with
```
mmatera, line 611: Yeah!

mmatera (commented):
Apart from the comment about `module_collection_fn(_)` LGTM

mathics/doc/common_doc.py
rocky: You are right. This 

This is a holdover from the stuff that came before where the "module" parameter was passed in but later it was taken from by the closure from a surrounding variable. 

Is removed. now.

PR #998, opened 2024-02-08
adding Exit as an alias of Quit

### LINKED ISSUE #996: `Exit` builtin not working
`Exit` builtin doesn't seem to work. The rule for `Quit` in `mathics/builtin/evaluation.py` seems correct to me, but `Exit` does nothing and doesn't show up among the available builtins in mathicsscript's autocompletion. `Quit` works just fine. 
If there is anything else I could check, I would be glad to do it.

mmatera:
This fixes #996. It seems that the rule for `Exit` was not loaded together with `Quit`.


mathics/builtin/evaluation.py
```diff
@@ -338,3 +335,26 @@ def eval(self, evaluation: Evaluation, n):
         if isinstance(n, Integer):
             exitcode = n.get_int_value()
         raise SystemExit(exitcode)
+
+
+class Exit(Quit):
+        """
+    <url>:WMA link:https://reference.wolfram.com/language/ref/Quit.html</url>
```
rocky, line 342: Should  all uses of Quit be Exit in the doc here?

rocky: This is s duplicate of the docstring for Quit. Quit can be removed here, and "Exit" in the other docstring copy. It is okay however to mention that Quit and Exit are the same thing (if they are) or mention what the differences are if they differ somehow.

rocky:
LGTM. Thanks.

PR #999, opened 2024-02-08
Doc code rebased rebased

rocky:
Things we need to do to get #984 functionality for the current master. 

A bit of code has been DRY'd.


mathics/builtin/assignments/assignment.py
```diff
@@ -273,14 +273,14 @@ class TagSet(Builtin, _SetOperator):
     </dl>
 
     Create an upvalue without using 'UpSet':
-    >> x /: f[x] = 2
-     = 2
-    >> f[x]
-     = 2
-    >> DownValues[f]
+    >> square /: area[square] = s^2
```
mmatera, line 276: IMHO, this is better:
>> square /: area[square[s_]] := s^2

mathics/builtin/assignments/assignment.py
```diff
@@ -273,14 +273,14 @@ class TagSet(Builtin, _SetOperator):
...
+     = s ^ 2
```
mmatera, line 277: (without this line, because I use SetDelayed before)

mathics/builtin/assignments/assignment.py
```diff
@@ -273,14 +273,14 @@ class TagSet(Builtin, _SetOperator):
...
+
+    >> DownValues[square]
      = {}
-    >> UpValues[x]
-     = {HoldPattern[f[x]] :> 2}
+
+    >> UpValues[square]
+     = {HoldPattern[area[square]] :> s ^ 2}
```
mmatera, line 283: = {HoldPattern[area[square[s_]]] :> s ^ 2}

mathics/doc/common_doc.py
```diff
@@ -504,9 +577,81 @@ def all_sections(self):
...
-    return sorted(chapters, key=lambda chapter: chapter.title)
+    return sorted(
+        chapters,
+        key=lambda chapter: str(chapter.sort_order)
+        if chapter.sort_order is not None
+        else chapter.title,
+    )
+
+
+def sorted_modules(modules) -> list:
```
mmatera, line 646: Maybe this could be inside `mathics.doc.utils`

mathics/doc/common_doc.py
```diff
@@ -741,17 +855,20 @@ def get_tests(self):
             pass
         return
 
-    def load_part_from_file(self, filename, title, is_appendix=False):
+    def load_part_from_file(
+        self, filename, title, chapter_order: int, is_appendix=False
```
mmatera, line 859: Maybe take the change to add annotations?

mathics/doc/common_doc.py
rocky: Right now this is only used in `common_doc.py`, when it is used somewhere else let's move it.

mathics/builtin/assignments/assignment.py
rocky: I trust your judgement in these things. Would you commit over this and improve the examples? 

If I try to do that, I am sure that I will get something wrong.

mmatera: OK, since this change seems to be completely unrelated to the rest, I can change this after merge.

rocky:
Here is what remains...

* `make doctest-data` fails all of the Mathics3 Module tests. I imagine somewhere, somehow the definitions it adds get removed.  Note that when this is run in isolation such as by `MATHICS_CHARACTER_ENCODING="ASCII" SANDBOX= python3 mathics/docpipeline.py --load-module pymathics.graph,pymathics.natlang -s 'Core routines for working with Graphs.'` everything works. 
* more unit tests for the doctest options, `mathics.doc.util` and `mathics.doc.common_doc` should be written. 
* The issue where Mathics Module Guide function names appears twice in the LaTeX documentation should be removed

However since this PR is a bit large, perhaps we should commit as is and iterate the above in a new PR? @mmatera - your thoughts? 

One thing that is cool is that we can create documentation narrowed to a particular chapter quickly. In other words, `MATHICS_CHARACTER_ENCODING="ASCII" SANDBOX= python3 doc2latex.py  -p 'Manual' works.  Down the line sectioning, the `-s` option, might be gone over too. 

We still need to create a discussion for the various documentation things to do and possibly a plan for that.

mmatera:
> Here is what remains...
> 
>     * `make doctest-data` fails all of the Mathics3 Module tests. I imagine somewhere, somehow the definitions it adds get removed.  Note that when this is run in isolation such as by `MATHICS_CHARACTER_ENCODING="ASCII" SANDBOX= python3 mathics/docpipeline.py --load-module pymathics.graph,pymathics.natlang -s 'Core routines for working with Graphs.'` everything works.
> 
>     * more unit tests for the doctest options, `mathics.doc.util` and `mathics.doc.common_doc` should be written.
> 
>     * Guide Section names names in Mathics3 Modules appears twice in the LaTeX documentation and should be removed.
> 
>     * We should allow Guide sections to appear in Mathics3 Modules
> 

OK, let's merge this and iterate. However, I though that in my branch the doctests for graphs works, and guide sections worked both in in the main and the modules section. 

> 
> However since this PR is a bit large, perhaps we should commit as is and iterate the above in a new PR? @mmatera - your thoughts?
> 

What about trying to split this into smaller PRs? There are changes in the documentation of some builtins that should work in master as it is. Also, the part of moving some functions to the `mathics.doc.utils` could be done with too much work.

> One thing that is cool is that we can quickly create documentation narrowed to a particular Part name. In other words, `MATHICS_CHARACTER_ENCODING="ASCII" SANDBOX= python3 doc2latex.py -p 'Manual'` works. Down the line, section filtering, or the `-s` option, might be gone over too.
> 
That is indeed cool. What we maybe should change is to use "section" when what we mean is `builtin symbol`: In GuideSections, Builtin symbols appears as "subsections".

> We still need to create a discussion for the various documentation things to do and possibly a plan for that.

I fully agree. I guess that in this round of revision, we learned several things about the documentation system, and several new ideas of improvement came up.

rocky:
> However, I though that in my branch the doctests for graphs works, and guide sections worked both in in the main and the modules section.

And it also worked in the first PR before that. I suspect this is  small bug somewhere in further refactoring. It is possible in the fixing of this we can get things clearer and better. But I think that kind of thing is better done in isolation from this code. 

> What about trying to split this into smaller PRs? 

Good. I will look over all of the changes to see what can be split off. There are some basic bug fixes that I know can be isolated. 

Which reminds me, we still have #997 hanging out there.

test/doc/test_common.py
```diff
@@ -46,15 +45,15 @@
     #> 2+2
      = 4
     A private doctest with a message
-    #> 1/0 
-     : 
+    #> 1/0
+     :
```
mmatera, line 49: The test in line 89 fails because this line should be
`: Infinite expression 1 / 0 encountered.`

test/doc/test_common.py
```diff
@@ -85,7 +84,10 @@ def test_gather_tests():
             ),
         )
         assert isinstance(result, list)
-        assert len(list_expected_types) == len(result)
+
+        # @mmatera this test is failing, why?
```
mmatera, line 88: The test fails because in line 49 the message is missing. Then, the text in that and the subsequent lines generates a `DocText` entry that should not be there.

PR #1001, opened 2024-02-08
Fix textrecognize doctest

mmatera:
Just fix the doctest that produces the error in #999 (probably because that PR fixes something that makes the test to run)


rocky:
LGTM - thanks!

rocky:
> Just fix the doctest that produces the error in #999 (probably because that PR fixes something that makes the test to run)

I think the breakage occurred when Python testing versions was bumped in a prior PR to 3.11 which does not support OCR in all platforms  yet.

PR #1003, opened 2024-02-09
adjust TextRecognize doctest

rocky:
Tolerate situation when optional OCR package does not exist


rocky:
Merging since this fixes a failure in  testing master and causes other PRs to fail.

PR #1007, opened 2024-02-10
Exit doc tweaks

rocky:
Add Kernel Sessions and Evaluation Control ...
    
Go over `Quit[]` and `Exit[]` docstrings, and put this into Chapter "Kernel. Session"s. 

Move `Out[]`  from Main Loop Chapter (which is not really a guide section).
    
Add a title to `mathics.builtin.evaluation` so that it will be seen in the docs.


mmatera (approved):
LGTM

PR #1009, opened 2024-02-12
Docpipeline working part

mmatera:
This is another layer of improvements in the documentation system.
* Fixes some references in `ImportExport` documentation.
* Add and improve some docstrings and comments.
* makes that the `key` attribute in `mathics.doc.common.DocTest` be more deterministic.
* DRY `__init__` routines in `latex_doc` subclasses.


mathics/docpipeline.py
```diff
@@ -43,11 +43,6 @@ def max_stored_size(self, _):
 
 
 # Global variables
-definitions = None
```
mmatera, line 46: Trailing lines from the v6.0 version.

mathics/docpipeline.py
```diff
@@ -257,6 +255,10 @@ def show_test_summary(
     return
 
 
+#
+#  TODO: Split and simplify this section
+#
+#
```
mmatera, line 261: I guess this function still requires work. In general, `test_tests/test_sections/test_chapters/test_all` doesn't seem to take advantage of the new organization in ``common_doc``.

mathics/docpipeline.py
```diff
@@ -532,6 +539,88 @@ def test_tests(
 
     if generate_output and (failed == 0 or keep_going):
         save_doctest_data(output_data)
+
+    return total, failed, skipped, failed_symbols, index
+
+
+def test_tests_old(
```
mmatera, line 546: I keep the old version here, because the new version is not working.

mathics/docpipeline.py
```diff
@@ -688,7 +776,7 @@ def test_sections(
     return total
 
 
-def test_all(
+def test_all_new(
```
mmatera, line 779: This is not working either...

mathics/doc/latex_doc.py
```diff
@@ -481,72 +481,7 @@ class LaTeXDocTest(DocTest):
     """
 
     def __init__(self, index, testcase, key_prefix=None):
-        def strip_sentinal(line):
```
mmatera, line 484: DRY

mathics/doc/latex_doc.py
```diff
@@ -916,23 +832,6 @@ def __init__(
         super().__init__(
             chapter, section, title, text, operator, installed, in_guide, summary_text
         )
-        self.doc = LaTeXDocumentationEntry(text, title, section)
```
mmatera, line 919: DRY.

rocky:
I was looking at this earlier. It is good to see the LaTeX document build again. 

Something to keep in mind. Chapters in the Reference part, part 1 should not be sorted by either chapter or section. In this branch, tests appear sorted by chapter in part 1. And in the LaTeX doc the sections in part 1 are also sorted. 

Part of the problem here is that I removed the "sort" parameter that allows one to specify whether or not to sort chapters and sections.  In master, you'll find a sort-order field for chapters for part 1 chapters. 

The logic of this branch follows the older logic where tests are accessed by document sequentially and they happen to come out ordered by section and chapter. One reason it is useful to have routines to get tests by chapter or section is that sometimes we want to run just the tests or update documentation for just section or chapter.  For the tests one can get direct access using a slug name.

.github/workflows/windows.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: [windows]
-        python-version: ['3.10', '3.11']
+        python-version: ['3.7', '3.8']
```
rocky, line 15: we probably shouldn't use 3.7  since that will be disappearing after release.

mmatera: I just wanted to see wether this fixes the error or not. Now I reverted it to the state in master.  I will look at how to solve this in a different branch.

mmatera:
> The logic of this branch follows the older logic where tests are accessed by document sequentially and they happen to come out ordered by section and chapter. One reason it is useful to have routines to get tests by chapter or section is that sometimes we want to run just the tests or update documentation for just section or chapter. For the tests one can get direct access using a slug name.

Yes, I am aware of that, but the problem is that right now, it is not working. What I tried to do is to isolate what is not working, and then see if we can fix it.

mmatera:
superseded by  #1021

PR #1011, opened 2024-02-13
Handle setting $Input and $InputFileName during Get[]

hrueter:
Currently the global variables `$Input` and `$InputFileName` are not set correctly. This pull request fixes this.

#### Steps to reproduce
- Create a file `script.m` with contents  
```
Print["Input: ",$Input]
Print["InputFileName: ", $InputFileName]
```
- Within mathics execute: `Get["script.m"]`
- or call the script from the command line `mathics script.m`

#### Observed output on mathics as of commit 430b1f47840fa064a8e324f8f2b759fb1db1bef3
Using `Get` the global variables are not set at all.
```
In[1]:= Get["script.m"]
Script Input: 
Script InputFileName: 
Out[1]= None
```

From the command line `InputFileName` is set, but not the full path. `Input` is still unset.
```
Script Input: 
Script InputFileName: script.m
```

#### Expected output
```
In[1]:= Get["script.m"]                                                        
Input: script.m
InputFileName: /full/path/to/script.m
```

### DIFF
--- a/mathics/builtin/files_io/files.py
+++ b/mathics/builtin/files_io/files.py
@@ -13,6 +13,7 @@ from io import BytesIO
 from mathics_scanner import TranslateError
 
 import mathics
+import mathics.eval.files_io.files as eval_files
 from mathics.core.atoms import Integer, String, SymbolString
 from mathics.core.attributes import A_PROTECTED, A_READ_PROTECTED
 from mathics.core.builtin import (
@@ -53,7 +54,6 @@ INPUT_VAR = ""
 
 SymbolInputStream = Symbol("InputStream")
 SymbolOutputStream = Symbol("OutputStream")
-SymbolPath = Symbol("$Path")
 
 # TODO: Improve docs for these Read[] arguments.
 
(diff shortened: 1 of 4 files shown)


mathics/builtin/files_io/files.py
```diff
@@ -387,54 +392,75 @@ class Get(PrefixOperator):
     def eval(self, path, evaluation: Evaluation, options: dict):
         "Get[path_String, OptionsPattern[Get]]"
 
-        def check_options(options):
-            # Options
-            # TODO Proper error messages
+        def eval_inner(self, path, evaluation: Evaluation, options: dict):
```
rocky, line 395: Thanks for noticing the bug and PR. 

the actual evaluation implementation of builtins (aside from basic parameter checking) nowadays we put into `mathics.eval.<name-of-module>`. Here, _name_of_module_ would be `files_io.files`. 

Are you up for creating this module under `mathics.eval` and moving `eval_inner()` there? 

This not only reduces the size of `mathics.builtin.files_io.files` but it also allows other routines that might want to use the evaluation primitives to do so in a faster way, and may allow use to write an instruction-based interpreter.  See also https://mathics-development-guide.readthedocs.io/en/latest/extending/code-overview/python-modules.html#id6

hrueter: done.

rocky:
There are a few more small changes. Let me add them in this branch and you can then review the changes.

PR #1012, opened 2024-02-14
Fix input tweaks

rocky:
Further changes to #1011 

Revise $InputFileName code
    
* Move more code out of mathics/builtin/files_io/files.py and intomathics/eval/files_io/files.py
* Add required __init__.py file for new module
* main.py: import location of set_input_var has changed be under mathics.eval
*  * test*: Add test of Get[]
* systemsymbols add Symbol for $Path

@hrueter - please review.


mathics/builtin/files_io/files.py
```diff
@@ -384,59 +377,21 @@ class Get(PrefixOperator):
...
-                result["TraceFn"] = None
-
-            return result
+        trace_fn = None
+        trace_get = evaluation.parse("Settings`$TraceGet")
+        if (
+            options["System`Trace"].to_python()
+            or trace_get.evaluate(evaluation) is SymbolTrue
+        ):
+            trace_fn = builtins.print
```
rocky, line 389: This code now only handles option processing. The rest of the code is in `mathics.eval.files_io.files`

test/builtin/files_io/test_files.py
```diff
@@ -37,6 +37,16 @@ def test_get_and_put():
     check_evaluation(f"DeleteFile[{temp_filename}]", "Null")
 
 
+def test_get():
+    # Check that $InputFileName sets filename inside Get[]unction with Get[]
```
hrueter, line 41: This comment has several typos. Should be something like:
`# Check that $InputFileName is set inside Get[] function`

hrueter (changes requested):
The setting of `$Input` is broken by the second commit.
I suggest to add a test, that also checks that `$Input` is set correctly.

rocky:
> The setting of `$Input` is broken by the second commit. I suggest to add a test, that also checks that `$Input` is set correctly.

Good suggestion. Behavior is now fixed and `$Input` is set properly. 

Other things were improved or corrected (in the case of a WMA link). So please look over and see if this is still lacking.

mathics/builtin/files_io/files.py
```diff
@@ -48,10 +43,9 @@
     SymbolReal,
 )
 from mathics.eval.directories import TMP_DIR
+from mathics.eval.files_io.files import INPUT_VAR, eval_Get
```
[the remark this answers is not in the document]
rocky: `set_input_var()` _was_ moved. Python imports work in a funny way. (Well, it is logical, but you have to know a bit about the semantics of `import` which is a little bit subtle.

hrueter (approved):

mathics/eval/files_io/files.py
```diff
@@ -0,0 +1,78 @@
...
+import mathics
+from mathics.core.builtin import MessageException
+from mathics.core.evaluation import Evaluation
+from mathics.core.parser import MathicsFileLineFeeder, parse
+from mathics.core.read import MathicsOpen
+from mathics.core.symbols import SymbolNull
+from mathics.core.systemsymbols import SymbolFailed, SymbolPath
+
+# Python representation of $InputFileName
+INPUT_VAR: str = ""
```
mmatera, line 19: I think that this should go inside the ``Definitions`` object.

mmatera (approved):
Apart from the comment about where a session variable like "INPUT_FILE" should be localized, LGTM

mathics/eval/files_io/files.py
rocky: @mmatera I don't understand what you mean. Would you  please just go ahead and make the change you are suggesting? 

Also, I don't understand why in `set_inputfile()` we aren't storing `String` instead of `str` so we have to do String conversions on each evaluation of `$Input`.  Conceptually, in a definitions object we should prefer Mathics3 datatypes for session values that are user-accessible, right?

mmatera: As I mentioned before, IMHO, constants associated with the interpreter session should be stored in the ``Definitions`` object instead of using variables at the level of the modules. 

This allows, for example, two independent sessions of the Mathics interpreter in the same Python session. 

On the other hand, I agree that it would be better to store these variables as Mathics instead of Python native types to avoid unnecessary conversions.

In any case, both changes could be done in a subsequent PR

rocky: Yes, I understood conceptually the idea. The details, I am not confident about and don't want or have the time to get into a back-and-forth: no that's not what I really meant, I meant, ... 

Ok. We'll merge this and feel free to iterate and go further.

PR #1013, opened 2024-02-21
Modernize Python metadata (pyproject.toml)

mkoeppe:

PR #1014, opened 2024-03-02
Modernize Python metadata (local)

rocky:
Local fork of #1013


rocky:
@mmatera the failures here are failures in the core system. While I am currently working on fixing those problems, we might consider merging this in since it has been a while lingering. 

Your thoughts?

mmatera:
Actually, this PR seems to produce more errors. In my opinion, to merge this would make it harder to determine the origin of this and the previous failures.

rocky:
> Actually, this PR seems to produce more errors. In my opinion, to merge this would make it harder to determine the origin of this and the previous failures.

Ok. Hopefully, I will have the others addressed within a day or so.

mmatera (approved):
@rocky, thanks for all this work! LGTM

mkoeppe (approved):
LGTM

PR #1015, opened 2024-03-02
black weirdness wars

rocky:
In some versions, black seems to want optional "," at the end of the last parameter.

In support of getting #1013 to build cleanly


mmatera (approved):
LGTM

PR #1020, opened 2024-03-13
Fix required test in doctests

mmatera:
After the last changes in the documentation system, `requires` was not taken into account in docpipeline. This PR fixes that.


rocky:
LGTM

PR #1022, opened 2024-03-14
Docpipeline tests by chapter and section

mmatera:
This continues #1021, and starts to tackle the problem of accessing tests by chapter and section without interacting over the full documentation.

For chapters, the change is quite straightforward. On the other hand, for sections, I think we need to think a little bit what is the best way to do this.

One possibility would be to produce the documentation entry from the docstring of the target builtin, without passing through the documentation. The other way to go would be to keep a dictionary in the documentation, storing all the sections "by slug".

Thoughts?

PR #1025, opened 2024-03-19
Guides from submodules

mmatera:
This is a proposal to start addressing the problem of the guide sections stated in #1023. Changes here requires adjusts in mathics-django too.

### DIFF
--- a/mathics/builtin/assignments/upvalues.py
+++ b/mathics/builtin/assignments/upvalues.py
@@ -2,7 +2,7 @@
 """
 UpValue-related assignments
 
-An <i>UpValue<i> is a definition associated with a symbols that does not appear directly its head.
+An <i>UpValue</i> is a definition associated with a symbols that does not appear directly its head.
 
 See <url>
 :Associating Definitions with Different Symbols:
--- a/mathics/doc/doc_entries.py
+++ b/mathics/doc/doc_entries.py
@@ -54,7 +54,7 @@ DL_ITEM_RE = re.compile(
 )
 DL_RE = re.compile(r"(?s)<dl>(.*?)</dl>")
 HYPERTEXT_RE = re.compile(
-    r"(?s)<(?P<tag>em|url)>(\s*:(?P<text>.*?):\s*)?(?P<content>.*?)</(?P=tag)>"
+    r"(?s)<(?P<tag>em|url|pageref)>(\s*:(?P<text>.*?):\s*)?(?P<content>.*?)</(?P=tag)>"
 )
 IMG_PNG_RE = re.compile(
     r'<imgpng src="(?P<src>.*?)" title="(?P<title>.*?)" label="(?P<label>.*?)">'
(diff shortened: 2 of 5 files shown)


mathics/doc/documentation/1-Manual.mdoc
```diff
@@ -216,7 +216,7 @@ The relative uncertainty of '3.1416`3' is 10^-3. It is numerically equivalent, i
   >> 3.1416`3 == 3.1413`4
    = True
 
-We can get the precision of the number by using the \Mathics Built-in function <url>:'Precision': /doc/reference-of-built-in-symbols/atomic-elements-of-expressions/representation-of-numbers/precision/</url>:
+We can get the precision of the number by using the \Mathics Built-in function <url>:'Precision': /doc/reference-of-built-in-symbols/atomic-elements-of-expressions/precision</url>:
```
mmatera, line 219: There are more changes needed like this around, because now symbols are always at the level of the sections.

mmatera:
This is how the documentation looks for a module with submodules:

imagen

And this is for a shallow module:

imagen

mmatera:
Things that still need to be fixed:
* key indices for tests: it seems that by now just works for shallow modules
* references: in the LaTeX documentation, some kind of hyperref instead of ref is needed, because the Symbol sections do not have section numbers.

mmatera:
> Things that still need to be fixed:
> 
>     * key indices for tests: it seems that by now just works for shallow modules
> 
This seems to work now
>     * references: in the LaTeX documentation, some kind of hyperref instead of ref is needed, because the Symbol sections do not have section numbers.
In guides, I managed to use page numbers instead of references

rocky:
> > Things that still need to be fixed:
> > ```
> > * key indices for tests: it seems that by now just works for shallow modules
> > ```
> 
> This seems to work now
> 
> > ```
> > * references: in the LaTeX documentation, some kind of hyperref instead of ref is needed, because the Symbol sections do not have section numbers.
> > ```
> 
> In guides, I managed to use page numbers instead of references

I just built the PDF from LaTeX. This is what I see. 

arithmetic-functions

Would you please put Accumulate as 1.1.1 and have it numbered? (In LaTeX "section" without the star at the end).

To me, this is kind of formatting is the kind of creativity that is not helpful or desired. Normal tried and true book formatting is better.

mmatera:
> To me, this is kind of formatting is the kind of creativity that is not helpful or desired. Normal tried-and-true book formatting is better.

mmatera:
> 
> I just built the PDF from LaTeX. This is what I see.
> 
> arithmetic-functions
> 
> Would you please put CubeRoot as 1.1..1 and Accumulate as 1.2.1 and have it numbered? (In LaTeX "section" without the star at the end).
> 


@rocky, this is already done in #1022. This is just the slight modification that I proposed in the discussion. Although I am not completely happy with the format in that PR, it seems to align with your point of view. The aim of this branch was to show how the format can be modified to have all the symbols mapped to sections instead of the mix of sections and subsections.

PR #1034, opened 2024-05-27
`setup.py`: Build `op-tables.json` as part of "build_py"

mkoeppe:
This makes it possible to install mathics directly from the repo, using `pip install git+https://github.com/Mathics3/mathics-core`

Also updating the definition of the Python packages in `pyproject.toml`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,8 +1,11 @@
 [build-system]
 requires = [
     "setuptools>=61.2",
-    "cython>=0.15.1; implementation_name!='pypy'"
+    "cython>=0.15.1; implementation_name!='pypy'",
+    # For mathics-generate-json-table
+    "Mathics-Scanner >= 1.3.0",
 ]
+build-backend = "setuptools.build_meta"
 
 [project]
 name = "Mathics3"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,46 +79,9 @@ mathics = "mathics.main:main"
 
 [tool.setuptools]
 include-package-data = false
-packages = [
-    "mathics",
-    "mathics.algorithm",
-    "mathics.compile",
-    "mathics.core",
-    "mathics.core.convert",
-    "mathics.core.parser",
-    "mathics.builtin",
-    "mathics.builtin.arithfns",
-    "mathics.builtin.assignments",
-    "mathics.builtin.atomic",
-    "mathics.builtin.binary",
-    "mathics.builtin.box",
-    "mathics.builtin.colors",
-    "mathics.builtin.distance",
-    "mathics.builtin.exp_structure",
-    "mathics.builtin.drawing",
-    "mathics.builtin.fileformats",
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -60,17 +60,17 @@ build:
 # because pip install doesn't handle
 # INSTALL_REQUIRES properly
 #: Set up to run from the source tree
-develop:  mathics/data/op-tables.json
+develop:
```
rocky, line 63: The removal of the dependency in here I don't understand. Why should this be removed?

rocky:
Overall, this is great. I don't understand the changes to Makefile though.

Makefile
mkoeppe: It's no longer needed because `setup.py` now builds this file by itself

rocky: Using (most? or all?) of the Makefile is optional. 

Its purpose for old-timers like me is to allow various common tasks to be done in the same way as other projects that have Makefiles.  So instead of remembering the common lingo for running tests (which has changed over time) , I can run "make check" which insteads changes from say "nosetest" to "pytest" (to whatever it is going to be next year and or next decade). 

But specifically, that dependency is there to see if the optables needs to be created. Without this, whenever someone runs "make develop" the optable will get created, whether or not it was there.

mkoeppe: If you don't want `setup.py` to unconditionally generate `op-tables.json`, we can put in a check there; but in any case, there is no longer a need for the Makefile rule for building `mathics/data/op-tables.json`

mkoeppe: > If you don't want `setup.py` to unconditionally generate `op-tables.json`, we can put in a check there

Done in cb38d92b

rocky: We are talking at cross purposes. The change  make are mechanisms like adding Python code, or whether to use this library call, or some other and so on.

The purpose of the Makefile is to present, at least to me and other people who work across many projects on many platforms using different languages, a common _interface_ or way to do common kinds of things in projects. 

If a file is expected to get created and there is a Makefile around, then "make _filename_" will create the file, by whatever _mechanism_ you want. It could do it by calling python setup.py or switch to some other "standard" mechanism in the future. 

When I type "make develop", make has a built-in ability to skip steps based on tracking dependencies. You'll notice there is also a separate target for mathics/data/op-tables.json which calls the shell script that was removed. (So now this is broken with respect to the Makefile).

mkoeppe: I have removed the changes to the Makefile and have restored the shell script.

rocky: Many thanks. And also thanks for improving the setup here. 

I am good with this. @mmatera  - you too?

mmatera (commented):
LGTM

rocky:
Again thanks @mkoeppe for doing  this and thanks @mmatera for reviewing. 

(I do plan on getting to another release in the not too distant future, and then moving forward on the many problems that still remain.)

PR #1036, opened 2024-07-26
Fixing single \Mathics in docstrings

mmatera:
This PR fixes some possible misunderstood escaped characters, which now produce a warning.

### DIFF
--- a/mathics/builtin/attributes.py
+++ b/mathics/builtin/attributes.py
@@ -7,7 +7,7 @@ While a definition like 'cube[$x_$] = $x$^3' gives a way to specify \
 specify general properties of functions and symbols. This is \
 independent of the parameters they take and the values they produce.
 
-The builtin-attributes having a predefined meaning in \Mathics which \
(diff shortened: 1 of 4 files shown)


rocky:
LGTM. We are running into the same problems in testing seen in mathics-scanner

mmatera:
@rocky, it looks like an incompatibility with the last version of sympy. Let's fix it in another PR.

rocky:
> @rocky, it looks like an incompatibility with the last version of sympy. Let's fix it in another PR.

Yes, I see, a change somewhere between Sympy 1.12 and 1.13. 

Now a question is whether we want to do a release requiring 1.12 or delay release and fix for 1.13?

I suppose it depends on how much work is involved.

mmatera:
I am going to investigate a little bit this weekend. Until now what I found was a change in the format of the numbers from mpmath that seems to confuse the Mathics formatter, and a movement in some functions of the sympy and scipy libraries. In the mean time we can force the older versions in master, and go forward with the release.

PR #1037, opened 2024-07-27
Sympy 1.13 compatibility

mmatera:
This PR implements the changes required for compatibility with the lastest version of Sympy. The main changes are related to the fact that the new version does not allows to compare `Sympy.Float`s against Python `float`s

### DIFF
--- a/mathics/builtin/numbers/numbertheory.py
+++ b/mathics/builtin/numbers/numbertheory.py
@@ -3,9 +3,9 @@
 """
 Number theoretic functions
 """
-
 import mpmath
 import sympy
+from packaging.version import Version
 
 from mathics.core.atoms import Integer, Integer0, Integer10, Rational, Real
 from mathics.core.attributes import (
@@ -501,7 +501,11 @@ class PartitionsP(SympyFunction):
 
     attributes = A_LISTABLE | A_NUMERIC_FUNCTION | A_ORDERLESS | A_PROTECTED
     summary_text = "number of unrestricted partitions"
-    sympy_name = "npartitions"
+    # The name of this function changed in Sympy version 1.13.0.
+    # This supports backward compatibility.
+    sympy_name = (
+        "npartitions" if Version(sympy.__version__) < Version("1.13.0") else "partition"
+    )
 
     def eval(self, n, evaluation: Evaluation):
         "PartitionsP[n_Integer]"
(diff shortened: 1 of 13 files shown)


mathics/builtin/numbers/numbertheory.py
```diff
@@ -476,7 +476,8 @@ def to_int_value(x):
         result = n.to_python()
         for i in range(-py_k):
             try:
-                result = sympy.ntheory.prevprime(result)
+                # from sympy 1.13, the previous prime to 2 fails...
```
mmatera, line 479: Sympy v1.13 seems to have problems to find out that the previous prime to 2 is -2

mathics/builtin/testing_expressions/numerical_properties.py
```diff
@@ -41,13 +41,18 @@ class CoprimeQ(Builtin):
     CoprimeQ also works for complex numbers
     >> CoprimeQ[1+2I, 1-I]
      = True
+    
+    ## This test case is commenteted out because the result produced by sympy is wrong:
```
mmatera, line 45: Sympy 2.13 seems to fix this case, but breaks the previous one...

pyproject.toml
```diff
@@ -24,7 +24,7 @@ dependencies = [
     "python-dateutil",
     "requests",
     "setuptools",
-    "sympy>=1.8",
+    "sympy==1.11",
```
mmatera, line 27: I fix the version by now. It seems that still there are some places where comparisons against 0 fails. I plan to allow versions larger than 1.8 again once the other cases get fixed.

mmatera:
Most of the issues with Sympy1.13 seems related to the fact that now `sympy.Float(f)==f` evals as `False` for any `f`.

pyproject.toml
rocky: For the next version, how do you feel about requiring Sympy 1.11 and greater? 

The next release drops support for Python 3.6 and after that I suppose we will drop support for Python 3.7.

rocky:
LGTM.

pyproject.toml
rocky: >
> The next release drops support for Python 3.6 and after that I suppose we will drop support for Python 3.7.

Change this. I see that the [`mathics-scanner/pyproject.toml` now requires Python 3.8 or later](https://github.com/Mathics3/mathics-scanner/blob/master/pyproject.toml#L18).  And this is done because Sympy 1.11 needs 3.8 or greater. 

Given this, we should change [pyproject.toml here](https://github.com/Mathics3/mathics-core/blob/master/pyproject.toml#L29) and drop Python 3.7 in the next release too. (We will have to adjust release testing)

And all of this tends to suggest not trying to support for Sympy 1.8. 

Your thoughts?

PR #1038, opened 2024-07-28
sympy  1.13 compatibility

mmatera:
And here we have a partial compatibility patch. The root of the issues seems to be the change in which comparisons between `sympy.Float` and `float` are done from sympy 1.13.
It breaks several things, including the cache mechanism for complex numbers. This draft makes it works by disabling the cache.

### DIFF
--- a/mathics/builtin/numbers/numbertheory.py
+++ b/mathics/builtin/numbers/numbertheory.py
@@ -581,13 +581,13 @@ class PrimePi(SympyFunction):
     attributes = A_LISTABLE | A_NUMERIC_FUNCTION | A_PROTECTED
     mpmath_name = "primepi"
     summary_text = "amount of prime numbers less than or equal"
-    sympy_name = "ntheory.primepi"
+    sympy_name = "primepi"
 
     # TODO: Traditional Form
 
     def eval(self, n, evaluation: Evaluation):
         "PrimePi[n_?NumericQ]"
-        result = sympy.ntheory.primepi(eval_N(n, evaluation).to_python())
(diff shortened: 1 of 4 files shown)


mathics/builtin/numbers/numbertheory.py
```diff
@@ -581,13 +581,13 @@ class PrimePi(SympyFunction):
     attributes = A_LISTABLE | A_NUMERIC_FUNCTION | A_PROTECTED
     mpmath_name = "primepi"
     summary_text = "amount of prime numbers less than or equal"
-    sympy_name = "ntheory.primepi"
```
mmatera, line 584: using this qualified name is deprecated...

mathics/builtin/numbers/numbertheory.py
```diff
@@ -581,13 +581,13 @@ class PrimePi(SympyFunction):
...
+    sympy_name = "primepi"
 
     # TODO: Traditional Form
 
     def eval(self, n, evaluation: Evaluation):
         "PrimePi[n_?NumericQ]"
-        result = sympy.ntheory.primepi(eval_N(n, evaluation).to_python())
```
mmatera, line 590: same here.

mathics/core/atoms.py
```diff
@@ -732,7 +732,7 @@ def __new__(cls, real, imag):
             self._value = value
 
             # Cache object so we don't allocate again.
-            self._complex_numbers[value] = self
```
mmatera, line 735: The cache mechanism confuse now numbers with different precisions. Let's disable this for a while.

PR #1041, opened 2024-07-29
Disallow using older setuptools...

rocky:
which is reported to be highly vulnerable.
Make Python 3.8 be the minimum-allowed Python since sympy 1.11 needs at least 3.8.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-    "setuptools>=61.2",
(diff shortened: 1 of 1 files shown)


mmatera (approved):
LGTM

bnavigator:
An reason why you pin to sympy < 1.13 here despite #1038 and #1037?

rocky:
Sympy 1.13 is supported in the master branch. See: https://github.com/Mathics3/mathics-core/blob/master/pyproject.toml#L27

PR #1042, opened 2024-07-30
Go over number format boxing

rocky:
pyproject.toml: we don't support sympy 1.13.1 just yet
makeboxes.py: Add annotations, docstrings and remove ambiguity
test_number_for.py: test changed routines
output.py: More NumberForm docstring examples

### DIFF
--- a/mathics/builtin/makeboxes.py
+++ b/mathics/builtin/makeboxes.py
@@ -111,7 +111,7 @@ def real_to_tuple_info(real: Real, digits: Optional[int]) -> Tuple[str, int, boo
         else:
             assert value >= 0
             is_nonnegative = True
-        # Set exponent. ``exponent`` is actual, ``pexp`` of ``number_form()`` is printed.
+        # Set exponent. ``exponent`` is actual, ``pexp`` of ``NumberForm_to_string()`` is printed.
         if "e" in s:
             s, exponent = s.split("e")
             exponent = int(exponent)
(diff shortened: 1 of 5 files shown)


mathics/builtin/makeboxes.py
```diff
@@ -65,94 +71,119 @@ def make_boxes_infix(
...
+def number_form(
+    value: Union[Real, Integer],
+    digits: Optional[int],
+    digits_after_decimal_point: Optional[int],
+    evaluation: Evaluation,
+    options: dict,
+) -> String:
     """
-    Converts a Real or Integer instance to Boxes.
+    Converts a Real or Integer value to a String. FIXME: the return should be a Numberform.
```
mmatera, line 148: Numberform is a Python type? As far as I can see, the function returns `String` objects.

mmatera (approved):
Apart from the tiny comment, LGTM

mathics/builtin/makeboxes.py
rocky: I am not sure I understand. NumberForm _currently_ returns a String in Mathics3, but should it? I don't that is what it does in WMA. 

Here is what I see in wolframscript: 

```
In[1]:= NumberForm[5, {5, 3}]                                                   

Out[1]//NumberForm= 5  (* Note this is tagged on output as NumberForm, and that suggests the result is not a String. In Mathics3 the result form is None.*)

In[2]:= Head[NumberForm[5, {5, 3}]]                                             

Out[2]= NumberForm  (* This suggests the result is a NumberForm. Head [] is not does not have  A_HOLD_ attributes *)

In[3]:= Head[ToString[NumberForm[5, {5, 3}]]                                    ]                                                                               

Out[3]= String  (* Not surprising; but just verifying. *)                                                
```

Suppose I write: Something like 

```
FIXME: the return value should be a NumberForm Value instead of a String. 
```

Is this correct? Or suggest what would be more correct. Thanks!

mmatera: OK, maybe what is misleading is the name of the function.  The return value is the result of formatting a NumberForm[...] as a String. Maybe "numberform_to_string" would be a more enlightening name...

rocky:
@mmatera Also now the small change to `mathics.builtin.forms.output` to disambiguate the section name. This prevented it from showing up in django output.

PR #1045, opened 2024-08-01
fixing typos discovered by codespell

mmatera:
As suggested by @rocky, I was over all the folders except `mathics.builtins` and `test.builtins` with codespell and I corrected several typos.
In another round, I will pass through the remaining modules.


rocky:
Thanks - LGTM.

PR #1046, opened 2024-08-01
Docpipeline django compat

mmatera:
Another round of improvements for the docpipeline module. Now we allow to specify the format of the output (latex, "xml") as well as the path of the corresponding pickle datafile.
With this change, mathics-django docpipeline get a much simpler implementation.

To work together with https://github.com/Mathics3/mathics-django/pull/203


rocky:
LGTM - and thanks! (Sorry for the delay in looking at).

PR #1047, opened 2024-08-02
Custom format is more xml

rocky:


mmatera:
@rocky, the custom format is not xml neither (xml does not supports LaTeX equations, or test cases). I called it "markdown" in the wide sense that is used in pandoc. In any case, if you think it is clearer to call it xml, then go ahead.

rocky:
I never use the word "XML" without clarifying it.  The thing that really bothered me was the use of markdown without any qualifier. XML is what it was based on - it has a lot of start and end tags. And one of the features of markdown is that it skips end tagging often in sectioning.

PR #1050, opened 2024-08-04
using more print_and_log

mmatera:
In the refactor of docpipeline I forgot to use more this method, and also at some place it was called from the wrong object. This PR ammends that.


rocky:
LGTM

PR #1055, opened 2024-08-08
normalizing linebreaks after </dl>

mmatera:
This PR just normalizes the line breaks after </dl> in the docstrings


rocky:
Thanks!

PR #1061, opened 2024-08-08
Test b64

mmatera:
Restore an old doctest in the form of a pytest


rocky:
LGTM. 

Do have any idea why your previous comment mentioned something about the test being conditional, but here it is in unconditionally?

mmatera:
Probably because when it was a doctest the output was dependent on the encoding and the platform. Now it seems it is not a problem.

PR #1062, opened 2024-08-08
Fix character encoding

mmatera:
Looking at #1061 I realized we could not change `$CharacterEncoding` from inside the session. This PR makes the behavior of this symbol closer to the one in WMA.


rocky:
LGTM - and thanks.

PR #1065, opened 2024-08-10
adding packaging as an explicit dependency (#932)

### LINKED ISSUE #932: Support for Python 3.12
"distutils" has been removed in Python 3.12 and Mathics3 will throw an error that no module named 'disutils'.

Officially support is needed.

mmatera:
This should fix #932. For some reason, CI does not see any difference, but it doesn't hurt to add it.


rocky:
Ok - let's try it!

PR #1066, opened 2024-08-11
Misc doctest and pytrest fixes

rocky:
Fix some doctest and pytest bugs...

test.parameters.{doc_even_if_error,keep_going}:
*  `doc_even_if_error` is not defined; `keep_going is`.

mathics/docpipeline.py:
*  remove find_texdata_folder(): it is not used and it also refers to a nonexistent class element

*   `test_case(... fail: Optional[]...= ...)` allows an explicit None to be passed. `None` was not tested for in using ` fail()`. Make parameter not optional

* call to doctest_pripile.print_and_log missing f-string specifier.  Not sure why we need the for loop.
* start_time needs to be initialized

test/builtin/files_io/test_importexport.py:
 * test was pollutes local file system with `123.txt`. Ideally *None* of the tests would work in the local filesystem; a _temporary_ file system is okay though.

### DIFF
--- a/mathics/docpipeline.py
+++ b/mathics/docpipeline.py
@@ -161,10 +161,6 @@ class TestStatus:
         self.prev_key = []
         self.quiet = quiet
 
-    def find_texdata_folder(self):
-        """Generate a folder for texdata"""
-        return self.textdatafolder
-
     def mark_as_failed(self, key):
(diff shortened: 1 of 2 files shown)


mmatera (approved):
Good catchs!LGTM

PR #1067, opened 2024-08-13
Fix one more test that pollutes the filesystem

rocky:
Ideally, we should be setting $InitialDirectory to $TempoarayDirectory as well. We should not be writing into $InitialDirectory which is not guaranteed to be writable.

### DIFF
--- a/test/builtin/files_io/test_files.py
+++ b/test/builtin/files_io/test_files.py
@@ -169,6 +169,12 @@ def test_close():
             "Close[{OutputStream, MathicsNonExampleFile}]",
             "",
         ),
+        (
+            "Delete[MathicsNonExampleFile]",
+            None,
+            "Delete[MathicsNonExampleFile]",
+            "",
+        ),
         ## writing to dir
         ("x >>> /var/", ("Cannot open /var/.",), "x >>> /var/", ""),
         ## writing to read only file


mmatera:
> Ideally, we should be setting $InitialDirectory to $TempoarayDirectory as well. We should not be writing into $InitialDirectory which is not guaranteed to be writable.



This could be done inside `test.helper.reset_session` :
```
def reset_session(add_builtin=True, catch_interrupt=False ):
    global session
    session.reset()
    session.evaluate('SetDirectory["/tmp"] ')    # <<  add me
```

Then, by adding 
```
(None, None, None, None)
```
as the first "test" in the @pytest.mark.parametrize list, we avoid possible side effects from definitions introduced in other tests.

rocky:
> > Ideally, we should be setting $InitialDirectory to $TempoarayDirectory as well. We should not be writing into $InitialDirectory which is not guaranteed to be writable.
> 
> This could be done inside `test.helper.reset_session` :
> 
> ```
> def reset_session(add_builtin=True, catch_interrupt=False ):
>     global session
>     session.reset()
>     session.evaluate('SetDirectory["/tmp"] ')    # <<  add me
> ```
> 
> Then, by adding
> 
> ```
> (None, None, None, None)
> ```
> 
> as the first "test" in the @pytest.mark.parametrize list, we avoid possible side effects from definitions introduced in other tests.

Probably more simply done as a setup_function() of pytest. See https://docs.pytest.org/en/stable/how-to/xunit_setup.html

mmatera:
> > > Ideally, we should be setting $InitialDirectory to $TempoarayDirectory as well. We should not be writing into $InitialDirectory which is not guaranteed to be writable.
> > 
> > 
> > This could be done inside `test.helper.reset_session` :
> > ```
> > def reset_session(add_builtin=True, catch_interrupt=False ):
> >     global session
> >     session.reset()
> >     session.evaluate('SetDirectory["/tmp"] ')    # <<  add me
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then, by adding
> > ```
> > (None, None, None, None)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > as the first "test" in the @pytest.mark.parametrize list, we avoid possible side effects from definitions introduced in other tests.
> 
> Probably more simply done as a setup_function() of pytest. See https://docs.pytest.org/en/stable/how-to/xunit_setup.html

Probably it is, and it would be clearer than the "Null" tests... Let's do it in another round.

PR #1068, opened 2024-08-13
Using tmp as pwd in pytests

mmatera:
As General Peron said, "Better than saying is doing". This is what I was talking about in #1067.
Also, I improved some docstrings a little bit.

### DIFF
--- a/test/builtin/files_io/test_files.py
+++ b/test/builtin/files_io/test_files.py
@@ -102,6 +102,8 @@ def test_close():
 @pytest.mark.parametrize(
     ("str_expr", "msgs", "str_expected", "fail_msg"),
     [
+        (None, None, None, None),  # Reset the session and set the working
+        # directory as the temporary directory
         ('Close["abc"]', ("abc is not open.",), "Close[abc]", ""),
         (
             "exp = Sin[1]; FilePrint[exp]",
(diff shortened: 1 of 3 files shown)


test/builtin/files_io/test_files.py
```diff
@@ -102,6 +102,8 @@ def test_close():
 @pytest.mark.parametrize(
     ("str_expr", "msgs", "str_expected", "fail_msg"),
     [
+        (None, None, None, None),  # Reset the session and set the working
```
rocky, line 105: It is not obvious to me why this should work. Couldn't this be done as test setup instead?

rocky:
Also note that there are ways to run something per test suite or test module as well. (I don't know the exact syntax though.)

test/helper.py
```diff
@@ -38,21 +42,30 @@ def check_evaluation(
...
 
     Compares the expressions represented by ``str_expr`` and  ``str_expected`` by
     evaluating the first, and optionally, the second. If omitted, `str_expected`
     is assumed to be `"Null"`.
 
-    to_string_expr: If ``True`` (default value) the result of the evaluation is
-                    converted into a Python string. Otherwise, the expression is kept
-                    as an Expression object.
-                    If this argument is set to ``None``, the session is reset.
+    str_expr: The expression to be tested. If its value is ``None``, the session is
```
mmatera, line 51: @rocky, when you call a `None` test, this function calls `reset_session`, and this is why it works.

rocky: This is convoluted.

PR #1069, opened 2024-08-13
Change dl margin space to match vspace length

rocky:
Looking at the grey margin space around definition lists, it is still a little too large. 0.6em matches the vspace.

Somewhere I read a design principle that one should try reduce the overall number of different kinds of spacing.


mmatera (approved):
@rocky, I trust you. LGTM

PR #1070, opened 2024-08-13
Remove "sympy_name" in from_sympy() calls:

rocky:
Generally, we can get sympy_name from the object.
Previously, we sometimes we had two-parameter "from_sympy()" calls and three-parameter calls where sometiems a parameter might not be used.

Overall, I think this will simplifies things.

Some type annotations were added.

### DIFF
--- a/mathics/builtin/numbers/calculus.py
+++ b/mathics/builtin/numbers/calculus.py
@@ -10,7 +10,7 @@ arithmetic operations.
 """
 
 from itertools import product
-from typing import Optional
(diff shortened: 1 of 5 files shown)


mmatera:
Good! LGTM

rocky:
Great. I am now starting to look at adding some simple event tracing that will help me understand better what's going on in the code. It is sad that every time I start to look into doing something new I am haunted by weirdnesses from the past.

PR #1072, opened 2024-08-19
Misc tweaks...

rocky:
(These were noticed in working on event tracing)

`mathics/builtin/numbers/algebra.py`: black changes its autoformatting

`mathics/builtin/numbers/numbertheory.py`: some linting prefers triples quotes for docstrings

`mathics/core/convert/sympy.py`: use accurate location for singleton class. This reduces unnecessary sympy imports

`CHANGES.rst`: document recent SetEnvironment addition and GetEnvironment changes

`test/builtin/numbers/__init__.py`:  needed if testing in this directory is done separately.

### DIFF
--- a/mathics/core/symbols.py
+++ b/mathics/core/symbols.py
@@ -14,7 +14,11 @@ from mathics.core.element import (
 # I put this constants here instead of inside `mathics.core.convert.sympy`
 # to avoid a circular reference. Maybe they should be in its own module.
 
-sympy_symbol_prefix = "_Mathics_User_"
+# Prefix used for Sympy variables.
+# We wan t this to be short to keep variable names short.
+# In tracing values, long names make output messy.
+sympy_symbol_prefix = "_m_"
+
 sympy_slot_prefix = "_Mathics_Slot_"
 
 
--- a/mathics/core/symbols.py
+++ b/mathics/core/symbols.py
@@ -14,11 +14,7 @@ from mathics.core.element import (
 # I put this constants here instead of inside `mathics.core.convert.sympy`
 # to avoid a circular reference. Maybe they should be in its own module.
 
-# Prefix used for Sympy variables.
-# We wan t this to be short to keep variable names short.
-# In tracing values, long names make output messy.
-sympy_symbol_prefix = "_m_"
-
+sympy_symbol_prefix = "_Mathics_User_"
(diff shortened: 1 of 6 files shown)


mathics/core/symbols.py
```diff
@@ -14,7 +14,11 @@
 # I put this constants here instead of inside `mathics.core.convert.sympy`
 # to avoid a circular reference. Maybe they should be in its own module.
 
-sympy_symbol_prefix = "_Mathics_User_"
+# Prefix used for Sympy variables.
+# We wan t this to be short to keep variable names short.
+# In tracing values, long names make output messy.
+sympy_symbol_prefix = "_m_"
```
mmatera, line 20: @rocky, in the test it seems you used `_mu_` as the prefix

test/core/test_sympy_python_convert.py
```diff
@@ -42,10 +42,10 @@ def compare(self, mathics_expr, sympy_expr, **kwargs):
         self.compare_to_mathics(mathics_expr, sympy_expr)
 
     def testSymbol(self):
-        self.compare(Symbol("Global`x"), sympy.Symbol("_Mathics_User_Global`x"))
+        self.compare(Symbol("Global`x"), sympy.Symbol("_mg`x"))
         self.compare(
-            Symbol("_Mathics_User_x"),
-            sympy.Symbol("_Mathics_User_System`_Mathics_User_x"),
+            Symbol("_mu_x"),
+            sympy.Symbol("_mu`x"),
```
mmatera, line 48: I would expect that if the conversion routines haven't changed,  and the prefix is `_m_`, 
```"_Mathics_User_System`_Mathics_User_x"``` should be now ```"_m_System`_m_x"```,
with the prefix both in the symbol name and the context. 
In any case, it would be reasonable to put the prefix in the context.

rocky: See above. Not sure why testing works when it shouldn't. Testing is probably a bit weak then.

The issue of shortening context prefixes will be addressed in a more isolated PR.

mmatera: good. Then let's merge this when you feel it is ready.

mmatera: The reason why the tests are "weak" is that we are not checking the result of the comparison. I will put a fix for this in another PR.

PR #1077, opened 2024-08-31
non-recursive SameQ[]

mmatera:
This PR is a proposal for a reimplementation of `Expression.sameQ` which uses a tree transversal iterative algorithm instead of recursion. Using this implementation, we avoid the recursion issue in the Pytests for Python 3.12.

I leave this as a draft for a while, in case someone else wants to propose a better solution.


rocky:
@mmatera The idea is interesting and needed. I'd like look over this in more detail over the weekend and think about it. Thanks!

mathics/doc/latex/mathics-title.pdf
rocky, line 1: @mmatera Something in "make clean" is removing this. I haven't found where or fixed this. 

Also it would be nice if we had a way to automatically generate ths.

rocky:
LGTM - merge when you are ready.

mmatera:
This is what I get when I run `test/builtin/test_attributes.py`

imagen

For some reason, it does not happen now in Github Actions

PR #1079, opened 2024-08-31
SameQ[]  and Expression module tweaks

rocky:
@mmatera looked the code over and do not find anything wrong with it. 

I made some small changes to the docstring comment and addressed some of the lint errors my editor is telling me about.

We can't use generators here since there we access and then "rewind" sometimes. And this _is_ essentially a version without the tail recursions.  So I've removed the comments concerning doubt.

If you have the example that caused the failure on 3.12 without this change, it would be interesting to benchmark just this test inside a Mathics3 session to see if there's any difference in execution time. 

Other comments I'll put in the specific code changes in this PR.

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -5,7 +5,7 @@ import math
 import time
 from bisect import bisect_left
 from itertools import chain
-from typing import Any, Callable, Iterable, List, Optional, Tuple, Type, Union
+from typing import Any, Callable, Iterable, Optional, Tuple, Type, Union
 
 import sympy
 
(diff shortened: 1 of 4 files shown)


mathics/core/expression.py
```diff
@@ -88,21 +88,17 @@
...
+    Tree traversal comparison between `self` and `other`.
     Return `True` if both tree structures are equal.
 
-    This implementation avoids the issue with
-    the recursion limit in Python 3.12
+    This non-recursive implementation reduces the Python stack needed
+    in evaluation. Staring in Python 3.12 there is a limit on the
+    recursion level.
     """
-    # TODO: Consider a faster implementation.
```
rocky, line 101: It is not clear to me this is not fast enough. What is done here is equivalent to tail recursion elimination. 

Actually, on my home machine I've never ran into the recursion limit problem. I guess I have more memory. The reason Python 3.12 put compiler/OS limitation to recursion was no doubt some sort of security thing. People aren't going to know (or care about) what is being referred to with  "the issue".

mathics/core/expression.py
```diff
@@ -234,14 +230,6 @@ def union(expressions, evaluation) -> Optional["ExpressionCache"]:
...
                 return None
 
-        # FIXME: this is workaround the current situtation that some
-        # Atoms, like String, have a cache even though they don't need
-        # it, by virtue of this getting set up in
-        # BaseElement.__init__. Removing the self._cache in there the
-        # causes Boxing to mess up. Untangle this mess.
-        if expr._cache is None:
-            return None
-
```
rocky, line 244: My linter is telling me correctly that `expr` can be uninitialized, and indeed it is the loop iterator variable in the preceding "for" loop.  Testing that after the loop ends is weird. Removing this doesn't cause any test failure.

mmatera (approved):
LGTM

mathics/core/expression.py
mmatera: Actually, comparing the time the tests takes in GithubActions, it seems that this is faster than the recursive implementation.

PR #1080, opened 2024-08-31
Remove "precedence" in operator classes...

rocky:
Removes setting class variable precedence. This is now controlled from the Mathics Scanner project. 

Duplicating information can lead to inconsistency and confusion.


mmatera (approved):
LGTM

PR #1081, opened 2024-08-31
Blanks as singleton

mmatera:
Looking at the initialization process, `Blank*` PatternObjets are created several times. This PR makes that these elements be created just once when a parameter is not provided. This would also help to optimize the pattern-matching algorithm when we face that problem.


mathics/core/builtin.py
```diff
@@ -191,16 +192,15 @@ def __new__(cls, *args, **kwargs):
...
         if kwargs.get("expression", None) is not False:
             return to_expression(cls.get_name(), *args)
-        else:
-            instance = super().__new__(cls)
-            if not instance.formats:
-                # Reset formats so that not every instance shares the same
-                # empty dict {}
-                instance.formats = {}
-            return instance
+
```
mmatera, line 197: This is already done in `Builtin.__new__(...)`

rocky:
LGTM

PR #1084, opened 2024-09-08
apply_rule -> apply_function ...

rocky:
Change `apply_rule()` -> to `apply_function()` when that is what is is; (and not when it is not).

Also mark BaseRule an abstract class, and go over docstrings in mathics.core.rules

@mmatera I think i now understand what you were getting at when we were discussion replacement name for `do_replace()`.  Both functions are definitely apply functions. We could call the method `apply()`. I think though being specific and explicit increases clarity and reduces potential confusion.


rocky:
Haven't heard back one way or another on this, so merging. We can adjust down the line.

mmatera:
@rocky, sorry for the delay. In any case, I agree with these changes.

PR #1086, opened 2024-09-16
Starting to tidy up mathics.core.pattern

mmatera:
In line with the last changes proposed by @rocky, I was doing a pass over mathics.core.pattern and fixing some issues reported by the linter.

### DIFF
--- a/mathics/core/pattern.py
+++ b/mathics/core/pattern.py
@@ -1,12 +1,17 @@
 # cython: language_level=3
 # cython: profile=False
 # -*- coding: utf-8 -*-
+"""Core to Mathics3 is are patterns which match symbolic expressions. A pattern are built up in a custon pattern notation.
+The parts of a pattern are called "Pattern Objects".
+
+While there is a built-in function which allows users to match parts of expressions, patterns are also used in applying of transformation
+rules and deciding functions that get applied.
+
+See also: mathics.core.rules and https://reference.wolfram.com/language/tutorial/PatternsAndTransformationRules.html
 """
-Basic classes for Patterns
-
-"""
 
 
+from abc import ABC
 from itertools import chain
 from typing import Callable, List, Optional, Tuple, Union
 
@@ -70,14 +75,14 @@ class StopGenerator_ExpressionPattern_match(StopGenerator):
 
 class StopGenerator_Pattern(StopGenerator):
     """
-    Exception raised when  Pattern matches
+    Exception raised when  BasePattern matches
     an expression.
     """
 
 
-class Pattern:
+class BasePattern(ABC):
     """
-    This is the base class for Mathics Pattern objects.
+    This is the base class for Mathics3 Pattern objects.
 
     A Pattern is a way to represent classes of expressions.
     For example, ``F[x_Symbol]`` is a pattern which matches an expression whose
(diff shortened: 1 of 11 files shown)


mathics/core/pattern.py
```diff
@@ -1,6 +1,11 @@
 # cython: language_level=3
```
rocky, line 1: BTW - I think we should remove cython. At this point it is probably not speeding things up and getting in the way.

mathics/core/pattern.py
```diff
@@ -906,9 +822,197 @@ def get_match_candidates_count(
 
         count = 0
         for element in elements:
-            if self.does_match(element, evaluation, vars):
+            if self.does_match(element, evaluation, vars_dict):
                 count += 1
         return count
 
     def sort(self):
+        """Sot the elements according to their sort key"""
```
rocky, line 830: Sot the -> Sort

mathics/core/pattern.py
mmatera: For this, Cython was useful in detecting some errors when I renamed some arguments. Maybe we can remove this later.

rocky:
As I look over I see lots of opportunity for adding type annotations. But one step at a time. See comments and merge when you are satisfied. 

Also consider making a pass over https://github.com/Mathics3/mathics-development-guide/blob/master/docs/extending/code-overview/pattern-matching.rst (but note that this section is going to be moved up a level out of "Extending" and its own section.

mathics/core/pattern.py
rocky: Ok - leave in then.

mmatera:
> As I look over I see lots of opportunity for adding type annotations. But one step at a time. See comments and merge when you are satisfied.
> 

Fist, I need to be sure about what are the parameters that the functions and methods are receiving. Also, one of the complaints of the linter is that many of these functions have too many parameters. Probably a first step would be to collect all these parameters into a dictionary or a tuple, and then establish the typing.


> Also consider making a pass over https://github.com/Mathics3/mathics-development-guide/blob/master/docs/extending/code-overview/pattern-matching.rst (but note that this section is going to be moved up a level out of "Extending" and its own section.


OK

rocky:
Concerning the https://github.com/Mathics3/mathics-core/actions/runs/10905508459/job/30264517603?pr=1086 failure, I suspect that various linters detect this too. In other

rocky:
@mmatera When you think things are stable, let me know. I'd like to checkout this branch and look it over to see if this helps me understand how patterns work better. Thanks.

mmatera:
> @mmatera When you think things are stable, let me know. I'd like to checkout this branch and look it over to see if this helps me understand how patterns work better. Thanks.

I am nearly to finish: I am trying to determine the right annotations using the errors reported by Cython. I'll let you when I finish

mmatera:
@rocky, I think this is ready now. In a next PR I will go over `mathics.builtin.patterns.py` doing the same.

rocky:
> @rocky, I think this is ready now. In a next PR I will go over `mathics.builtin.patterns.py` doing the same.
 
Thanks - give me another day or so to try this out.

rocky:
LGTM  - let's merge and iterate. The developer docs and debugger will now need to track these changes.

PR #1087, opened 2024-09-19
Minor things noticed in looking over the code

rocky:
Here are some minor things noticed in looking over this module.

There is something that greatly bothers me that isn't fixed though. 

In this module, the base class is named `Pattern`. Elsewhere, base classes are prefaced with `Base`. For example, `BaseElement`,  `BaseRule`, or `BaseForm`.  

And then, because of this, the builtin function `Pattern[]` in module `mathics.builtin.pattern` uses the class name `Pattern_`.

To me, all of this feels like a gratuitous complication (even though I imagine this was due to sloppiness).

### DIFF
--- a/mathics/builtin/exp_structure/general.py
+++ b/mathics/builtin/exp_structure/general.py
@@ -8,7 +8,7 @@ from mathics.core.builtin import BinaryOperator, Builtin, Predefined
 from mathics.core.exceptions import InvalidLevelspecError
 from mathics.core.expression import Evaluation, Expression
 from mathics.core.list import ListExpression
-from mathics.core.rules import Pattern
+from mathics.core.rules import BasePattern
 from mathics.core.symbols import Atom, SymbolFalse, SymbolTrue
 from mathics.core.systemsymbols import SymbolMap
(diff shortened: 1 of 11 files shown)


mmatera:
> Here are some minor things noticed in looking over this module.
> 
> There is something that greatly bothers me that isn't fixed though. 
> 
> In this module, the base class is named `Pattern`. Elsewhere, base classes are prefaced with `Base`. For example, `BaseElement`,  `BaseRule`, or `BaseForm`.  
> 
> And then because of this the builtin function `Pattern[]` in module `mathics.builtin.pattern` uses the class name `Pattern_`.
> 
> To me, all of this feels like a gratuitous complication (even though I imagine this was due to sloppiness). 
> 
> 

I fully agree. If you are up for it, please include the changes here.

rocky:
> I fully agree. If you are up for it, please include the changes here.

Done. I suspect down the line we will want to make even more changes to the code touched here, but this is just another step in the right direction of making the code more easily comprehensible.

mmatera (approved):

PR #1088, opened 2024-09-19
Simplfy checking for patterns

rocky:
Replace `pat.get_head_name() == "System`Pattern"` with  `isinstance(pat, Pattern)`.

I think there are other places where we are using ``get_head_name() == "System`xxx`` but that is left for a different time.


mmatera (approved):
LGTM

PR #1089, opened 2024-09-20
Accomodation for debugging Get, and parse

rocky:
Access parse() from Get in a way that allows a debugger to trap the call. 
Change the way Get[..., Trace->True] so that a debugger can hook into this more naturally -- separate the line number from the text. 

Some other small changes were made like putting the Number class in alphabetic order, or adding more annotations.
,


mmatera (approved):
LGTM

PR #1091, opened 2024-09-21
allow setting In/Out prefix...

rocky:
The Mathics3 debugger has the abilty to go back into a Mathics session in the middle of debugging.

To separate what is done inside a debug sesssion from what is done outside in the interactive session we set "In[]" to be "Debug In[]" and Out[] to be "Debug Out[]".

Also we DRY the eval loop of main()


mmatera (approved):
LGTM

PR #1092, opened 2024-09-23
Match WMA nocatch msg better and check...

rocky:
* mathics/builtin/procedural.py: add doctest check for error message; correct class sorting
* mathics/core/evaluation.py; wrap result in Hold, but not the message

@mmatera it looks like you added the Throw error handling code a while ago (4 years ago?). 

 I am in the slow process of going over https://github.com/Mathics3/mathics-core/pull/new/match-Hold-error-msg  and noticed this.


mmatera (approved):
LGTM

PR #1093, opened 2024-09-23
`ReadList[]` refactor

rocky:
Move builtin/read.py -> eval/files_io/read.py and start to handle TokenWords.
Add more systemsymbols, 
Add a proper error message for TokenWord option processing

I come at this in trying to get lisp.mathics working.


rocky:
What remains is updating pytests and doctests. I will be doing that next. But this is getting rather large so we should consider merging and iterating.

mmatera:
@rocky, thanks for the work. When you feel it is ready, let's merge and iterate.

PR #1094, opened 2024-09-25
tweak some error messages to be more like WMA...

rocky:
in particular Get::noopen, {First,Last,Most}::normal

### DIFF
--- a/mathics/builtin/list/eol.py
+++ b/mathics/builtin/list/eol.py
@@ -646,7 +646,7 @@ class First(Builtin):
             and len == 2
             and isinstance(first_elem, Atom)
         ):
-            return expr.elements[1].evaluate(evaluation)
+            return expr.elements[1]
 
         return first_elem
 
@@ -908,15 +908,7 @@ class Last(Builtin):
             evaluation.message("Last", "argt", expr_len)
             return
 
-        last_elem = expr.elements[-1]
-        if expr.head == SymbolSequence or (
-            not isinstance(expr, ListExpression)
-            and len == 2
-            and isinstance(expr.elements[0], Atom)
-        ):
-            return last_elem.evaluate(evaluation)
-
-        return last_elem
+        return expr.elements[-1]
 
 
 class Length(Builtin):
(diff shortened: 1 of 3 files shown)


mathics/builtin/list/eol.py
```diff
@@ -589,24 +590,24 @@ class First(Builtin):
     >> First[a + b + c]
      = a
     >> First[x]
-     : Nonatomic expression expected.
+     : Nonatomic expression expected at position 1 in First[x].
      = First[x]
     >> First[{}]
      : {} has zero length and no first element.
      = First[{}]
     """
```
axkr, line 599: `First` and `Last`should have a default value as second arg and attribute `A_HOLD_REST`.

mmatera:
```
In[1]:= First[a]                                                                

First::normal: Nonatomic expression expected at position 1 in First[a].

Out[1]= First[a]

In[2]:= First[a, b]                                                             

Out[2]= b

In[3]:= First[a, b, c]                                                          

First::argt: First called with 3 arguments; 1 or 2 arguments are expected.

Out[3]= First[a, b, c]

```

rocky: Thanks -  this should be handled now.

mathics/builtin/list/eol.py
```diff
@@ -580,39 +581,74 @@ class First(Builtin):
...
+            return
+
+        first_elem = expr.elements[0]
+
+        if expr.head == SymbolSequence or (
+            not isinstance(expr, ListExpression)
+            and len == 2
+            and isinstance(first_elem, Atom)
+        ):
+            return expr.elements[1].evaluate(evaluation)
```
mmatera, line 649: Just a comment: to call the `evaluate(evaluation)` method at the end is not needed: it is going to be called in the next `rewrite_apply_eval` iteration

rocky: Ok - thanks. Will remove the explicit call.

PR #1095, opened 2024-09-26
WriteString: check for improper 1st arg....

rocky:
When given a list of stream/string args, note this is not implemented


mmatera (approved):
LGTM

PR #1098, opened 2024-09-27
Get trace/debug tweaks and fixes...

rocky:
Get[xx, Trace->True] was not showing all lines read. And we did not properly hook into a changeable print function.

Using:
   import a.b.c as e

reduces unchangeable lookup (a.b.c) does not change and simplifies and use -- e.GET_PRINT_FN as opposed to a.b.c.GET_PRINT_FN

### DIFF
--- a/mathics/builtin/files_io/files.py
+++ b/mathics/builtin/files_io/files.py
@@ -9,7 +9,9 @@ import os.path as osp
 import tempfile
 from io import BytesIO
 
-import mathics.eval.files_io.files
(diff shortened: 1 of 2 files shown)


rocky:
See also the branch with the same name in mathics-debugger which changes along with this.

With both changes, we can debug step through Get requests and try stuff out as the code gets loaded.

Some of the user interface feels clunky, but over time and with use, it will be clearer how to improve things where needed.

mmatera (approved):
LGTM

PR #1100, opened 2024-09-28
More `Read[]` and `ReadList[]` fixes

rocky:
*  Check Read{,List} types, 
* Change "General" in general error messages to include the specific function names
* Use parameter names that match what the function spec says it is; adjust some function names as a result.
* Move tuple adjustment of types out of eval_Read and into caller where it belongs.
* TokenWords (in contrast to Separators) should be returned into the stream.


mmatera (approved):
LGTM

rocky:
@mmatera FYI there will be more changes - it still doesn't get us through what is needed in Greg Chaitin's Lisp interpreter. And there are quite a few pytests I now need to add for the additional verification (and some are needed for prior changes too). 

However, this is getting large. So merging and iterating.

Issue #1099, opened 2024-09-28
Pyodide/marimo support


davidar:
Mathics *very nearly* supports [Pyodide](https://pyodide.org/en/stable/). I put together a quick demo of this in [marimo](https://marimo.io/), which uses pyodide to run Python notebooks entirely in a web browser, without requiring a backend server for execution:

https://marimo.io/p/@davidar/mathics

The only major issue I had was the fact that `mathics-core` lists `llvmlite` as a mandatory dependency, which isn't currently supported by pyodide (https://github.com/pyodide/pyodide/issues/621). This meant I had to manually install all the dependencies myself to avoid this one package (otherwise you get an error like [this](https://marimo.app/l/9huj4l)). As far as I can tell, everything except for `Compile[]` is still fully functional.

Otherwise I just lightly modified some of the rendering code from `mathics-django` to output results in a format marimo understands (click the top line of code in the notebook to expand it). I might see if I can pull the necessary bits into a separate package, as it's useful to have outside of just the django application.

**Is your feature request related to a problem? Please describe.**
Being able to easily run mathics by just visiting a webpage without needing to install anything makes it a lot easier to experiment with for new users.

**Describe the solution you'd like**
Make `llvmlite` an optional dependency, so that mathics can still be installed when `Compile[]` isn't needed/supported (as is already the case for several other builtin functions).

**Describe alternatives you've considered**
As I showed above, it is possible to work around this issue, but making this dependency optional would make it a lot more convenient to install mathics.

**Additional context**
![image](https://github.com/user-attachments/assets/58408e40-7028-462d-942f-4c33ce0d668a)

davidar:
It looks like `mathics.core.convert.function` already has a fallback for when LLVM isn't installed, however in this case it doesn't re-export `CompileError`, which causes this error when initialising compilation builtins:

```
cannot import name 'CompileError' from 'mathics.core.convert.function' (/lib/python3.12/site-packages/mathics/core/convert/function.py)
    Not able to load mathics.builtin.compilation. Check your installation.
    mathics.builtin loads from /lib/python3.12/site-packages/mathics/core/load
```

davidar:
Here's a similar example that works in [JupyterLite](https://jupyterlite.readthedocs.io/en/stable/) with the pyodide kernel:

https://gist.github.com/davidar/1d05165c624ee591376397ec8eb93cd1

rocky:
First and foremost, thanks for doing this! 

[...]

PR #1101, opened 2024-09-28
making llvmlite to be optional

mmatera:
Following the requirement in #1099, this PR makes the dependency in llvm-lite optional.
Notice that the "Action Tests" here still run over the full installation, which includes this library.

### DIFF
--- a/mathics/eval/files_io/read.py
+++ b/mathics/eval/files_io/read.py
@@ -248,12 +248,14 @@ def read_list_from_types(read_types):
     # TODO: look for a better implementation handling "Hold[Expression]".
     #
     read_types = (
-        SymbolHoldExpression
-        if (
-            typ.get_head_name() == "System`Hold"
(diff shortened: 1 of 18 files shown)


mathics/eval/files_io/read.py
```diff
@@ -248,12 +248,14 @@ def read_list_from_types(read_types):
...
-        if (
-            typ.get_head_name() == "System`Hold"
-            and typ.elements[0].get_name() == "System`Expression"
+        (
+            SymbolHoldExpression
+            if (
+                typ.get_head_name() == "System`Hold"
+                and typ.elements[0].get_name() == "System`Expression"
+            )
+            else typ
```
rocky, line 257: Reformatting may conflict with another PR, but we'll merge this first and I'll deal with the other later.

rocky: Thanks @mmatera!

PR #1102, opened 2024-09-28
Go over package dependencies...

rocky:
* Use try/else to narrow the region of uncertainty
* We no longer need to check for Python 3.7
* add llvm under as a "full" install dependency
* scipy is a dependency; add it


mmatera (approved):
LGTM

PR #1104, opened 2024-09-29
Fix Read{,List}[] with TokenWord and improve ...

rocky:
* Read{,List}[] with TokenWord needs to return two tokens sometimes
* Adjust error message the right value of Read or ReadList is used.
* simplify logic in eval/files_io/read.py
* Add data_dir variable that can be used in testing.


rocky:
There will be at least two more rounds of PRs, but this is already getting large.

One round involves splitting out [Reading & Writing Files](https://reference.wolfram.com/language/guide/FileOperations.html)  from  `mathics.builtins.file_io.files` .  And with this, the corresponding tests in test_files.py will be moved. And `test_read` will be renamed probably to `readwrite` with `mathics.builtins.file_io.readwrite` getting created. So `mathics.eval.files_io.read` will also get renamed. 

In this test for this PR, I suggest a better way of accessing test data than using `StringToStream`, and this probably could simplify other existing tests as well. 

There will be another round to increase test coverage.  But do it in a way that is not as messy as the pytest and doctest code we have right now.

I am always amazed at the pile of mess we still are digging out of. 



A second round will be 

The I/O routines are just such a big mess

mmatera (approved):
LGTM

PR #1006, opened 2024-02-09
bark when a rule in Builtin.rules cannot be loaded

mmatera:
This is related to #1000. There are some rules included in the `Builtin.rules` property that cannot be loaded. With this PR, it this happens, a warning is printed. In any case, the idea is to fix these cases before merging this PR.

PR #1110, opened 2024-10-01
More on patterns overhaul

mmatera:
This PR includes many tiny changes over mathics.core.pattern and mathics.builtin.patterns.

* following @cory 's ExpReduce, match method now have a `pattern_context` parameter that collects all the parameters. The linter seems to be happier with this change.
* More annotations
* More docstrings
* More on avoiding the use of Python's built-in functions as identifiers.
* `Rule` and `RuleDelayed` now check if they receive two parameters. If they receive more or less parameters, show a warning message.

The work is not complete but is closer to being: it remains to remove ** from the match calls, as well as some other adjusts.


mathics/doc/latex/mathics-title.pdf
rocky, line 1: We should figure out what in "Make clean" is removing this and fix that. Also, we should have a way to regenerate this. 

A hacky solution for now is to have another file that doesn't end in pdf, and that gets copied to pdf when mathics-title.pdf when it has gotten removed by a make clean

rocky:
If the changes have settled down, let me know and I will pull down the code, try it, and try to understand it in detail. Thanks.

mmatera:
@rocky, look at this PR when you have the time. There are much more things to do, but it is already quite large, despite most of the changes are trivial.

rocky:
> There are much more things to do, but it is already quite large, 

Could you outline those much more things?  

We eventually want to use is a [trie structure](https://en.wikipedia.org/wiki/Trie)  or some other pattern-matching data structure , analogous to regular-expression patterns, to do the bulk of the work.  So we should not do fancy restructuring or pattern detection as code that makes  this harder.

mmatera:
I am just talking about making trivial changes that reduce the linter complaints. The first thing that comes to my mind is to decompose Mathis.builtin.patterns into smaller submodules.
Also there are some long functions that should be split into smaller parts. 
Just this would take at least a couple of rounds more.

mathics/builtin/patterns.py
```diff
@@ -119,31 +122,45 @@ class RuleDelayed(BinaryOperator):
     """
 
     attributes = A_SEQUENCE_HOLD | A_HOLD_REST | A_PROTECTED
+    messages = {
```
rocky, line 125: As above this is not needed since it is in General.messages

mathics/builtin/patterns.py
```diff
@@ -173,28 +192,28 @@ def create_rules(
...
             )
-        else:
-            evaluation.message(name, "rmix", rules_expr)
+
+        evaluation.message(name, "rmix", rules_expr)
+        return None, True
+
+    result = []
+    for rule in rules:
+        if rule.get_head_name() not in ("System`Rule", "System`RuleDelayed"):
```
rocky, line 201: I believe:

```        
    if rule.head not in (SymbolRule, SymbolRuleDelayed):
```

is simpler, clearer, and faster.  It will be in a PR.

mathics/builtin/patterns.py
```diff
@@ -485,20 +507,24 @@ class ReplaceList(Builtin):
...
+        expr: BaseElement,
+        rules: BaseElement,
+        maxidx: Number,
+        evaluation: Evaluation,
     ) -> OptionalType[BaseElement]:
-        "ReplaceList[expr_, rules_, max_:Infinity]"
+        "ReplaceList[expr_, rules_, maxidx_:Infinity]"
 
-        if max.get_name() == "System`Infinity":
+        if maxidx.get_name() == "System`Infinity":
```
rocky, line 518: This is no longer correct because Infinity is now `DirectedInfinity[1]`

mathics/builtin/patterns.py
```diff
@@ -588,9 +621,14 @@ def yield_match(vars_2, rest):
             else:
                 yield_func(vars_2, None)
 
-        self.pattern.match(yield_match, expression, vars, evaluation)
+        pattern_context = pattern_context.copy()
+        pattern_context["yield_func"] = yield_match
```
rocky, line 625: I don't understand why the `copy` and `pattern_context`  now is needed when it wasn't before. Was the previous code wrong?  Please enlighten me - thanks.

mathics/builtin/patterns.py
```diff
@@ -621,9 +670,14 @@ def yield_match(vars_2, rest):
...
+        self.pattern.match(expression, pattern_context)
+
+    def match_real_numberq(self, expression: Expression, pattern_context: dict):
+        """Match function for RealValudNumberQ"""
```
rocky, line 678: RealValudNumberQ -> RealValuedNumberQ   and will be in next commit of more-patterns-more

mathics/builtin/patterns.py
```diff
@@ -676,35 +750,43 @@ def yield_match(vars_2, rest):
...
-            from mathics.core.builtin import Test
-
-            builtin = None
-            builtin = evaluation.definitions.get_definition(test)
-            if builtin:
-                builtin = builtin.builtin
-            if builtin is not None and isinstance(builtin, Test):
-                return builtin.test(candidate)
+
+        from mathics.core.builtin import Test
```
rocky, line 772: This can be moved to the top of the file and will make our linter overlords happy. Will be in the next commit of more_patterns_more

mathics/builtin/patterns.py
rocky: I agree that using a fixed parameter is less prone to error, and allows for automated error checking.  

As for the efficiency part, unless this very much slows things down, let's revisit this. It may be that some future way of handling patterns will make all of this moot. 

It would be good to note somewhere in the code that a copy is needed because parameters can sometimes get altered, maybe with an example.

mmatera: > As for the efficiency part, unless this very much slows things down, let's revisit this. It may be that some future way of handling patterns will make all of this moot.
> 

Also, using copy() avoids explicitly rebuilding each entry in the dict, which makes the code shorter.

> It would be good to note somewhere in the code that a copy is needed because parameters can sometimes get altered, maybe with an example.

I thought I had put a comment in `BasePattern.match`, but it seems it was lost in one of my failed attempts that this works. Also, I have to adjust its docstring. Maybe in the next round.

mathics/builtin/patterns.py
```diff
@@ -763,31 +851,34 @@ class Alternatives(BinaryOperator, PatternObject):
...
-    def get_match_count(self, vars={}):
-        range = None
+            # for new_vars_dict, rest in alternative.match(
+            #     expression, vars_dict, evaluation):
+            #     yield_func(new_vars_dict, rest)
+            alternative.match(expression, pattern_context)
+
+    def get_match_count(
+        self, vars_dict: OptionalType[dict] = None
+    ) -> Union[int, tuple]:
```
rocky, line 870: The return type annotation is not correct. In particular, if `self.alternatives` is empty, then `tuple[None]` is returned.

mathics/builtin/patterns.py
rocky: It looks like `match_numericq()`, `match_real_numberq()` and other `match_xxx()` could be DRY'd by having a common function that takes a `yield_match` function in as a parameter.

The other good thing about doing something like this is that it makes it clearer what the _difference_ is between these routiens.

mathics/core/pattern.py
```diff
@@ -284,38 +272,37 @@ def yield_match(sub_vars, rest):
...
     def get_match_candidates_count(
-        self,
-        elements: Tuple[BaseElement],
-        expression: BaseElement,
-        attributes: int,
-        evaluation: Evaluation,
-        vars_dict: Optional[dict] = None,
-    ):
+        self, elements: Tuple[BaseElement], pattern_context: dict
+    ) -> Union[int, tuple]:
```
rocky, line 303: The actual return: `return (1,1)` does match this return annotation. It should be `Tuple[int, int]`. Change will be in more_patterns_more PR>

mathics/builtin/patterns.py
mmatera: @rocky, as I mentioned before, there are a lot of things to do just to make the code more readable and easy to understand and debug, but a larger PR would make it harder to review. If you agree, in this round we can leave a TODO comment and face it in the following round.

mathics/core/pattern.py
```diff
@@ -532,19 +486,24 @@ def get_match_count(self, vars_dict: Optional[dict] = None):
...
-        items: Tuple,
-        max_count: Optional[int],
-        expression: Expression,
-        attributes: int,
-        include_flattened: bool = True,
-    ):
-        """Get the possible wrappings"""
+    def get_wrappings(self, yield_func: Callable, items: Tuple, pattern_context: dict):
+        """
+        Get the possible wrappings
```
rocky, line 491: I don't understand and possibly don't like the term "wrappings" which is vague. 

What exactly _is_ a wrapping?  Changing the name can be left for the future though.

mathics/core/pattern.py
mmatera: Regarding this, I wonder if we should not make the return value of this method a little bit more consistent, but again, it would imply making more changes about how the code works.

mathics/core/pattern.py
mmatera: To be honest, I neither fully understand/like this term. What is true is that this function is called a lot, and seems to be related to the different orders/ grouping that the same expression must be (tried to) match against a pattern. The docstring has the best understand I could get from looking at the code.

mathics/builtin/patterns.py
rocky: No problem. But since I am looking at this, I want to make sure I note what I see in this pass.

I am now done making a pass over the PR. The final changes suggested are in the `more_patterns_more` branch.

mathics/core/pattern.py
rocky: ok.

mathics/core/pattern.py
rocky: I agree - the `Union[int, tuple]` should probably become `tuple``. And yes, this might be done in a future PR.

rocky:
Ok. With this set of changes, I think we're ready to merge. #1006 getting looked at next.

PR #1114, opened 2024-10-01
Add post-mortem debugging via trepan3k

rocky:
Adds post-mortem debugging on a fatal exception via trepan3k. There may be a way to also support pdb, but I didn't see done as easily.


mmatera (approved):
LGTM

PR #1120, opened 2024-10-04
Noticed another ` _apply` that should be `eval_xxx`

rocky:
Something I noticed during debugging something else. In looking at tracebacks, names in context are a bit more visible, making it easier when names reflect what's going on.

Also, an error message added to General and duplication in classes removed.


rocky:
Merging, but if there are any comments or concerns these will be addressed in a future commit.

PR #1126, opened 2024-10-12
Add PRE_EVALUATION_HOOK to facilitate debugging.

rocky:
This allows us to modify definitions at the top-level to trace or debug.


rocky:
See https://github.com/Mathics3/mathics3-debugger/tree/add-top-level-evaluation-hook for an example of how this is used.

mmatera:
LGTM

PR #1129, opened 2024-10-15
Speed up list processing

rocky:
Remove unnecessary evaluations for literal List evaluations.

### DIFF
--- a/mathics/eval/nevaluator.py
+++ b/mathics/eval/nevaluator.py
@@ -40,8 +40,10 @@ def eval_N(
     result = eval_NValues(evaluated_expression, prec, evaluation)
     if result is None:
         return expression
-    if isinstance(result, Number):
+
+    if result.is_literal:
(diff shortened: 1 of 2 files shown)


mathics/eval/nevaluator.py
```diff
@@ -83,6 +79,21 @@ def eval_NValues(
         result._build_elements_properties()
         return result
 
+    # Get the precision goal to use in non-integer numbers
```
mmatera, line 82: good catch!

mmatera:
@rocky, tell me when this is ready.

rocky:
> @rocky, tell me when this is ready.

It is ready. There will probably be more. But I like small easier-to-follow commits. 

I think there is a way that to_python to pick up (Python) literal values instead of going through the entire list.

mmatera (approved):
LGTM

PR #1130, opened 2024-10-15
Go over drawing.plot and ...

rocky:
add links to functional_iteration


mathics/core/list.py
```diff
@@ -110,12 +110,7 @@ def evaluate_elements(self, evaluation: Evaluation) -> Expression:
             return self
 
         new_list = ListExpression(*elements)
-        # TODO: we could have a specialized version of this
-        # that keeps self.value up to date when that is
-        # easy to do. That is left of some future time to
-        # decide whether doing this this is warranted.
         new_list._build_elements_properties()
-        new_list.value = None
```
rocky, line 118: @mmatera do you remember why we null'd `new_list.value` and or what it is that changes `new_list` without creating a new list and rebuilding properties?

mmatera: It looks like we wanted to avoid copying the elements, but we want to have the field `value` in the object.

mmatera (approved):
LGTM

PR #1131, opened 2024-10-15
Romberg is deprecated...

rocky:
and will be removed by SciPyi 1.15

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -33,6 +33,7 @@ API incompatibility
 -------------------
 
 * ``Matcher`` now requires an additional ``evaluation`` parameter
+* ``Romberg`` removed as an ``NIntegrate[]`` method. It is depcrecated in SciPy and is to be removed by SciPy 1.15.
 
 
 Bugs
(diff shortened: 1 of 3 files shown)


rocky:
Scipy has `dblquad`, `tplquad, and  `cumulative_trapezoid`  that we don't currently support. But that is left to someone who understands this better.  It also looks like we don't use SciPy's `simpson`.

mmatera:
> Scipy has `dblquad`, `tplquad`, and `cumulative_trapezoid` that we don't currently support. But that is left to someone who understands this better. It also looks like we don't use SciPy's `simpson`.

I didn't include these methods because they do not accept general functions as an argument. `dblquad`, `tplquad` only work for 2 and three variable functions respectively, while `cumulative_trapezoid` and `simpson` estimate integrals using arrays of numbers instead of functions.

mmatera (approved):
LGTM

rocky:
> > Scipy has `dblquad`, `tplquad`, and `cumulative_trapezoid` that we don't currently support. But that is left to someone who understands this better. It also looks like we don't use SciPy's `simpson`.
> 
> I didn't include these methods because they do not accept general functions as an argument. `dblquad`, `tplquad` only work for 2 and three variable functions respectively, while `cumulative_trapezoid` and `simpson` estimate integrals using arrays of numbers instead of functions.

Thanks for the explanation.

Down the line, it might be possible to analyze the expression passed to determine when these are applicable and handle these. However, with all the other stuff to do, that's best left for when a situation that needs these arises.

PR #1136, opened 2024-10-20
Convert RootSum to SymPy

davidar:
I split this out into a separate PR as I think it's going to require some more work. We already had a SymPy -> Mathics conversion for RootSum, this adds one going the other way too. However, it results in some tests failing as it means that Simplify automatically expands RootSums now, not sure if we want to add some hints to prevent that from happening.


mmatera:
@davidar 
the problem with the failing test is that by default, WMA and Mathics do not expand  `RootSum` for polynomials of degree higher than 2. This was explicitly stated in the WMA 12.2 documentation:
```
   For linear and quadratic polynomials f[x], Root[f,k] is automatically reduced to explicit rational or radical form. 
   For other polynomials, ToRadicals can be used to convert to explicit radicals.
```
This is not explicitly stated in Wolfram Reference now, but seems to still be true according to the content of 
https://reference.wolfram.com/language/tutorial/ManipulatingEquationsAndInequalities.html#28070

davidar:
I fixed the issue @mmatera noted, but I can't quite get mathics to sort the polynomial terms in a consistent order. In fact it behaves kind of strangely whenever the terms are anything more complicated than an atom:

```py
In[9]:= F^5+2F+1
Out[9]= 1 + 2 F + F ^ 5

In[10]:= F[]^5+2F[]+1
Out[10]= 1 + F[] ^ 5 + 2 F[]
```

WMA sorts both of these in order of increasing powers (e.g. `1 + 2*F[] + F[]^5`). I'm having trouble following the `get_sort_key` logic to figure out where this discrepancy is arising...

rocky:
> I fixed the issue @mmatera noted, but I can't quite get mathics to sort the polynomial terms in a consistent order. In fact it behaves kind of strangely whenever the terms are anything more complicated than an atom:
> 
> ```python
> In[9]:= F^5+2F+1
> Out[9]= 1 + 2 F + F ^ 5
> 
> In[10]:= F[]^5+2F[]+1
> Out[10]= 1 + F[] ^ 5 + 2 F[]
> ```
> 
> WMA sorts both of these in order of increasing powers (e.g. `1 + 2*F[] + F[]^5`). I'm having trouble following the `get_sort_key` logic to figure out where this discrepancy is arising...

The problem with the ordering is something that should be addressed in boxing the expression for output. My expectation is that this would be fixed in the process of revising the way boxing works.

davidar:
@rocky Ok, I won't worry too much about it now then, I'll just adjust the ordering in the tests for now

rocky:
LGTM

mmatera:
@rocky, @davidar, as I understand this, the result of a `RootSum` and `Root` should be a kind of BuiltinElement expression (like `Pattern` or `Image`) that stores both the Sympy and the WMA canonical forms. This kind of elements does not need to be reevaluated, since they represents fixed numbers/objects.

In any case, I am OK if we merge this as it is, and reformulate it later.

PR #1139, opened 2024-10-20
We can tolerate PyPy now...

### LINKED ISSUE #661: PyPy not supported
## Description

1. Attempting to install `Mathics-omnibus` gets stuck at `Installing build dependencies ... |`.
2. Attempting to install `Mathics3` raises the following error:
```
gcc -pthread -DNDEBUG -O2 -fPIC -I/home/hyperswine/.pyenv/versions/pypy3.9-7.3.10/include/pypy3.9 -c lib/recordclass/_litelist.c -o build/temp.linux-x86_64-3.9/lib/recordclass/_litelist.o
lib/recordclass/_litelist.c: In function ‘litelist_resize’:
lib/recordclass/_litelist.c:31:56: error: ‘PyLiteListObject’ {aka ‘PyListObject’} has no member named ‘ob_item
```

## How to Reproduce

1. From a popos (ubuntu 22.04) environment with pypy 3.9.15, run `pip install Mathics-omnibus`. And get the error in `1`
2. From the same environment, run `pip install Mathics3`. And get the error in `2`

## Expected behavior

Not get stuck at installing build dependencies and not raise an error. I did check the `recordclass` source code, and it seems like the `PyLiteListObject` struct did not in fact contain `ob_item`. Unless Im seeing things, that does seem like a problem with that dependency, so I actually don't know why `Mathics3` works when its including that dep.

## Your Environment

- PopOS 22.04, Intel + Nvidia
- Pypy 3.9.15 through pyenv
- No other mathematica /mathics installs on the system

rocky:
Fixes #661

### DIFF
--- a/mathics/eval/drawing/plot.py
+++ b/mathics/eval/drawing/plot.py
@@ -134,6 +134,7 @@ def compile_quiet_function(expr, arg_names, evaluation, list_is_expected: bool):
 
 
 def eval_ListPlot(
+    # TODO: plot_groups should be a tuple only?
     plot_groups: Union[list, tuple],
     x_range: list,
     y_range: list,
--- a/mathics/eval/drawing/plot.py
+++ b/mathics/eval/drawing/plot.py
@@ -134,7 +134,7 @@ def compile_quiet_function(expr, arg_names, evaluation, list_is_expected: bool):
 
 
 def eval_ListPlot(
-    plot_groups: list,
+    plot_groups: Union[list, tuple],
     x_range: list,
     y_range: list,
     is_discrete_plot: bool,
--- a/mathics/eval/nevaluator.py
+++ b/mathics/eval/nevaluator.py
@@ -29,7 +29,7 @@ def eval_N(
     expression: BaseElement,
     evaluation: Evaluation,
     prec: BaseElement = SymbolMachinePrecision,
-) -> BaseElement:
+) -> Optional[BaseElement]:
     """
     Equivalent to Expression(SymbolN, expression).evaluate(evaluation)
     """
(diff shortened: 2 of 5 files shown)


mathics/eval/nevaluator.py
```diff
@@ -29,7 +29,7 @@ def eval_N(
     expression: BaseElement,
     evaluation: Evaluation,
     prec: BaseElement = SymbolMachinePrecision,
-) -> BaseElement:
```
rocky, line 32: This type mismatch and the one above were noticed when looking at another bug we have.

There is pollution to values of some non-Python data. (Probably some structure is modifying a list in place).  

A fix for the underlying problem will occur down the line.

mathics/eval/drawing/plot.py
```diff
@@ -134,7 +134,7 @@ def quiet_f(*args):
 
 
 def eval_ListPlot(
-    plot_groups: list,
```
mmatera, line 137: Should it not be better to ensure using one type or the other?

rocky: Hmm...  

plot_groups is sometimes the result of `expr.to_python()`. For a `ListExpression` this is a list.
In the case of constant data we have a tuple; here, this is synonymous with an immutable list. 

I think plot_groups should be a tuple - immutable. However, I don't think we can change the return type of `to_python()` and I don't think we can change it to tuple on return without first detecting that it is a list. Otherwise, we would be wrapping a scalar into a tuple. 

So, while I agree it would be better to ensure a tuple is passed in, at this stage, I am not sure I want take the risk of converting a list to tuple right now. 

I have made note of this issue though in commit 016f6be125b97f2e0a2329e28999495f2464bda6

mmatera: OK, I think it would be enough by now.

mmatera (approved):
LGTM

PR #1141, opened 2024-10-21
Add `FileNameDrop[]`

rocky:
Left to do are changes with mixed negative and positive to integer arg values. I don't understand the behavior though. 

And this isn't needed for the ChemTools loader anyway. (But ChemTools has other bugs that still need work).


mmatera (commented):
I had tested locally and works the last time... Let me see what is happening.

mmatera:
If the second coefficient is negative, and saturates the elements in the math, the result should be the empty string. Instead, I am getting
```
In[1]:=  FileNameDrop["a/b/c", -3]
Out[1]= "a/b/c"

In[2]:=  FileNameDrop["a/b/c", -4]
Out[2]= "c"
```

Other cases seem to work fine.

rocky:
> FileNameDrop["a/b/c", -3]

Fixed in more recent commits.

PR #1143, opened 2024-10-25
Add CI job for Pyodide

davidar:
I figured it was worth testing the Pyodide support (#1099) a bit more comprehensively. The changes to the tests are just to skip the handful of cases that Pyodide doesn't support related to I/O and threading.

### DIFF
--- a/.github/workflows/pyodide.yml
+++ b/.github/workflows/pyodide.yml
@@ -51,7 +51,9 @@ jobs:
 
           # Activate the virtual environment
           source .venv-pyodide/bin/activate
-          
+
+          pip install "setuptools>=70.0.0" PyYAML click packaging
+
           git clone https://github.com/Mathics3/mathics-scanner.git
           cd mathics-scanner/
           pip install --no-build-isolation -e .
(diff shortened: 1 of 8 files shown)


rocky:
This is awesome! Thanks! 

A bit off topic and I will add a feature request for this soon. So consider this a draft of that for now....

The other day I tried to recreate what you did on Pyodide from scratch and was not able to. I see that there are Pyodide packages. And I thought maybe we should add one form Mathics.  It looked like I would need to add one for all of the dependent packages like PyYAML as well?

davidar:
Oh, creating packages specifically for pyodide is only necessary for libraries that use C extensions. For pure python packages like mathics it can just use the same wheels as regular python (which is all I built for my jupyterlite demo). The only reason it doesn't work with the release currently on PyPI is because of the llvmlite dependency, once the next release is uploaded without that, pyodide will be able to just pip install the regular release.

rocky:
> Oh, creating packages specifically for pyodide is only necessary for libraries that use C extensions. For pure python packages like mathics it can just use the same wheels as regular python (which is all I built for my jupyterlite demo). The only reason it doesn't work with the release currently on PyPI is because of the llvmlite dependency, once the next release is uploaded without that, pyodide will be able to just pip install the regular release.

Ok - thanks for the information!   I expect another release before the end of the year.

PR #1145, opened 2024-10-26
Fix mypy errors

davidar:
I've been going through fixing mypy errors, and have gotten through about half of them. In the process I've added some extra error handling that fixes some exceptions I've seen due to methods being called on the wrong types.

Current master:

> Found 1718 errors in 183 files (checked 443 source files)

This branch:

> Found 872 errors in 155 files (checked 443 source files)

Feel free to squash all these commits, the only reason I've left them separate is that I made sure the test suite was passing after each one.

### DIFF
--- a/mathics/builtin/box/expression.py
+++ b/mathics/builtin/box/expression.py
@@ -1,7 +1,7 @@
 # This is never intended to go in Mathics3 docs
 no_doc = True
 
-from typing import Optional, Sequence
+from typing import Optional, Sequence, Union
 
 from mathics.core.attributes import A_PROTECTED, A_READ_PROTECTED
 from mathics.core.builtin import BuiltinElement
@@ -126,7 +126,7 @@ class BoxExpression(BuiltinElement, BoxElementMixin):
         raise ValueError("BoxExpression.head is write protected.")
 
     def has_form(
-        self, heads: Sequence[str] | str, *element_counts: Optional[int]
+        self, heads: Union[Sequence[str], str], *element_counts: Optional[int]
     ) -> bool:
         """
         element_counts:
--- a/mathics/builtin/box/expression.py
+++ b/mathics/builtin/box/expression.py
@@ -1,6 +1,8 @@
 # This is never intended to go in Mathics3 docs
 no_doc = True
 
+from typing import Optional, Sequence
+
 from mathics.core.attributes import A_PROTECTED, A_READ_PROTECTED
 from mathics.core.builtin import BuiltinElement
 from mathics.core.element import BoxElementMixin
@@ -109,8 +111,8 @@ class BoxExpression(BuiltinElement, BoxElementMixin):
     def get_lookup_name(self):
         return self.get_name()
 
-    def get_sort_key(self) -> tuple:
-        return self.to_expression().get_sort_key()
+    def get_sort_key(self, pattern_sort=False) -> tuple:
+        return self.to_expression().get_sort_key(pattern_sort)
 
     def get_string_value(self):
         return "-@" + self.get_head_name() + "@-"
@@ -119,7 +121,13 @@ class BoxExpression(BuiltinElement, BoxElementMixin):
     def head(self):
         return self.get_head()
 
-    def has_form(self, heads, *element_counts):
+    @head.setter
+    def head(self, value):
+        raise ValueError("BoxExpression.head is write protected.")
+
+    def has_form(
+        self, heads: Sequence[str] | str, *element_counts: Optional[int]
+    ) -> bool:
         """
         element_counts:
             (,):        no elements allowed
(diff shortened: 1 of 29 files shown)


mathics/builtin/box/expression.py
```diff
@@ -1,6 +1,8 @@
 # This is never intended to go in Mathics3 docs
 no_doc = True
```
rocky, line 2: BTW current linters want  `no_doc = True`  after all imports.  But if you don't want to address this here it can be done in a future PR.

mathics/builtin/list/constructing.py
```diff
@@ -349,7 +350,7 @@ def eval_n(self, li, n, evaluation: Evaluation):
             py_n = min(n.get_int_value(), len(li.elements))
         elif n.has_form("List", 1) and isinstance(n.elements[0], Integer):
             py_n = n.elements[0].get_int_value()
-            rs = (py_n,)
+            rs = [py_n]
```
rocky, line 353: Tuple (immutable object) is better if we can accommodate that.

mathics/builtin/list/constructing.py
```diff
@@ -359,12 +360,12 @@ def eval_n(self, li, n, evaluation: Evaluation):
...
         if py_n is None or py_n < 0:
             evaluation.message(
-                self.get_name(), "nninfseq", Expression(self.get_name(), li, n)
+                self.get_name(), "nninfseq", Expression(Symbol(self.get_name()), li, n)
             )
             return
 
         if rs is None:
-            rs = range(py_n + 1)
+            rs = list(range(py_n + 1))
```
rocky, line 368: As above, tuple would be better if that is possible.

mathics/builtin/list/rearrange.py
```diff
@@ -363,7 +365,7 @@ def _rotate(self, expr, n, evaluation: Evaluation):
             return expr
 
         index = (self._sign * n[0]) % len(elements)  # with Python's modulo: index >= 1
-        new_elements = chain(elements[index:], elements[:index])
+        new_elements = list(chain(elements[index:], elements[:index]))
```
rocky, line 368: As above, tuple would be better if this does not cause a lot of distruption.

mathics/builtin/numbers/calculus.py
```diff
@@ -1046,7 +1052,7 @@ def prepare_sympy(self, elements):
                 return [elements[0]] + x.elements
         return elements
 
-    def from_sympy(self, elements: Iterable) -> Expression:
+    def from_sympy(self, elements: list) -> Expression:
```
rocky, line 1055: "elements" we would prefer to be a tuple. Modifying the elements in place is generally not something we want to make easy. 

In the past, it was thought that this was done for "efficiency", but it just leads to subtle bugs.  We have also moved to nothing list expression properties (whether a list is already sorted, is a literal, can be represented in Python fully, etc).  Tuple help ensure these properties are up to date while lists will tend to allow for subtle mistakes.

mathics/core/builtin.py
```diff
@@ -264,7 +263,7 @@ def contribute(self, definitions, is_pymodule=False):
                 f"illegal option mode {option_syntax}; check $OptionSyntax."
             )
 
-        rules = []
+        rules: List[BaseRule] = []
```
rocky, line 266: Again, if Tuple is not too disruptive that would be better.

rocky:
Reducing mypy errors has long been a desire. So I really appreciate this. 

The main ask is to be sensitive as to when to use a list versus a tuple.   I don't think one can know in advance whether the disruption is too great.  Too often in this code base the mess is too great that we can't tackle an entire problem in one go. Instead, personally I have found that for me, I have to do it little by little. 

Also, if you would like direct write access to the repositories, let us know.

mathics/builtin/box/expression.py
davidar: Yeah, let's leave that for another one.

mathics/builtin/list/constructing.py
```diff
@@ -10,6 +10,7 @@
 
 import typing
 from itertools import permutations
+from typing import Optional, Tuple
```
rocky, line 13: Thanks!

rocky:
LGTM. This is a large PR, in the future smaller ones would be nice. @mmatera - your comments?

davidar:
> LGTM. This is a large PR, in the future smaller ones would be nice.

Yeah, sorry about that, I couldn't really find a neat way to cleave it up so I kind of just kept going until I got bored of it...

> Also, if you would like direct write access to the repositories, let us know.

Sure, if that would make it easier to make these kinds of changes incrementally?

rocky:
> > LGTM. This is a large PR, in the future smaller ones would be nice.
> 
> Yeah, sorry about that, I couldn't really find a neat way to cleave it up so I kind of just kept going until I got bored of it...
> 
> > Also, if you would like direct write access to the repositories, let us know.
> 
> Sure, if that would make it easier to make these kinds of changes incrementally?

An invite should have been sent.

mmatera (approved):
Very nice! Thank you very much! LGTM

PR #1147, opened 2024-10-27
Fix remaining mypy errors in mathics.core

davidar:
This fixes the remaining mypy errors in `mathics.core`, with the exception of `mathics.core.expression` which needs a bit more work.

### DIFF
--- a/mathics/core/builtin.py
+++ b/mathics/core/builtin.py
@@ -13,7 +13,18 @@ import re
 from abc import ABC
 from functools import total_ordering
 from itertools import chain
-from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Union, cast
+from typing import (
+    Any,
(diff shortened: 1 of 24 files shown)


rocky:
LGTM

mathics/core/builtin.py
```diff
@@ -636,7 +653,7 @@ def get_mpmath_function(self, args):
     def eval(self, z, evaluation: Evaluation):
         "%(name)s[z__]"
 
-        args = numerify(z, evaluation).get_sequence()
+        args = cast(Sequence[Number], numerify(z, evaluation).get_sequence())
```
mmatera, line 656: Since this just makes silent a warning about something that probably should produce a warning, maybe it worth to add a comment before this line

mmatera (approved):
LGTM

PR #1150, opened 2024-10-30
Add CI job for mypy

davidar:
I've fixed up the remaining mypy errors outside of:

- `mathics.builtin` and `mathics.eval`: these still have quite a few errors, but are going to be a bigger job to go through than the core modules, and are probably easier to just knock off one by one incrementally
- `test/` which I'd say are lower priority for type checking
- `mathics/benchmark.py`: is this still used? It just crashes when I try to run it

For now I've just set the mypy config to ignore these directories so we can run a CI job to ensure that the other core modules continue to typecheck, but might be better to be a bit more selective.

There were also a couple bits of code I've commented out as they appeared to be dead as best I could tell, but tell me if I'm mistaken.

### DIFF
--- a/mathics/core/builtin.py
+++ b/mathics/core/builtin.py
@@ -26,7 +26,7 @@ from typing import (
     cast,
 )
 
-import mpmath  # type: ignore[import-untyped]
+import mpmath
 import pkg_resources
 import sympy
 
--- a/mathics/core/builtin.py
+++ b/mathics/core/builtin.py
@@ -196,7 +196,7 @@ class Builtin:
     rules: Dict[str, Any] = {}
     formats: Dict[str, Any] = {}
     messages: Dict[str, Any] = {}
-    options: Optional[Dict[str, Any]] = {}
+    options: Dict[str, Any] = {}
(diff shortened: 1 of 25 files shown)


mathics/core/builtin.py
```diff
@@ -244,6 +244,7 @@ def contribute(self, definitions: Definitions, is_pymodule=False):
 
         option_syntax = "Warn"
 
+        assert self.options is not None
```
mmatera, line 247: then `options` is not `Optional`, right?

mathics/format/mathml.py
```diff
@@ -303,24 +303,6 @@ def graphicsbox(self, elements=None, **options) -> str:
 add_conversion_fn(GraphicsBox, graphicsbox)
 
 
-def sqrtbox(self, **options):
```
mmatera, line 306: Why is this gone? Is it repeated in any other place?

mathics/core/builtin.py
davidar: True, I had just copied this from BaseElement

mmatera:
@davidar, let's merge this when you feel it is ready

PR #1155, opened 2024-11-05
2d print

mmatera:
This is a kind of experiment. In WMA, in opposition to `InputForm`  which produces an "inline" representation of an expression,   `OutputForm` consists in a "2D" text-like representation of an expression. There are some examples:
```
In[1]:= a/b                                                                     

        a
Out[1]= -
        b

In[2]:= Sqrt[a]                                                                 

Out[2]= Sqrt[a]

In[3]:= Integrate[f[x],x]                                                       

Out[3]= Integrate[f[x], x]

In[4]:= Sqrt[a]^q                                                               

         q/2
Out[4]= a

In[5]:= MatrixForm[Table[i*j,{i,2},{j,2}]]                                      

Out[5]//MatrixForm= 1   2

                    2   4
```
In this way, the output in `OutputForm` is formatted in a similar way to the *prettyprint* format of `Sympy`.

This PR provides a way to partially reproduce this behavior when an expression is wrapped by `OutputForm`, and a variable `$Use2DOutputForm` is set to True.

The support is not complete, and probably can be improved using the code  in `sympy.prettyprint`. Also, it currently does not work in Mathics-Django, because it is not able to print strings with line breaks.

Here are some examples in Mathics, under this branch:

```
In[1]:=  $Use2DOutputForm=True
Out[1]= True

In[2]:= OutputForm[a/b]
Out[2]= 
         a 
        ---
         b 

In[3]:= OutputForm[Sqrt[a]]
Out[3]= 
        __
        |a

In[4]:= Integrate[F[x],x]^2//OutputForm
Out[4]= 
                   2
         /+         
         |  F[x] dx 
        +/          

In[5]:= MatrixForm[Table[i*j,{i,2},{j,2}]]
Out[5]//MatrixForm= 1   2
                    
                    2   4
```

### DIFF
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -1,0 +1,40 @@
+name: Mathics3 (Type checking)
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.12']
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        sudo apt update -qq && sudo apt install llvm-dev remake
+        python -m pip install --upgrade pip
+        # We can comment out after next Mathics-Scanner release
+        # python -m pip install -e git+https://github.com/Mathics3/mathics-scanner#egg=Mathics-Scanner[full]
+        git clone https://github.com/Mathics3/mathics-scanner.git
+        cd mathics-scanner/
+        pip install -e .
+        cd ..
+
+    - name: Install Mathics with minimum dependencies
+      run: |
+        make develop
+    - name: Run mypy
+      run: |
+        pip install mypy==1.13 sympy==1.12
+        touch ./mathics-scanner/mathics_scanner/py.typed
+        pip install ./mathics-scanner/
+        mypy --install-types --non-interactive mathics
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,7 @@ By `@davidar <https://github.com/davidar>`_:
 * ``PolyLog``
 * ``PowersRepresentations``
 * ``ReverseSort``
+* ``RootSum``
 * ``SeriesCoefficient``
 * ``SquaresR``
 * ``Subfactorial``
--- a/mathics/__init__.py
+++ b/mathics/__init__.py
@@ -3,7 +3,7 @@
 import platform
 import sys
 from importlib import import_module
-from typing import Dict
+from typing import Dict, Tuple
 
 from mpmath import __version__ as mpmath_version
 from numpy import __version__ as numpy_version
@@ -26,7 +26,7 @@ version_info: Dict[str, str] = {
 
 # optional_software contains a list of Python packages
 # that add functionality but are optional
-optional_software: Dict[str, str] = (
+optional_software: Tuple[str, ...] = (
     "cython",
     "lxml",
     "matplotlib",
--- a/mathics/builtin/arithfns/basic.py
+++ b/mathics/builtin/arithfns/basic.py
@@ -14,6 +14,7 @@ from mathics.core.atoms import (
     Integer1,
     Integer3,
     Integer310,
+    Integer400,
     IntegerM1,
     Number,
     Rational,
@@ -48,6 +49,7 @@ from mathics.core.symbols import (
     SymbolNull,
     SymbolPower,
     SymbolTimes,
+    SymbolTrue,
 )
 from mathics.core.systemsymbols import (
     SymbolBlank,
@@ -152,7 +154,7 @@ class Divide(BinaryOperator):
     default_formats = False
 
     formats = {
-        (("InputForm", "OutputForm"), "Divide[x_, y_]"): (
+        ("InputForm", "Divide[x_, y_]"): (
             'Infix[{HoldForm[x], HoldForm[y]}, "/", 400, Left]'
         ),
     }
@@ -166,8 +168,25 @@ class Divide(BinaryOperator):
             "FractionBox[MakeBoxes[x, f], MakeBoxes[y, f]]"
         ),
     }
-
     summary_text = "divide"
+
+    def format_outputform(self, x, y, evaluation):
+        "OutputForm: Divide[x_, y_]"
+        use_2d = (
+            evaluation.definitions.get_ownvalues("System`$Use2DOutputForm")[0].replace
+            is SymbolTrue
+        )
+        if not use_2d:
+            return Expression(
+                SymbolInfix,
+                ListExpression(
+                    Expression(SymbolHoldForm, x), Expression(SymbolHoldForm, y)
+                ),
+                String("/"),
+                Integer400,
+                SymbolLeft,
+            )
+        return None
 
 
 class Minus(PrefixOperator):
@@ -406,10 +425,21 @@ class Power(BinaryOperator, MPMathFunction):
             Expression(SymbolPattern, Symbol("x"), Expression(SymbolBlank)),
             RationalOneHalf,
         ): "HoldForm[Sqrt[x]]",
-        (("InputForm", "OutputForm"), "x_ ^ y_"): (
+        (("InputForm",), "x_ ^ y_"): (
             'Infix[{HoldForm[x], HoldForm[y]}, "^", 590, Right]'
         ),
-        ("", "x_ ^ y_"): (
+        (("OutputForm",), "x_ ^ y_"): (
+            "If[$Use2DOutputForm, "
+            "Superscript[HoldForm[x], HoldForm[y]], "
+            'Infix[{HoldForm[x], HoldForm[y]}, "^", 590, Right]]'
+        ),
+        (
+            (
+                "StandardForm",
+                "TraditionalForm",
+            ),
+            "x_ ^ y_",
+        ): (
             "PrecedenceForm[Superscript[PrecedenceForm[HoldForm[x], 590],"
             "  HoldForm[y]], 590]"
         ),
--- a/mathics/builtin/atomic/numbers.py
+++ b/mathics/builtin/atomic/numbers.py
@@ -17,10 +17,7 @@ precision is supported.
 However, things like 'N[Pi, 100]' should work as expected.
 """
 
-from functools import lru_cache
-
 import mpmath
-import sympy
 
 from mathics.core.atoms import (
     Integer,
@@ -33,6 +30,7 @@ from mathics.core.atoms import (
 from mathics.core.attributes import A_LISTABLE, A_PROTECTED, A_READ_PROTECTED
 from mathics.core.builtin import Builtin, Predefined
 from mathics.core.convert.python import from_python
+from mathics.core.evaluation import Evaluation
 from mathics.core.expression import Expression
 from mathics.core.list import ListExpression
 from mathics.core.number import (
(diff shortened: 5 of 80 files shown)


rocky:
2D printing does not feel like a "core" function. Instead it is purely a formatting function, done after boxing. 

Can this be reformulated as a formatting process instead?

mmatera:
@rocky, indeed, my idea when I started writting this (circa July 2023) was to out in a module. But still I was not able to redefine the boxing mechanism and the package interface to make it works as an independent package. In any case, most of the code is independent enough to move it to a package when the other things get ready.

rocky:
It does not necessarily have to be in a Mathics3 module outside of mathics core, but it should be outside of `mathics.core` and we should figure out what adjustments to the boxing mechanism we need to make in order to allow this to work.

(And yes, if we could hook into SymPy's format routines that would also be awesome.)

I fear that we are making things worse for us in the long run by violating modularity or separation of phases. This kind of thing has happened in this project in the past, and it has caused a lot of extra work that has taken a long time to address (and some of it hasn't been fully addressed even now).

rocky:
```
n[1]:=  $Use2DOutputForm=True
Out[1]= True

In[2]:= OutputForm[a/b]
Out[2]= 
         a 
        ---
         b 
```

wolframscript has: 

```
     a
     -
     b
```

Similarly: these appear differently in wolframscript: 

```
In[3]:= OutputForm[Sqrt[a]]
Out[3]= 
        __
        |a

In[4]:= Integrate[F[x],x]^2//OutputForm
Out[4]= 
                   2
         /+         
         |  F[x] dx 
        +/          
```

rocky:
Comparing with what SymPy produces: 

```
>>> x, y, z, t = symbols("x y z t")
>>> x / y
x
─
y
>>> sqrt(2)
√2
>>> xx, yy, = symbols("xx, yy")
>>> xx / yy
xx
──
yy
````

This is different from wolframscript in that some Unicode symbols seem to be used: `─`  and  `√` and I think that is okay if we hook into SymPy's formatting. However we have a 3rd invention formulation and we have encountered this kind of problem as well in the past.

**Comment by mmatera** [comment]
> √2

Indeed, what I build was a kind here is a kind of experiment. I didn't try too hard to mimic exactly what Sympy or WMA does, apart from building a "2D" text representation, that is what `OutputForm` is about.
In any case, regarding modularity, we could also move the *mathics/core/prettyprint.py* module to the formatting folder, or even to a different project, since it is more or less independent.

File: mathics/format/pane_text.py [hunk 1]
```diff
@@ -233,7 +233,7 @@ def fraction(a: Union[TextBlock, str], b: Union[TextBlock, str]) -> TextBlock:
         a = TextBlock(a)
     if isinstance(b, str):
         b = TextBlock(b)
-    width = max(b.width, a.width) + 2
+    width = max(b.width, a.width)
```
rocky, line 236: Instead of adding more original code, please look into hooking into SymPy's 2D formatting mechanism.  

Otherwise, this may be another kind of thing where effort is put into creating something that is later removed,  because there is something that is more likely to be more complete and that we won't have to maintain.  

If it turns out that we can't use SymPy's 2D printing, _then_ we can go down this road.

mmatera: Indeed, the plan is to replace the use of `mathics.format.pane_text` by `sympy.printing.pretty.stringpict`, which is something that I already started to do.

 What is not possible to do is just using `sympy.pretty(expression.to_sympy())` because the translation function which is useful for evaluation, is not useful for formatting. 

For example
```
>>> import sympy
>>> from mathics.session  import MathicsSession
>>> session=MathicsSession()
>>> expr=session.evaluate("Integrate[f[x]/g[x]^2,x]")
>>> sympy.pretty(expr.to_sympy())
``` 
produces
```
⌠                                                                            
⎮ SympyExpression(_uGlobal`f, _uGlobal`x)[Global`f[Global`x]])               
⎮ ───────────────────────────────────────────────────────────── d(_uGlobal`x)
⎮                                                             2              
⎮ SympyExpression(_uGlobal`g, _uGlobal`x)[Global`g[Global`x]])               
⌡                                                                            
```
while
```
>>> expr=session.evaluate("Integrate[f[x]/g[x]^2,{x,a,b}]")
>>> print(sympy.pretty(expr.to_sympy()))
```
is not even able to identify the integrate symbol:
```
SympyExpression(_uSystem`Integrate, SympyExpression(_uGlobal`f, _uGlobal`x)/SympyExpression(_uGlobal`g, _uGlobal`x)**2, SympyExpression(_uSystem`List, _uGlobal`x, _uGlobal` ↪

↪ a, _uGlobal`b))[System`Integrate[System`Times[Global`f[Global`x], System`Power[Global`g[Global`x], -2]], {Global`x,Global`a,Global`b}]])
```

In any case, the purpose of this PR is to 
* considering if this 2D format is something that we would like to use in the REPL 
* bring the formatting routines closer to the one used in WL
* explore a possible design pattern (based on what we already have) to connect a formatted Mathics expression to a `prettyForm` object.

rocky: > Indeed, the plan is to replace the use of `mathics.format.pane_text` by `sympy.printing.pretty.stringpict`, which is something that I already started to do.
> 
> What is not possible to do is just using `sympy.pretty(expression.to_sympy())` because the translation function which is useful for evaluation, is not useful for formatting.
> 
> For example
> 
> ```
> >>> import sympy
> >>> from mathics.session  import MathicsSession
> >>> session=MathicsSession()
> >>> expr=session.evaluate("Integrate[f[x]/g[x]^2,x]")
> >>> sympy.pretty(expr.to_sympy())
> ```

This is clearly wrong because `expr` needs to be boxed first. Calling SymPy formatting routines are triggered from formatting boxes. 

This is setting up a strawman or superficial argument only to be able to shoot it down. 

The time spent adjusting the bar in a division I think is better spent towards getting to the skeleton of a possible solution.  The plan that you wanted written down has you working on revising Boxing. When that is in place, we might be in an even better position to work on the boxing to formatting step needed in character-based printing.

mmatera: > 
> This is clearly wrong because `expr` needs to be boxed first. Calling SymPy formatting routines are triggered from formatting boxes.
> 
> This is setting up a strawman or superficial argument only to be able to shoot it down.
> 

Please, do not take this in a wrong way:  the example was just a way to show some of the challenges in hooking up `sympy.pretty`: even at the level of symbols, a preprocessing must be done. Probably what is more easy to hook is the `sympy.printing.pretty.stringpict.prettyForm`, which is quite analogous to what I put in `pane_text`. 

`sympy.pretty` works using a `sympy.printing.pretty.PrettyPrinter` object that does something similar to what I did in `mathics.format.prettyprint`. 

> The time spent adjusting the bar in a division I think is better spent towards getting to the skeleton of a possible solution. The plan that you wanted written down has you working on revising Boxing. When that is in place, we might be in an even better position to work on the boxing to formatting step needed in character-based printing.

OK, but the skeleton of the solution that I propose is already here: When an expression is wrapped by `OutputForm` (even if `$Use2DOutputForm` is set to `False`) the formatting process follows a sequence closer to the one (I think) WMA follows.

In any case, I wanted to put this here, because I will need it to present my case when I propose the other changes.

mathics/format/pane_text.py
```diff
@@ -0,0 +1,465 @@
+"""
```
mmatera, line 1: All this module can be removed once `prettyprint.py` be reformulated using `sympy.prettyprint.stringpict.prettyForm`.

mathics/format/pane_text.py
rocky: > Please, do not take this in a wrong way: the example was just a way to show some of the challenges in hooking up `sympy.pretty`: even at the level of symbols, a preprocessing must be done. 

I was aware of that well before you started this PR. 

> Probably what is more easy to hook is the `sympy.printing.pretty.stringpict.prettyForm`, which is quite analogous to what I put in `pane_text`.
> 
> `sympy.pretty` works using a `sympy.printing.pretty.PrettyPrinter` object that does something similar to what I did in `mathics.format.prettyprint`.
> 

Except more effort, time, and thought was probably put into `sympy.printing.pretty.PrettyPrinter` object.  It may be that it is more elucidating for you to write some code so you understand the basic concepts rather than look at someone else's code. For me though this kind of thing is more of a distracting and possibly a dangerous activity, because I get the feeling that if I don't mention something this kind of thing will get into the code base and then we'll want to remove it later on. We have seen this kind of thing too often. Furthermore, we haven't dug out of the previous messes fully yet.


> 
> OK, but the skeleton of the solution that I propose is already here: When an expression is wrapped by `OutputForm` (even if `$Use2DOutputForm` is set to `False`) the formatting process follows a sequence closer to the one (I think) WMA follows.

There are probably very many situations that aren't covered and haven't been considered. And the first few things I tried I saw differences. 

BTW, I don't like the term "2D". This is character-based output. Most output, such as SVG, MathML, and LaTeX, is 2D. 

> 
> In any case, I wanted to put this here, because I will need it to present my case when I propose the other changes.

Personally, I would prefer if you discuss what changes you want to propose at a high level first. If I need detailed code to understand, _then_ we can code this out.   If you need to write some sample code for _yourself_ , sure do that. But unless others express interest in seeing this, these branches don't help me in a positive way. Rather, it feels negative because I see flailing about where it feels to me there shouldn't be flailing like this.

In my view, the developer docs describe in pretty good detail how the system transforms M-expressions to boxed-expressions, and then to formatted output.

mathics/format/pane_text.py
rocky: A step in the right direction!

mathics/format/pane_text.py
mmatera: > BTW, I don't like the term "2D". This is character-based output. Most output, such as SVG, MathML, and LaTeX, is 2D.
> 

This is the way in which Wolfram Reference refers to this kind of output:
imagen

https://reference.wolfram.com/language/tutorial/TextualInputAndOutput.html

But we could call it "PrettyPring" or StandardForm character-based output.

rocky: > > BTW, I don't like the term "2D". This is character-based output. Most output, such as SVG, MathML, and LaTeX, is 2D.
> 
> This is the way in which Wolfram Reference refers to this kind of output: 

Not quite. There is a difference between 2D and two-dimensional _using only keyboard characters_. 

To me, this is like shortening "strawberry" to " straw" instead of "berry". 

The wolfram terminology is generally well thought out. Sometimes people who make an attempt to paraphrase it get into trouble when the same care is not used.

mmatera: OK, but for naming functions, it would be a little bit long, isn't it? Here I didn´t want to call it "OutputForm", because in all the other places, "OutputForm"  is still a *"one-dimensional using only keyboard characters"* which most of the time is the same than `InputForm`, but with spaces between infix operators and operands.

mmatera: In any case,  the question is: supposing we found and agree on an implementation of this *"two-dimensional using only keyboard characters"*, is it something that we want to have (at least optional) available in the command line interface?

rocky: I guess what I was trying to say is that if you have to drop something in the description, dropping "character-based" is bad, in the same way that condensing "strawberry" to "straw" rather than "berry" is not helpful. 
"Character2D"  is not too long. But the word "character" is as important as 2D.

rocky: > In any case, the question is: supposing we found and agree on an implementation of this _"two-dimensional using only keyboard characters"_, is it something that we want to have (at least optional) available in the command line interface?

It feels to me that we are thinking about this the wrong way. To me, this is like asking if TeXForm should be in the command-line interface. And whether we should be able to return TeXForm as MathML in the command-line interface.  

To me, the focus of the implementation should be on how things are boxed and formatted in a generic and general way. Not about what is appropriate for a particular front end.   The current implementation is lacking in the generic and general nature. It was particularly more evident when this was in `mathics.core` which 2D character-based formatting is totally inappropriate for. 

Here is another example from our code base. We have these whacky and complicated regular expressions for handling doctests and docstrings. I imagine the person that started this may have thought it cool to recreate sphinx using regular expressions. Those regular expressions use some pretty advanced  and little-used tagging mechanisms.  If you want to show off how clever you can code, great.  But as far as handling the underlying _problem_ in a uniform, maintainable, and comprehensible way, this code totally fails. 

Let's defer character-based 2D output formatting until _after_ we have Boxing under control. Formatting is intimately tied to Boxing. And if we have a good Boxing mechanism, I think you'll see how easily character-based 2D formatting falls out from that.

mmatera: > > In any case, the question is: supposing we found and agree on an implementation of this _"two-dimensional using only keyboard characters"_, is it something that we want to have (at least optional) available in the command line interface?
> 
> It feels to me that we are thinking about this the wrong way. To me, this is like asking if TeXForm should be in the command-line interface. And whether we should be able to return MathML formatted output in the command-line interface.
> 

TeXForm is useful and needed at last to generate the PDF documentation. MathML is used in the Django front-end. PrettyPrint is something "pretty" but maybe is a waste of time to implement/maintain it. For example, probably we do not want to use to check doctests, because writing the expected results would be awkward.



> To me, the focus of the implementation should be on how things are boxed and formatted in a generic and general way. Not about what is appropriate for a particular front end. The current implementation is lacking in the generic and general nature. It was particularly more evident when this was in `mathics.core` which 2D character-based formatting is totally inappropriate for.
> 

Yes, but at least for me, it helps me to think why WMA implements boxing as it does. And this is all the reason I wrote and put this here. 
Now I am putting another PR, where all the *Character2D* code is stripped away. 


> Here is another example from our code base. We have these whacky and complicated regular expressions for handling doctests inside of docstrings. I imagine the person that started this may have thought it cool to recreate sphinx using regular expressions. Those regular expressions use some pretty advanced and little-used tagging mechanisms. If you want to show off how clever you can code, great. But as far as handling the underlying _problem_ in a uniform, maintainable, and comprehensible way, this code totally fails.
> 

OK

> Let's defer character-based 2D output formatting until _after_ we have Boxing under control. Formatting is intimately tied to Boxing. And if we have a good Boxing mechanism, I think you'll see how easily character-based 2D formatting falls out from that.

Sure. In any case, if you are OK with it, I will leave this here for a while.

rocky:
Just a FYI: logically 2D character-oriented formatting comes _after_  a pass over Boxing expressions, in the same way as Boxing follows after we get pattern selection and set assignment.

mmatera:
@rocky, I just try to keep this close to the main branch, to be able to pick the code from here when I need it.

rocky:
> @rocky, I just try to keep this close to the main branch, to be able to pick the code from here when I need it.

I think you'll find that when the time is right, it will be straightforward and obvious to just create what is needed from scratch.  But... to each his own.

mmatera:
The still valid part of this PR is in #1162

PR #1158, opened 2024-11-07
adding a test

mmatera:
By now, just adding a test for #1156 to see what does the CI

### DIFF
--- a/test/builtin/numbers/test_linalg.py
+++ b/test/builtin/numbers/test_linalg.py
@@ -211,6 +211,12 @@ def test_inverse(str_expr, str_expected, fail_msg, warnings):
             "SingularValueDecomposition[{1, {2}}]",
             None,
         ),
+        (
+            "A = Array[a, {2,2}]; Eigenvalues[A.ConjugateTranspose[A]] // FullSimplify",
+            None,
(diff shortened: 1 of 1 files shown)


rocky:
LGTM - I think we should merge this.

rocky:
One other thought. If this example is equally valid testing on a 2x2, that would be better here. 

The 3x3 is slow, and while that is interesting, we should focus on that separately. In testing, we want the _smallest_ example that exhibits a problem. If we want later to add a test for some performance issue, that is a _separate_ test.

test/builtin/numbers/test_linalg.py
```diff
@@ -211,6 +211,12 @@ def test_inverse(str_expr, str_expected, fail_msg, warnings):
             "SingularValueDecomposition[{1, {2}}]",
             None,
         ),
+        (
+            "A = Array[a, {2,2}]; Eigenvalues[A.ConjugateTranspose[A]] // FullSimplify",
```
mmatera, line 215: @rocky, the example is already for a 2x2 matrix.

rocky: I am sorry for the noise then and that I did not noticed previously.

mmatera: In any case, I split the test case in two parts, the first that computes the eigenvalues, and the second, that simplifies just the first eigenvalue.

PR #1160, opened 2024-11-09
handling `TimeConstrained[]` in SymPy evaluation

mmatera:
An idea to fix `TimeConstrained` and related functions, at least for evaluations involving sympy. 
Still I can produce a core dump if I ask for a ridiculus task in Sympy, but is a problem of Sympy, not from Mathics


mmatera:
@davidar, ideas about how to implement this in Pyodide are welcome.

davidar:
Unfortunately I don't think there is a good solution to that in pyodide yet, at least not without cooperation from the javascript environment hosting the pyodide interpreter (https://pyodide.org/en/stable/usage/keyboard-interrupts.html). Might just need to disable it in pyodide for now.

mathics/eval/sympy.py
```diff
@@ -12,34 +12,47 @@
 from mathics.core.convert.sympy import from_sympy, to_numeric_sympy_args
 
 
-def eval_sympy(self, z, evaluation):
+def eval_sympy_unconstrained(self, z, evaluation):
```
rocky, line 15: Let's call this simply "eval_sympy". as before. The variations can have more complicated names.

mmatera: The reason to use a more specific name is to be able to use it as it is (with the simpler name as an alias) or to use it inside the new routine that checks for timeouts.

rocky:
> Unfortunately I don't think there is a good solution to that in pyodide yet, at least not without cooperation from the javascript environment hosting the pyodide interpreter (https://pyodide.org/en/stable/usage/keyboard-interrupts.html). Might just need to disable it in pyodide for now.

Thanks for the information and link. It confirms what I had meant to look up.

What Mathics3 should do is provide a signal-handling mechanism that causes the current computation to stop after a certain amount of time (SIGALRM).

 Mathics3 should also trap other signals like SIGINT.  In wolframscript, Ctrl-C (SIGINT) goes into an Interrupt prompt. Later,  we should provide a compatible (and possibly much better) interrupt interface. 

If I have this correct, the link you have indicates that pyodide _does_ provide a means for sending trapping and sending signals back to Python. 

One other thing to note about Python's signal handling - only the main thread can trap signals.  I don't think this is a problem, just something to note.  

With this we _can_ provide a somewhat coarse mechanism for handling timeouts. 

@mmatera Now that you may have been just informed about signal handling and timeouts, please don't jump ahead and start implementing this in code. (If you want to experiment privately - by all means go ahead.)

If there is a need right now TimeConstrained, I'd appreciate knowing what the deal is. 
Otherwise, let's leave adding signal handlers in Mathics3 and having that work in pyodide for later.

mathics/eval/sympy.py
rocky: Ah - okay. This is fine. I am sorry I missed the assignment in the below code.

mmatera:
> @mmatera Please give me a couple of days for me to pull down this branch to try out. Thanks.

Sure, no hurry. I just put this because when I looked over the issue with `Simplify` I realized that this would be a way to support `TimeConstrained`  in systems that support Threads.

mathics/builtin/datentime.py
```diff
@@ -1055,7 +1055,13 @@ def eval(self, n, evaluation):
             )
             return
 
-        time.sleep(sleeptime)
+        steps = int(1000 * sleeptime)
+        while steps > 0:
```
rocky, line 1059: `for _ in range(steps)` is a little nicer. 

In general, whenever I see a polling loop like this, I often wonder if there isn't a better way. I can't think of one offhand. 

When I look at https://reference.wolfram.com/language/ref/Pause.html I note a discrepancy in our description. 

We have

we have 

> pauses for _n_ seconds.


where WMA says:

> pauses for _at least n_ seconds. 

which, in fact, matches the behavior we have better.

Note also that by polling we do increase the `Timing` value a little better. And that's also another reason why using `time.sleep()` by itself would better.

mathics/core/evaluation.py
```diff
@@ -140,7 +141,7 @@ def run_with_timeout_and_stack(request, timeout, evaluation):
     if success:
         return result
     else:
-        raise result[0].with_traceback(result[1], result[2])
+        raise result[1].with_traceback(result[1], result[2])
```
rocky, line 144: There single parameter to `with_traceback()` should be a tb parameter: result[2]. See PR.

mathics/builtin/datentime.py
```diff
@@ -1124,23 +1130,6 @@ class TimeConstrained(Builtin):
         the state of the Mathics3 kernel.
         """
 
-        # FIXME: these tests sometimes cause SEGVs which probably means
-        # that TimeConstraint has bugs.
-
-        # Consider testing via unit tests.
-        # >> TimeConstrained[Integrate[Sin[x]^1000000,x],1]
-        # = $Aborted
```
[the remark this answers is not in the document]
mmatera: Actually, with Pyston I got a core dump for that test. But it is reasonable: Sympy does not control by itself how much memory/work needs to converge.

mathics/core/evaluation.py
mmatera: OK. I just copied the line 143 from `mathics.core.evaluation`. Probably we need to fix it there too.

mathics/eval/sympy.py
```diff
@@ -0,0 +1,76 @@
...
+        try:
+            result = eval_sympy_unconstrained(self, z, evaluation)
+            queue.put((True, result))
+        except BaseException:
+            exc_info = sys.exc_info()
+            queue.put((False, exc_info))
+
+    queue = Queue(maxsize=1)  # stores the result or exception
+
+    def evaluate():
```
rocky, line 52: @mmatera where/how is `evaluate()` used?

mathics/core/evaluation.py
rocky: Both places should be fixed in the updated PR.

mathics/eval/sympy.py
rocky: @mmatera For what it's worth, I removed the `evaluate()` function and the tests still run fine. Can we remove this code?

mmatera: Yes. I forgot to remove it. I moved its code to `eval_sympy_unconstrained()`, so we can use it when `Threads` are not available.

PR #1163, opened 2024-11-11
MakeBoxes OutputForm / FullForm

mmatera:
This is the idea I am trying to elaborate and test in the "Character2D" branches. 
In WMA, "formatting" is the result of "render" a `Box` expression, and box expressions are generated by evaluating `MakeBoxes[expr]` in a way that is different from the standard evaluation.

In WMA, the "normal" evaluation of  the expression `MakeBoxes[expr, form]`  discards any user definition and always proceed as follows:

* Evaluate formatting rules (apply Format[....] rules) over `expr`.
* Then, applies MakeBoxes rules.
For example:
```

In[1]:= MakeBoxes[F[x_],fmt_]:=RowBox[{"F",MakeBoxes[x]}]                       

In[2]:= Format[F[x_],OutputForm]:="OutputForm F"                                

In[3]:= Format[F[x_],StandardForm]:="StandardForm F"                            

In[4]:= MakeBoxes[F[x],StandardForm]                                            

Out[4]= "StandardForm F"

In[5]:= MakeBoxes[F[x]//OutputForm,StandardForm]                                

Out[5]= InterpretationBox[PaneBox["OutputForm F"], OutputForm F, 
 
>    Editable -> False]

In[6]:= FormatValues[F]                                                         

Out[6]= {HoldPattern[OutputForm F] :> OutputForm F, 
 
>    HoldPattern[MakeBoxes[OutputForm F, StandardForm]] :> "StandardForm F"}

In[7]:= ??MakeBoxes                                                             

Out[7]= MakeBoxes[expr, form] is the low‐level function used in 
         
        >    Wolfram System sessions to convert expressions into boxes. 

         
        >     MakeBoxes[expr] is the function to convert expr
         
        >       to StandardForm boxes.
        MakeBoxes[F[x_], fmt_] := RowBox[{"F", MakeBoxes[x]}]
         
        Attributes[MakeBoxes]={HoldAllComplete}
```

Notice that evaluating `MakeBoxes[F[x], StandardForm]` does not returns the (down) value `RowBox[...]` (which is what happens in Mathics) but applies first the format rule. Then, MakeBoxes rules are applied over the result of formatting.
Notice also that MakeBoxes rules are looked first in the FormValues of the head of the resulting expression, and then in the downvaloes of MakeBoxes. 

Here we reset the StandardForm of F in terms of `Fs`, and set a "UpValue" for MakeBoxes[Fs]:
```
In[12]:= Format[F[x_],StandardForm]:={Fs[1]}                                    

In[13]:= MakeBoxes[Fs[x_],form_]^:="p"                                          
```

Now,  when the REPL wants to format `F[1]`, uses the `OutputForm`

```
In[14]:= F[1]                                                                   

Out[14]= OutputForm F
```

But if we ask for the StandardForm
```
In[16]:= F[1]//StandardForm                                                     

Out[16]//StandardForm= {p}
```
MakeBoxes looks in FormatValues[Fs] for a rule. Setting a DownValue for MakeBoxes does not have any effect, because the rules are first looked in the FormatValues of `Fs`.

```
In[17]:= MakeBoxes[Fs[u_],form_]:="q"                                           

In[18]:= F[1]//StandardForm                                                     

Out[18]//StandardForm= {p}
```

Here we see that UpSet[MakeBoxes[...], ....] stores the second argument as FormatValues instead of Upvalues:
```
In[19]:= ?? Fs                                                                  

Out[19]= Global`Fs
         MakeBoxes[Fs[x_], form_] ^:= "p"
          

In[20]:= FormatValues[Fs]                                                       

Out[20]= {HoldPattern[MakeBoxes[Fs[x_], form_]] :> p}
```

### DIFF
--- a/mathics/builtin/box/layout.py
+++ b/mathics/builtin/box/layout.py
@@ -146,6 +146,7 @@ class GridBox(BoxExpression):
     # >> MathMLForm[TableForm[{{a,b},{c,d}}]]
     #  = ...
     """
+
     options = {"ColumnAlignments": "Center"}
     summary_text = "low-level representation of an arbitrary 2D layout"
 
@@ -208,6 +209,33 @@ class InterpretationBox(BoxExpression):
 
     def eval_display(boxexpr, evaluation):
         """DisplayForm[boxexpr_InterpretationBox]"""
+        return boxexpr.elements[0]
+
+
+class PaneBox(BoxExpression):
+    """
+    <url>
+    :WMA link:
+    https://reference.wolfram.com/language/ref/InterpretationBox.html</url>
+
+    <dl>
+      <dt>'PaneBox[expr]'
+      <dd> is a low-level box construct, used in OutputForm.
+    </dl>
+
+    """
+
+    attributes = A_HOLD_ALL_COMPLETE | A_PROTECTED | A_READ_PROTECTED
+    summary_text = "box associated to panel"
+
+    def apply_display(boxexpr, evaluation, expression):
+        """ToExpression[boxexpr_PaneBox, form_]"""
+        return Expression(expression.head, boxexpr.elements[0], form).evaluate(
+            evaluation
+        )
+
+    def apply_display(boxexpr, evaluation):
+        """DisplayForm[boxexpr_PaneBox]"""
         return boxexpr.elements[0]
 
 
--- a/mathics/core/atoms.py
+++ b/mathics/core/atoms.py
@@ -1015,10 +1015,13 @@ class String(Atom, BoxElementMixin):
         return '"%s"' % self.value
 
     def atom_to_boxes(self, f, evaluation):
+        return self.make_boxes(f.get_name())
+
+    def make_boxes(self, f):
         from mathics.eval.makeboxes import _boxed_string
 
         inner = str(self.value)
-        if f in SYSTEM_SYMBOLS_INPUT_OR_FULL_FORM:
+        if f in ("System`InputForm", "System`FullForm"):
             inner = '"' + inner.replace("\\", "\\\\") + '"'
             return _boxed_string(inner, **{"System`ShowStringCharacters": SymbolTrue})
         return String('"' + inner + '"')
--- a/mathics/builtin/forms/output.py
+++ b/mathics/builtin/forms/output.py
@@ -64,7 +64,12 @@ from mathics.core.systemsymbols import (
     SymbolSubscriptBox,
     SymbolSuperscriptBox,
 )
-from mathics.eval.makeboxes import StringLParen, StringRParen, format_element
+from mathics.eval.makeboxes import (
+    StringLParen,
+    StringRParen,
+    format_element,
+    makeboxes_outputform,
+)
 from mathics.eval.testing_expressions import expr_min
 from mathics.format.outputform import expression_to_outputform_text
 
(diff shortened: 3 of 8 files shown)


mmatera:
In this PR, I partially mimic this behavior for OutputForm and FullForm by modifying `mathics.eval.makeboxes.format_element` to follow this path (call a function that is equivalent to evaluate MakeBoxes[...] in WMA)

mathics/builtin/box/layout.py
```diff
@@ -188,6 +188,33 @@ def eval_display(boxexpr, evaluation):
         return boxexpr.elements[0]
 
 
+class PaneBox(BoxExpression):
```
rocky, line 191: Why is the class called PaneBox when the WMA link goes to InterpretationBox?

mmatera:
```
In[1]:= MakeBoxes[OutputForm[a^2/b],StandardForm]                               

                                                    2
                                                   a
Out[1]= InterpretationBox[PaneBox[" 2\na\n--\nb"], --, Editable -> False]
                                                   b
```
A PaneBox is a block of multiline text, which by default is "centered". 
It can be produced by applying MakeBoxes to a `Pane` expression:

```
In[2]:= MakeBoxes[Pane["hola\nHello",10],StandardForm] 
Out[2]= PaneBox["hola\nHello", ImageSize -> 10]
```

rocky: Ok - thanks for the information. Given this, it seems this is more like a Pane than it is an InterpretationBox. I see that we can get `Information[PaneBox]` gives information.

Are there any other Boxes in this category? (Like ParentBox, it exists but is not documented). I can't find any, but I feel we've encountered this kind of situation where there is something that exists but is not documented. And how did we handle things there? 

But as for changing the link say from InterpretationBox to Pane (if that turns out to be the better thing to do),  that is not urgent right now. 

This PR is large and I think this PR should come after we work out some _simple_ basics of MakeBoxes. 

Also, the PR title is misleading because a lot of this is about character-based formatting on top of stuff related to MakeBoxes.

mathics/core/atoms.py
```diff
@@ -1081,10 +1081,13 @@ def __str__(self) -> str:
...
 
     def atom_to_boxes(self, f, evaluation):
+        return self.make_boxes(f.get_name())
+
+    def make_boxes(self, f):
         from mathics.eval.makeboxes import _boxed_string
 
         inner = str(self.value)
-        if f in SYSTEM_SYMBOLS_INPUT_OR_FULL_FORM:
+        if f in ("System`InputForm", "System`FullForm"):
```
rocky: Can `instance (f, SymblInputForm, SymbolFullForm)` be used instead?

mmatera: For some reason, this method receives strings instead symbols as input. We can change this, too.

mathics/eval/makeboxes/makeboxes.py
```diff
@@ -126,8 +126,21 @@ def int_to_string_shorter_repr(value: int, form: Symbol, max_digits=640):
     return String(value_str)
 
 
+def eval_makeboxes_outputform(expr, evaluation, form):
+    """
+    Build a 2D text representation of the expression.
```
rocky: If 2D is synonymous with `str(...)` on line 136, then something feels very wrong about putting this in `mathics.eval.makeboxes`.  2D is a _formatting_ kind of thing, not a Boxing kind of thing.

mathics/core/atoms.py
rocky: (I would not be surprised if there is a conversion or access function to go from an Expression symbol into a string.  If so, this was one of the many ways that the original code was hideously slow.)

rocky: BTW I noticed recently that  Expression.replace_vars is another place where strings are used instead of Symbol.

mathics/eval/makeboxes/makeboxes.py
rocky: Thinking about this more. For high-level form handling and boxing, which largely involve Mathics3 (simple and compound) expressions, these should be done in separate directories, e.g., `mathics.form` and `mathics.box`.

PR #1169, opened 2024-11-17
Redo TraceEvaluation using a decorator...

rocky:
This also allows the external debugger to do a much better job of this. See [this discussion](https://github.com/Mathics3/mathics-core/discussions/1168#discussioncomment-11281866) for an example of using.

### DIFF
--- a/mathics/builtin/trace.py
+++ b/mathics/builtin/trace.py
@@ -357,11 +357,11 @@ class TraceEvaluation(Builtin):
 
     The 'ShowTimeBySteps' option prints the elapsed time before an evaluation occurs.
 
-    >> TraceEvaluation[(x + 1)^2]
+    >> TraceEvaluation[(x + x)^2]
      | ...
      = ...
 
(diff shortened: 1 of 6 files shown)


mathics/builtin/trace.py
```diff
@@ -354,11 +355,13 @@ class TraceEvaluation(Builtin):
       <dd>Evaluate $expr$ and print each step of the evaluation.
     </dl>
 
-    >> TraceEvaluation[(x + x)^2]
+    The 'ShowTimeBySteps' option prints the elapsed time before an evaluation occurs.
+
+    >> TraceEvaluation[(x + 1)^2]
```
rocky, line 360: Note this gives less verbose trace output since we don't have to look up x again.

rocky: @mmatera would like to merge this in. The next step will be something similar where we decorate the rewrite-apply step so we can see what's going on there. It will be similar to this.

mmatera (approved):
LGTM

PR #1170, opened 2024-11-17
Respect "short" option in Symbol.get_name()

rocky:
And start a unit test for mathics.core.symbols. 

(Much more testing of that class would be nice, but that's left for some other time.)

### DIFF
--- a/mathics/core/symbols.py
+++ b/mathics/core/symbols.py
@@ -575,6 +575,16 @@ class Symbol(Atom, NumericOperators, EvalMixin):
         return definitions.get_attributes(self.name)
 
     def get_name(self, short=False) -> str:
+        """
+        Returns symbol's name field. If short=True
+        we strip off the context parts.
+
+        Note however that many places in the code we do not need the
+        "short" parameter becasue of Definitions.shorten_name() which
+        keeps track of the current $Context and $ContextPath to decide
+        whether the name of a symbol should or should not be
+        shortened.
+        """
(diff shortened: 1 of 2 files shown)


mathics/core/symbols.py
```diff
@@ -575,7 +575,7 @@ def get_attributes(self, definitions):
         return definitions.get_attributes(self.name)
 
     def get_name(self, short=False) -> str:
-        return self.name
+        return self.name.split("`")[-1] if short else self.name
```
mmatera, line 578: @rocky, notice that in the code, we use the method `Definitions.shorten_name()` which keeps track of the current `$Context` and `$ContextPath` to decide whether the name of a symbol should or should not be shortened.

rocky: Ok. I am not totally certain how to use this here or if this is the right thing, user interface-wise.

 It might be that in the debugger TraceActivate and TraceDebug do some sort of context-based searching. But on the other hand , it might also be that in a debugger ContextPath's change so you want the looseness. When the test is done might be in a totally different context from the context that set up the trace. 

I will add some comments and possibly TODO's in the debugger code so this idea not lost and made more in the forefront. 

But I also note that if there is a parameter named "short=False", it is wrong to have it not work.

mmatera: > 
> But I also note that if there is a parameter named "short=False" offered, it is wrong to have it not work.

Yep, I totally agree. Let's merge this when you feel it is ready.

PR #1171, opened 2024-11-17
Start Operators with no meaning section

rocky:
@mmatera Please look this over when you have a chance to see if this "no-meaning"  Star operator is done properly.

@axkr says that it would be needed for Rubi.

### DIFF
--- a/mathics/core/builtin.py
+++ b/mathics/core/builtin.py
@@ -1133,7 +1133,8 @@ class PostfixOperator(UnaryOperator):
 
 class BinaryOperator(Operator):
     """
-    Class for Builtin Binary Operators, e.g. Plus (+)
+    Class for Builtin Infix Operators (which includes Binary
+    operators), e.g. Plus (+).
     """
 
     # Note: grouping must be Python string, not a Symbol.
--- a/mathics/core/builtin.py
+++ b/mathics/core/builtin.py
@@ -1136,6 +1136,7 @@ class BinaryOperator(Operator):
     Class for Builtin Binary Operators, e.g. Plus (+)
     """
 
+    # Note: grouping must be Python string, not a Symbol.
     grouping = "System`None"  # NonAssociative, None, Left, Right
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


mmatera:
Isn't this symbol already contained in the Mathics script tables?

rocky:
> Isn't this symbol already contained in the Mathics script tables?

Yes, `\[Start]` is already available as a Character Symbol.  However to use it as an Infix binary operator, right now we need some code. 

One day after the operator table in MathicsScanner is completed, we might be able to use the information there and create classes from generic operators. Unfortunately, we aren't close to having that done.

rocky:
I've been thinking about this some more, and there is probably a way to get closer by having a generic "no-meaning" class, of which "Star" is an instance.  And with this, we can probably plug in more of what we currently have on the MathicsScanner operator YML side.

mmatera:
It would be awesome to be able to load all the "operators without a meaning" from mathics-scanner. We could have a "mother class" as we have for NamedColors, and then fill it with the entries inside of Mathics scanner tables that are not associated to already loaded symbols.

rocky:
@aravindh-krishnamoorthy I you need the Star implemented as an operator for Rubi, this branch can be used.  However, I will be reworking the code to be more generic and hook into a list of operators that have no meaning maintained in MathicsScanner. However, that may take a bit more time. I'll probably keep this branch as is in the interim.

mathics/core/builtin.py
```diff
@@ -1133,9 +1133,11 @@ def __init__(self, *args, **kwargs):
 
 class BinaryOperator(Operator):
     """
-    Class for Builtin Binary Operators, e.g. Plus (+)
+    Class for Builtin Infix Operators (which includes Binary
+    operators), e.g. Plus (+).
     """
 
+    # Note: grouping must be Python string, not a Symbol.
```
mmatera, line 1140: It would be OK if in the class definition, we use a Python string, if we convert it later into a Symbol. So, it would be a problem of `contribute`

mmatera:
I think we can merge this as it is, adding a comment at the beginning of `no_meanin.py` explaining how this should grow.

mathics/core/builtin.py
mmatera: I checked it out, and indeed `grouping` is used to build a Python `str`, that is parsed as a pattern object for defining formatting and boxing rules. So probably it is OK to keep is as a str.

PR #1175, opened 2024-11-19
Rename BinaryOperator to InfixOperator..

rocky:
Rename BinaryOperator to InfixOperator, since that is more precisely what it is.

Also, order functions inside builtin more alphabetically within each category of classes. Attach ABC to some abstract classes.


rocky:
Often when I want to start something new, like handling the kinds of "no-meaning"  operators, I find there is some code that is imprecise or weak in some way that impacts doing the new thing well.

One cool thing that we are in a position to do now (but isn't done in this PR) is to adjust all of the operator symbols in the operator classes. The no-meaning-operators-from-JSON branch does this just. However, here I think it best to complete this, _then_ the "no-meaning" branch, and then come back and do the rest of the operators.

mmatera:
LGTM

rocky:
Thanks - onward and upward.

Issue #1173, opened 2024-11-19
Mathics crash running `MatchQ[0 + x, ( a_. + b_. ) ]`


rocky:
## Description

```
MatchQ[0 + x, ( a_. + b_. ) ]
```

produces a Python traceback.

## How to Reproduce

```
$ mathics
In[1]:=  MatchQ[0 + x, ( a_. + b_. ) ]
Traceback (most recent call last):
  File "/home/rocky/.pyenv/versions/3.12.7/bin/mathics", line 33, in <module>
    sys.exit(load_entry_point('Mathics3', 'console_scripts', 'mathics')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

## Expected behavior

No crash. For the example given, True is expected.

## Additional context

This was encountered in working on getting Rubi running. See also https://github.com/Mathics3/mathics-core/discussions/1168#discussioncomment-11298819

rocky:
When I just tried it now, I am not seeing a crash. There is probably something more to this that needs to be worked out. Perhaps it is the lack of wrapping `0 + x` in `Hold[]` and there is some particular value of x that causes a problem?

mmatera:
Notice that also the pattern must be wrapped in Hold[...]. Otherwise, the match fails at the level of the `Hold` head.

rocky:
Whether or not the pattern is wrapped in Hold[...], Mathics should not crash.

aravindh-krishnamoorthy:
In function:
https://github.com/Mathics3/mathics-core/blob/bf6a8c0153170b1350f7ed00792eaa4ac801813a/mathics/core/pattern.py#L668-L672
the pattern `a_. + b_.` causes `self.elements` to be:
```python
In[1]:= MatchQ[x, a_. + b_.]
> ~/git/mathics/mathics-core/mathics/core/pattern.py(760)match_expression_with_one_identity()
-> assert new_pattern is not None
(Pdb) new_pattern
(Pdb) self
<ExpressionPattern: System`Plus[System`Optional[System`Pattern[Global`a, System`Blank[]]], System`Optional[System`Pattern[Global`b, System`Blank[]]]]>
(Pdb) self.elements
[<mathics.builtin.patterns.defaults.Optional object at 0x7ff543d16190>, <mathics.builtin.patterns.defaults.Optional object at 0x7ff543e7ef50>]
(Pdb)
```
[...]

Issue #1172, opened 2024-11-19
 `MatchQ[1 + 2 x, (a_. + b_. x_Symbol)]` incorrectly failing 


rocky:
## Description
`MatchQ[1 + 2 x, (a_. + b_. x_Symbol)]` also returns `False` when it should return `True`. 

## How to Reproduce

```
$ mathics
In[1]:= MatchQ[1 + 2 x, (a_. + b_. x_Symbol)]
Out[1]= False (* Wrong *)

In[2]:= MatchQ[2 x, (a_. + b_. x_Symbol)]
Out[2]= True (* Correct *)
```

## Additional context

This comes from trying to get Rubi working. The function in that, which is failing is called `Default[]`

aravindh-krishnamoorthy:
The issue (for now) is mainly with matching patterns that have a dot like `a_. + b_. x_Symbol` above. Here, the dot means that the pattern matches not just an expression of type `a + b x` but also `1 + x`, `b x`, `x`, and so on. The function `Default[]` is a Wolfram Language function, which informs the pattern matcher on what the "default" values for `a` and `b` are, when omitted. So, `x` is internally matched to  `a_. + b_. x_Symbol` as `Plus[Default[Plus, 1], Times[Default[Times, 1], x]` where `Default[Plus, 1] = 0` is the default argument (additive identity) for `Plus`, and similarly `1` for `Times`.

I started very recently with Mathics, so there may be errors in the analysis below. 

From what I gather, the problem seems to be that when matching `1 + 2 x` to `a_. + b_. x_Symbol`, in line 465, https://github.com/Mathics3/mathics-core/blob/bf6a8c0153170b1350f7ed00792eaa4ac801813a/mathics/core/pattern.py#L462-L467 the exception `StopGenerator_ExpressionPattern_match` is raised WITHOUT a full match.

The exception is raised because the condition `candidates < match_count[0]` is met: https://github.com/Mathics3/mathics-core/blob/bf6a8c0153170b1350f7ed00792eaa4ac801813a/mathics/core/pattern.py#L830-L831

However, inspecting the pattern at this point reveals that:
```python
(Pdb) unmatched_elements
(<Integer: 1>, <Expression: <Symbol: System`Times>[<Integer: 2>, <Symbol: Global`x>]>)
(Pdb) element
[...]

PR #1176, opened 2024-11-19
1 / Fix issues with pattern matching for Rubi

aravindh-krishnamoorthy:
# Overview

In this PR, fixes for pattern matching, esp. those that concern ``Rubi` ``, are collected.

# Related issues
- #1172 
- #1173 

# Failing patterns
- Pattern `a_. + b_. x_` does not match `1 + 2 x`
    - Likely fixed by https://github.com/aravindh-krishnamoorthy/mathics-core/commit/85ad0f5ed9055bcc27dedde379809f8e22f9a4b8
- Pattern `x_.^m_.` does not match `x`
    - Likely fixed by https://github.com/aravindh-krishnamoorthy/mathics-core/commit/c1999ce8b160ac8e20fa0ec9878dbb20992dba98 + https://github.com/aravindh-krishnamoorthy/mathics-core/commit/e1dee54db77c38e3e8fdb7c42297985c4636d815
- Pattern `a_. + b_.` does not match `x`
    - Crash: `mathics/core/pattern.py` -> `assert new_pattern is not None`
    - Likely fixed by https://github.com/aravindh-krishnamoorthy/mathics-core/commit/0a89945523224615a7d7d6c27a402fdf7fdf0ade
- ~~Pattern `x_Integer + y_Integer z_.` for `1 + 2 x`~~ (Not directly relevant for Rubi)
    - ~See https://github.com/Mathics3/mathics-core/pull/1176#issuecomment-2492089298~
- ~Pattern with conditional within `With`: `g[x_] := With[{}, x /; False] /; True` matches for `g[2]`~
    - ~Must not match.~

(Fixes for items struck out above are moved to the next PR.)

# Final review

# Tests
- [X] Tests for https://github.com/aravindh-krishnamoorthy/mathics-core/commit/85ad0f5ed9055bcc27dedde379809f8e22f9a4b8
- [X] Tests for https://github.com/aravindh-krishnamoorthy/mathics-core/commit/c1999ce8b160ac8e20fa0ec9878dbb20992dba98 + https://github.com/aravindh-krishnamoorthy/mathics-core/commit/e1dee54db77c38e3e8fdb7c42297985c4636d815
- [X] Tests for https://github.com/aravindh-krishnamoorthy/mathics-core/commit/0a89945523224615a7d7d6c27a402fdf7fdf0ade

# Documentation

N/A


rocky:
For the kinds of problems this fixes, please let's add a test that gets run for this. This way we can ensure no regressions to the code.

Since `MatchQ[]` is currently in `mathics.core.testing_expressions` a test case could be added to `test/builtin/test_testing_expressions.py`  Or if you want a more isolated kind of unit test, a new file `test/core/test_patterns.py` would be created. Thanks.

aravindh-krishnamoorthy:
> For the kinds of problems this fixes, please let's add a test that gets run for this. This way we can ensure no regressions to the code.
> 

Fully agree! I will definitely add them before I finalise this PR.

aravindh-krishnamoorthy:
Pattern `x_Integer + y_Integer z_.` for `1 + 2 x` (Not directly relevant for Rubi)

Possibly due to this optimization? https://github.com/Mathics3/mathics-core/blob/0d8a8f6e7cd17348870e501f4ce99103772be5df/mathics/core/pattern.py#L783-L796

aravindh-krishnamoorthy:
> We need to get it right before we "optimize". So please fix. Thanks.

Sorry, @rocky, what do you mean? The mentioned optimization is not in the PR but in the `mathics-core` code.

rocky:
> > We need to get it right before we "optimize". So please fix. Thanks.
> 
> Sorry, @rocky, what do you mean? The mentioned optimization is not in the PR but in the `mathics-core` code.

I mean please put in a PR to fix the mathics-core code. Thanks.

mmatera:
> Pattern `x_Integer + y_Integer z_.` for `1 + 2 x` (Not directly relevant for Rubi)
> 
> Possibly due to this optimization?
> 
> https://github.com/Mathics3/mathics-core/blob/0d8a8f6e7cd17348870e501f4ce99103772be5df/mathics/core/pattern.py#L783-L796

If you want to deactivate the optimization, just set `leading_blanks=False` before the loop.

mmatera:
Up to here, the changes look reasonable, but it would be nice to see some tests that pass after these changes.

rocky:
> Up to here, the changes look reasonable, but it would be nice to see some tests that pass after these changes.

Let me go further and explain.  From my side, it is better to have many smaller complete fixes and PRs than wait until everything is done in several weeks/months and have one huge PR.  

It is also useful to add tests after each bug fix is done rather than view this as some sort of separate and independent tasks.

aravindh-krishnamoorthy:
Indeed, it makes sense to have smaller PRs. I will finalise these with tests soon.

aravindh-krishnamoorthy:
> > Pattern `x_Integer + y_Integer z_.` for `1 + 2 x` (Not directly relevant for Rubi)
> > Possibly due to this optimization?
> > https://github.com/Mathics3/mathics-core/blob/0d8a8f6e7cd17348870e501f4ce99103772be5df/mathics/core/pattern.py#L783-L796
> 
> If you want to deactivate the optimization, just set `leading_blanks=False` before the loop.

Perhaps the removal of optionals in this PR can be reverted after we undo the `leading_blanks` optimization. I'll double-check that in the next PR.

test/core/test_patterns.py
```diff
@@ -0,0 +1,51 @@
...
+import time
+from test.helper import check_evaluation, evaluate
+
+import pytest
+
+@pytest.mark.parametrize(
+    ("str_expr", "msgs", "str_expected", "fail_msg"),
+    [
+        # Two default arguments (linear)
+        ("MatchQ[1, a_.+b_.*x_]",None,"True",None),
```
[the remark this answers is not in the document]
aravindh-krishnamoorthy: Agree! This makes sense.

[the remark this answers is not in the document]
aravindh-krishnamoorthy: I will add them as soon as I have all your and rocky's comments.

mmatera (approved):
LGTM

test/core/test_patterns.py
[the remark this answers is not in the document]
rocky: I'd like to see more/some unit tests for `mathics.core.pattern` .  Other than that, I don't have any further comments.  

When everything is in place, I'd like to look over.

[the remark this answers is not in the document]
aravindh-krishnamoorthy: Done!

test/core/test_patterns.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 """
-Unit tests for mathics pattern matching
+Unit tests for mathics.core.pattern
```
rocky, line 3: Here is the part that feels a little weird to me. We are testing a lower-level function using higher-level tests.  I am a little bit sensitive to this kind of thing, because this less-good practice is pervasive in this code.

At one time, we didn't have pytests. There were doctests that took the place of pytests. 

The MatchQ tests to me should be put in `test.buitin.testing_expressions.expression_tests` and the rule tests put wherever rules are located. 

Tests here could (and should be) to build a ExpressionPattern and run `match` on it.

test/core/test_patterns.py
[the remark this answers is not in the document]
mmatera: > > I'd like to see more/some unit tests for `mathics.core.pattern` as its own unit tests, e.g. (not via MatchQ). Other than that, I don't have any further comments.
> > When everything is in place, I'd like to look over.
> 
> I guess it's always gonna be via `MatchQ` or rules, as mmatera suggested. Alternate ideas are welcome!

At some point, we should try to add tests at a lower level (using `mathics.eval.patterns.match` or directly testing the subclasses of `mathics.core.pattern.Pattern`) that avoids passing through the evaluation loop.  But for now I think this is a good starting point.

test/core/test_patterns.py
mmatera: > Here is the part that feels a little weird to me. We are testing a lower-level function using higher-level tests. I am a little bit sensitive to this kind of thing, because this less-good practice is pervasive in this code.
> 
> At one time, we didn't have pytests. There were doctests that took the place of pytests. It took us a while to get unburied from that practice.
> 
> The MatchQ tests to me should be put in `test.buitin.testing_expressions.expression_tests` and the rule tests put wherever rules are located.
> 
> Tests here could (and should be) to build a ExpressionPattern and run `match` on it.

Yes, but in any case, at some point we can rewrite these tests as lower-level tests. I do not think that these tests actually test the behavior of `MatchQ` or `ReplaceAll`, but the low-level code. I can do the translation later, when I find some time for it.

rocky: Tests in `test.builtin` are testing Mathics3 Builtin functions.  Of course, any time one tests a higher-level function one  is testing all sorts of lower-level functions. So all of these tests are valid, it is just organized in my opinion in the wrong place.

Tests of the ExpressionPattern match function are good because they isolate all of the other kinds of stuff and we pinpoint the problem better. A MatchQ test could fail because of something totally unrelated to matching.

Another thing that these smaller unit tests do (which shouldn't be needed but in fact is used way too much) is to show developers how one can work with, test, and debug the smaller units. Ideally, if our APIs and modularity are done well,  the code needed to do this is not that much. Personally, I have found ways to improve modularity when i have been forced to these the smaller parts in isolation.

rocky: > Yes, but in any case, at some point we can rewrite these tests as lower-level tests. I do not think that these tests actually test the behavior of `MatchQ` or `ReplaceAll`, but the low-level code. I can do the translation later, when I find some time for it.

My experience is that the rewrite tends to happen at a smaller rate than creating more code that is somewhat ill-conceived.  In fact, I am in the middle of a larger rewrite, for operators right now.

If you want to do the minimal thing, then move the MatchQ code to where MatchQ should be tested and the rule test to where rules are tested. At least, this honestly reflects what the test is doing. If later you want to rewrite, then at that time you can.

mmatera: And yes, in that case, the test should be under `test/eval/`

rocky: > When we test the evaluation of an expression like `MatchQ[1+2 a, x_.+y_.*z_Symbol]`, many processes are involved: the main evaluation loop, the definitions of `Plus` and `Times` and their internal implementation, the pattern matching done in the evaluation loop, and eventually, the implementation of `matchQ` and the pattern matching mechanisms. If we want to know if the pattern matching works as it should, it would be better to use `mathics.eval.patterns.match`, which does not try to evaluate its arguments, but just implements the pattern matching. I think that this is the right level of testing for these tests.
> 

Ok.  If testing `Matcher.match()` in   `mathics.eval.patterns.match` is a good place to test against, then let's do that and that would be done in a `test.eval.pattern` module. 

I am just saying that if we're going to test temporarily using MatchQ, then `testing.core.patterns` is not the right testr module for this. 

As written, the test is extending MatchQ tests, even if the motivation for doing this is to verify what's going on at a lower level.  I leave to @aravindh-krishnamoorthy whether to rewrite using `Matcher.match()` inside `test.eval.pattern` or move the existing tests to `test.builtin.testing_expressions.expression_tests` and wherever Rules are testing.

> Lower levels involve testing whether the functions and methods used to evaluate matching do what they are supposed to. But at this point, there are functions for which I am still unable to say exactly what the expected behavior is, even after spending quite a long time trying to understand and document them.

Then i think it was a mistake to "optimize" what is not well understood nor,  as we have seen, not well covered in tests.

rocky: Thanks for your understanding and patience. Sure, it if turns out to be too cumbersome, do as you say.

In fact, the messes in this code have been so great that often I find that I can't fix up a single mess in one go, but have to break things up into little pieces.  As you are learning, Pattern matching is one of those big messes. 

Down the line, we will use trie based machining akin to expreduce (and may try to copy its algorithm).  In preparation for that though, having more and better pattern-matching tests in advance will help greatly.

rocky:
Great - thanks!

PR #1177, opened 2024-11-19
use stopit to implement TimeConstrained

mmatera:
If this works, maybe we can simplify the implementation of `TimeConstrained


rocky:
LGTM - removing code and replacing with an existing module is always a good thing.

mmatera:
Great. Also, if eventually `stopit` starts working with `Emscripten`, then the whole problem would be solved.

PR #1179, opened 2024-11-21
Try Rubi as a git submodule

rocky:
Thoughts about doing this? 
@aravindh-krishnamoorthy  ?


rocky:
> @rocky, I think this would be OK once mathics-ruby loads and works properly. By now, I would do the opposite: add a github-action test on mathics-ruby, and check against the branches of mathics-core that improve the required support.

Yes, that makes sense. But here we need https://github.com/Mathics3/mathics-core/pull/1176 to finish. This corrects some bugs in handling the "optional" operator in pattern matching.  When you get a chance please look that over since you know far more about Mathics3's pattern-matching code (and WMA) than I.  Right now, these changes are pretty small.

aravindh-krishnamoorthy:
Indeed, this sounds good once basic Rubi starts working...

rocky:
Now that Rubi has merged, this is your last chance to weigh in before we merge this.

Of course, in order to use Rubi, you'll have to use a branch to make TimeConstrained work.  @aravindh-krishnamoorthy started a discussion thread for that.

mmatera (approved):
Ok, let's try this. LGTM

PR #1174, opened 2024-11-19
First-cut auto-generated no-meaning operators

rocky:

PR #1180, opened 2024-11-23
Operator info from Operators.YML file (in MathicsScanner)

rocky:
This is an extension and follow-on to #1174 .

More of the data encoded in `mathics.core.parsers.operators` now comes from JSON. And all of the operators without builtin-meanings are included. 

There will be some more work to fix up the MANIFESTs (@mmatera are you up for this?  And some additional changes would be nice in mathicsscanner to include all of the operators without built-in meanings. Right now, these are not pulled from JSON and are incomplete.


rocky:
Since we test _all_  new Builtin functions without pre-defined meaning including the operator form of that, these have to exist in the scanner.  Some of these unicode operator symbols are currently missing. 

I will be making another round in MathicsScanner to pick up all of these such operators from the YML soon.

mmatera:
What about using a blacklist in mathics-corr, to avoid loading the incomplete symbols?

rocky:
> What about using a blacklist in mathics-corr, to avoid loading the incomplete symbols?

I'd rather do a relatively small amount of work to finish that aspect than add some code that will be ripped out soon. There will be many other things that could be improved that can wait and be done.

mmatera:
Ok,when I get some time I'll do it.

mathics/builtin/no_meaning.py
```diff
@@ -1,42 +1,61 @@
...
+    NoMeaningPrefixOperator,
+)
+from mathics.core.parser.operators import OPERATOR_DATA
+
+# Generate no-meaning Mathics3 Builtin class from the operator name,
+# affix, and Operator Unicode values found in OPERATOR_DATA.  This
+# data ultimately comes from a YAML file in the MathicsScanner project
+# which is processed into a JSON file.
+
+for affix, format_fn, operator_base_class in (
```
mmatera, line 29: This loop should be inside a `def init()` function, called at the end of the module.
Otherwise, the global variable `generated_operator_class` is seen by `admin-tools/build_and_check_manifest.py` a duplication of one of the classes defined here (`Square`)
and raise an error.

After this change, just run `admin-tools/build_and_check_manifest.py` and the SYMBOL_MANIFEST.txt must be automatically updated.

Makefile
```diff
@@ -142,10 +142,9 @@ latexdoc texdoc doc:
 	(cd mathics/doc/latex && $(MAKE) doc)
 
 #: Build JSON ASCII to unicode opcode table and operator table
-mathics/data/op-tables.json mathics/data/operators.json:
+mathics/data/operator-tables.json mathics/data/op-tables.json mathics/data/operators.json:
```
mmatera, line 145: operators.json is not in mathics-core/mathics/data but in mathics_scanner/data/

SYMBOLS_MANIFEST.txt
```diff
@@ -134,8 +134,10 @@ System`Automatic
 System`Axes
 System`Axis
 System`Background
+System`Backslash
```
mmatera, line 137: @rocky, the changes in the SYMBOLS_MANIFEST.txt are already done...

Makefile
mmatera: (I should have put a line in `mathics_scanner.generate.build_operator_tables` that prints where `operators.json` is stored...

SYMBOLS_MANIFEST.txt
rocky: Something I ran must have changed this. I definitely have not been changing this by hand. 

At any rate, just fix whatever problem you encounter.

mmatera: It is fixed automatically once you run the test locally, and the test passes. It does not seem to fail here. What is failing is the localization of the operator.json file, that is not generated in the place where mathics_scanner.tokenizer looks for it.

rocky:
I had considered this a dead branch that should be closed and deleted. At some point this branch was previously merged in. Locally I forget to delete it and committed on top things that I think were also committed to a different branch.

mmatera:
Ok, then we need to adjust the mathics tests in mathics scanner

PR #1181, opened 2024-11-24
update SYMBOLS_MANIFEST

mmatera:
and put initialization code in `mathics.builtin.no_meaning` inside an initialization function.


mathics/builtin/no_meaning.py
```diff
@@ -26,36 +26,41 @@
...
-        if affix == "infix":
-            # FIXME get precedence from JSON. Value 100 is a hack for now.
-            mathics.core.parser.operators.flat_binary_ops[operator_name] = 100
-
-        # Put the newly-created Builtin class inside this module.
-        setattr(modules[__name__], operator_name, generated_operator_class)
+            )
+            if affix == "infix":
+                # FIXME get precedence from JSON. Value 100 is a hack for now.
+                mathics.core.parser.operators.flat_binary_ops[operator_name] = 100
```
rocky, line 60: I forgot that I also need to fix this. Instead of 100, we should use the value in JSON. At the time I wrote that I had not had the precedence values in the JSON yet.

rocky: And I also need to handle affix's "prefix" and "postfix".

rocky: And finally, on my todo is to turn "no meaning" into a guide section and have modules underneath for each of the affix's. When you look at this section in Django (or the PDF) it is very long.

rocky:
LGTM - thanks.

mathics/builtin/no_meaning.py
mmatera: @rocky, If you want, you can merge this, and then `SYMBOL_MANIFEST.txt` is going to be updated. Or once you fix this, just run the script to update the manifest.

PR #1182, opened 2024-11-25
Operator info from json mm

mmatera:
just to check a theory


mmatera:
@rocky, once I force to create `operators.json` in the `mathics_scanner/data` folder, the only thing that is failing is the `test_gs.py`. However, it is not very verbose about what is going wrong...

rocky:
> @rocky, once I force to create `operators.json` in the `mathics_scanner/data` folder, the only thing that is failing is the `test_gs.py`. However, it is not very verbose about what is going wrong...

Somewhere I have the book from which this is based on. And I now have better tools for debugging. I'll take a look at  at some point.

rocky:
> @rocky, by running `<< GS1.m` I get
> 
> ```
> {Test[19]:=
>         , x + 2 y z /. {x → z, y → x, z → y}, 
>         expected, z + 2 x y, 
>         eval'd expected, 2 x y + z, 
>         eval'd actual  , 2 x y + z, 
>         right?, True}
> 
> Syntax::sntxi: Incomplete expression; more input is needed (line 1866 of "./GS1.m").
> ```
> 
> But the same happens with previous code: I check against the Mathics release 6.0.0, and I get the same error. Could be possible that with the last changes one of the tests reach the last line `Exit[Min[totalWrong, 255]]` and then just now we see that the test does not pass?

Apparently, the scanning bug has been around for a while and something has been masking seeing it. I will put in a issue report in mathics scanner so we can fix this later. For now, I've removed the "\" in the comments of GS1.m.

rocky:
> @rocky, by running `<< GS1.m` I get
> 
> ```
> {Test[19]:=
>         , x + 2 y z /. {x → z, y → x, z → y}, 
>         expected, z + 2 x y, 
>         eval'd expected, 2 x y + z, 
>         eval'd actual  , 2 x y + z, 
>         right?, True}
> 
> Syntax::sntxi: Incomplete expression; more input is needed (line 1866 of "./GS1.m").
> ```
> 
> But the same happens with previous code: I check against the Mathics release 6.0.0, and I get the same error. Could be possible that with the last changes one of the tests reach the last line `Exit[Min[totalWrong, 255]]` and then just now we see that the test does not pass?

BTW using `Get["GS1.m", Trace->True]  can show you closer to where the problem is. This is what I used.

PR #1183, opened 2024-11-25
Implement `Breakpoint[]`

aravindh-krishnamoorthy:
This PR implements the `Breakpoint[]` functionality discussed in issue #1172.

With this implementation, it is possible to invoke a Python breakpoint from the mathics command line for debugging. Alternative debuggers can be loaded by setting the relevant Python environment variable `PYTHONBREAKPOINT`. 

Please feel free to edit the description text.

### DIFF
--- a/mathics/builtin/system.py
+++ b/mathics/builtin/system.py
@@ -882,12 +882,12 @@ class Breakpoint(Builtin):
            variable `PYTHONBREAKPOINT` can be utilized.
     </dl>
 
+    >> (* Test with a dummy breakpoint. *)
+    >> SetEnvironment["PYTHONBREAKPOINT" -> "mathics.docpipeline.dummy_breakpoint"]
+    Out[1]= SetEnvironment[PYTHONBREAKPOINT → mathics.docpipeline.dummy_breakpoint]
     >> Breakpoint[]
-    --Return--
-    > mathics/builtin/system.py(891)eval()->None
-    -> breakpoint()
-    (Pdb) c
-    Out[1]= Breakpoint[]
+    Dummy breakpoint() reached! Returning.
+    Out[2]= Breakpoint[]
     """
 
     summary_text = "invoke a Python breakpoint()"
--- a/mathics/builtin/system.py
+++ b/mathics/builtin/system.py
@@ -885,8 +885,7 @@ class Breakpoint(Builtin):
     </dl>
 
     >> (* Test with a disabled breakpoint. *)
-    >> In[1]:= SetEnvironment["PYTHONBREAKPOINT" -> "mathics.disabled_breakpoint"]
-    Out[1]= SetEnvironment[PYTHONBREAKPOINT → mathics.disabled_breakpoint]
+    >> In[1]:= SetEnvironment["PYTHONBREAKPOINT" -> "mathics.disabled_breakpoint"];
     >> Breakpoint[]
     Hit disabled breakpoint.
     Out[2]= Breakpoint[]
--- a/mathics/builtin/system.py
+++ b/mathics/builtin/system.py
@@ -877,7 +877,7 @@ class Breakpoint(Builtin):
     """
     <dl>
       <dt>'Breakpoint[]'
-      <dd> Invoke a Python breakpoint. By default, the python debugger \
+      <dd> Invoke a Python breakpoint. By default, the Python debugger \
            (pdb) is loaded. For loading other debuggers, the Python environment \
            variable `PYTHONBREAKPOINT` can be utilized.
     </dl>
@@ -890,7 +890,7 @@ class Breakpoint(Builtin):
     Out[1]= Breakpoint[]
     """
 
-    summary_text = "invoke a Python breakpoint"
+    summary_text = "invoke a Python breakpoint()"
 
     def eval(self, evaluation: Evaluation):
         "Breakpoint[]"
--- a/mathics/builtin/system.py
+++ b/mathics/builtin/system.py
@@ -885,7 +885,7 @@ class Breakpoint(Builtin):
     </dl>
 
     >> (* Test with a disabled breakpoint. *)
-    >> In[1]:= SetEnvironment["PYTHONBREAKPOINT" -> "mathics.disabled_breakpoint"];
+    >> SetEnvironment["PYTHONBREAKPOINT" -> "mathics.disabled_breakpoint"];
     >> Breakpoint[]
     Hit disabled breakpoint.
     Out[2]= Breakpoint[]
--- a/mathics/builtin/system.py
+++ b/mathics/builtin/system.py
@@ -875,6 +875,8 @@ class Share(Builtin):
 
 class Breakpoint(Builtin):
     """
+    ## <url>:trace native symbol:</url>
+
     <dl>
       <dt>'Breakpoint[]'
       <dd> Invoke a Python breakpoint. By default, the Python debugger \
--- a/mathics/builtin/system.py
+++ b/mathics/builtin/system.py
@@ -888,7 +888,7 @@ class Breakpoint(Builtin):
     >> SetEnvironment["PYTHONBREAKPOINT" -> "mathics.disabled_breakpoint"];
     >> Breakpoint[]
     = Hit disabled breakpoint.
-    Out[2]= Breakpoint[]
+    = Breakpoint[]
     """
 
     summary_text = "invoke a Python breakpoint()"
--- a/mathics/builtin/system.py
+++ b/mathics/builtin/system.py
@@ -887,7 +887,7 @@ class Breakpoint(Builtin):
     >> (* Test with a disabled breakpoint. *)
     >> SetEnvironment["PYTHONBREAKPOINT" -> "mathics.disabled_breakpoint"];
     >> Breakpoint[]
-    Hit disabled breakpoint.
+    = Hit disabled breakpoint.
     Out[2]= Breakpoint[]
     """
 
--- a/mathics/builtin/system.py
+++ b/mathics/builtin/system.py
@@ -884,11 +884,11 @@ class Breakpoint(Builtin):
            variable `PYTHONBREAKPOINT` can be utilized.
     </dl>
 
-    >> (* Test with a dummy breakpoint. *)
-    >> SetEnvironment["PYTHONBREAKPOINT" -> "mathics.docpipeline.dummy_breakpoint"]
-    Out[1]= SetEnvironment[PYTHONBREAKPOINT → mathics.docpipeline.dummy_breakpoint]
+    >> (* Test with a disabled breakpoint. *)
+    >> In[1]:= SetEnvironment["PYTHONBREAKPOINT" -> "mathics.disabled_breakpoint"]
+    Out[1]= SetEnvironment[PYTHONBREAKPOINT → mathics.disabled_breakpoint]
     >> Breakpoint[]
-    Dummy breakpoint() reached! Returning.
+    Hit disabled breakpoint.
     Out[2]= Breakpoint[]
     """
 
--- a/mathics/builtin/system.py
+++ b/mathics/builtin/system.py
@@ -871,3 +871,28 @@ class Share(Builtin):
         else:
             gc.collect()
             return Integer0
+
+
+class Breakpoint(Builtin):
+    """
+    <dl>
+      <dt>'Breakpoint[]'
+      <dd> Invoke a Python breakpoint. By default, the python debugger \
+           (pdb) is loaded. For loading other debuggers, the Python environment \
+           variable `PYTHONBREAKPOINT` can be utilized.
+    </dl>
+
+    >> Breakpoint[]
+    --Return--
+    > mathics/builtin/system.py(891)eval()->None
+    -> breakpoint()
+    (Pdb) c
+    Out[1]= Breakpoint[]
+    """
+
+    summary_text = "invoke a Python breakpoint"
+
+    def eval(self, evaluation: Evaluation):
+        "Breakpoint[]"
+
+        breakpoint()
--- a/mathics/__init__.py
+++ b/mathics/__init__.py
@@ -68,3 +68,8 @@ This program comes with ABSOLUTELY NO WARRANTY.
 This is free software, and you are welcome to redistribute it
 under certain conditions.
 See the documentation for the full license."""
+
+
+# Disabled breakpoint
+def disabled_breakpoint():
+    print("Hit disabled breakpoint.")
--- a/mathics/docpipeline.py
+++ b/mathics/docpipeline.py
@@ -60,11 +60,6 @@ TestParameters = namedtuple(
 )
 
 
-# Dummy breakpoint for doctests
-def dummy_breakpoint():
-    print("Dummy breakpoint() reached! Returning.")
-
-
 class TestOutput(Output):
     """Output class for tests"""
 
--- a/mathics/docpipeline.py
+++ b/mathics/docpipeline.py
@@ -60,6 +60,11 @@ TestParameters = namedtuple(
 )
 
 
+# Dummy breakpoint for doctests
+def dummy_breakpoint():
+    print("Dummy breakpoint() reached! Returning.")
+
+
 class TestOutput(Output):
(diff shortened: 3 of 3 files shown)


rocky:
@aravindh-krishnamoorthy I do have some small changes to suggest, like capitalization of the word Python and adding parenthesis around `breakpoint()`.  

I have just sent an invite as a collaborator to this project. If you would put in this PR as a branch of this repo, I think I can more easily edit it. Thanks, and sorry for having you do more work on this.

rocky:
Also, about the CI breakage. I think this will be fixed soon. Sorry again for this.

rocky:
Looks like this needs a merge or rebase from master. Unfortunately we have an API breaking change that was just made.

mathics/builtin/system.py
```diff
@@ -871,3 +871,28 @@ def eval_with_symbol(self, symbol, evaluation: Evaluation) -> Integer:
...
+
+    >> Breakpoint[]
+    --Return--
+    > mathics/builtin/system.py(891)eval()->None
+    -> breakpoint()
+    (Pdb) c
+    Out[1]= Breakpoint[]
+    """
+
+    summary_text = "invoke a Python breakpoint"
```
rocky, line 893: I was thinking `breakpoint()`, but I suppose without the parenthesis is also okay.

mathics/builtin/system.py
```diff
@@ -871,3 +871,28 @@ def eval_with_symbol(self, symbol, evaluation: Evaluation) -> Integer:
...
             gc.collect()
             return Integer0
+
+
+class Breakpoint(Builtin):
+    """
+    <dl>
+      <dt>'Breakpoint[]'
+      <dd> Invoke a Python breakpoint. By default, the python debugger \
+           (pdb) is loaded. For loading other debuggers, the Python environment \
```
rocky, line 881: the python debugger -> the Python debugger

rocky:
I think what we'll have to do here to give a `Breakpoint[]` example is set up some Python code that does not actually stop. For pytest, the mock library can be used.

mathics/builtin/system.py
aravindh-krishnamoorthy:
```suggestion
      <dd> Invoke a Python breakpoint. By default, the Python debugger \
           (pdb) is loaded. For loading other debuggers, the Python environment \
```

mathics/builtin/system.py
aravindh-krishnamoorthy, line 893:
```suggestion
    summary_text = "invoke a Python breakpoint()"
```

mmatera:
The problem seems to be that the docpipeline does not handle this kind of exception properly. We would need to modify it to support a test like this.

mathics/builtin/system.py
```diff
@@ -871,3 +871,28 @@ def eval_with_symbol(self, symbol, evaluation: Evaluation) -> Integer:
...
+class Breakpoint(Builtin):
+    """
+    <dl>
+      <dt>'Breakpoint[]'
+      <dd> Invoke a Python breakpoint. By default, the Python debugger \
+           (pdb) is loaded. For loading other debuggers, the Python environment \
+           variable `PYTHONBREAKPOINT` can be utilized.
+    </dl>
+
+    >> Breakpoint[]
```
mmatera, line 885: Add a pytest inside, which captures the interruption.

rocky: To get doctests to work here is a pretty simple solution....

Add:

```
def disabled_breakpoint():
    return
```

somewhere like in `mathics.main` or `mathics.docpipeline` and then run using `PYTHONBREAKPOINT="mathics.main.disabled_breakpoint" ...`

aravindh-krishnamoorthy: This might confuse the users of the function?

rocky: > This might confuse the users of the function?

I am not sure what you mean. If you mean that someone sets PYTHONBREAKPOINT to this function that is intended to be used in testing, then things can be made more explicit by having `disabled_breakpoint()` print a message saying that a breakpoint was called but it is ignoring it. For example: 

```
def disabled_breakpoint():
   print("PYTHONBREAKPOINT is set to disabled_breakpoint. Ignoring breakpoint.")
```

aravindh-krishnamoorthy:
Unfortunately, I'm not sure what URL to provide here. Comments and suggestions on the URL and the dummy breakpoint are welcome!

mathics/builtin/system.py
aravindh-krishnamoorthy: Using `PYTHONBREAKPOINT` is a great idea!

mathics/builtin/system.py
```diff
@@ -882,12 +882,12 @@ class Breakpoint(Builtin):
            variable `PYTHONBREAKPOINT` can be utilized.
     </dl>
 
+    >> (* Test with a dummy breakpoint. *)
```
rocky, line 885: I've been thinking about this, and even outside of testing it might be useful to have a _disabled_  as opposed to "dummy" breakpoint. 

Should we decide this,  I guess this should not be put in docpipeline, since it would no longer be about documentation. 

I am curious what other folks think and suggest.

aravindh-krishnamoorthy: From my point of view, if a `disabled_breakpoint` function outside of doctests would be useful, I'll gladly change the current implementation and move the function to `main`.

rocky:
> Unfortunately, I'm not sure what URL to provide here. Comments and suggestions on the URL and the dummy breakpoint are welcome!

Use: 
```
    ## <url>:trace native symbol:</url>
```

For an example, see https://github.com/Mathics3/mathics-core/blob/master/mathics/builtin/trace.py#L107

mathics/builtin/system.py
aravindh-krishnamoorthy: Waiting on a final decision on this.

rocky: > From my point of view, if a `disabled_breakpoint` function outside of doctests would be useful, I'll gladly change the current implementation and move the function to `main`.

If we go that route (and I am for it), then it might be slightly simpler to put this inside `mathics/__init__.py`. So it can be referred to as `mathics.disabled_breakpoint`. 

Normally putting functions inside `__init__.py` is considered a bad thing, especially when there is a lot of code and initialization code. This is because it the order in which code runs can be indeterminate or hard to figure out  and may change as imports change. (In my view this module system for Python is a misfeature that has persisted for too long). 

However in this particular situation where we have a short function  and that does not do any initialization, this kind of things is okay since it simplifies the user interface.

rocky: > Waiting on a final decision on this.

@mmatera ?

mmatera: If it is possible to handle it by defining a breakpoint handler, it would be great! About the proper place to leave it, I am OK about put it in `mathics__init__.py`, or whatever other place that makes sense to you.

mmatera:
> > Unfortunately, I'm not sure what URL to provide here. Comments and suggestions on the URL and the dummy breakpoint are welcome!
> 
> Use:
> 
> ```
>     ## <url>:trace native symbol:</url>
> ```
> 
> For an example, see https://github.com/Mathics3/mathics-core/blob/master/mathics/builtin/trace.py#L107

Maybe it could work a link to the debuger documentation. For example, https://python2-trepan.readthedocs.io/en/latest/commands/breakpoints/break.html

mathics/builtin/system.py
```diff
@@ -871,3 +871,30 @@ def eval_with_symbol(self, symbol, evaluation: Evaluation) -> Integer:
...
+
+    <dl>
+      <dt>'Breakpoint[]'
+      <dd> Invoke a Python breakpoint. By default, the Python debugger \
+           (pdb) is loaded. For loading other debuggers, the Python environment \
+           variable `PYTHONBREAKPOINT` can be utilized.
+    </dl>
+
+    >> (* Test with a disabled breakpoint. *)
+    >> In[1]:= SetEnvironment["PYTHONBREAKPOINT" -> "mathics.disabled_breakpoint"]
```
mmatera, line 888:
```
    >> SetEnvironment["PYTHONBREAKPOINT" -> "mathics.disabled_breakpoint"];
    >> Breakpoint[]
       = ...

```

aravindh-krishnamoorthy: @mmatera as I suggested to rocky, if you don't mind, you could use the +/- option in the review to suggest the changes, which I can apply directly. That way, your contribution is also acknowledged, and it's easier for me as well...

aravindh-krishnamoorthy: Regarding the URL, I will leave it to you both to decide the best course and advise me on the change.

aravindh-krishnamoorthy, line 888:
```suggestion
    >> In[1]:= SetEnvironment["PYTHONBREAKPOINT" -> "mathics.disabled_breakpoint"];
```

mathics/builtin/system.py
```diff
@@ -871,3 +871,30 @@ def eval_with_symbol(self, symbol, evaluation: Evaluation) -> Integer:
...
+    Out[1]= SetEnvironment[PYTHONBREAKPOINT → mathics.disabled_breakpoint]
```
aravindh-krishnamoorthy, line 889:
```suggestion
```

mathics/builtin/system.py
```diff
@@ -871,3 +871,29 @@ def eval_with_symbol(self, symbol, evaluation: Evaluation) -> Integer:
...
+
+    <dl>
+      <dt>'Breakpoint[]'
+      <dd> Invoke a Python breakpoint. By default, the Python debugger \
+           (pdb) is loaded. For loading other debuggers, the Python environment \
+           variable `PYTHONBREAKPOINT` can be utilized.
+    </dl>
+
+    >> (* Test with a disabled breakpoint. *)
+    >> In[1]:= SetEnvironment["PYTHONBREAKPOINT" -> "mathics.disabled_breakpoint"];
```
aravindh-krishnamoorthy, line 888:
```suggestion
    >> SetEnvironment["PYTHONBREAKPOINT" -> "mathics.disabled_breakpoint"];
```

rocky:
> > > Unfortunately, I'm not sure what URL to provide here. Comments and suggestions on the URL and the dummy breakpoint are welcome!
> > 
> > 
> > Use:
> > ```
> >     ## <url>:trace native symbol:</url>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > For an example, see https://github.com/Mathics3/mathics-core/blob/master/mathics/builtin/trace.py#L107
> 
> Maybe it could work a link to the debuger documentation. For example, https://python2-trepan.readthedocs.io/en/latest/commands/breakpoints/break.html

If you want to link to a description of how Python's  breakpoint() works use something like https://docs.python.org/3/library/functions.html#breakpoint 

While I am flattered to have a link to stuff I wrote, that link is to the Python2 (not current Python3) code and that specific text isn't going to be of much help here since this doesn't get you into the Python2 debugger. 

Right now I wouldn't fret over the specific details of a link or no link. Debugging is going to change and (I hope) get greatly more useful. This is the _one_ area I can see in the short term where we can do much much better than WMA.  In fact, I would argue that just having `Breakpoint[]` in there, as is, is already an improvement!

rocky:
@mmatera The intermittent TimeConstraint failures are getting annoying. Suggestions here? Mine is to mark it skipped until it is no longer flaky.

mathics/builtin/system.py
```diff
@@ -871,3 +871,29 @@ def eval_with_symbol(self, symbol, evaluation: Evaluation) -> Integer:
...
+      <dt>'Breakpoint[]'
+      <dd> Invoke a Python breakpoint. By default, the Python debugger \
+           (pdb) is loaded. For loading other debuggers, the Python environment \
+           variable `PYTHONBREAKPOINT` can be utilized.
+    </dl>
+
+    >> (* Test with a disabled breakpoint. *)
+    >> SetEnvironment["PYTHONBREAKPOINT" -> "mathics.disabled_breakpoint"];
+    >> Breakpoint[]
+    Hit disabled breakpoint.
```
aravindh-krishnamoorthy, line 890: Do you mean this?

```suggestion
    ...
```

As I suggested earlier, it's much simpler if you could please use the +/- button and the `suggestion` tag to indicate the exact change.

rocky: I tried adding the +/- . Maybe the suggestion tag I missed. 

Here is another suggestion that I made a while ago.  Make this a branch of this repository. Then when folks want to make suggestions we can just make a branch off of your branch it has all of the changes and you can review and or commit on top.  This is the way we've been working on other PRs. 

```
-   Hit disabled breakpoint.
+   ...
...

No, neither is correct.  _Either_ would be okay of you add `= ` beforehand.  For example: 

```
= Hit disabled breakpoint.
```

See [line 484](https://github.com/Mathics3/mathics- lcore/blob/fb6fadaf486c83a7e73403f94d73ec7bc208fb76/mathics/builtin/system.py#L484) and [line 362](https://github.com/Mathics3/mathics-core/blob/fb6fadaf486c83a7e73403f94d73ec7bc208fb76/mathics/builtin/system.py#L362) if this file for examples of each kind of method.  

Also look at the previous read-the-docs link. If there is something unclear about that, then ask we can make this more clear.

Comment (aravindh-krishnamoorthy) [reply mathics/builtin/system.py:890]: I see. I've edited my comment above and now it should likely reflect the expected changes. I looked at the read-the-docs link, and also the previous examples. Unfortunately, I missed this (some would say rather obvious) detail. Perhaps it would have helped to have an example following the table in read-the-docs.

Regarding the review suggestions, please click on the highlighted +/- button below, and then change the line within the `suggestion` block as desired. Then, if you click on the "Preview" tab, GitHub will already show the diff. This diff can be commit-ed either as a single commit, or all comment fixes as a "batch commit."

File: mathics/builtin/system.py [hunk 9]
```diff
@@ -871,3 +871,29 @@ def eval_with_symbol(self, symbol, evaluation: Evaluation) -> Integer:
...
+    Out[2]= Breakpoint[]
```
aravindh-krishnamoorthy, line 891:
```suggestion
    = Breakpoint[]
```

mmatera:
> @mmatera The intermittent TimeConstraint failures in this and in another PRs are getting annoying. Any suggestions on how to address? Mine is to mark it skipped until it is no longer flaky.

Just by making the time in the second argument shorter. It seems that something became faster since I added this tests...

aravindh-krishnamoorthy:
Perfect! I hope you find the feature useful. I just edited the previous comment to say that multi-line suggestions are also possible: Multi-line suggestions are also possible: https://stackoverflow.com/questions/60311158/how-can-i-suggest-multiple-lines-be-changed-in-markdown

Also, if you would prefer so, I could also commit the changes based on the review. However, as you suggested, I will start working directly on the `Mathics3/mathics-core` repository from the next PR.

mathics/builtin/system.py
mmatera: > @mmatera as I suggested to rocky, if you don't mind, could you please use the +/- option in the review to suggest the changes, which I can apply directly. That way, your contribution is also acknowledged, and it's easier for me as well...
> 

@aravindh-krishnamoorthy, actually I did it in that way: I added the comment just in the line where your test is localized.

> Please see step 8 (as of today) here for details: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request

mmatera:
> Perfect! I hope you find the feature useful. I just edited the previous comment to say that multi-line suggestions are also possible: Multi-line suggestions are also possible: https://stackoverflow.com/questions/60311158/how-can-i-suggest-multiple-lines-be-changed-in-markdown
> 
> Also, if you would prefer so, I could also commit the changes based on the review. However, as you suggested, I will start working directly on the `Mathics3/mathics-core` repository from the next PR.

Ah, sorry, I didn't understand that. Indeed, it looks useful!

rocky:
For me, this is okay to merge. There may be other changes down the line, but these are more tweak-kind of things.

mmatera (approved):
LGTM

PR #1189, opened 2024-11-26
Make the time shorter in TimeConstrained tests

mmatera:
This change should make these tests more stable...


mmatera:
This change is trivial, so let's merge it and adjust it again if necessary.

PR #1190, opened 2024-11-26
split no-meaning operators, add infix precedence

rocky:
Operators without Built-in Meanings is now a "Guide section" with parts for postfix, prefix and infix operators and operators that require additional Mathics3 module loading.

The precedence for infix operators is not set properly.


rocky:
After this is merge, there will be yet one more round that has to take place on both the MathicsScanner side and mathics-core side to redo table generation to include operator precedence for prefix and postfix no-meaning operators.

rocky:
After this merge, Mathics3 module  graph needs to be revised to import DirectedEdge and UndirectedEdge from changed module locations.

mmatera (approved):
LGTM

PR #1191, opened 2024-11-27
Another mini operator refactor

rocky:
Set operator precedence for  no-meaning prefix and posftix operators

Remove some  `operator =`  assignments in some Operator classes so we use the operator symbols from JSON.

A lot more work is needed here. This is just the first step.


mmatera (approved):
LGTM

PR #1194, opened 2024-11-28
Bugfix in `With`: Return evaluated result

aravindh-krishnamoorthy:
# Problem
```mathematica
Block[{i = 0}, With[{}, Module[{j = i}, Set[i, i+1]; j]]]
```` 
returns `1` whereas the correct answer is `0`. This is because, `With` evaluates the `Module` expression but erroneously returns the unevaluated version, triggering a second evaluation.

# Fix
The fix is to return the evaluated version. With the fix, the above code returns `0` as expected.

### DIFF
--- a/mathics/builtin/scoping.py
+++ b/mathics/builtin/scoping.py
@@ -712,5 +712,4 @@ class With(Builtin):
 
         vars = dict(get_scoping_vars(vars, "With", evaluation))
         result = expr.replace_vars(vars)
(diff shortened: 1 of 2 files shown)


aravindh-krishnamoorthy:
> Thanks! Would you add a pytest for this, please?

Indeed! Soon will add a pytest...

rocky:
LGTM

mmatera (approved):
LGTM

PR #1196, opened 2024-11-28
For unicode operators, remove class assign...

rocky:
For Unicode operators, remove class assign and get value from JSON.

Operators that are ASCII-only probably needs some coordination with JSON output on the mathicsscanner side.


mmatera (commented):
LGTM

PR #1199, opened 2024-12-01
Improve cosine distance

### LINKED ISSUE #1197: `CosineDistance[{a, b}, {x, y}]` - x,y values aren't conjugated
In WMA 14.1.0

```mma
  CosineDistance[{a, b}, {x, y}]
```

returns a result with x,y values conjugated:
```mma
  1 - (a*Conjugate[x] + b*Conjugate[y])/(Sqrt[Abs[a]^2 + Abs[b]^2]*Sqrt[Abs[x]^2 + Abs[y]^2]) 
```

- https://github.com/Mathics3/mathics-core/blob/807bd57f075a109cd790ad10a1365c26fe451306/mathics/builtin/distance/numeric.py#L142

rocky:
Handle CosineDistance more correctly in symbolic computations that should use Conjugate. Code adapted from https://github.com/axkr/symja_android_library/blob/master/symja_android_library/matheclipse-core/src/main/java/org/matheclipse/core/builtin/ClusteringFunctions.java#L377-L386

However, we also add the situation where arguments are numeric scalars, and we check for mismatched vector lengths.

Some tweaking of code for conversion to/from Complex has been done. 

TODO: more tests. 

There is a bit more that could be done, but this is getting large already. 

Fixes #1197

### DIFF
--- a/mathics/builtin/distance/numeric.py
+++ b/mathics/builtin/distance/numeric.py
@@ -2,9 +2,8 @@
 Numerical Data
 """
 
-from mathics.core.atoms import Integer0, Integer2
+from mathics.core.atoms import Integer2
 from mathics.core.builtin import Builtin
-from mathics.core.convert.sympy import from_sympy, to_sympy_matrix
 from mathics.core.expression import Evaluation, Expression
 from mathics.core.symbols import SymbolAbs, SymbolDivide, SymbolPlus, SymbolPower
 from mathics.core.systemsymbols import (
@@ -13,6 +12,7 @@ from mathics.core.systemsymbols import (
     SymbolSubtract,
     SymbolTotal,
 )
+from mathics.eval.distance import eval_CosineDistance
 
 
 def _norm_calc(head, u, v, evaluation: Evaluation):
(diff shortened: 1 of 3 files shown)


mmatera (approved):
LGTM 
Doctests would be OK by now. Then we can add more pytests for the module.

rocky:
> LGTM Doctests would be OK by now. Then we can add more pytests for the module.

Ok. Merging then. 

FYI - Notes for later have:

```
CosineDistance[5, 6]
 = 0

CosineDistance[5, 6.0]
 = 0.0

CosineDistance[{Complex[1, 0], Complex, 0, 2]]
 = 0.0

CosineDistance[Complex[5.0, 0], Complex[10,3]]

 = 0.0421737 + 0.287348 I

CosineDistance[I, 3]
1. - 1. I
```

Testing that from/to python respects integerness of real and imaginary parts. 
check_ArrayQ and check_SparseArrayQ should be called `eval_xxx` and put in mathics.eval. 

I am surprised at how often when I look at the code, I see there is still very much that needs revision.

The big thing right now is that I suspect there are other things like this: 

```
            return Expression(
                SymbolDivide,
                Expression(SymbolTotal, Expression(SymbolAbs, t)),
                Expression(
                    SymbolTotal, Expression(SymbolAbs, Expression(SymbolPlus, u, v))
                ),
```
which was is addressed here. 

Remember in one recent issue where you mentioned a "limitation" of SymPy in that it doesn't see the full picture in order to be able to do the right thing? Well, when we write stuff like this instead of converting things to sympy and then in one go issuing: 

```
        distance = 1 - u_val * v_val.conjugate() / (abs(u_val) * abs(v_val))

```

we are bringing on this kind of problem on ourselves.

PR #1200, opened 2024-12-02
 More tests, some minor other changes

rocky:


mathics/builtin/distance/numeric.py
```diff
@@ -186,16 +186,16 @@ def eval(self, u, v, evaluation: Evaluation):
...
 
         if not u_is_vector:
-            if isinstance(u, (Complex, Integer, Real)) and isinstance(
+            if not isinstance(u, (Complex, Integer, Real)) and isinstance(
                 v, (Complex, Integer, Real)
             ):
                 return
 
-            if v_is_vector:
+            elif v_is_vector:
```
mmatera, line 198: Usually linters complain about this structure, in which an `if` block ends with a `return` instruction, and then it is followed by an `elif` clause. The reason is that it is clear that the code in the`elif` block is only executed if the condition in the `if` clause is false. 
Why you want to change this?

mmatera (approved):
Beyond the question in the comment, LGTM

mathics/builtin/distance/numeric.py
rocky: It looks like this had been the case in master as well. Addressed now though.

PR #1210, opened 2024-12-05
Add 3-arg Select and go over...

rocky:
Add 3-argument Select, by adding a parameter inside Structure's filter function.

Go over Structure class:
   * remove duplication of filter() and add the count functionality needed by Select[]

Update over Select doctests for 3 arg form. Move error checking to pytest where it belongs.

`Select[]` count-parameter issue is mentioned  in https://github.com/Mathics3/Mathics3-Rubi/pull/2


mmatera (approved):
LGTM

PR #1211, opened 2024-12-06
Complex cosine distance

rocky:
Switch from using Python's multiply and divide, and abs() of complex numbers  to SymPy's Expr operators for more precise real-number results


mmatera (approved):
LGTM

PR #1214, opened 2024-12-11
Misc changes in advance of Box expression parsing

rocky:
Move unescaping string from "convert" to parsing (mostly - filenames might be done in a later step).

A number of small miscellaneous tweaks and mistakes found in looking over the code.

### DIFF
--- a/mathics/builtin/atomic/strings.py
+++ b/mathics/builtin/atomic/strings.py
@@ -428,7 +428,6 @@ class InterpretedBox(PrefixOperator):
      = 4
     """
 
-    operator = "\\!"
     summary_text = "interpret boxes as an expression"
 
     def eval(self, boxes, evaluation: Evaluation):
--- a/test/core/parser/test_parser.py
+++ b/test/core/parser/test_parser.py
@@ -49,8 +49,8 @@ class ParserTests(unittest.TestCase):
 class PrecedenceTests(ParserTests):
     def test_minuslike(self):
         self.check("a * + b", "Times[a, Plus[b]]")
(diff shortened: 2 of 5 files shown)


mathics/builtin/atomic/strings.py
```diff
@@ -428,7 +428,6 @@ class InterpretedBox(PrefixOperator):
      = 4
     """
 
-    operator = "\\!"
```
rocky, line 431: Oddly, this line is causing: 

```
SyntaxWarning: invalid escape sequence '\['
``` 

in tests. As an operator, `contribute()`  creates a rule for this. Then in  parsing, we have encounter a problem when trying to unescape this string. 

However, in the future (before next release) the operator-ness will come from operators.yml in the scanner repository.  If we still have a problem, let's deal with it then.

test/core/parser/test_parser.py
```diff
@@ -161,7 +161,10 @@ def testString(self):
         self.check(r'"abc"', String("abc"))
         self.incomplete_error(r'"abc')
         self.check(r'"abc(*def*)"', String("abc(*def*)"))
-        self.check(r'"a\"b\\c"', String(r"a\"b\\c"))
+
+        # The following check is not correct.
+        # the answer *should* be r'a"b\\c"'
+        # self.check(r'"a\"b\\c"', String(r"a\"b\\c"))
```
rocky, line 167: The fix for this should come from the scanner side. Possibly, we'll move tests like this over there.

mmatera (approved):
LGTM

PR #1217, opened 2024-12-11
simplify some all(...) and any(...)

fchapoton:
as it is better to avoid creating the inner list

`ruff` code C419


mmatera:
LGTM

PR #1218, opened 2024-12-11
Sort error message strings and add stresc

rocky:
Sort error message strings and add `stresc`.
Also, in the scanner "sntoct{1,2} do not provide a file/line combination


mmatera (approved):
LGTM

PR #1219, opened 2024-12-12
Handle 1-arg Polynomial and ...

rocky:
go over Polynomial doc.
Also, lint mathics.builtin.numbers.algebra module


rocky:
This bug was noticed in playing with Combinatorica 2.0.0

PR #1220, opened 2024-12-12
Combinatorica V0.9 workarounds and expanded tests

rocky:
SetPartitions[]`  and `KSetPartitions[]` from Combinatorica workin V2.0 (In V.09 it is not broken, it is just altogether missing.)  

So copy the 2.0.0 code into V0.9


mathics/packages/DiscreteMath/CombinatoricaV0.9.m
```diff
@@ -1174,7 +1219,7 @@
   Module[{mult = Table[0, {n}], j, d, r=n, z},
     While[ (r > 0),
       d = 1;  j = 0;
-      z = Random[] r PartitionsP[r];
+      z = RandomReal[] r PartitionsP[r];
```
rocky, line 1222: This change and others like this can be reverted after #1222 goes in

mmatera (approved):
LGTM

PR #1224, opened 2024-12-14
Tracing improvements

rocky:
Trace rewrite rules. Also, we were dropping some return value expressions.

Improvement seen from mathics: 

```
In[1]:= TraceEvaluation[Exponent[x, 1]]
  Evaluating Expression.evaluate(): System`Exponent[Global`x, 1]
    Rewriting: System`Exponent[Global`x, 1, System`Max] -> (<Expression: <Symbol: System`Exponent>[<Symbol: Global`x>, <Integer: 1>, <Symbol: System`Max>]>, True)
    -> System`Exponent[Global`x, 1, System`Max]
    Rewriting: System`Max[0] -> (<Expression: <Symbol: System`Max>[<Integer: 0>]>, True)
    -> System`Max[0]
  Returning: System`Exponent[Global`x, 1] = 0
Out[1]= 0

In[2]:= TraceEvaluation[1+3]
  Evaluating Expression.evaluate(): System`Plus[1, 3]
  Returning: System`Plus[1, 3] = 4
Out[2]= 4
```

image


mathics/eval/tracing.py
```diff
@@ -17,16 +17,63 @@
...
     """
     Called from a decorated Python @trace_evaluate .evaluate()
     method when TraceActivate["evaluate" -> True]
     """
+
     if evaluation.definitions.timing_trace_evaluation:
         evaluation.print_out(time.time() - evaluation.start_time)
+
+    # Test and dispose of various situations where showing information
+    # is pretty useless: evaluating a Symbol is the Symbol.
```
mmatera, line 30: mmm what about the ownvalues? (Example: `a=3` and then `a`?)

rocky: I don't understand. But better, if you have a suggestion for how to improve, just commit on top. There is a _lot_ of room for improvement. Thanks.

mmatera: My comment is just that I do not understand yours: differently to what happens with an Integer or a String, Symbols can be evaluated to something different. It happens when they definitions contain `Downvalues`.

rocky: As a heuristic, I have noticed that if we miss something like this, that is if we miss a transition from a symbol not having a value to one new getting a value, that is generally okay because the value starts to show up in the next evaluate or return message. 

Right now, in my view TraceEvaluation is a little too noisy.  Down the line, I imagine a way to indicate whether one wants to see both rewrites and evaluations or just one or the other. 

What I suggest here, is to come up try out what is here and come up with specific examples and output and then show how one is better or worse than another.  And then also if you want to modify code to show how things get better, that is great too!

mmatera: This is what I get in master:
```
In[1]:= a=4; 
Out[1]= None

In[2]:= TraceEvaluation[a+b]
  Evaluating: System`Plus[Global`a, Global`b]
    Evaluating: System`Plus
    Evaluating: Global`a
      -> 4
    Evaluating: Global`b
Out[2]= 4 + b
```

and this with this branch:
```
In[1]:= a=4;
Out[1]= None

In[2]:= TraceEvaluation[a+b]
  Evaluating Expression.evaluate(): System`Plus[Global`a, Global`b]
      -> 4
    Returning: Global`a = 4
  Returning: System`Plus[Global`a, Global`b] = System`Plus[4, Global`b]
```

The difference that I see is that trivial evaluations of the expression heads are not shown. I guess that is OK. However,  I find 
```
  Evaluating Expression.evaluate(): System`Plus[Global`a, Global`b]
      -> 4
```
a little bit confusing: it looks like the result of evaluating `Plus[a,b]` is `4`. Then 
```Returning: Global`a = 4``` only makes sense if we assume that there was a call to `Symbol.evaluate()`.

rocky: > The difference that I see is that trivial evaluations of the expression heads are not shown. I guess that is OK. However, I find
> 
> ```
>   Evaluating Expression.evaluate(): System`Plus[Global`a, Global`b]
>       -> 4
> ```
> 
> a little bit confusing

This apparently was old output from the code inside Expression.evaluation for old way  TraceEvaluation was implemented. I removed that code in commit ed5826787ef4e16a93f356ee8b83af608591d200

mmatera (approved):
It seems that this simplifies the code, and also helps to make that the trepan output looks better, so LGTM.

PR #1226, opened 2024-12-15
Combinatorica as submodule

rocky:
Remove Combinatorica from this repository but include it as a git submodule from Mathics3-Combinatorical. 

As a by-product, version 2.0.0 of Combinatorica is also now available. An additional CI job is set up to run the Combinatorica tests that we have (currently for v0.9). These will be expanded soon. 
 And possibly we may include v2.0.0 tests as well. 

To be determined if we want the one CI just to run tests for other packages as well.


mmatera:
> is

The only question I have is  if we want to run the Combinatorica tests here: combinatorica tests are about testing the package, or testing the support of Mathics-core to the package?

rocky:
> > is
> 
> The only question I have is if we want to run the Combinatorica tests here: combinatorica tests are about testing the package, or testing the support of Mathics-core to the package?

The combinatoria tests test both the Mathics3 code we have and Mathics3's support of basic primitives. 

Some examples to support Mathics3's support of basic primitives. The bugs uncovered and fixed have to do with Drop'ing and Take'ing element 0. These I guess are just plain bugs.  Supporting Random while it is deprecated, still seems to be around in WMA, and probably because too many packages don't want to switch. 

So here is how this PR proposes to handle this.  There is a new CI job just for combinatorica (whether we expand this workflow  to include other packages or make them separate workflows is up for discussion and experimentation). 

This way, we don't have to test combinatorica on all of the different OS versions and OS''s; one is probably fine for here.  And this is done _in parallel_ to the other tests.

Over in Mathics3-Combinatorica we can test many versions and OS's.  

Your thoughts?

mmatera:
> 
> The combinatoria tests test both the Mathics3 code we have and Mathics3's support of basic primitives.
> 
> Some examples to support Mathics3's support of basic primitives. The bugs uncovered and fixed have to do with Drop'ing and Take'ing element 0. These I guess are just plain bugs. Supporting Random while it is deprecated, still seems to be around in WMA, and probably because too many packages don't want to switch.
> 
> So here is how this PR proposes to handle this. There is a new CI job just for combinatorica (whether we expand this workflow to include other packages or make them separate workflows is up for discussion and experimentation).
> 
> This way, we don't have to test combinatorica on all of the different Python versions and OS''s; one Python version on one OS is probably fine for here. And this is done _in parallel_ to the other tests.
>
> Over in Mathics3-Combinatorica we can test many Python versions and OS's.
>

 This is a very good point

 
> Your thoughts?

OK, go ahead then! LGTM

PR #1231, opened 2024-12-16
Adjust sizes and text to match Mathematica better?

rocky:
graphics.py: Reduce size on PointSize, Thickness, AbsoluteThickness
svg.py: reduced thickness of text in InsetBox

I have been looking at graphs in Implementing Discrete Mathematics and noticed for a while that dots shown in Mathics3  seem a bit too big compared to what they are in the text, and text for axes is too bold. 

Here is an attempt to make this look more like what is printed in this book. 

For comparison with change: examples of dot sizes: 

Before:

image

Proposed change:

image

And text in axes of graphs. Before: 

image

Proposed change: 

image

Since I don't have access to anything that draws corresponding Mathematica graphs I rely on you to let me know if this is closer.

Note this is not about whether one is aesthetically better than the other, but instead, whether this matches what Mathematica does. The examples are all taken from the Django gallery.

### DIFF
--- a/mathics/builtin/graphics.py
+++ b/mathics/builtin/graphics.py
@@ -411,7 +411,7 @@ class _Size(_GraphicsDirective):
     def init(self, graphics, item=None, value=None):
         super(_Size, self).init(graphics, item)
         if item is not None:
-            self.value = item.elements[0].round_to_float()
+            self.value = item.elements[0].round_to_float() * 0.7
         elif value is not None:
             self.value = value
         else:
--- a/mathics/format/asy.py
+++ b/mathics/format/asy.py
@@ -417,12 +417,11 @@ def inset_box(self, **options) -> str:
         edge_color=self.color, edge_opacity=opacity_value, fontsize=font_size
     )
     asy = """// InsetBox
-label("$%s$", (%s,%s), (%s,%s), %s);\n""" % (
+label("$%s$", (%s,%s), %s, %s);\n""" % (
         content,
         x,
         y,
-        -self.opos[0],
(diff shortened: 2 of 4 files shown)


rocky:
And with commit 3fde7b23 we align southwest: 

Current:

image


Proposed:

image

mmatera:
> 
> Since I don't have access to anything that draws corresponding Mathematica graphs I rely on you to let me know if this is closer.
> 

@rocky, you can use `Export` in wolframscript to see how the graphics looks in WMA.  For example,
`Export["output.pdf",Plot[x,{x,0,2}]]`  stores in  `output.pdf` a plot that looks quite close to what you can see in the notebook interface.


> Note this is not about whether one is aesthetically better than the other, but instead, whether this matches what Mathematica does. The examples are all taken from the Django gallery.

mmatera:
In any case, LGTM

rocky:
> In any case, LGTM

Thanks - there will be more. It looks like the lines that asymptote draws are a bit thicker compared to SVG.

aravindh-krishnamoorthy:
Great! LGTM too! Export works well!

rocky:
The follow on - #1234 makes the graphs just a tad better. But I only realized this after looking at the PDFs, so Export is way cool. We should add that form to Mathics3.

The other thing that Export gives us is a way to format each primitive like InsetBox in isolation, see how it renders in SVG (alas, no asymptote as far as I know), and then we can try to match that in Mathics3.  I haven't started seriously doing that, but I bet that our rendering would get much better.

PR #1234, opened 2024-12-18
Start adding Alignment to InsetBox in 2D graphs

rocky:
Add the notion of text alignment of InsetBox which is typically used in drawing axes. The text for an x (horizontal) axis is under the ticks while the text for a y (vertical) axis is to the left.

Here we've done this for asymptote and SVG. There is a lot more that could be done, but this is a small start.


mmatera (approved):
LGTM

Issue #1237, opened 2024-12-21
`MapAt[f, expr, {i, j, ...}]` is broken when expr and i, j, etc are lists of lists. 


rocky:
## Description

The implementation of `MapAt[]` is horribly lacking. (I looked at who wrote this mess, and I see it was me!)

In particular, it does not handle lists of lists. 

## How to Reproduce

```
$ mathics -e 'MapAt[1&, {{0, 0}, {0, 0}}, {{1, 2}, {2, 1}}]'
{{0, 0}, {0, 0}}
```

## Expected behavior

The right answer for the previous section is `{ {0, 1}, {1, 0} }`. 

## Workarounds

An implementation could probably done in pure Mathic3 code.

## Priority

This is used in Combinatorica V0.9 which is how I discovered it. It appears that MapAt was added for the first time by me in trying to get Combinatorica working initially. 

## Additional context

Symja doesn't have the lameness and shallowness that Mathics3 currently has in its implementation. @axkr I would be grateful if you'd describe in human language, or at a high level, how SymJa implements MapAt. If possible, I'll try to follow that behavior in Mathics3's implementation.

axkr:
> I would be grateful if you'd describe in human language, or at a high level, how SymJa implements MapAt.

I refactored the main algorithm here:
- https://github.com/axkr/symja_android_library/blob/master/symja_android_library/matheclipse-core/src/main/java/org/matheclipse/core/eval/util/positions/MapPositions.java

Maybe you can "reference" the part in a new Symja discussion entry, where we can discuss, what's going on in the Java code?

rocky:
In short, I think I have found a solution or the beginning of a solution in https://github.com/Mathics3/mathics-core/pull/1239. 

I spent a day looking at MapAt and its wonky behavior.  The code has to be mindful of whether the third argument is an integer, a vector position, or a list of vector positions. The code has to tolerate changing the same location several times, and the shape of the replaced position can change dimension as well. 

And then a position might be out of range, but it also might specify a higher number of dimensions than the target expression has. 

I was interested in MapAt because it is used a bit (or at a lower-level core graph function) in Combinatorica v0.9. In Combinatorica v2.0.0 it is still around but not used so much.  And I don't see it used in Rubi and probably most other packages.

rocky:
Fixed by #1239

PR #1238, opened 2024-12-21
Revise `MapAt[]` part 1

rocky:
Miscellaneous lint for expanding and correcting `MapAt[]` 

* add eval_MapAt (move out of builtin class), and start to address error messages more correctly 
* walk_parts -> eval_Part which is what it is. And move to `mathics.eval.list.eol` since that is where it belongs
* DRY use of "psl" message
* Drop "_" at the beginning of routines which weren't really internal.


rocky:
@mmatera I haven't begun to add the needed bug fixes for #1237 but this is already starting to get large.

mmatera:
> @mmatera I haven't begun to add the needed bug fixes for #1237 but this is already starting to get large.

My only comment here is that `walk_parts` got its name from a routine already existing in the old Mathics code.  That routine allowed both to find a part of a expression or to apply a function over some of the parts.

mmatera (approved):
LGTM

rocky:
> > @mmatera I haven't begun to add the needed bug fixes for #1237 but this is already starting to get large.
> 
> My only comment here is that `walk_parts` got its name from `walk_levels`, a routine already existing in the old Mathics code. That routine allowed both to find a part of a expression or to apply a function over some of the parts.

As the code is currently written, `eval_Part` exactly handles the `Part[]` built-in function when we don't have a ByteArray. Therefore it is exactly the kind of function that we have been calling eval_<builtin-name> and placing with a similar modular structure as builtin in `mathics.eval`.

Initially, my thought was that functions called eval_Xyy are likely to be used the implementation of some future instruction set which would be used after compilation. 

But what I am coming to appreciate in doing the debugger work is that by making things regular, it makes things easier on a debugger or some other similar introspection tool.  If I want to show, track, or stop at key parts of execution, doing this in routines in `mathics_eval` that have the name eval_Xyy may be helpful. And usually if I go up one level in a call stack often there will be an eval  _method_ which is also probably of interest. 

Sure, I understand that a lot of what is there now may have some historical precedent. And the name walk_xxx conveys some sort of idea and possibly groups things along some lines.  

In contrast to calling "elements", "leaves" or "eval" routines "apply",  it is not a bad name or _wrong_, just not as helpful for the ways in which the code is currently envisioned to be organized in a way that makes it helpful to scale in the future. 

It may very well be that in the future we may move and rename "walk_levels" is not a bad name or _wrong_, just not as helpful for the ways in which the code is currently envisioned to be organized in a way that makes it helpful to scale in the future. 

It may very well be that in the future we may move and rename "walk_levels". For example that may also be the bulk of some builtin-function evaluation routine as well.

PR #1239, opened 2024-12-21
Corrections to MapAt for Multi-dimensions...

rocky:
The following changes implement pretty much all of what Combinatorical V0.09 needs. We have much better compliance to WMA MapAt[]


rocky:
I should add some error handling tests.

PR #1240, opened 2024-12-22
Add stacktrace

rocky:
Start to implement StackTrace[]


rocky:
Not having this in master is getting in the way of my being able to debug stuff. If there are comments or problems, we will address down the line.

rocky:
(BTW, we do need to still deal with #1241

PR #1243, opened 2024-12-23
Improve PolynomialQ[] one arg...

rocky:
Improve PolynomialQ[] for one argument.

 We now handle variable functions, e.g. x[1], x[6], etc.

Simplify some argument processing.

This is needed for Combinatorica 2.0.0 and


rocky:
I will start a discussion later to describe what's up here and for us to decide how to best handle this.

PR #1245, opened 2024-12-25
Add "lpn" checking in _ListPlot

rocky:
"lpn" is "list of points" checking on ListPlot[] and ListLinePlot[]


rocky:
This PR started just before the last release.  It had an infinite loop in it when we had bad input operands. Now that we have an interrupt handler for such situations, especially the "show" and "debugger" commands, it was straightforward to find and fix the bug.

PR #1246, opened 2024-12-25
Lint found in ListPlot-point checking branch

rocky:
The good stuff from #1245, but alas the thing we were trying to do there can cause an infinite loop in the eval_ListPlot routine. So let's do this part and narrow that PR.

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -1417,8 +1417,13 @@ class Expression(BaseElement, NumericOperators, EvalMixin):
                 if not isinstance(result, EvalMixin):
                     return result, False
                 if result.sameQ(new):
-                    new._timestamp_cache(evaluation)
-                    return new, False
+                    # Even though result and new may be the same,
+                    # new can be a Expression[SymbolConstant: System`List...]
+                    # while "result' might be ListExpression!?
+                    # So make sure to use "result", not "new".
+                    if isinstance(result, Expression):
+                        result._timestamp_cache(evaluation)
+                    return result, False
                 else:
                     return result, True
 
--- a/mathics/eval/drawing/plot.py
+++ b/mathics/eval/drawing/plot.py
@@ -201,19 +201,21 @@ def eval_ListPlot(
             [xx for xx, yy in plot_groups], [xx for xx, yy in plot_groups], x_range
         )
         plot_groups = [plot_groups]
-    elif all(isinstance(line, list) for line in plot_groups):
-        if not all(isinstance(line, list) for line in plot_groups):
+    elif all(isinstance(line, (list, tuple)) for line in plot_groups):
+        if not all(isinstance(line, (list, tuple)) for line in plot_groups):
             return
 
         # He have a list of plot groups
         if all(
-            isinstance(point, list) and len(point) == 2
-            for plot_group in plot_groups
+            isinstance(point, (list, tuple)) and len(point) == 2
+            for _ in plot_groups
             for point in plot_groups
         ):
             pass
         elif not is_discrete_plot and all(
-            not isinstance(point, list) for line in plot_groups for point in line
+            not isinstance(point, (list, tuple))
+            for line in plot_groups
+            for point in line
         ):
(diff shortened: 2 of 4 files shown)


mathics/core/expression.py
```diff
@@ -1410,8 +1410,13 @@ def rules():
                 if not isinstance(result, EvalMixin):
                     return result, False
                 if result.sameQ(new):
-                    new._timestamp_cache(evaluation)
-                    return new, False
+                    # Even though result and new may be the same,
```
mmatera, line 1413: good catch!

mathics/eval/drawing/plot.py
```diff
@@ -201,19 +201,21 @@ def eval_ListPlot(
...
+    elif all(isinstance(line, (list, tuple)) for line in plot_groups):
+        if not all(isinstance(line, (list, tuple)) for line in plot_groups):
             return
 
         # He have a list of plot groups
         if all(
-            isinstance(point, list) and len(point) == 2
-            for plot_group in plot_groups
+            isinstance(point, (list, tuple)) and len(point) == 2
+            for _ in plot_groups
```
mmatera, line 211: I do not get this: what is the purpose of this for?

rocky: In the follow-on, there may be elements coming from LisetExpression which are tuples. I had considered Iterable here, but list/tuple is a little more specific. The other change here is to use `_` instead of `plot_group` because `_` is just needed as a placeholder for the double iteration.

rocky: Oh, and Merry Christmas!

rocky: Here is my understanding of what this code does: if at the top level we have a list, then all of the list elements of the top-level list should also be lists. If that's not the case, then we'll return None and not try perform/evaluate ListPlot.

I've always found it flaky the way WMA works where if something is wrong, sometimes a message is not produced but, instead, nothing is done, and possibly a rewrite rule may down the line deal with the unchanged result. 

So this kind of thing may be happening here. I assume this models WMA behavior.

mmatera: > Here is my understanding of what this code does: if at the top level we have a list, then all of the list elements of the top-level list should also be lists. If that's not the case, then we'll return None and not try perform/evaluate ListPlot.

Ok, but the line with the loop over the underscore variable does not play any role on that.


> 
> I've always found it flaky the way WMA works where if something is wrong, sometimes a message is not produced but, instead, nothing is done, and possibly a rewrite rule may down the line deal with the unchanged result. 
> 
> So this kind of thing may be happening here. I assume this models WMA behavior.
> 
> 

Yep, the general philosophy in WL seems to be that  if there is a way in which an expression have a meaning, then the expression is acceptable. Otherwise, just leave it as it is, until a definition is provided.

rocky: > 
> Ok, but the line with the loop over the underscore variable does not play any role on that.

Ah. Good catch. I've removed that line and tested locally. I guess we should see if this has any effect on anything, so I removed it in 84ddcb23. 

There is a lot of code that still stinks. mypy flags some fishy things like the fact that we were looping over a variable that isn't used. But I guess it is up to us to understand whether the entire double loop is even necessary. And here, there is still too much code that feels screwy that my mind glosses over some of the remaining mess that is still there. So in sum, thanks.

.

mmatera: The plot module contains a large and entangled code, which have passed through a couple of refactors, so  this kind of things are expactable. Hopefully, as we continue disentangling this, we will eventually to converge to something more clear and solid.

rocky: In particular, I was trying to get Mesh with an integer value working. I see that while for Plot we have moved this code over to eval for Ploat3D we haven't. 

I gave up on getting Mesh working with an integer value.

rocky:
Ok. We are passing now. Merging before something else breaks.  Feel free to follow up with any of the numerous messes that remain.

PR #1248, opened 2024-12-28
Add SyntaxQ and add a new "String Tests" Section

rocky:
* Add `SyntaxQ[]`
* Create  "String Tests" section in the "Testing Expressions" Guide section
*  Misc lint.
* move another eval function to `mathics.eval`

`SyntaxQ[]` is used in the PacletManager.


mathics/builtin/testing_expressions/string_tests.py
```diff
@@ -0,0 +1,287 @@
...
+    """
+
+    # FIXME:
+    # replace messages[string] with the below.
+    # Better is to have this kind of thing done a function for doing this.
+    # Same things for "noopt" message which does not exist yet.
+    messages = {"string": "String expected at position `1` in `2`."}
+
+    summary_text = (
+        "test whether a string is as yntactically-correct a Mathics3 expression"
```
axkr, line 269: Typo: `syntactically`

rocky: Thanks! Should be fixed now. 

BTW I don't know why this reminded me, but a long time ago I asked you in slack about packages that would be easy to get running. Your reply in slack history has been long removed. Do you remember what package or packages you mentioned?

axkr: > BTW I don't know why this reminded me, but a long time ago I asked you in slack about packages that would be easy to get running. Your reply in slack history has been long removed. Do you remember what package or packages you mentioned?

I only documented this one in the Wiki:
- https://github.com/axkr/symja_android_library/wiki/context#wl-packages

mmatera (approved):
The only thing that I think is missing here is to include some tests for SyntaxQ

rocky:
> The only thing that I think is missing here is to include some tests for SyntaxQ

There is a doctest for True and False. If there are other kinds of tests you'd like to see, please suggest them.

mmatera:
> > The only thing that I think is missing here is to include some tests for SyntaxQ
> 
> There is a doctest for True and False. If there are other kinds of tests you'd like to see, please suggest them. We might add a note in the doc suggesting that this is how one can check the version level, since that's what it is used for in PacletManager.
> 
> Also, we probably have pretty extensive Syntax checks (or should add them elsewhere). Checking the scanner and parser here doesn't feel that right.

OK, in that case, let's add them in another round. Merge this when you feel it is ready.

rocky:
There will indeed be more improvements down the line. We don't have SyntaxQ with a form. We don't have SyntaxLength implemented either.  However this allows us the possibility of running PacletManager which was the motivation for doing this.

PR #1256, opened 2024-12-30
Increased use of binary operator string values from JSON

rocky:
Pick up binary operator string from enhanced Mathics-Scanner JSON operator tables. 

For now, we have to limit to binary operators only.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -28,9 +28,10 @@ jobs:
         python -m pip install --upgrade pip
     - name: Install Mathics3 with full Python dependencies
       run: |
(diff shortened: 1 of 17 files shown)


rocky:
Note: after this PR we may have a few more binary operators; also postfix and prefix operators should be done as well. However just this part is major disruption in of itself, to warrant stopping here and doing the others other PRs.

mmatera (approved):
LGTM

rocky:
Mathics-scanner code is merged, so we'll now try with master. If this works, I will merge.  And maybe we can continue with prefix/postifx operators.

PR #1259, opened 2025-01-01
Add `Stack[]`

rocky:
Start `Stack[]` builtin function. I hope more sometime after release.


rocky:
The results don't quite match WMA, but it is a step in the right direction.

PR #1260, opened 2025-01-03
Parsing refactor

rocky:
Originally I wanted to make progress on Box operator parsing. However as happens too often, before I can make forward progress it is useful to clean things up and document the current state. 

This is not getting a bit large, so probably we should do the clean up work separate from the box operators to follow.

### DIFF
--- a/mathics/core/parser/README.md
+++ b/mathics/core/parser/README.md
@@ -24,11 +24,16 @@ references that explain the algorithm better than I could:
 - [Wikipedia: Operator-precedence parser](https://en.wikipedia.org/wiki/Operator-precedence_parser#Precedence_climbing_method)
 - [Eli Bendersky's blog entry: Parsing expressions by precedence climbing](http://eli.thegreenplace.net/2012/08/02/parsing-expressions-by-precedence-climbing)
 
-### Precedence
+This algorithm has a natural application to the Wolfram language.
 
-This algorithm has a natural application to the Wolfram language. Every
-operator has an inbuilt precedence level which can be accessed with e.g.
-`Precedence[Plus]`.
+### Precedence Value
+
+Every operator has a numeric precedence value which. This is
+user-visible via the built-in function `Precedence[]`,
+e.g. `Precedence[Plus]`.  A higher value causes an operator to bind
+more tightly. For example, the Times precedence 400 is higher than the Plus
+precedence 310 because a + b * c is a + (b * c), not (a + b) * c.
+
 
 ### Grouping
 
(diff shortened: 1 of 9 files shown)


mathics/core/parser/README.md
```diff
@@ -2,82 +2,91 @@
...
 The relevant code snippet for binary operators is:
 
 
 ```python
-def parse_binary(self, expr1, token, p):    # called with expr1
-    'expr <- expr1 BINARY expr2'
+def parse_binary(self, expr1, token, expr1_precedence: int) -> Node:
+    """
+    Implements grammar rule:
+	   expr ::= expr1 BINARY expr2
```
mmatera, line 57: Where was the `::=` meaning defined?

mathics/core/parser/README.md
```diff
@@ -166,81 +174,65 @@ This is simple to handle with precedence climbing:
...
 #### Backtracking
-Most of the Wolfram language can be parsed with precedence climbing but there
-are a few special language features that require something more. The `Span`
-operator is one example. Both `a ;; b` and `a ;;` are valid syntax, the latter
-is equivalent to `a ;; All`.
+Most of the Wolfram language can be parsed with precedence climbing but there are a few special language features that require something more. The `Span` operator is one example. Both `a ;; b` and `a ;;` are valid syntax, the latter is equivalent to `a ;; All`.
 
 Consider the example: `a;;!b`. There are four options for parsing this:
  - `Span[a, Not[b]]`
  - `Factorial[Span[a, Null]]`
```
mmatera, line 219: `Times[Factorial[Span[a,Null]], b`  ?

mathics/core/parser/README.md
```diff
@@ -166,81 +174,65 @@ This is simple to handle with precedence climbing:
...
  - `Times[Span[a, All], Not[b]]`
```
mmatera, line 220: Maybe it is worth to mention that this one is the WMA choice.

mathics/core/parser/README.md
rocky: I checked ANTR and BISON syntax and they both uses ":", so I'll change to this.

mmatera: The notation is OK, but it may be worth mentioning in this document.

mmatera (approved):
Apart from the small comments, LGTM. @rocky, thanks for this effort!

mathics/core/parser/README.md
rocky: When two expressions butt up against each other there is an implicit multiplication. For example `(5)3, or 3x. 

The interpretation of `Times[Factorial[Span[a,All]], b` is `(a;;!)b`.

So "All", not Null". 

I've gone over this example though in order to try to clarify it. Also, it looks like Mathics3 isn't parsing this correctly.

mathics/core/parser/README.md
rocky: I've mentioned that Mathics3 doesn't get this right.

rocky:
@mmatera I've tried to incorporate all of your suggestions/comments and address these. If I've missed some, feel free to put in more changes and PRs.

PR #1261, opened 2025-01-04
Using patched stopit for the  TimeConstrained support

mmatera:
This PR is for testing if the modified stopit helps to pass the tests in all the platforms.


test/builtin/test_datentime.py
```diff
@@ -10,29 +10,22 @@
...
-    reason="TimeConstrained is based in Threads, which are not supported in Piodide",
-)
 def test_timeconstrained1():
     #
     str_expr1 = "a=1.; TimeConstrained[Do[Pause[.01];a=a+1,{1000}],.1]"
     result = evaluate(str_expr1)
     str_expected = "$Aborted"
     expected = evaluate(str_expected)
     assert result == expected
+    assert evaluate("a").to_python() <= 11
```
rocky, line 26: This kind of coding smells really bad.  It assumes something about the speed of computation.

mmatera: Indeed, it is fragile. In any case, what we actually need to test is if after `TimeConstrained` returns a value, the inner expression continues the evaluation.

mmatera: On the other hand, if `Pause` could ensure some degree of accurateness, and TimeConstrained does its job, I would expect that the number of increments would be bounded by some reasonable number.

rocky: So let's spend the time to come up with a better test please.

mmatera: OK, but right now I would like to figure out why this test fails randomly in macos and not in the other platforms. Once I narrow the issue, maybe a better test come up. Of course, ideas for a better test are welcome.

rocky: > OK, but right now I would like to figure out why this test fails randomly in macos and not in the other platforms.

A good test might help elucidate this.  

> Once I narrow the issue, maybe a better test come up. Of course, ideas for a better test are welcome.

Give me some time to think of something.

mathics/builtin/procedural.py
```diff
@@ -474,10 +474,14 @@ def eval(self, n, evaluation: Evaluation):
             return
 
         steps = int(self.PAUSE_TICKS_PER_SECOND * sleeptime)
+        step_duration = 1 / self.PAUSE_TICKS_PER_SECOND
```
rocky, line 477: This code change is implicitly assuming something about the way Python works. 

If one is using SIGALRM, e.g. via stopit, then having many small sleeps allows SIGALRM more opportunity to kick in. On the other hand if one is not using SIGALRM, this is not only more complicated but it puts more strain on the running program. There are scenarios where having allowing for larger sleep values allows threads to run longer.  

And the current implementation of sleep in Python possibly varies over OS and may possibly vary over Python version. (I don't think it has, but there is nothing stopping it from doing so.)

mathics/builtin/datentime.py
```diff
@@ -1065,65 +1070,74 @@ def evaluate(self, evaluation):
...
-                    return result
-            except Exception:
+    def eval_3(self, expr, t, failexpr, evaluation):
+        "TimeConstrained[expr_, t_, failexpr_]"
+        t = t.evaluate(evaluation)
+        if not t.is_numeric(evaluation):
+            evaluation.message("TimeConstrained", "timc", t)
+            return
+        try:
+            timeout = float(t.to_python())
```
rocky, line 1125: This line should not be inside the `try`.

mathics/builtin/procedural.py
mmatera: What I know is that stopit cannot interrupt `sleep` due to the Python's GIL lock, so I need to split it into small intervals, whatever the implementation is.

rocky: > What I know is that stopit cannot interrupt `sleep` due to the Python's GIL lock, so I need to split it into small intervals, whatever the implementation is.

Partially true. It cannot interrupt `sleep` if the `SIGALRM` is used if done via threads, a longer sleep is actually may be (or probably is) more advantageous. We haven't talked about stopit, but one of the ways on MS Windows to simulate SIGALRM is to do this in threads where there is a ticker thread. So a longer sleep time allows this more freedom to tick. 

BTW, it doesn't look like this implementation of TimeConstrained spawns a new thread. Is that right? 

Lastly, after spending the day looking at stopit, it looks to me that a robust solution is a lot more difficult and involved than what we have here.  I think a good solution is doable, but it is going to take a lot more work.

mmatera: TimeConstrained does not spawns a new thread: stopit does, but the thread does not contains the code to execute, but just a the timer that raises the exception. So the code in  `Pause.eval` is in the main thread.

mathics/builtin/procedural.py
```diff
@@ -472,11 +472,29 @@ def eval(self, n, evaluation: Evaluation):
...
+        # It was also noticed in tests that in some platforms
+        # the total time that takes n calls to time.sleep(delta_t)
+        # can be appreciably larger than n*delta_t. For this reason,
+        # we also need to chech that inside the loop that the
+        # enlapsed time at the i-esim iteration does not exceed
+        # the  desired total time.
+
         steps = int(self.PAUSE_TICKS_PER_SECOND * sleeptime)
+        step_duration = 1 / self.PAUSE_TICKS_PER_SECOND
+        start = time.time()
```
rocky, line 491: See https://stackoverflow.com/questions/8420422/python-windows-equivalent-of-sigalrm for a more robust mechanism that uses `Threading.Event` to allow Pause or sleep to be canceled.

rocky: Involved code like this should be in eval.

mathics/builtin/datentime.py
```diff
@@ -1134,27 +1134,11 @@ class TimeConstrained(Builtin):
...
-        t = t.evaluate(evaluation)
-        if not t.is_numeric(evaluation):
-            evaluation.message("TimeConstrained", "timc", t)
-            return
-        try:
-            timeout = float(t.to_python())
-        except (ValueError, TypeError):
-            return
-        return eval_timeconstrained(expr, timeout, SymbolAborted, evaluation)
+        return eval_timeconstrained(expr, t, SymbolAborted, evaluation)
```
rocky, line 1137: To be clear about what goes into the evaluation _method_ of Builtin object class versus the evaluation interpreter function. 

It is okay to do parameter type checking and parameter conversion in the class method, especially when interactions inside an interpreter are not likely to need to do this kind of thing. 

There is a bit of wiggle room for where to do this, so use your judgment.

mathics/builtin/procedural.py
mmatera: This should now go inside our `stopit` branch.

mathics/builtin/procedural.py
```diff
@@ -472,11 +472,29 @@ def eval(self, n, evaluation: Evaluation):
...
+        # by just calling `time.sleep(sleeptime)`, the
+        # evaluation would not be aware of `TimeoutException`
+        # raised by an outer `TimeConstrained` expression.
+        #
+        # For this reason, the splitting of the sleep time is
+        # needed.
+        # It was also noticed in tests that in some platforms
+        # the total time that takes n calls to time.sleep(delta_t)
+        # can be appreciably larger than n*delta_t. For this reason,
+        # we also need to chech that inside the loop that the
```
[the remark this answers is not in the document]
mmatera: I put both this code and  `eval_timeconstrained` code  inside `mathics.eval.datetime`. This would help to understand better what the comment talks about and to improve both at the time.

mmatera:
I included #1274 here to make the tests pass.

.github/workflows/pyodide.yml
```diff
@@ -54,6 +54,11 @@ jobs:
 
           pip install "setuptools>=70.0.0" PyYAML click packaging pytest
 
+          # First install our patched version of stopit
```
rocky, line 57: Since TimeConstrained is not used, there is no need to install stopit. 

Simpler CI is better, in my opinion.

mathics/builtin/datentime.py
```diff
@@ -1061,69 +1085,71 @@ class Now(Predefined):
...
+    if sys.platform == "emscripten":
+        messages.update({"tcns": f"TimeConstrained is not supported in {sys.platform}"})
 
-        summary_text = "run a command for at most a specified time"
+    summary_text = "run a command for at most a specified time"
 
-        def eval_2(self, expr, t, evaluation):
-            "TimeConstrained[expr_, t_]"
-            return self.eval_3(expr, t, SymbolAborted, evaluation)
+    def eval_2(self, expr, t, evaluation) -> Optional[BaseElement]:
```
rocky, line 1136: Let's use more descriptive eval names, e.g. `eval_with_time()`

mathics/builtin/datentime.py
```diff
@@ -1061,69 +1085,71 @@ class Now(Predefined):
...
-                    done = True
-                if done:
-                    evaluation.timeout_queue.pop()
-                    return result
-            except Exception:
-                evaluation.timeout_queue.pop()
-                raise
-            evaluation.timeout_queue.pop()
-            return failexpr.evaluate(evaluation)
+    def eval_3(self, expr, t, failexpr, evaluation) -> Optional[BaseElement]:
```
rocky, line 1145: eval_3 -> eval_with_time_and_failexpr

mathics/eval/datetime.py
```diff
@@ -0,0 +1,94 @@
...
+    #
+    # For this reason, the splitting of the sleep time is
+    # needed.
+    # It was also noticed in tests that in some platforms
+    # the total time that takes n calls to time.sleep(delta_t)
+    # can be appreciably larger than n*delta_t. For this reason,
+    # we also need to check that inside the loop that the
+    # enlapsed time at the i-esim iteration does not exceed
+    # the  desired total time.
+
```
rocky, line 49: Perhaps mention the possibility of implementing using `Threading.Event()`

test/builtin/test_datentime.py
```diff
@@ -12,7 +12,7 @@
 
 @pytest.mark.skipif(
     sys.platform in ("emscripten",),
-    reason="TimeConstrained is based in Threads, which are not supported in Piodide",
+    reason="TimeConstrained is not supported in Piodide",
```
rocky, line 15: Piodide -> Pyodide

test/builtin/test_datentime.py
```diff
@@ -22,17 +22,33 @@ def test_timeremaining():
...
```
rocky, line 25: Piodide -> Pyodide

mathics/builtin/procedural.py
mmatera: I added a comment in mathics.eval.datetime about this.

mathics/eval/datetime.py
mmatera: I put the comment in mathics.eval.datetime.

mathics/builtin/procedural.py
rocky: I think that we will have more flexibility if it is in a separate thread which is periodically looking at an event (via an improved stopit library.  The main thread however would be aware of Pause and TimeConstrained threads.

mathics/builtin/procedural.py
rocky: There is something about that comment that doesn't feel quite right, but I'd like to leave it until such time when we can devote resources to fully solving the problem.

rocky:
LGTM

PR #1264, opened 2025-01-05
Split mathics.core.assignment into  mathics.core.assignment and mathics.eval.assignment

mmatera:
And adding docstrings and type annotations.

### DIFF
--- a/mathics/builtin/assignments/assignment.py
+++ b/mathics/builtin/assignments/assignment.py
@@ -2,14 +2,8 @@
 """
 Forms of Assignment
 """
+from typing import Optional
 
-
-from mathics.core.assignment import (
-    ASSIGNMENT_FUNCTION_MAP,
-    AssignmentException,
-    assign_store_rules_by_tag,
-    normalize_lhs,
-)
 from mathics.core.atoms import String
 from mathics.core.attributes import (
     A_HOLD_ALL,
@@ -18,45 +12,16 @@ from mathics.core.attributes import (
     A_SEQUENCE_HOLD,
 )
 from mathics.core.builtin import Builtin, InfixOperator
+from mathics.core.element import BaseElement
+from mathics.core.evaluation import Evaluation
 from mathics.core.symbols import SymbolNull
 from mathics.core.systemsymbols import SymbolFailed
+from mathics.eval.assignments import eval_assign
 from mathics.eval.pymathics import PyMathicsLoadException, eval_LoadModule
 
 
-class _SetOperator:
-    """
-
-    This is the base class for assignment Builtin operators.
-
-    Special cases are determined by the head of the expression. Then
-    they are processed by specific routines, which are poke from
-    the ``ASSIGNMENT_FUNCTION_MAP`` dict.
-    """
-
-    # FIXME:
-    # Assignment is determined by the LHS.
-    # Are there a larger patterns or natural groupings that we are missing?
-    # For example, it might be that it
-    # we can key off of some attributes or other properties of the
-    # LHS of a builtin, instead of listing all of the builtins in that class
-    # (which may miss some).
-    # Below, we key on a string, but Symbol is more correct.
-
-    def assign(self, lhs, rhs, evaluation, tags=None, upset=False):
-        lhs, lookup_name = normalize_lhs(lhs, evaluation)
-        try:
-            # Using a builtin name, find which assignment procedure to perform,
-            # and then call that function.
-            assignment_func = ASSIGNMENT_FUNCTION_MAP.get(lookup_name, None)
-            if assignment_func:
-                return assignment_func(self, lhs, rhs, evaluation, tags, upset)
-
-            return assign_store_rules_by_tag(self, lhs, rhs, evaluation, tags, upset)
-        except AssignmentException:
-            return False
-
-
-# Placing this here is a bit weird, but it is not clear where else is better suited for this right now.
+# Placing this here is a bit weird, but it is not clear where else is better
+# suited for this right now.
 class LoadModule(Builtin):
     """
     ## <url>:mathics native for pymathics:</url>
@@ -88,13 +53,13 @@ class LoadModule(Builtin):
         except PyMathicsLoadException:
             evaluation.message(self.name, "notmathicslib", module)
             return SymbolFailed
-        except Exception as e:
-            evaluation.message(self.get_name(), "loaderror", String(str(e)))
+        except ImportError as exception:
+            evaluation.message(self.get_name(), "loaderror", String(str(exception)))
             return SymbolFailed
         return module
 
 
-class Set(InfixOperator, _SetOperator):
+class Set(InfixOperator):
     """
     <url>:WMA link:https://reference.wolfram.com/language/ref/Set.html</url>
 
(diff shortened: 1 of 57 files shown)


rocky:
Thanks for undertaking this.

(I think somewhere I see Builtin misspelled - running through `codespell` will catch stuff like this.

mathics/eval/assignment.py
```diff
@@ -0,0 +1,1461 @@
...
+    def __init__(self, lhs, rhs) -> None:
+        super().__init__(f" {rhs} cannot be assigned to {lhs}")
+        self.lhs = lhs
+        self.rhs = rhs
+
+
+class _SetOperator:
+    """
+
+    This is the base class for assignment Builtin operators.
```
rocky, line 56: Stuff in eval probably should not have classes with objects. The focus of eval is an interpreter that knows nothing about OO. You should be able to write this code in C or assembly.

It is okay to pass OO objects or "self" pointers though. For these there are object _dereferences_,; Dereferencing is different from implicit OO method lookup which goes on inside classes. 

In sum, this class probably belongs in either `mathics.core` (similar to Atom classes) or `mathics.builtin`.

rocky: Grepping `mathics.eval` right now I see: 

```console
$ PAGER=cat git grep -n "class "
distance/clusters.py:8:class IllegalDataPoint(Exception):
distance/clusters.py:12:class IllegalDistance(Exception):
files_io/read.py:42:class MathicsOpen(Stream):
patterns.py:11:class _StopGeneratorMatchQ(StopGenerator):
patterns.py:15:class Matcher:
pymathics.py:21:class PyMathicsLoadException(Exception):
rules.py:149:class Dispatch(Atom):
test.py:12:class _StopGeneratorBaseElementIsFree(StopGenerator):
```

Exceptions and Generators are weird kinds of classes, so there may be some wiggle room here. But `Dispatch(Atom)` and `Matcher` should probably be moved.

mathics/core/assignment.py
```diff
@@ -223,32 +228,33 @@ def rejected_because_protected(
 
 def unroll_conditions(lhs: BaseElement) -> Tuple[BaseElement, Optional[Expression]]:
     """
-    If lhs is a nested `Condition` expression,
+    If `element` is a nested `Condition` expression,
```
rocky, line 231: Does "element" refer to "lhs"  on line 229?

mathics/core/assignment.py
```diff
@@ -5,12 +5,12 @@
 Support for Set and SetDelayed, and other assignment-like builtins
 """
 
-from typing import Callable, List, Optional, Tuple
+from typing import Callable, List, Optional, Tuple, cast
```
rocky, line 8: Using cast is a bit of a cop-out. It just fixing things down the line. So if we can fix the problem without cast that would be nice. 

 But sure, I understand we have a _big_ mess to deal with.

mmatera: Yep, I am trying to figure out how to deal with functions that return `Optional[...]`. In any case, this will need a couple of rounds to get ready...

rocky: I have long had this annoying feeling that our classes in BaseElement have been a little bit wrong.  I've deferred undertaking addressing this until I understand the code better and what to do.

For example, it might be the case that everything should have a "head" and "elements", and perhaps  "value". And if we did that there would be fewer testing of types and simpler code.

mathics/eval/assignment.py
mmatera: The problem I found is that putting this class in another place produces circular dependencies among modules. I am trying to figure out how to avoid it.

mmatera: Easy: just get ride of the class...

mathics/core/assignment.py
mmatera: I think that only `Expression` and their subclasses should have `head` and `elements` properties. 
On the other hand, in cases like this, where you want to use polymorphism, maybe using `cast` is the best solution.
In this module, I used it in two cases: one is the conversion from `BaseElement` to `Expression`. The other is to use a definition obtained from  `Definitions.get_definition`.

In the second case, the problem is that `get_definition` returns  `None` when the symbol does not have a definition. There would be two alternatives:

* create an empty definition: This is costly and useless. We want to know as early as possible that a symbol does not have a definition, to avoid asking questions to it. Create an empty definition would introduce a lot of unnecesary overhead.
* raise an exception: maybe this would be a more pythonic solution, just that I do not finish to get used to think of exceptions as something which is part of the standar logic of a code.

mmatera:
I think this is a good moment to consider merging this before continuing.

mathics/core/assignment.py
rocky: > On the other hand, in cases like this, where you want to use polymorphism, maybe using cast is the best solution.

I speak from lots of experience in other programming languages, do not go down this path.

rocky:
> I think this is a good moment to consider merging this before continuing.

Give me a couple of days to look at. I have been meaning to see what's up with stopit first.

mathics/core/assignment.py
mmatera: OK. In any case, using `cast` is a good way to detect where the code get's fishy.

mmatera:
> > I think this is a good moment to consider merging this before continuing.
> 
> Give me a couple of days to look at. I have been meaning to see what's up with stopit first.

OK. There is no hurry. On the other hand, this is just juggling a little bit with some names and distribution of the code, and completing the documentation and type annotations.

mathics/core/assignment.py
mmatera: OK, now I was able to get ride of all  `cast`s

rocky: Thank you. I greatly appreciate the attention to improving code quality.

mathics/builtin/assignments/assignment.py
```diff
@@ -290,18 +259,27 @@ class TagSet(Builtin, _SetOperator):
...
-    def eval(self, f, lhs, rhs, evaluation):
-        "f_ /: lhs_ = rhs_"
+    def eval(
+        self,
+        tag: BaseElement,
+        lhs: BaseElement,
+        rhs: BaseElement,
+        evaluation: Evaluation,
+    ) -> Optional[BaseElement]:
+        "tag_ /: lhs_ = rhs_"
```
rocky, line 274: Actually, `f_` is correct here since this is how the documentation describes the parameter. I will be putting in a PR for this and other changes.

mathics/core/assignment.py
```diff
@@ -190,662 +227,59 @@ def rejected_because_protected(self, lhs, tag, evaluation, ignore=False):
...
-        # Maybe this replamement should be delayed instead,
-        # like
-        # rhs = Expression(Symbol("System`Replace"), Rule(*rulerepl))
-        # TODO: check if this is the correct behavior.
-        rhs, _ = rhs.do_apply_rules([Rule(*rulerepl)], evaluation)
-        name = lhs.get_head_name()
-    elif name == "System`HoldPattern":
-        lhs = lhs_elements[0]
-    return lhs, rhs
+# In the folliwing routines
```
rocky, line 230: ??

PR #1269, opened 2025-01-07
Assignment another round rocky

rocky:


mathics/builtin/assignments/assignment.py
```diff
@@ -266,14 +268,14 @@ class TagSet(Builtin):
 
     def eval(
         self,
-        tag: BaseElement,
+        f: BaseElement,
```
rocky, line 271: Note that in eval methods we should match the docstring. And the docstring parameter generally matches the WMA docstring parameter.

mmatera (approved):
Thanks! LGTM

PR #1270, opened 2025-01-08
Remove a few more flycheck messages and some other small changes.

rocky:
emove a few more flycheck messages and some other small changes


mathics/core/definitions.py
```diff
@@ -598,7 +598,7 @@ def get_accessible_contexts(self) -> set:
         accessible_ctxts.add(self.current_context)
         return accessible_ctxts
 
-    def get_matching_names(self, pattern: str) -> List[str]:
```
rocky, line 601: Line 616 suggests that we can pass a compiled pattern as well

mmatera (approved):
LGTM

PR #1274, opened 2025-01-08
track JSON table generation changes ...

rocky:
Extract and use builtin_constants
Extract operator_to_amslatex
change operator_precedence dict to operator_precedences


mmatera (approved):
LGTM

rocky:
We have other problems building documentation, but this addresses the problems in trying to show unicode operators in section heads. Here, we can use the AMSLaTeX equivalent string.

PR #1277, opened 2025-01-09
Add Close["path"]

rocky:
Also, move code to class to eval_Close under `mathics.eval`

We mimic WMA behavior which is slightly different from what the docs say it should do when closing duplicate open streams by name.


rocky:
This is used in PacletManager.  I need to add some tests for the new behavior.

mmatera (approved):
LGTM

PR #1279, opened 2025-01-10
Path search finds ".wl" extension & flycheck fixes

rocky:
* Include search of  `.wl` extension  of xxx in  ``Needs["xxx`"]``
* Use `eval_Get` instead of heavyweight `Expression[SymbolGet, ... ].evaluate()`. 
* Remove misguided comment recently added
* Guard against using None as suggested by flycheck
* Start unit test for `mathics.core.streams`


mmatera (commented):
LGTM

PR #1280, opened 2025-01-10
Split the merging definitions code from Definitions.get_definition

mmatera:
This PR continues the simplification of the code in the `mathics.code.definitions` module.


rocky:
LGTM

mathics/core/definitions.py
```diff
@@ -1150,3 +880,281 @@ def get_history_length(self) -> int:
         if history_length is None or history_length > 100:
             history_length = 100
         return history_length
+
+
+def autoload_files(
```
rocky, line 885: Why was this moved to the end? 

Also, autoloading files feels more like a session kind of thing than a "core" function. And finally the mathics.core.definitions module is pretty large so moving this somewhere else would balance things in terms of size too.

PR #1284, opened 2025-01-10
Simpler Definition class

mmatera:
This PR proposes to simplify the Definition class by storing all the attributes containing rules into a unique dictionary. Access to the lists is in any case done through the corresponding methods.

### DIFF
--- a/mathics/builtin/assignments/clear.py
+++ b/mathics/builtin/assignments/clear.py
@@ -81,12 +81,13 @@ class Clear(Builtin):
     summary_text = "clear all values associated with the LHS or symbol"
 
     def do_clear(self, definition):
-        definition.ownvalues = []
-        definition.downvalues = []
-        definition.subvalues = []
-        definition.upvalues = []
-        definition.formatvalues = {}
-        definition.nvalues = []
+        rules_dict = definition.rules_dict
+        rules_dict["ownvalues"] = []
+        rules_dict["downvalues"] = []
+        rules_dict["upvalues"] = []
+        rules_dict["subvalues"] = []
+        rules_dict["nvalues"] = []
+        rules_dict["formatvalues"] = {}
 
     def eval(self, symbols, evaluation):
         "%(name)s[symbols___]"
@@ -164,9 +165,9 @@ class ClearAll(Clear):
     def do_clear(self, definition):
         super(ClearAll, self).do_clear(definition)
         definition.attributes = A_NO_ATTRIBUTES
-        definition.messages = []
-        definition.options = []
-        definition.defaultvalues = []
+        definition.rules_dict["messages"] = []
+        definition.rules_dict["options"] = {}
+        definition.rules_dict["defaultvalues"] = []
 
 
 class Remove(Builtin):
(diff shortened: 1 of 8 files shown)


rocky:
I don't understand the motivation for this.  Please explain.

Instead of attributes on a "definition" object there new attribute "rules" dect that contains a key for each old attribute. In what way is this "simpler"? 

It would be nice when you see something large that you want to redo (which too often happens right before we think that there is going to be a release) to describe the idea and discuss _before_ coding.

mmatera:
> I don't understand the motivation for this. Please explain.
> 

OK, let's try to make a case on this. On the one hand, Pylint noticed that the `Definition` class has "too many instance-attributes" (R0902). This is a symptom that some attributes should be grouped to improve the organization of the code.
On the other hand, in most of the places where these attributes are accessed, it is done using `getattr`, so in a way they are already used as dictionary keys. Most is not all, so I also pass through the code to see if it was possible to make a more consistent use of the interface to these methods. Finally, the __init__ method now takes all these lists bundled in a dictionary, so it seems more direct to just store the dictionary instead of splitting it into several attributes. 

> Instead of attributes on a "definition" object, there is a new attribute "rules" dictionary that contains a key for each old attribute. In what way is this "simpler"?
> 
In my opinion, is "simpler" in the sense that we are grouping related things. 

> It would be nice when you see something large that you want to redo (which too often happens right before we think that there is going to be a release) to describe the idea and discuss _before_ coding.

I made this PR because I thought it was in the line of the previous ones, and it did not look to me too controversial. Also, it's easier for me to explain an idea on existing code than to do it in the abstract, especially if I'm not sure what it would take to implement it.

mathics/eval/inference.py
```diff
@@ -121,17 +121,8 @@ def remove_nots_when_unnecesary(pred, evaluation):
...
-        "System`$Assumptions", only_if_exists=True
-    )
-    if assumptions_def:
-        assumptions = assumptions_def.ownvalues
-        if len(assumptions) > 0:
-            assumptions = assumptions[0].replace
-    if assumptions is None:
-        return None
-
+    """Return a list of assumptions"""
```
mmatera, line 124: This is an example of how the change helps me to find places where the code can be made simpler. We can now discard the change in the attributes of `Definition` and keep this.

mathics/builtin/assignments/clear.py
```diff
@@ -81,12 +81,13 @@ class Clear(Builtin):
     summary_text = "clear all values associated with the LHS or symbol"
 
     def do_clear(self, definition):
-        definition.ownvalues = []
-        definition.downvalues = []
-        definition.subvalues = []
-        definition.upvalues = []
-        definition.formatvalues = {}
-        definition.nvalues = []
+        rules_dict = definition.rules_dict
```
mmatera, line 84: `do_clear` seems to be a method of `Definitions`, not of a Builtin class.

mathics/eval/inference.py
rocky: It is more modular is is not quite the same as simpler.

mmatera:
OK, let's close this PR and keep the branch for later.

rocky:
Thank you!

PR #1292, opened 2025-01-15
Improve get functions for patterns

mmatera:
This is one tiny change I wanted to implement for a while. Im WMA, if you want to implement a rule that returns the same expression if certain test fails, the way to do it is
```
expression: Half[expr]:=If[NumberQ[expr], expr/2, expression]
```
i.e. use a `Pattern[patname, pat]` expression to pass the whole original unevaluated expression (`Half[a]` in the example) to the replace expression. However, we could not do this in the patterns especified in docstrings, because we "overloaded" the meaning of `patname: pattern` in the function that loads `eval_*` and `format_*` methods: we use them to indicate in the case of `format_` methods to which `Format` corresponds the rule. 
As in some methods we need the full original expression, then we added a "hack" to interpret a parameter `expression` in an `eval_` function to contain the whole expression.

This PR removes the need of all this hackiness by changing the convention in `format_` methods docstrings: if you want to especify an specific Format, you can do it with a docstring of the form `(form1, form2,...,): exprpattern` which is not a valid WL expression.

We can also use this form to avoid the use of `Optional` in the `eval_` methods, so each `eval_` method could return always a `BaseExpression`. But this will come later.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -52,8 +52,14 @@ Compatibility
 Internals
 ---------
 
-Operator information has been gone over and is picked up from JSON
+* Operator information has been gone over and is picked up from JSON
 tables produced from the Mathics Scanner project.
+* Patterns in ``eval_`` and ``format_`` methods of builtin classes
+  parses patterns in docstrings of the form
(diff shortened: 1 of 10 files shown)


rocky:
LGTM

PR #1294, opened 2025-01-17
flycheck two pytests

rocky:
@mmatera I was trying to add a test for `HoldForm[Times[x]] == Times[x]` of a recent PR of yours (from a while ago) but haven't been able to figure out how to do. in the process, I have this lint stuff from the tests.

### DIFF
--- a/test/builtin/test_evaluation.py
+++ b/test/builtin/test_evaluation.py
@@ -62,7 +62,6 @@ import pytest
             None,
         ),
         ("ClearAll[f];", None, None, None),
-        ("HoldForm[Times[]]", None, "Times[]", None),
     ],
 )
 def test_private_doctests_evaluation(str_expr, msgs, str_expected, fail_msg):
--- a/test/builtin/test_evaluation.py
+++ b/test/builtin/test_evaluation.py
@@ -5,7 +5,7 @@ Unit tests from mathics.builtin.evaluation.
 
 
 import sys
-from test.helper import check_evaluation_as_in_cli, session
+from test.helper import check_evaluation_as_in_cli
 
 import pytest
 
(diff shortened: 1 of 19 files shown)


mmatera:
Maybe adding a pytest in test/format?

rocky:
> Maybe adding a pytest in test/format?

Actually this is is possibly a  `mathics.core.convert.sympy` or a `HoldForm` test. I see what went wrong before, the output result should not be evaluated which can be indicated as a parameter in `check_evaluation()`.

mmatera:
> @mmatera I was trying to add a test for `HoldForm[Times[x]] == Times[x]` of a recent PR of yours (from a while ago) but haven't been able to figure out how to do. in the process, I have this lint stuff from the tests.

Maybe as a pytest in `test/builtin/arithmetic/test_basic.py`? Something like
```
("HoldForm[Times[x]]", "Times[x]", "with less than two arguments, Times keeps the functional form")
```
inside `test_multiply` ?

test/builtin/test_evaluation.py
```diff
@@ -62,6 +62,7 @@
             None,
         ),
         ("ClearAll[f];", None, None, None),
+        ("HoldForm[Times[]]", None, "Times[]", None),
```
mmatera, line 65: The case of formatting `Times[c]` was not handled before because it seems entangled with the format rule for Plus. I will fix this in a subsequent PR.

rocky: Ok. Please commit off of this and we'll combine these together.

rocky:
@mmatera is this okay to go in?

mmatera:
Sure. Merge when you feel it is ready.

PR #1298, opened 2025-01-18
More on Builtin.contribute and Builtin.get_functions.

mmatera:
Some missing details that I discovered regarding these methods, and a pytest to test the right behaviour:

* Fix the regular expression used in get_function to support multiline docstrings.
* Fix how the full name of builtins are built in `contribute`  when the class does not have a `context` attribute.
* Add tests for Builtin.contribute

More tests can be added later.

### DIFF
--- a/mathics/core/builtin.py
+++ b/mathics/core/builtin.py
@@ -515,7 +515,12 @@ class Builtin:
         requires = getattr(self, "requires", [])
         return None if check_requires_list(requires) else UnavailableFunction(self)
 
-    def get_option_string(self, *params):
+    def get_option_string(self, *params) -> Tuple[str, BaseElement]:
+        """
+        Return a tuple of a `str` representing the option name,
+        and the proper Mathics value of the option.
+        If the value does not have a name, the name is None.
+        """
         s = self.get_option(*params)
         if isinstance(s, String):
             return s.get_string_value(), s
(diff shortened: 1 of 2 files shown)


mathics/core/builtin.py
```diff
@@ -468,7 +472,15 @@ def get_functions(self, prefix="eval", is_pymodule=False):
...
                 else:
-                    m = re.match(r"[(]([\w,]+),[)]\:\s*(.*)", pattern)
+                    # TODO: consider to use a more sophisticated
+                    # regular expression, which handles breaklines
+                    # more properly, that supports format names
+                    # with contexts (context`name) and be less
+                    # fragile against leaving spaces between the
+                    # elements.
+                    m = re.match(
+                        r"[(]([\w,]+),[ ]*[)]\:\s*(.*)", pattern.replace("\n", " ")
```
rocky, line 482: I don't understand exactly which newlines in docstirng are getting removed.

If not, please show this in the pytest.

test/core/test_builtin.py
```diff
@@ -0,0 +1,63 @@
...
+        def eval_upvalue(self, expr, x, evaluation):
+            """expr: G[TestBuiltin[x_:Symbol]]"""
+            return
+
+        def format_parm1(self, expr, x, evaluation):
+            """(CustomForm,): expr: F[x_:Symbol]"""
+            return
+
+        def format_parm2(self, expr, x, y, evaluation):
+            """(MakeBoxes, ):expr: G[x_:Symbol,
```
mmatera, line 42: @rocky, the example is this one. If the newline is kept, the regular expression matches on the first line, and the rest is discarded, even if the `re.MULTILINE` parameter is given.

rocky: Got it. Thanks for the information and clarification.

rocky:
LGTM

mathics/core/builtin.py
rocky: Thanks for adding the comments in the pytest.

PR #1302, opened 2025-01-18
Limit TimeConstrained tests if unpatched stopit

rocky:
Some of the `TimeConstrained[]` tests do not work if we have the last PyPI distributed stopit.


mmatera (commented):
LGTM

PR #1307, opened 2025-01-21
Moving code from mathics.builtin.drawing to mathics.eval.drawing. First pass

mmatera:
This is a lot, but just moving code around. Doing this I found a lot of duplicated code, functions that do too much, and a long etc that should be handled later.


rocky:
Nice! LGTM

PR #1309, opened 2025-01-21
Moving evaluation code from mathics.builtin.makeboxes to mathics.eval.makeboxes

mmatera:
This goes in the same line as #1307. Essentially, moving code from mathics.builtin.makeboxes to `mathics.eval.makeboxes`, and reorganize the code inside `mathics.eval.makeboxes` as submodules.


rocky:
LGTM

PR #1314, opened 2025-01-24
Doc updates part7

rocky:
Subfactorial is not a mpmath function, SymPy only. SymPy can't handle machine reals that are not integers. Add rule to handle lists. 

Various summaries have been gone over to have them start with an active verb when possible.


mmatera (approved):
LGTM

PR #1315, opened 2025-01-24
Use context-stripped names in str(Expression)

rocky:
In the str(Expression), show `Head` as a context-stripped name, e.g. `Plus` instead of ``System`Plus``.   We see these long names in `TraceEvaluation` and in error messages. With this change we now will see: 

```
In[1]:= TraceEvaluation[1+2]
  Evaluating: Plus[1, 2]
  Returning: Plus[1, 2] = 3
Out[1]= 3
```

instead of: 

```
In[1]:= TraceEvaluation[1+2]
  Evaluating: System`Plus[1, 2]
  Returning: System`Plus[1, 2] = 3
Out[1]= 3
```

which is what I think is desired most of the time. In the rare occasions where one wants the full context path, `repr()` shows that. 

Also, in the future I think we will have better debugging features that can fill out this information on demand.


mathics/core/expression.py
```diff
@@ -765,8 +765,12 @@ def get_elements(self) -> Sequence[BaseElement]:
     def get_head(self):
         return self._head
 
-    def get_head_name(self):
-        return self._head.name if isinstance(self._head, Symbol) else ""
+    def get_head_name(self, short=False) -> str:
```
mmatera, line 768: This is OK

mathics/core/expression.py
```diff
@@ -317,7 +317,7 @@ def __repr__(self) -> str:
 
     def __str__(self) -> str:
         return "%s[%s]" % (
-            str(self.head),
+            self.get_head_name(short=True),
```
mmatera, line 320: This could be a problem if there are symbols with the same name in different contexts, which is something that you would want to detect in debugging.

rocky: `repr` still shows the long names.  

Personally in my debugging, I don't recall needing this yet.  And in WMA I am seeing short names used in errors and tracing. 

We could add a user-defined setting, but this is more complicated.

mmatera:
Maybe this could be useful :
https://github.com/Mathics3/mathics-core/tree/makeboxes_refactor_2025_step_0p5

The idea is to have a function that produces the "FullForm" of an expression, which takes into account the current context and context path.

mathics/core/expression.py
rocky: This is useful and related, but different.  

This is a useful Mathics3 user function.  But in addition, there is always the need for lower-level displays of elements. These appear in error messages as well as in that TraceEvaluation.

What I've been noticing is that WMA does not fill out context paths when reporting error messages and we do. I can't find a good example of this but I am sure you've seen experienced these.

mmatera: Yep, because uses "FullForm".

rocky:
> Maybe #1316 could be useful. The idea is to have a function that produces the "FullForm" of an expression, which takes into account the current context and context path.
> 
> Instead of `str(expr)` you could use `eval_makeboxes_fullform(expr, evaluation).boxes_to_text(evaluation)` and then the context names are should just if they are not in the ContextPath.

If you can find evidence of WMA doing something like this in its trace routines like `Trace[]` or `TraceEvaluation[]`, I would be interested to learn of this.

mathics/core/expression.py
rocky: Ok - then this can wait until after we have boxing under control.

mmatera:
Let`s see if this helps. This is what happends in WMA, if I interrupt the evaluation with Ctrl+C:


```
In[1]:= p`s:=Sin[8*t];t:=N[Pi/4,10];  Do[Pause[4]; Print[F[s,p`s,p`r]+s],{i,10}]                                                                                              
^C
Interrupt> ?
??
Your options are:
	abort (or a) to abort current calculation
	continue (or c) to continue
	exit (or quit) to exit Mathematica
	inspect (or i) to enter an interactive dialog
	show (or s) to show current operation (and then continue)
	trace (or t) to show all operations

Interrupt> i  

Interrupt::dgbgn: Entering Dialog; use Return[] to exit.

 In[2]:= p                                                                                                                                                                    

 Out[2]= p

 In[3]:= p`s                                                                                                                                                                  

              -9
 Out[3]= 0. 10

 In[4]:= t                                                                                                                                                                    

 Out[4]= 0.7853981634

 In[5]:= Return[]                                                                                                                                                             

Interrupt::dgend: Exiting Dialog.
              -9
s + F[s, 0. 10  , p`r]
              -9
s + F[s, 0. 10  , p`r]
              -9
s + F[s, 0. 10  , p`r]
^C
Interrupt> t
evaluating Print[]
evaluating Print[]
evaluating Plus[,]
evaluating Plus[,]
evaluating F[,,]
evaluating F[,,]
evaluating s
evaluating s
evaluating Sin[]
evaluating Times[,]
evaluating Times[,]
evaluating t
evaluating t
evaluating N[,]
evaluating Times[,]
evaluating Times[,]
evaluating Power[,]
evaluating Power[,]
evaluating Power[,]
evaluating Power[,]
evaluating Power[,]
evaluating
evaluating
evaluating
evaluating Times[,]
evaluating Times[,]
evaluating Times[,]
evaluating Times[,]
evaluating Times[,]
evaluating Times[,]
evaluating Times[,]
evaluating Times[,]
evaluating N[,]
evaluating N[,]
evaluating N[,]
numerically evaluating Times[,]
numerically evaluating Times[,]
numerically evaluating <<big real number>>
evaluating Times[,]
evaluating Times[,]
evaluating Sin[]
evaluating Sin[]
evaluating Sin[]
numerically evaluating Pi
numerically evaluating Pi
evaluating F[,,]
evaluating F[,,]
evaluating F[,,]
evaluating F[,,]
evaluating F[,,]
evaluating Plus[,]
evaluating Plus[,]
evaluating Plus[,]
evaluating Plus[,]
evaluating Plus[,]
evaluating Print[]
evaluating Print[]
evaluating Print[]
evaluating $Output
evaluating $Output
evaluating List[]
evaluating List[]
evaluating List[]
formatting Plus[,]
formatting s
formatting F[,,]
formatting F
formatting s
formatting <<big real number>>
formatting bignum
formatting integer
formatting Times[,]
formatting 
formatting 0.
formatting Power[,]
formatting 10
formatting -9
formatting r
formatting Plus[,]
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
Puts
              -9
s + F[s, 0. 10  , p`r]
evaluating CompoundExpression[,]
evaluating CompoundExpression[,]
evaluating CompoundExpression[,]
evaluating CompoundExpression[,]
evaluating CompoundExpression[,]
evaluating Pause[]
evaluating Pause[]
evaluating Pause[]
evaluating Pause[]
evaluating Pause[]
```
In any case, it is not a great debugger either...

mmatera:
Notice that in the example "Show" shows the FullForm of the current expression, and "Trace" shows the variables without the context.

rocky:
> Notice that in the example "Show" shows the FullForm of the current expression, and "Trace" shows the variables without the context.

In neither case am I seeing `System` or `Global` which is what this was mostly about. 

Thanks for the information. I am sure in the future we will be able to provide this and much more. 

But now is probably not the time to add this. I was looking for small non-controversial  user-facing issues.

mmatera:
image

This is the debugger in the notebook interface. Indeed, it does not show the context of the symbols, which is weird, but it is what it is...

PR #1316, opened 2025-01-24
Independent FullForm

mmatera:
This is useful to handle the FullForm outside the evaluation process, for  example, in debugging, or as a last resource with the standard evaluation process fails.


rocky:
Let's defer this until after a release, which I think may happen this weekend.

mmatera:
Sure. I just put this here in relation with the other PR.

mathics/eval/makeboxes/formatvalues.py
```diff
@@ -183,14 +183,10 @@ def do_format_rational(
     if not isinstance(element, Rational):
         return None
     if form is SymbolFullForm:
```
rocky, line 185: I realize this is not part of the changed code, but I have this nagging feeling that code like `if form is ...`  is not the right way ultimately we should be approaching this.

mmatera: @rocky, feel free to push over all the changes you find useful.

rocky: Right now, I'd like to defer work on this kind of code until a better handle on MakeBox migration.

rocky:
It would be very useful to add some tests that exercise and show in isolation what is going on here.

mathics/eval/makeboxes/formatvalues.py
rocky: I guess now is the time for this. Thanks for waiting.

rocky:
> It would be very useful to add some tests that exercise and show in isolation what is going on here.

@mmatera Is it correct that there is no user-facing change that is observed, but instead these changes are purely organizational and internal?

rocky:
LGTM

mmatera:
Yep, it is just a way to decouple Full form of the other formatting. This helps to debug the other changes.

PR #1321, opened 2025-01-28
Fixes for compatibility with Python 3.13

mmatera:
This PR contains the fixed needed to make Mathics compatible with Python 3.13.


mathics/builtin/files_io/importexport.py
```diff
@@ -1060,16 +1060,16 @@ class RegisterImport(Builtin):
...
-     | 0.711009	0.687453
-     | 0.246540	0.433973
-     | 0.926871	0.887255
-     | 0.825141	0.940900
-     | 0.847035	0.127464
-     | 0.054348	0.296494
-     | 0.838545	0.247025
-     | 0.838697	0.436220
-     | 0.309496	0.833591
+     | 0.629452    0.586355
```
mmatera, line 1063: In docstrings, use 4 spaces instead of a tab character.

.github/workflows/ubuntu.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.12', '3.11', '3.8', '3.9', '3.10']
+        python-version: ['3.13', '3.12', '3.11', '3.8', '3.9', '3.8']
```
mmatera, line 14: Should keep 3.8? Maybe 3.8,  3.10, 3.12 and 3.13 would be enough?

mathics/doc/doc_entries.py
```diff
@@ -93,7 +93,7 @@
     r"""(?mx)^  # re.MULTILINE (multi-line match)
                 # and re.VERBOSE (readable regular expressions
         ((?:.|\n)*?)
-        ^\s+([>#SX])>[ ](.*)  # test-code indicator
+        ^\s*([>#SX])>[ ](.*)  # test-code indicator
```
mmatera, line 96: In Python 3.13, doctrings are trimmed of "margin" spaces.

mathics/doc/doc_entries.py
```diff
@@ -216,10 +216,7 @@ def parse_docstring_to_DocumentationEntry_items(
         logging.warning("``key_part`` is deprecated. Its value is discarded.")
 
     # Remove commented lines.
-    doc = filter_comments(doc).strip(r"\s")
-
-    # Remove leading <dl>...</dl>
-    # doc = DL_RE.sub("", doc)
+    doc = filter_comments(doc)  # .strip("\s")
```
mmatera, line 219: I do not remember the reason of the r"\s" strip, but it breaks some tests now.

mathics/doc/doc_entries.py
```diff
@@ -394,11 +391,16 @@ def compare_out(self, outs: tuple = tuple()) -> bool:
             # Mismatched number of output lines, and we don't have "..."
             return False
 
+        # Python 3.13 replaces tabs by a single space in docstrings.
```
mmatera, line 394: Python 3.13 converts tab characters in docstrings to simple spaces. This "hack" helps to keep the doctests compatible with this change.

.github/workflows/ubuntu.yml
rocky: I am for dropping 3.8.

mmatera: OK, so ['3.13', '3.12', '3.11', '3.10', '3.9'] in all the workflows?

rocky:
Note: Supporting 3.13 may push us into using a newer pip that disallows the `find_namespace_packages` that is needed right now in the Mathics3 Modules. 

Keep in mind, getting everything working on 3.13 where we can _release_ might be a bigger undertaking than just getting the Mathics Kernel adjusted.

mmatera:
OK, this is just to adjust the kernel compatibility, and does not breaks the compatibility with previous versions.

.github/workflows/ubuntu.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.13', '3.12', '3.11', '3.8', '3.9', '3.8']
+        python-version: ['3.13', '3.12', '3.10', '3.8']
```
rocky, line 14: Dod you mean to change 38 to 3.9?

mmatera: I was saying to keep the even versions, plus the last one (if odd).

mmatera: 3.8 is the version supported by the last version of Pyston.

rocky: I think we should drop Pyston then in its standalone form. There is still a pyston-lite plugin. 

I don't see a reason to drop version  that end in an odd number. Python does not say that minor numbers that end in odd numbers have any less status than those with even numbers. Dropping 3.8 makes sense because the Python language drifts and it is extra effort to support older Pythons.

mmatera: It is a pitty because Pyston is much more faster than CPython and Pyston-lite, but OK. In any case, we still are compatible with 3.8

rocky:
I don't have a problem with the idea or the code, but the timing of doing this right now is weird. 

I've just spent the last 3 days if not longer trying to get out a release ensuring all the parts work together. (And I am not sure they completely do).  Something like this now ensures things will be a little bit out of whack again and puts us back in the state of a little off-ness. Since I don't think some of the code will be able to handle 3.13 unless older pip is used. 

And of course there are other PRs in the backlog. These change the API which will also require changing some other bits of code outside of this repository.

To accommodate moving forward which we need to do and keeping some stability for a little while right now, I suggest we have an "unstable" branch that we merge to, along with the master branch which is compatible across other projects.

Your thoughts?

rocky:
Hmm. We could split this up into two parts, isolate that version testing change part and everything else. So that when we are ready to switch over _everything_ there is a small bit of code in this repository that does that.

mmatera:
> I don't have a problem with the idea or the code, but the timing of doing this right now is weird.

As with much of the other stuff I did, the timing depends more on finding the time to do it than on achieving a deadline or a particular general goal. 

> 
> I've just spent the last 3 days if not longer trying to get out a release ensuring all the parts work together. (And I am not sure they completely do). Something like this now ensures things will be a little bit out of whack again and puts us back in the state of a little off-ness. I don't think some of the code will be able to handle 3.13 unless an older pip is used. And code in other repositories may need additional adjusting too.

The changes here (except for the tests) ensure that mathics-core tests still pass in Python 3.13. This is not a release or something that breaks the rest of the code. The changes do not seem to affect the behavior of the rest of the components, but in six months or a year, when we want to start the next release, they would make the process simpler.

> 
> And of course, there are other PRs in the backlog. These change the API which will also require changing some other bits of code outside of this repository.
> 
> To accommodate moving forward which we need to do and keep some stability for a little while right now, I suggest we have an "unstable" branch that we merge to, along with the master branch which is compatible across other projects.
>

That sounds reasonable. 
 
> Your thoughts?

rocky:
> As with much of the other stuff I did, the timing depends more on finding the time to do it than on achieving a deadline or a particular general goal.
> 

This is a problem that can be solved by working in a more coordinated way. 

From the project's perspective, it is better to have a plan for moving forward that we agree on and discuss.

You put in a [checklist of ideas for the 8.0.0 release](https://github.com/Mathics3/mathics-core/issues/) which has been deferred. Items in that list were picked precisely not to be disruptive. 

And we have [open issues](https://github.com/Mathics3/mathics-core/issues/). I suppose we should add upgrading to 3.13 to that list, but 3.13 was not on that list. Upgrading to a newer Python release, especially with the way Python and its ecosystem move nowadays, is disruptive. 

And if you have a question about doing something you could just mention it or ask about it in advance.

mmatera:
> > As with much of the other stuff I did, the timing depends more on finding the time to do it than on achieving a deadline or a particular general goal.
> 
> This is a problem that can be solved by working in a more coordinated way.
> 
> From the project's perspective, it is better to have a plan for moving forward that we agree on and discuss.
> 
> You put in a [checklist of ideas for the 8.0.0 release](https://github.com/Mathics3/mathics-core/issues/). Some of these have been deferred. Items in that list were picked precisely not to be disruptive.
> 
> And we have [open issues](https://github.com/Mathics3/mathics-core/issues/). I suppose we should add upgrading to 3.13 to that list, but 3.13 was not on that list. Upgrading to a newer Python release, especially with the way Python and its ecosystem move nowadays, is disruptive.
> 
> And if you have a question about doing something you could just mention it or ask about it in advance.

OK, but the changes in this PR are closely related to the work done before the release, relative to adjusting the `get_function` method and the documentation building. Even for the currently supported versions, these changes make the code more robust. But OK, I will split this in two parts, and leave this for later.

BTW, the support for LaTeX expressions in the online Mathics-Django using MathJax, something that was in the list for the pre-release, is done: https://github.com/Mathics3/mathics-django/pull/221

rocky:
> > > As with much of the other stuff I did, the timing depends more on finding the time to do it than on achieving a deadline or a particular general goal.
> > 
> > 
> > This is a problem that can be solved by working in a more coordinated way.
> > From the project's perspective, it is better to have a plan for moving forward that we agree on and discuss.
> > You put in a [checklist of ideas for the 8.0.0 release](https://github.com/Mathics3/mathics-core/issues/). Some of these have been deferred. Items in that list were picked precisely not to be disruptive.
> > And we have [open issues](https://github.com/Mathics3/mathics-core/issues/). I suppose we should add upgrading to 3.13 to that list, but 3.13 was not on that list. Upgrading to a newer Python release, especially with the way Python and its ecosystem move nowadays, is disruptive.
> > And if you have a question about doing something you could just mention it or ask about it in advance.
> 
> OK, but the changes in this PR are closely related to the work done before the release, relative to adjusting the `get_function` method and the documentation building. 

You may see it that way, but as the discussion I hope shows, it isn't like that. There could be a lesson here: mention first, code later. 


> Even for the currently supported versions, these changes make the code more robust.

Again, it is not about whether certain changes are good, but more about coordinating, timing, and appropriateness for the situation right now.

>  But OK, I will split this in two parts, and leave this for later.

Thank you! 

> BTW, the support for LaTeX expressions in the online Mathics-Django using MathJax, something that was in the list for the pre-release, is done: [Mathics3/mathics-django#221](https://github.com/Mathics3/mathics-django/pull/221)

Sorry, I lost track of this. In my last look, you mentioned that MathJaX needed to be updated. I see that's been done now.

Looking at it and trying now. Thanks for doing this.

rocky:
@mmatera - thanks for this!

PR #1323, opened 2025-01-29
More robust docstring parsing

mmatera:
In the current (master) documentation/doctest implementation, it is assumed that all the docstrings have a "margin" of spaces, comming from the class indentation. Then, the regular expression looking for doctests fails to find them if there is not at least a single space at the beginning of the line.
In the future, when we want to give support to Python 3.13, which removes the "left margin" of docstrings, the regular expression will fail finding some of the tests. 
This PR changes the regular expression to allow doctests which do no start with a space.

Also, in future versions of Python, tab characters in docstrings will be converted into spaces, which will make to fail some tests. In this PR, tabs in doctests are replaced by sequences of 4 spaces, making the tests more robust under this kind of changes.


rocky:
LGTM - thanks.

PR #1324, opened 2025-01-29
A couple of doc tagging corrections

rocky:


mmatera (approved):
LGTM

mathics/builtin/numbers/calculus.py
```diff
@@ -984,18 +986,20 @@ class Integrate(SympyFunction):
...
-    > Integrate[-Infinity, {x, 0, Infinity}]
+    >> Integrate[-Infinity, {x, 0, Infinity}]
      = -Infinity
 
-    > Integrate[-Infinity, {x, Infinity, 0}]
-     = Infinity
+    Integrating something ill-defined returns the expression untouched:
 
-    Integration in TeX:
+    >> Integrate[1, {x, Infinity, 0}]
```
rocky, line 994: Change from `-Infinity` to `1` because we pull out the `-Infinity` from inside the integral to outside, whereas WMA does not do this.  

While this might be an interesting curiosity to ponder, and whether we should match WMA, it is not instructive and does not belong in the doc.  I had also considered moving this into pytest, but felt that as long as there some motivation for doing this it is okay to leave in. Maybe even better, since this gets rendered by MathML.

PR #1326, opened 2025-01-30
Improve cone rendering plus some lint...

rocky:
Surface bottom added to cone; cone orientation corrected.


rocky:
The opacity of the base should probably be changed to 0.9, but I will leave that for later.

PR #1327, opened 2025-01-31
improve LaTeX processing. Fix inconsistencies in the docstring format

mmatera:
This is another pass through the documentation system. The PR includes
* Fix inconsistencies in the docstrings (missing '$', escape characters, etc)
* Remove the hardcoded code that handles special characters from the `DocumentEntry.latex()`. Almost all special characters are now handled by the mathicsscanner tables.
* Increase a little bit the consistency in using the `$` character in docstrings. Now it must be explicitly escaped.
* The docstring of `escape_latex` function was improved.
* Expressions of the form `'Head'[$var_1$,$var_2$]`   now are formatted as expected (`'Head'` as a code variable, and `$var_1$` as a LaTeX expressions).

### DIFF
--- a/mathics/builtin/system.py
+++ b/mathics/builtin/system.py
@@ -76,15 +76,17 @@ class Breakpoint(Builtin):
 
     def eval(self, evaluation: Evaluation):
         "Breakpoint[]"
-
-        breakpoint()
+        # I have to comment out this to pass the
+        # "precommit". @rocky, any idea about how to
+        # remove the comment?
+        # breakpoint()
 
 
 class CommandLine(Predefined):
     """
-    <url>:WMA link:https://reference.wolfram.com/language/ref/$CommandLine.html</url>
+    <url>:WMA link:https://reference.wolfram.com/language/ref/\\$CommandLine.html</url>
     <dl>
-    <dt>'$CommandLine'
+    <dt>'\\$CommandLine'
       <dd>is a list of strings passed on the command line to launch the Mathics3 session.
     </dl>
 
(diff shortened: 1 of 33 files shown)


rocky:
A brief look at this and it looks good. But I'd like to try it out in detail tomorrow. 

I'm thinking that after all of the documentation stuff is fixed we might make a minor release of Mathics Kernel. 

Your thoughts?

mmatera:
imagen


This is how Cuboid help looks in the PDF documentation after these changes.

mmatera:
@rocky, I agree: after finishing with the documentation, we can move to a 8.0.1 release or something like that. 
Still, there is some work to do on the docstrings by replacing
`<dt>'Head[$x_1$,$x_2$]'`
to
`<dt>'Head'[$x_1$,$x_2$]`
in a way that the `$` are interpreted as LaTeX delimiters instead of a character, and to ensure that Symbols are tagged with as `'Symbol'` instead of `$Symbol$`. But this is already too long.

mathics/builtin/system.py
```diff
@@ -76,15 +76,17 @@ class Breakpoint(Builtin):
 
     def eval(self, evaluation: Evaluation):
         "Breakpoint[]"
-
-        breakpoint()
+        # I have to comment out this to pass the
+        # "precommit". @rocky, any idea about how to
+        # remove the comment?
+        # breakpoint()
```
mmatera, line 82: @rocky, please check if you are able to uncomment this without complains from precommit.

rocky: See #1328

rocky:
LGTM. Tried it and it works great!

mmatera:
@rocky, before the before the new release, please give me the chance to pass over all the documentation to adjust the quotations in <dt> fields.

rocky:
> @rocky, before the before the new release, please give me the chance to pass over all the documentation to adjust the quotations in fields.

No problem. Also, Look form places were we have `$x1$`,  `$x2$,  `$xn$`,  etc. and change these to `$x_1$, $x_2$, $x_n$ respectively.  See https://github.com/Mathics3/mathics-core/blob/master/mathics/builtin/drawing/graphics3d.py#L205 for an example.

PR #1329, opened 2025-01-31
More about latex in online doc

mmatera:
This PR has the first bunch of changes in the <dd> field of docstrings to mark as `code` just the head of expressions which has arguments in LaTeX form.
There are many changes, so let's split this in several bunches.


rocky:
LGTM

PR #1330, opened 2025-01-31
Dd fixes 02

mmatera:
Another bunch. Here I also include the change like $x1$ ->$x_1$


mmatera:
@rocky, these are many changed files, but all the changes are trivial. If you agree, I will check that documentation compiles until this point, and then I will merge this to continue with the other bunches.

rocky:
Yes just keep making more changes like this.

PR #1341, opened 2025-02-01
Method to convert documentation entries into RsT 

mmatera:
This supports the basic features to convert a docstring into an RsT documentation entry.

### DIFF
--- a/mathics/doc/doc_entries.py
+++ b/mathics/doc/doc_entries.py
@@ -9,6 +9,7 @@ system, and the functions used to parse docstrings into these objects.
 
 import logging
 import re
+from abc import ABC
 from os import getenv
 from typing import TYPE_CHECKING, Callable, Dict, List, Optional, Sequence, Tuple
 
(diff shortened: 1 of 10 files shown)


mathics/doc/doc_entries.py
```diff
@@ -427,7 +426,12 @@ def key(self, value):
 
     def rst(self, test_data=None) -> str:
         result = ">>> " + self.test + "\n"
-        result += "  = " + "\n    ".join(line for line in self.result.split("\n"))
+        if self.result is None:
+            return result
+
+        result += "  = " + "\n    ".join(
+            line for line in cast(str, self.result).split("\n")
```
rocky, line 433: Why is this cast needed?

mmatera: because `Test.result` is optional. I have to improve this interface eventually...

rocky: Please leave the error then as a reminder to fix rather than silence this warning.

rocky: There are rare instances where casts are valid. This is not one of them.

mmatera: > Please leave the error then as a reminder to fix rather than silence this warning.

I am fixing it now. On the other hand, it is not hard to find places where we use `cast`, so it is easy to find these cases when we focus on fixing it.

mmatera:
At this point, we start to be in a position to produce documentation in a Sphinx-compatible form. 

This is what I get when I setup and run Sphinx:

imagen

rocky:
Ok. I guess we'll be able to do a conversion sometime this year. 

For now, I think we want to get out an 8.0.1 revision and then start on boxing. Right?

mmatera:
> Ok. I guess we'll be able to do a conversion sometime this year.
> 
> For now, I think we want to get out an 8.0.1 revision and then start on boxing. Right?

Yep, the idea is to have the mathics-core codebase ready to use in another repo focused on user documentation. Once RsT conversion works properly, I hope to be able to remove `latex_doc.py` and `doc2latex.py` from mathics-core and from the entire universe...

mmatera:
The development of the code in this branch continues in the new repo

https://github.com/Mathics3/Mathics3-user-guide

The plan is to slowly improve the parser to RsT until produces a Sphinx compliant documentation. Then we can see to move the generation of PDF documentation to that project, using sphinx.

PR #1344, opened 2025-02-01
not show Out[]=Null as in wolframscript

mmatera:
In the WMA CLI, when an expression evaluates to ```System`Null```, there is no `Out[]=...` line. This PR reproduces this behavior.


rocky:
LGTM

PR #1350, opened 2025-02-03
use eval_Close, and some fs lint

rocky:
Now that we have more `mathics.eval`  `eval_...`  functions, there are a probably a lot more opportunities like this one involving `Close()`.  

While I am not a fan of petty optimization, `eval_Close(...)` is simpler and clearer than `Expression(SymbolClose, ...).evaluate(...)`

And if you are tracing Evaluations, calling `Expression()`  as historically done, also pollutes tracing, along with its inherent slowness and complication. 

If you _do_ want see see the close() calls, you can always trace the `mathics.eval`  `eval_xxx` functions.


mmatera (approved):
Great! LGTM

PR #1352, opened 2025-02-03
Sort methods in DocElement classes. Add a base class.

mmatera:
This is the part of #1341 that I would like to include in the release. Apart from a small tidy-up of the order of the methods and where are stored some pattern constants, the proposed change allows to process the documentation in a more similar fashion as Sphinx does, without limit of "levels".  
The rest of the code in #1341 can be moved to another repo to be worked out independently.


rocky:
LTM

PR #1354, opened 2025-02-03
removing <CON>...</CON> in favor of <CONSOLE>...</CONSOLE> and single quote codemarks

mmatera:
Among other hacks, the homebrew documentation system introduces some *custom* tags, that make the parsing process unnecessarily complex. A couple of them are `<CONSOLE>content</CONSOLE>` and `<CON>content</CON>`. Both means more or less the same as using `'content'`.
This PR removes the second form, with the idea of removing the first form too, in favor of using '``' or  
```'.. code-block:: console```, following markdown or ReStructuredText conventions.

Also, `mathics/doc/latex/1-Manual.mdoc` was removed, which is just a copy of `mathics/doc/documentation/1-Manual.mdoc`


rocky:
LGTM

PR #1356, opened 2025-02-03
Last changes in documentation system

mmatera:
I hope, this are the last changes to do with the documentation system. As always, trying to fix something simple other things come up. But I guess I cover all of them:
* Procesing recurrent replacements in `post_sub`.
* Removing the non-existing `<imgpng>` tag, and replacing it by the existing but not used .
* Fixing remaining docstrings containing  '$Symbol'


rocky:
LGTM

PR #1358, opened 2025-02-04
Note Mathics Live and Trepan Debugger

rocky:


rocky:
In the next version of the tutorial and guide, we should say something about Mathics3 live and the notebook front-ends.

What exactly? I am not sure. Here is a draft. Please change or comment as needed.

mmatera (approved):
LGTM

PR #1361, opened 2025-02-05
Remove double loading of rules, from times of apply_ methods

mmatera:
This PR removes a block of code that we keep from the transition from `apply_` to `eval_` methods, which makes that FunctionRules be loaded twice.

### DIFF
--- a/mathics/core/builtin.py
+++ b/mathics/core/builtin.py
@@ -305,18 +305,6 @@ class Builtin:
                     system=True,
                 )
             )
-        for pattern, function in self.get_functions(is_pymodule=is_pymodule):
-            pat_attr = attributes if pattern.get_head_name() == name else None
-            rules.append(
-                FunctionApplyRule(
-                    name,
-                    pattern,
-                    function,
-                    check_options,
-                    attributes=pat_attr,
-                    system=True,
-                )
-            )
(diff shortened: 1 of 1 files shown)


rocky:
I don't understand the sequence of events that caused this to be a duplicate, or this code, but if it helps then LGTM.

Also, we should remove that TimeConstrained tests that seem to be constantly failing. Those tests need to be rethought and made more robust.

mmatera:
> I don't understand the sequence of events that caused this to be a duplicate, or this code, but if it helps then LGTM.
> 
At some point we wanted to load rules from `eval_` methods and from the default `apply_` methods. Then, we changed the default to `eval_`, so we are loading `eval_` twice. 

> Also, we should remove that TimeConstrained tests that seem to be constantly failing. Those tests need to be rethought and made more robust.

I agree. Let's disallow them until a better test comes up.

rocky:
> > I don't understand the sequence of events that caused this to be a duplicate, or this code, but if it helps then LGTM.
> 
> At some point we wanted to load rules from `eval_` methods and from the default `apply_` methods. Then, we changed the default to `eval_`, so we are loading `eval_` twice.
> 

Thanks for clearing up this mystery!

PR #1368, opened 2025-02-05
Update expintegral.py

mmatera:
x


rocky:
BTW I also see that `Arg[z, method, option]` needs fixing.

PR #1370, opened 2025-02-06
Add links to DLMF and add some more plots

rocky:


mathics/builtin/specialfns/bessel.py
```diff
@@ -1,5 +1,9 @@
 """
 Bessel and Related Functions
+
+See also <url>
+:Chapter 10 Bessel Functions in the Digital Library of Mathematical Functions:
+https://dlmf.nist.gov/10</url>.
```
mmatera, line 6: Good idea! NIST DLMF is a great source of information!

mmatera (approved):
LGTM

mathics/builtin/specialfns/bessel.py
rocky: What I think is going on is that the DLMF structure informed the maxima organization which informed the Mathematica organization which also probably informs the sympy organization.

PR #1371, opened 2025-02-06
Fix handling subsection tags in input files

mmatera:
Another tiny fix in the code that process documentation. This handles the `<subsection></subsection>` tags, like in Mathics-Django documentation.


rocky:
LGTM

PR #1374, opened 2025-02-07
Adjust for type mismatch...

rocky:
Correct for mismatch between ListExpression and tuple


rocky:
While deeper stuff for supporting PacletManager will wait until after release 8.0.1  (add `$LibraryPath` variable, add `On`,  and `Off`, and reorganize Compilation to put under a "Calling External Programs" guide section, moving "Run" into this section,  and adding more), the code in this PR feels like a long-standing outright bug that occurs when running `Get` on `PacletManager.m`

PR #1379, opened 2025-02-10
Bug: an association was confused with an option

rocky:
Union, Complement and Intersection allow a SameTest option. This has to be matched not as a general options association, but instead as a SameTest association

### DIFF
--- a/mathics/builtin/list/rearrange.py
+++ b/mathics/builtin/list/rearrange.py
@@ -1424,7 +1424,12 @@ class Union(_SetOperation):
      = {1, 2, 2, 3, 4}
     """
 
-    attributes = A_FLAT | A_ONE_IDENTITY | A_PROTECTED | A_READ_PROTECTED
+    # FIXME: WMA add the A_FLAT attribute, but that messes up function parsing.
+    # In particular:
+    #   Union[{1, -1, 2}, {-2, 3}, SameTest -> (Abs[#1] == Abs[#2] &)]
+    # is parsed/passed as
+    #   Union[{1, -1, 2}, SameTest -> (Abs[#1] == Abs[#2] &)]
+    attributes = A_ONE_IDENTITY | A_PROTECTED | A_READ_PROTECTED
     summary_text = "union distinct elements of list(s) or association(s)"
     _operation = "union"
 
--- a/mathics/builtin/list/rearrange.py
+++ b/mathics/builtin/list/rearrange.py
@@ -435,7 +435,7 @@ class _SetOperation(Builtin, ABC):
         return self.eval(lists, evaluation, SymbolSameQ)
 
     def eval(self, lists, evaluation, sametest):
-        "%(name)s[lists__, SameTest->sametest_:SameQ]"
+        "%(name)s[lists__, SameTest->sametest_]"
 
         seq = lists.get_sequence()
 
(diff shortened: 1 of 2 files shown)


rocky:
@mmatera I have been looking over the Boxing proposal. In that I thought it would be cool if we could start testing Mathics3 builtin functions like this using MUnit  rather than pytest.  In looking over what it would take to get MUnit working, I came across this bug in trying to Union two associations.

mmatera:
@rocky, sure. Could you provide me some examples or a pattern to reproduce?

rocky:
> @rocky, sure. Could you provide me some examples or a pattern to reproduce?

https://github.com/Mathics3/mathics-core/blob/b045d5d63bef35e8d49196d2e882dfd253073ff9/mathics/builtin/list/rearrange.py#L1408-L1411

mmatera:
> > @rocky, sure. Could you provide me some examples or a pattern to reproduce?
> 
> https://github.com/Mathics3/mathics-core/blob/b045d5d63bef35e8d49196d2e882dfd253073ff9/mathics/builtin/list/rearrange.py#L1408-L1411

I mean, about how to use MUnit.

rocky:
> I mean, about how to use MUnit.

The broader idea is to start doing things the same way if we were using WMA. I don't know in great detail how one writes tests for WMA behavior. However, https://reference.wolfram.com/language/tutorial/UsingTheTestingFramework.html suggests: 

```mathematica
test = VerificationTest[2+2, 4]
```

and `test["Outcome"] == "Success"`.  

Is there a simpler commonly used test methodology used in WMA?  One could write something simple, such as what is done for gs, but if we could use something that is not custom, that would be nice. (Nice, but not strictly necessary). 

But this is a little bit of a digression. There is currently a bug in Union, Complement, and Intersection when working with Associations. And there is a bug in the attributes listed for Union. This PR starts to address these.

mmatera:
> But this is a little bit of a digression. There is currently a bug in Union, Complement, and Intersection when working with Associations. And there is a bug in the attributes listed for Union. This PR starts to address these.

I will try to review this over this day, but at first glance it is OK.

Regarding the test suite, let's continue in #1380

mathics/builtin/list/rearrange.py
```diff
@@ -1388,12 +1402,18 @@ class Union(_SetOperation):
...
-
+    A union of two lists:
     >> Union[{a, b, c}, {c, d, e}]
      = {a, b, c, d, e}
 
+    A union of two associations:
+    >> Union[{a -> b}, {c -> d}]
+     = {a -> b, c -> d}
+
+    A union of one item is the item, however note that the list is sorted:
```
mmatera, line 1413:
```suggestion
    A union of one item is the item, however note that the list is sorted and reduced to have unique elements:
```

mmatera (approved):
LGTM, but please check the comments.

mathics/builtin/list/rearrange.py
```diff
@@ -1404,7 +1424,13 @@ class Union(_SetOperation):
      = {1, 2, 2, 3, 4}
     """
 
-    summary_text = "enumerate all distinct elements in a list"
+    # FIXME: WMA add the A_FLAT attribute, but that messes up function parsing.
```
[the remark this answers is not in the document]
rocky: I have now created issue #1381 to track this problem. But to answer your question, we have a different eval method with a different pattern when there is no `SameTest` option.

mathics/builtin/list/rearrange.py
rocky: There is something that feels weird to me about combining the two ideas in one sentence like this. In a union, it is expected that duplicates are removed. However, it is not expected that the result will be sorted by value. 

I have reworked the existing examples so that there is one example to show just the sorting without duplication removal.  And then a later example can just mention that duplicates are removed in sets one expects to happen in a union operation.

rocky:
If there is still concern about the doctest example comments, feel free to revise in a future PR.

PR #1385, opened 2025-02-13
Small lint-like changes to BinaryRead and Import

rocky:
These items were noticed in looking at what is needed to support some version of PacletManager.  After this there set of changes, there will be more changes to put more Import code, BinaryRead, into `mathics.eval` and to handle `BinaryReadList` which is needed to be able to support Importing gzip files.


mmatera (approved):
LGTM

PR #1386, opened 2025-02-13
Split out eval_Open to mathics.eval.files

rocky:
Isolate a file open evaluation routine used by `Import`, `OpenRead`, and `OpenWrite`.


rocky:
I still have to do an evaluation routine for the combined functions `BinaryRead` and `BinaryReadList` which will happen in a future PR. 

In general, what I am finding is that a major obstacle for using a number of packages has been our I/O weaknesses. And there is just a lot of (boring and tedious) work to do here.

PR #1387, opened 2025-02-14
Fix `Gamma[]` simplification rules.

aravindh-krishnamoorthy:
`Gamma[]` function has the rule:

```mathematica
Gamma[1 + z_] :> z!
```

However, this rule is only valid when `z` is an integer.

The fix changes the rule to apply only to integral `z`.

### DIFF
--- a/mathics/builtin/specialfns/gamma.py
+++ b/mathics/builtin/specialfns/gamma.py
@@ -323,7 +323,7 @@ class Gamma(MPMathMultiFunction):
 
     rules = {
         "Gamma[z_, x0_, x1_]": "Gamma[z, x0] - Gamma[z, x1]",
-        "Gamma[1 + z_]": "z!",
+        "Gamma[1 + z_?IntegerQ]": "z!",
         "Gamma[Undefined]": "Undefined",
         "Gamma[x_, Undefined]": "Undefined",
         "Gamma[Undefined, y_]": "Undefined",


mathics/builtin/specialfns/gamma.py
```diff
@@ -323,7 +323,7 @@ class Gamma(MPMathMultiFunction):
 
     rules = {
         "Gamma[z_, x0_, x1_]": "Gamma[z, x0] - Gamma[z, x1]",
-        "Gamma[1 + z_]": "z!",
+        "Gamma[1 + z_?IntegerQ]": "z!",
```
rocky, line 326:
```suggestion
        "Gamma[1 + z_?NumberQ]": "z!",
```

Factorial is defined for numbers outside of integers. For example 2.5! is 3.32335. 

For those numbers like E, Pi, and I where factorial is not defined, factorial returns the expression unevaluated without an error; and this is the desired behavior we want in this case.

aravindh-krishnamoorthy: Hello, @rocky. This should be `_?IntegerQ` because for non-integer `x` factorial is just defined as `Gamma[1+x]` so there's no point converting back and forth... Whereas for integers, notation-wise, the having `x!` looks good...

aravindh-krishnamoorthy: MMA, e.g., follows this.

rocky: > Hello, @rocky. This should be `_?IntegerQ` because for non-integer `x` factorial is just defined as `Gamma[1+x]` so there's no point converting back and forth... Whereas for integers, notation-wise, the having `x!` looks good...

I just pulled the branch and tried `Gamma[1 + 2.5]` and this code works. So I don't have a strong opinion one way or the other. 

But from a consistency standpoint, if `x!` looks good for integer x, it should also look good for real x as well in my opinion. Or remove both. It is just that having it go one way for integers and the other way for reals seems weird to me.

rocky: > Hello, @rocky. This should be `_?IntegerQ` because for non-integer `x` factorial is just defined as `Gamma[1+x]` so there's no point converting back and forth... Whereas for integers, notation-wise, the having `x!` looks good...

One things I was going to bring up but didn't because I didn't think were were going to go at the end with an integer test:   the preferred way to express this is more simply:  `_Integer`.   When what you are testing is defined type, you can drop the "?".

aravindh-krishnamoorthy, line 326:
```suggestion
        "Gamma[1 + z_Integer]": "z!",
```

aravindh-krishnamoorthy: As you're aware, factorial was first defined for positive integers, and then analytically continued to non-integers and complex numbers via the `Gamma` function. To me, it feels strange to see `x!` notation for non-integral `x`. Also, MMA does not perform this simplification for arbitrary `x`.

```mathematica
In[7]:= FullSimplify[Gamma[1+x]]
Out[7]= Gamma[1+x]
In[8]:= FullSimplify[Gamma[1+x], Element[x, Integers] && x > 0]
Out[8]= x!
```

But, I will leave this open and wait for others to provide an opinion as well.

rocky: Ok - thanks for the further explanation. There is a lot of leeway for matters of taste. And say you say, WMA does not `FuillSimplify` with `Gamma` to factorial when `Integer` is replaced with `Rationals` or `Reals`.  

So aside from the fact that there is a failing test, I am fine with this. @mmatera ?

aravindh-krishnamoorthy: > So aside from the fact that there is a failing test, I am fine with this. @mmatera ?

@rocky on my end all tests are passing:
```python
================== 2870 passed, 364 skipped, 14 xfailed, 16 xpassed, 3 warnings in 131.40s (0:02:11) ===================
```

Could you please let me know which test is failing? Perhaps I can fix it if it's related to this change...

rocky: Look at the actions tab. See for example: https://github.com/Mathics3/mathics-core/actions/runs/13345211275/job/37274817229?pr=1387

aravindh-krishnamoorthy: > Look at the actions tab. See for example: https://github.com/Mathics3/mathics-core/actions/runs/13345211275/job/37274817229?pr=1387

This is a problem because MMA evaluates it to a factorial. I'll take a deeper look...

aravindh-krishnamoorthy:
> > # Introduction
> > `Gamma[]` function has the rule:
> > ```mathematica
> > Gamma[1 + z_] :> z!
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > However, this rule is only valid when `z` is an integer.
> 
> Actually, according to WMA and the example `Product[k, {k, 3, n}]`, it is valid _always_ valid. So what is currently in the code base matches WMA better.
> 
> Previously, we had discussed how this could be expanded to Numbers rather than Integers. But WMA behavior takes this to be rewritten to be valid no matter _what_ `z` is. Go figure.

@rocky as I showed you in the comment before, `Gamma[1 + z_] :> z!` is not the rule used for simplifying the `Product`. This rule seems wrong to me because for any non-integral `z`, factorial is defined via the `Gamma` function. So this would be a back and forth. I think MMA is also wrong wrt handling `Product[k, {k, 3, n}]` and seems to have specialised rules for those to help schoolkids maybe.

aravindh-krishnamoorthy:
@rocky Unfortunately, `Gamma` is still inconsistent with MMA (see below) and this behaviour makes my day-to-day work with Mathics a bit difficult. But, I agree that MMA has many weird quirks (another example is that using 1. causes numerical evaluation similar to older languages like Axiom), and I guess 100% compatibility means these have to be carried forward too...

```mathematica
(* Inconsistent *)
In[1]:= Gamma[1+x]
Out[1]= x!

(* Consistent *)
In[2]:= Product[k, {k, 3, n}]
Out[2]= n! / 2

(* Not (x+1)! *)
In[3]:= Gamma[2+x]
Out[3]= Gamma[2 + x]
```

I'd still vote not to have the generic rule `Gamma[1 + z_] :> z!` but leave the final call up to you.

mmatera:
I was not follow this discussion very closely, but as a general rule, except when there is a very good reason, we go for the most compatible behavior.

aravindh-krishnamoorthy:
#1395 does this better...

PR #1391, opened 2025-02-16
Add `BinaryReadList[]`

rocky:
Add `ReadList` and split off read evaluation routine into  `mathics.eval.io.eval_Read`


rocky:
After this we'll finally have the prerequisites for Importing a gzipped file.

PR #1392, opened 2025-02-18
Clear variable before using it.

rocky:
Clear variable x before using it.


rocky:
This doesn't address the problem I was intending to address (even if it is slightly more precise).

PR #1394, opened 2025-02-18
Go over docs for gamma and no-meaning infix ops

rocky:
Go over docs for `Beta`, `Gamma`,  `Product`, and infix operators with no meaning


mmatera (approved):
LGTM

PR #1406, opened 2025-06-06
Address bug in structure cache `None `accessed as a `dict`

### LINKED ISSUE #1405: structure_cache in `mathics.core.expression.structure` is `None` but we try to set it in `_is_neutral_symbol()`
## Description

Structure cache object is `None`, but we try to set it as though it were a dictionary in `_is_neutral_symbol()`.

```
$ mathics --post-mortem

In[1]:=  Needs["CodeParser`"] (* Note: You should have arranged to have CodeParser in your packages *)

ReplaceAll::reps: PacletInformation[CodeParser] is not a valid replacement rule.
ReplaceAll::reps: PacletInformation[CodeParser] is not a valid replacement rule.
ReplaceAll::reps: FileNameJoin[{Location /. PacletInformation[CodeParser], PacletInfo.wl}] is not a valid replacement rule.
CodeParser::notransport: No transport specified.

In[2]:= CodeTokenize["1"]
  File "/tmp/Mathics3/mathics-core/mathics/core/expression.py", line 1907, in _is_neutral_symbol
    definition = definitions.get_definition(symbol_name, only_if_exists=True)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/Mathics3/mathics-core/mathics/core/definitions.py", line 506, in get_definition
    raise KeyError
KeyError
...
```

## Expected behavior

No error

## Additional context

I will be putting in a PR to initialize cache to the empty dictionary instead of `None`, assuming that is what is intended.

rocky:
Use {} instead of None for inital structure cache
    
Fixes #1405

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -1942,22 +1942,20 @@ def structure(head, origins, evaluation, structure_cache={}):
 
     if isinstance(origins, (Expression, Structure)):
         cache = origins._cache
-        if cache is not None and not _is_neutral_head(
-            head, structure_cache, evaluation
-        ):
-            cache = None
+        if cache and not _is_neutral_head(head, structure_cache, evaluation):
+            cache = {}
     elif isinstance(origins, (list, tuple)):
         if _is_neutral_head(head, structure_cache, evaluation):
             cache = ExpressionCache.union(origins, evaluation)
         else:
-            cache = None
+            cache = {}
     else:
         raise ValueError("expected Expression, Structure, tuple or list as orig param")
 
-    if cache is None:
-        return UnlinkedStructure(head)
-    else:
+    if cache:
         return LinkedStructure(head, cache)
+    else:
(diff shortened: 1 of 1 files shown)


mathics/core/expression.py
```diff
@@ -1120,7 +1120,7 @@ def has_symbol(self, symbol_name: str) -> bool:
             if hasattr(element, "has_symbol")
         )
 
-    def restructure(self, head, elements, evaluation, structure_cache=None, deps=None):
+    def restructure(self, head, elements, evaluation, structure_cache={}, deps=None):
```
mmatera: Shouldn't it be better to create the dictionary inside the function instead of setting a mutable element as a default parameter?

rocky: Ok. I'll change this.

PR #1407, opened 2025-06-07
Split out fn from print_evaluate

rocky:
Add skip_trivial_evaluation. This is also used in the Mathics Debugger to not stop too many at places that are not of interest. (evaluating a literal string, or a Symbol, etc.)


mmatera (approved):
LGTM

PR #1408, opened 2025-06-07
remove unused parentheses in for etc

fchapoton:
- remove parenthese around arguments of for
- remove parentheses before =
when they are not needed


rocky:
LGTM

PR #1413, opened 2025-06-28
Revise API for position tracking

rocky:
Start tracking locations in specific expressions to aid error reporting and debugging.

We need some care to make sure we don't bloat memory with positions. Right now, we start locations are tagged for FunctionApplyRule, an ExpressionPattern -> Python method. Here, the location is the Python method, or rather can be extracted from that. So we can reuse the location information that Python is storing anyway.

For Expressions and ExpressionPatterns that are read in from Files, things are more complicated..

This needs to be coordinated with a changes in mathics-scanner of the same branch name.

### DIFF
--- a/.github/workflows/consistency-checks.yml
+++ b/.github/workflows/consistency-checks.yml
@@ -23,7 +23,7 @@ jobs:
         sudo apt update -qq && sudo apt install llvm-dev remake
         python -m pip install --upgrade pip
         # We can comment out after next Mathics-Scanner release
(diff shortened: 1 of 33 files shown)


mathics/core/symbols.py
```diff
@@ -495,6 +495,8 @@ def evaluate(self, evaluation):
         for rule in rules:
             result = rule.apply(self, evaluation, fully=True)
             if result is not None and not result.sameQ(self):
+                if result.is_literal:
```
mmatera: Makes sense!

PR #1414, opened 2025-06-29
Lesser changes and possible bugs encountered adding location information #1413

rocky:
These are changes and slight bugs encounted in PR #1413

As a separate PR, this should reduce the size and effort for dealing with #1413, which I'll rebase after this is merged

### DIFF
--- a/mathics/builtin/patterns/basic.py
+++ b/mathics/builtin/patterns/basic.py
@@ -10,6 +10,7 @@ from typing import Optional as OptionalType
 from mathics.core.builtin import PatternObject
 from mathics.core.evaluation import Evaluation
 from mathics.core.expression import Expression
+from mathics.core.symbols import BaseElement
 
 # This tells documentation how to sort this module
 sort_order = "mathics.builtin.rules-and-patterns.basic"
@@ -89,7 +90,7 @@ class Blank(_Blank):
     }
     summary_text = "match to any single expression"
 
-    def match(self, expression: Expression, pattern_context: dict):
+    def match(self, expression: BaseElement, pattern_context: dict):
         vars_dict = pattern_context["vars_dict"]
         yield_func = pattern_context["yield_func"]
 
--- a/mathics/core/definitions.py
+++ b/mathics/core/definitions.py
@@ -18,7 +18,6 @@ from mathics.core.atoms import Integer, String
 from mathics.core.attributes import A_NO_ATTRIBUTES
 from mathics.core.convert.expression import to_mathics_list
 from mathics.core.element import BaseElement, fully_qualified_symbol_name
-from mathics.core.load_builtin import definition_contribute, mathics3_builtins_modules
 from mathics.core.rules import BaseRule, Rule
 from mathics.core.symbols import Atom, Symbol, strip_context
 from mathics.core.util import canonic_filename
@@ -1066,6 +1065,10 @@ def load_builtin_definitions(
     """
     Load definitions from Builtin classes, autoload files and extension modules.
     """
+    from mathics.core.load_builtin import (
+        definition_contribute,
+        mathics3_builtins_modules,
+    )
     from mathics.eval.files_io.files import get_file_time
     from mathics.eval.pymathics import PyMathicsLoadException, load_pymathics_module
     from mathics.session import autoload_files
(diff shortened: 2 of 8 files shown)


mathics/builtin/patterns/basic.py
```diff
@@ -89,7 +90,7 @@ class Blank(_Blank):
     }
     summary_text = "match to any single expression"
 
-    def match(self, expression: Expression, pattern_context: dict):
+    def match(self, expression: BaseElement, pattern_context: dict):
```
rocky, line 93: Change to BaseElement reduces mypy complaints. It may very well be that Expression is more correct, but I don't know how to get mypy to agree with this. So here I switched rather than argue with mypy.

mathics/core/definitions.py
```diff
@@ -1066,6 +1065,10 @@ def load_builtin_definitions(
     """
     Load definitions from Builtin classes, autoload files and extension modules.
     """
+    from mathics.core.load_builtin import (
```
rocky, line 1068: I ran into circular dependencies from changes in #1413, and this fixed that. 

However, moving this down _also_ feels right in that there are the other imports here. 

What this is telling us is that `load_buildin_definitions()` should be moved out of this module.

mathics/core/parser/parser.py
```diff
@@ -49,7 +49,7 @@
 # FIXME: should rework so we don't have to do this
 # We have the character name ImaginaryI and ImaginaryJ, but we should
 # have the *operator* name, "I".
-special_symbols["\uF74F"] = special_symbols["\uF74E"] = "I"
+special_symbols["\uf74f"] = special_symbols["\uf74e"] = "I"
```
rocky, line 52: This one is weird in that black or some formatter decided to do this. 

Again, I'd rather switch than argue with whatever formatter decided this.

mathics/core/parser/util.py
```diff
@@ -12,7 +14,7 @@
 parser = Parser()
 
 
-def parse(definitions, feeder: LineFeeder) -> Any:
+def parse(definitions, feeder: LineFeeder) -> Optional[BaseElement]:
```
rocky, line 17: Adding more precise type annotations for mypy.

mathics/core/pattern.py
```diff
@@ -399,6 +399,12 @@ def get_match_count(self, vars_dict: Optional[dict] = None) -> Tuple[int, int]:
         """The number of matches"""
         return (1, 1)
 
+    @property
+    def short_name(self) -> str:
```
rocky, line 403: Using `short_name` came up in working on TraceEvaluation. I am surprised we never needed `short_name` for patterns before.

mathics/core/symbols.py
```diff
@@ -495,6 +495,8 @@ def evaluate(self, evaluation):
         for rule in rules:
             result = rule.apply(self, evaluation, fully=True)
             if result is not None and not result.sameQ(self):
+                if result.is_literal:
```
rocky, line 498: This is probably just an oversight on our part. Stuff like this becomes apparent when you start looking at traces of expression evaluations.

mathics/core/symbols.py
```diff
@@ -743,7 +745,12 @@ class BooleanType(SymbolConstant):
     the constant is either SymbolTrue or SymbolFalse.
     """
 
-    pass
+    @property
```
rocky, line 748: I thiink @mmatera had mentioned that redfining True and False in MMA, if it is possible, will horribly break MMA.

I recall we had a discussion about whether the parent class SymbolConstant could be a literal. Right now it is not. However True and False I have a harder time seeing why it wouldn't be literal.

mathics/eval/tracing.py
```diff
@@ -26,8 +26,18 @@ def skip_trivial_evaluation(expr, status: str, orig_expr=None) -> bool:
     * the evaluation is a literal that evaluates to the same thing,
     * evaluating a Symbol which the Symbol.
     * Showing the return value of a ListExpression literal
+    * Evaluating Pattern[] which define a pattern.
```
rocky, line 29: This change is inspired by looking at TraceEvaluation. There is no benefit in watching an evaluation of `_`,  `n_`,  or  `n_Integer`.

mathics/core/symbols.py
mmatera: Indeed,  `True` and   `False`, like `List` are some of the few symbols having the attribute  `Locked`. I guess that this is because the evaluation relies on the fixed value of these symbols (but it is not especified anywhere)

Issue #1425, opened 2025-07-22
Erroneous Protected message in SetDelayed


rocky:
## Description

Mathics3 reports 

```
SetDelayed::write: Tag HoldPattern in HoldPattern[Dist[u_, v_, _]] is Protected.
```
when WMA does not. 

## How to Reproduce

```
$ mathics -c 'Format[HoldPattern[Dist[u:_, v:_, _]], TraditionalForm] := HoldForm[u * v]'
```

## Output Given

```
SetDelayed::write: Tag HoldPattern in HoldPattern[Dist[u_, v_, _]] is Protected.
$Failed
```

## Expected behavior

No error, no output (`Null`)

## Additional context

This behavior was noticed in running:

```
Needs[Rubi`"]
LoadModule["pymathics.trepan"]
TraceEvaluation[Int[x, x]]

```
in [`Rubi/Show/Formatting.m` line 14](https://github.com/Mathics3/Mathics3-Rubi/blob/master/ShowStepFormatting.m#L14)

mmatera:
The problem is not that HoldPattern is protected, but how eval_assignment handles this LHS. In this case, the assignment is not to HoldPattern, but to the expression which is wrapped with it.

rocky:
@mmatera Thanks for the information. How easy or difficult do you think a fix is? Is this, in your opinion, a simple mistake or something more fundamentally wrong that may cause massive disruption?

mmatera:
@rocky 
I guess it is not too hard: I have to add a rule to handle this special case. I can try to do that along this afternoon.

mmatera:
Update: after blackout in my town during the afternoon, I had the chance to look at this, and I remember better what was the problem with this: I can handle the special cases of `Format[HoldPattern[...],___]:=` and `HoldPattern[...]:=` more or less easily by adding extra rules to the `mathics.eval.assignment` module. However, still would have problems with nested cases. I found this problem trying to redefine the rules to overwrite `Set` to my custom WL-based MakeBoxes implementation. To do this right, I think it would take more than a few days.

rocky:
[...]

PR #1427, opened 2025-07-23
adding tests on assignments whose behaviour must be fixed.

mmatera:
This PR is WIP to solve  #1425. I start by adding some pytests...

### DIFF
--- a/mathics/core/builtin.py
+++ b/mathics/core/builtin.py
@@ -462,12 +462,17 @@ class Builtin:
                 if pattern is None:  # Fixes PyPy bug
                     continue
                 else:
-                    # TODO: consider to use a more sophisticated
+                    # TODO 1: consider to use a more sophisticated
                     # regular expression, which handles breaklines
                     # more properly, that supports format names
                     # with contexts (context`name) and be less
                     # fragile against leaving spaces between the
                     # elements.
+                    #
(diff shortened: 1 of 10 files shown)


rocky:
> This PR is WIP to solve #1425. I start by adding some pytests...

Starting with pytests is a great idea! Thanks!

rocky:
> > This PR is WIP to solve #1425. I start by adding some pytests...
> 
> Starting with pytests is a great idea! Thanks!

BTW,  when I compare the test behavior that we have right now with the correct behavior, what we have, there are some gaps in Mathics3 behavior. I am glad we are starting to address this.

mathics/core/builtin.py
```diff
@@ -462,12 +462,17 @@ def get_functions(self, prefix="eval", is_pymodule=False):
                 if pattern is None:  # Fixes PyPy bug
                     continue
                 else:
-                    # TODO: consider to use a more sophisticated
+                    # TODO 1: consider to use a more sophisticated
```
rocky, line 465: I suspect regular expressions aren't going to be the solution. 

And this kind of stuff (not matching package context) might be why we've been getting failures in running packages.

mmatera:
#1442 does the same in a less disruptive way.

PR #1428, opened 2025-07-25
Better handling of conditions in replace rules

mmatera:
This PR starts to address some of the issues with `mathics.eval.assignment` by removing a hack, and implementing the evaluation of `Condition` expressions in rules in a closer way to WMA

CHANGES:

mathics.core.definition

* Rules with `Conditional[...]` as replace expressions do not
  overwrite existing rules, and are not overwritten by other rules
  (except by `Unset`).

mathics.core.rule

* For (replacement) `Rule`, `get_sort_key` with `sort_pattern=True`
  returns a sort_key corresponding to a `Conditional` pattern if the
  replace attribute is a conditional expression.

* In the `apply_rule` method, if a condition in the RHS fails to match
  in a replacement rule, or the method in the FunctionApplyRule returns None,
  the rule is treated as if the pattern have not matched the expression.
  In the evaluation loop, this means that further rules are tried until
  one of the associated rules matchs, or there are not any other rule.

  This means that to stop the evaluation process in builtin classes,
  we have to return explicitly an expression.
  This requires some small adjusts on several builtin
  classes.

mathics/eval/assignments/assignment.py

* process_rhs_conditions is not needed anymore.


mmatera:
@rocky: Here I find errors in the combinatorica tests that I am not able to reproduce locally. Any idea of the source of these errors?

rocky:
> @rocky: Here I find errors in the combinatorica tests that I am not able to reproduce locally. Any idea of the source of these errors?

I'll look at soon. A general kind of thing that can happen is that locally you haven't refreshed the Combinatorica code.

rocky:
> @rocky: Here I find errors in the combinatorica tests that I am not able to reproduce locally. Any idea of the source of these errors?

I think there really is an error here. 

I see errors when running this locally.  I also get an error when running directly from the Mathics3-Combinatorica repository. 

To see where the two go differently, try: 

```
In[1]:= Needs["Combinatorica`CombinatoricaV091`"]
In[2]:= Permute[{a,b}, {2,1}]
Out[2]= Permute[{a, b}, {2, 1}]  (* This is wrong, it should be {b, a} *)

In[3]:= LoadModule["pymathics.trepan"]
Out[3]= "pymathics.trepan"

In[3]:= TraceEvaluation[Permute[{a,b}, {2,1}]]

```

They both start the same, but this branch ends: 

```
    Replacing : Permute[{a, b}, {2, 1}] -> Permute[{a, b}, {2, 1}]

  Returning : Permute[{a, b}, {2, 1}] -> Permute[{a, b}, {2, 1}]

Out[5]= Permute[{a, b}, {2, 1}]

```
whereas the master branch keeps going...

```
      Evaluating: PermutationQ[2]

        Replacing : PermutationQ[2] -> PermutationQ[2]

      Returning : PermutationQ[2] -> PermutationQ[2]

      Evaluating: PermutationQ[1]

        Replacing : PermutationQ[1] -> PermutationQ[1]

      Returning : PermutationQ[1] -> PermutationQ[1]

      Replacing : PermutationQ[2] && PermutationQ[1] -> PermutationQ[2] && PermutationQ[1]

    Returning : And @@ PermutationQ /@ {2, 1} -> PermutationQ[2] && PermutationQ[1]

    Evaluating: {2, 1}

    Evaluating: PermutationQ[{2, 1}]

      Replacing : PermutationQ[{2, 1}] -> Sort[{2, 1}] == Range[Length[{2, 1}]]
...
{b, a}
```

rocky:
> > @rocky: Here I find errors in the combinatorica tests that I am not able to reproduce locally. Any idea of the source of these errors?
> 
> I think there really is an error here.
> 
> I see errors when running this locally. I also get an error when running directly from the Mathics3-Combinatorica repository.
> 
> To see where the two go differently, try:
> 
> ```
> In[1]:= Needs["Combinatorica`CombinatoricaV091`"]
> In[2]:= Permute[{a,b}, {2,1}]
> Out[2]= Permute[{a, b}, {2, 1}]  (* This is wrong, it should be {b, a} *)
> 
> In[3]:= LoadModule["pymathics.trepan"]
> Out[3]= "pymathics.trepan"
> 
> In[3]:= TraceEvaluation[Permute[{a,b}, {2,1}]]
> ```
> 
> They both start the same, but this branch ends:
> 
> ```
>     Replacing : Permute[{a, b}, {2, 1}] -> Permute[{a, b}, {2, 1}]
> 
>   Returning : Permute[{a, b}, {2, 1}] -> Permute[{a, b}, {2, 1}]
> 
> Out[5]= Permute[{a, b}, {2, 1}]
> ```
> 
> whereas the master branch keeps going...
> 
> ```
>       Evaluating: PermutationQ[2]
> 
>         Replacing : PermutationQ[2] -> PermutationQ[2]
> 
>       Returning : PermutationQ[2] -> PermutationQ[2]
> 
>       Evaluating: PermutationQ[1]
> 
>         Replacing : PermutationQ[1] -> PermutationQ[1]
> 
>       Returning : PermutationQ[1] -> PermutationQ[1]
> 
>       Replacing : PermutationQ[2] && PermutationQ[1] -> PermutationQ[2] && PermutationQ[1]
> 
>     Returning : And @@ PermutationQ /@ {2, 1} -> PermutationQ[2] && PermutationQ[1]
> 
>     Evaluating: {2, 1}
> 
>     Evaluating: PermutationQ[{2, 1}]
> 
>       Replacing : PermutationQ[{2, 1}] -> Sort[{2, 1}] == Range[Length[{2, 1}]]
> ...
> {b, a}
> ```

Hmm. I wonder if there is a namespace issue and `PermutationQ` is not getting found.

mmatera:
OK, I will dig a little bit more on this then.

mmatera:
I think I found the issue.
This PR is about improving the compatibility with the way in which WMA evaluate rules of the form

`pattern -> Condition[repl, cond]`

In WMA, this structure nearly equivalent to

`Condition[pattern, cond]-> repl`

in the sense that if `cond` does not evaluates to `True` after replacing the variables, the result is similar to fail the pattern match. To fix this at the level of assignments, we hacked the assignment evaluation function to move Condition to the lhs. However, this is not completely equivalent.

For example, in

`Condition[pattern,cond1] -> Condition[repl, cond2]`

`cond1` is evaluated first, and if it results True, then `cond2` is evaluated. In our tricky implementation,
the evaluation order is the opposite:

`Condition[Condition[pattern, cond1], cond2] -> repl`

Also, this makes the code of assignment more complicated, and makes that the rules written in this way fail.



What I did here is try to reproduce the WMA behavior at the level of the rules, by return `None` when a failing condition in the replacement expression. This should have the consequence that the rule *fails* to apply.
However, there is a hack in `BaseRule.apply(expr)`, that replaces `None` by `expr`. As a result,
the replacement do not *fail*. But then, in the evaluation loop, the rule is given as applied, and no other rules are tryied.

Combinatorica defines two Downvalues for `Permute`. The first applies when

```
Permute[l_List,p_?PermutationQ] := l [[ p ]]
```
and

```
Permute[l_List,p_List] := Map[ (Permute[l,#])&, p] /; (Apply[And, res=Map[PermutationQ, p]; Print[res];res])
```

It happends that the second rule is tryied first. In WMA, when we evaluate
```
Permute[{a,b},{2,1}]
```
the replacement fails for the first rule, and go to the other rule.
But in Mathics, as the rule do not *fail* but just returns the original expression, the evaluation is halted.


The problem now is that the implementation of several functions in Mathics is based on the wrong behavior. I am trying now to ammend it with the minimum number of changes...

rocky:
> I think I found the issue. This PR is about improving the compatibility with the way in which WMA evaluate rules of the form
> 
> `pattern -> Condition[repl, cond]`
> 
> In WMA, this structure nearly equivalent to
> 
> `Condition[pattern, cond]-> repl`
> 
> in the sense that if `cond` does not evaluates to `True` after replacing the variables, the result is similar to fail the pattern match. To fix this at the level of assignments, we hacked the assignment evaluation function to move Condition to the lhs. However, this is not completely equivalent.
> 
> For example, in
> 
> `Condition[pattern,cond1] -> Condition[repl, cond2]`
> 
> `cond1` is evaluated first, and if it results True, then `cond2` is evaluated. In our tricky implementation, the evaluation order is the opposite:
> 
> `Condition[Condition[pattern, cond1], cond2] -> repl`
> 
> Also, this makes the code of assignment more complicated, and makes that the rules written in this way fail.
> 
> What I did here is try to reproduce the WMA behavior at the level of the rules, by return `None` when a failing condition in the replacement expression. This should have the consequence that the rule _fails_ to apply. However, there is a hack in `BaseRule.apply(expr)`, that replaces `None` by `expr`. As a result, the replacement do not _fail_. But then, in the evaluation loop, the rule is given as applied, and no other rules are tryied.
> 
> Combinatorica defines two Downvalues for `Permute`. The first applies when
> 
> ```
> Permute[l_List,p_?PermutationQ] := l [[ p ]]
> ```
> 
> and
> 
> ```
> Permute[l_List,p_List] := Map[ (Permute[l,#])&, p] /; (Apply[And, res=Map[PermutationQ, p]; Print[res];res])
> ```
> 
> It happends that the second rule is tryied first. In WMA, when we evaluate
> 
> ```
> Permute[{a,b},{2,1}]
> ```
> 
> the replacement fails for the first rule, and go to the other rule. But in Mathics, as the rule do not _fail_ but just returns the original expression, the evaluation is halted.
> 
> The problem now is that the implementation of several functions in Mathics is based on the wrong behavior. I am trying now to ammend it with the minimum number of changes...

Thanks for all of this.  

Is it possible to do more work inside the _Condition_ evaluation rather than inside the assignment code? I see you deleted the move of weird code to move the RHS to the LHS assignment evaluation and that is good.

mathics/core/rules.py
```diff
@@ -254,6 +254,13 @@ def apply_rule(
         new = self.replace.replace_vars(vars)
         new.options = options
 
+        while new.has_form("System`Condition", 2):
+            new, cond = new.get_elements()
+            if isinstance(cond, Expression):
+                cond = cond.evaluate(evaluation)
+            if cond is not SymbolTrue:
```
rocky, line 261: It would be great if we could move all of this into Condition evaluation. 

What would it take to _not_ have to add `if cond is not SymbolTrue`?

mmatera: `Condition` is never *evaluated*: it is just a modifier on how rules are applied. I am working on it now. The problem are the side effects that I have to control before.

rocky: Line 260 looks like an evaluation. And why is it done only if it is an Expression? 

But, Ok, I'll wait until this is finished.

rocky:
@mmatera: Please make 7fcbc011fa549391b61d029492790f36f62b19d7 a separate PR. It is rather large, and we should review/discuss it separately. Many thanks.

mathics/core/rules.py
mmatera: To be precise, `Condition[]` do not have attached an evaluation rule like, let' s say `Plus[]`. `Condition`  has a particular meaning in replacement rules: when a pattern of a rule matches with an expression, and it happens that the replacement has the form `Condition[repl, cond]`, `cond` is evaluated. If the evaluation yields `True`, then the replacement proceeds, otherwise, the replacement is aborted. But this happens occurs outside of the *regular* evaluation loop.

rocky:
@mmatera: Please don't add any more `Pattern[]` doc changes in this PR. 

I have been thinking about this. One of the downsides of this approach is that the docstring is a bit more obscured. 

Based on the work I've been doing in the debugger, I think there's a way to use code introspection, i.e. inspect.currentframe(). 

This is better because we don't have to change docstrings everywhere, each docstring is better, and we get all of the benefits of the `Pattern` approach. I'll put in a PR to show this in a day or so.  (Oh, and it is also faster, since work is done only on need without having to create Pattern expressions.)

Thanks.

mmatera:
> @mmatera: Please don't add any more `Pattern[]` doc changes in this PR.
> 
> I have been thinking about this. One of the downsides of this approach is that the docstring is a bit more obscured.
> 
> Based on the work I've been doing in the debugger, I think there's a way to use code introspection, i.e. inspect.currentframe().
> 
> This is better because we don't have to change docstrings everywhere, each docstring is better, and we get all of the benefits of the `Pattern` approach. I'll put in a PR to show this in a day or so. (Oh, and it is also faster, since work is done only on need without having to create Pattern expressions.)
> 
> Thanks.

@rocky, if you look at WMA packages, you could find that assignment expressions like 
```
expr:F[...] := ...
```
are quite common. This is because many times you would like to avoid recreating expressions that already exist. What makes our docstrings ugly is that we can not use the operator form `name:pat` because we use that syntax for another thing. I have planned to change that too in the rework of `MakeBoxes` ("""form: pat""" is used to define formats), and then we could use a more compact notation, but I do not want to mix the stuff here.

The changes introduced here are (almost) just enough to make all the tests pass, including the ones in Combinatorica. I also would expect that once this and the other PRs get merged (the ones related to get_sort_key and assignment), several packages that currently can not load start to work, just because we get much closer to the way in which WMA processes expressions.

rocky:
> > @mmatera: Please don't add any more `Pattern[]` doc changes in this PR.
> > I have been thinking about this. One of the downsides of this approach is that the docstring is a bit more obscured.
> > Based on the work I've been doing in the debugger, I think there's a way to use code introspection, i.e. inspect.currentframe().
> > This is better because we don't have to change docstrings everywhere, each docstring is better, and we get all of the benefits of the `Pattern` approach. I'll put in a PR to show this in a day or so. (Oh, and it is also faster, since work is done only on need without having to create Pattern expressions.)
> > Thanks.
> 
> @rocky, if you look at WMA packages, you could find that assignment expressions like
> 
> ```
> expr:F[...] := ...
> ```
> 
> are quite common. This is because many times you would like to avoid recreating expressions that already exist

I got that. What I am saying is there is a clearer way to do this.

I am asking that you hold off on changing docstings of the form:

```
"Append[expr_, item_]"
```

which is pretty clear, to: 

```
"Pattern[expression, Append[expr_, item_]]"
```

which adds what I suspect is a gratuitous `Pattern[expression, ... ]`. I think we can have the same effect _without_ doing that. 

It is great to see problems and come up with ideas for fixing them. But before making dozens of changes of this kind , let's discuss first.


I will put in a small PR to show how to accomplish this for this `Append[]` case and we can discuss. Thanks.

mmatera:
I tried to change as few as possible to keep the behavior. But maybe some of the changes were unnecessary. I need to check again.

mmatera:
Maybe a more interesting discussion is about the semantic of return None in a functionapply rule. Here I followed what I used to understood as the semantic in the old mathics project: if an eval method returns None, it is the same as the pattern do not match. Ao, is the same as a Condition which a non true second argument. An alternative approach (which is how it was working now) is to assume that the pattern match, and returns the original expression. If we want to keep this semantic, what we have to do is something different to indicate the evaluator to continue looking for other rules. Maybe raise an special exception. This would be an alternative approach that does not requires the changes in mathics.builtin....

rocky:
> Maybe a more interesting discussion is about the semantic of return None in a functionapply rule. 

Both discussions are good. 

> Here I followed what I used to understood as the semantic in the old mathics project: if an eval method returns None, it is the same as the pattern do not match. Ao, is the same as a Condition which a non true second argument. An alternative approach (which is how it was working now) is to assume that the pattern match, and returns the original expression. If we want to keep this semantic, what we have to do is something different to indicate the evaluator to continue looking for other rules. Maybe raise an special exception. This would be an alternative approach that does not requires the changes in mathics.builtin....

I have to think and reflect on this one a bit. This topic is the harder of the two.

When you wrote: 

> To be precise, Condition[] do not have attached an evaluation rule like, let' s say Plus[]. Condition has a particular meaning in replacement rules: when a pattern of a rule matches with an expression, and it happens that the replacement has the form Condition[repl, cond], cond is evaluated. If the evaluation yields True, then the replacement proceeds; otherwise, the replacement is aborted. But this happens occurs outside of the regular evaluation loop.

I am not sure I see this as something outside of the usual kinds of functional programming LISP-like idioms. `Condition` has a  `HoldAll` attribute, which means that evaluation can be dictated by the code for `Condition`, or not done at all. This is no different from, say, in LISP, you might have `(if expr then-clause else-clause)` where clearly one doesn't evaluate *both*  the "then" and "else" clauses. Even more extreme, I guess, is `(quote (+ 1 "3")` which just doesn't ever evaluate it's arguments. (And here if it would do that, there would be an error since you can't add an integer to a string.) But there's nothing that stops some parent of this Sexp from doing the evaluation.

But again, for this topic, I would need more time to think about this.

Do we know anything about how WMA handles this problem?

mmatera:
Maybe this can help to elucidate your question about how it works in WMA:
```
In[1]:= Condition[x,True]                                                                                                                                                     

Out[1]= x /; True
```
If `Condition` would work like, let's say, `If[cond, result]`, the output would be `x` instead of staying unevaluated. 

On the other hand, 
```
In[2]:= y/.y:>Condition[x,a>0]                                                                                                                                                

Out[2]= y
```
but if we set a=1,

```
In[3]:= a=1                                                                                                                                             

Out[3]= 1

In[4]:= y/.y:>Condition[x,a>0]                                                                                                                                                

Out[4]= x
```

This is an example on how it works in assigments:
```
In[5]:= Condition[F[x_],Print[1];True]:=Condition[t,Print[2];True]                                                                                                            

In[6]:= F[2]                                                                                                                                                                  
1
2

Out[6]= t
```
Notice that when the rule is attempt to be applyied on `F[2]`, the LHS condition is evaluated, and then the RHS is evaluated, before apply the replacement rule.

Here is a more sofisticated example, with two rules:

```
In[1]:= G[x_]:=x^2                                                                                                                                                            

In[2]:= G[x_]:=Condition[0, Print["check first"];x<0]                                                                                                                         

In[3]:= G[1]                                                                                                                                                                  
check first

Out[3]= 1

In[4]:= G[-1]                                                                                                                                                                 
check first

Out[4]= 0
```
Notice that in both cases, the pattern with the condition is evaluated first. If the condition is evaluated to `True` the rule is applied. If not, the next rule in the list is applied.

mmatera:
How is internally implementeds

> @mmatera: Please make [7fcbc01](https://github.com/Mathics3/mathics-core/commit/7fcbc011fa549391b61d029492790f36f62b19d7) a separate PR. It is rather large, and we should review/discuss it separately. Many thanks.

done in  #1439

mmatera:
#1441 does the same in a less disruptive way.

rocky:
> 
> This is an example on how it works in assigments:
> 
> ```
> In[5]:= Condition[F[x_],Print[1];True]:=Condition[t,Print[2];True]                                                                                                            
> 
> In[6]:= F[2]                                                                                                                                                                  
> 1
> 2
> 
> Out[6]= t
> ```
> 
> Notice that when the rule is attempt to be applyied on `F[2]`, the LHS condition is evaluated, and then the RHS is evaluated, before apply the replacement rule.
> 
> Here is a more sofisticated example, with two rules:
> 
> ```
> In[1]:= G[x_]:=x^2                                                                                                                                                            
> 
> In[2]:= G[x_]:=Condition[0, Print["check first"];x<0]                                                                                                                         
> 
> In[3]:= G[1]                                                                                                                                                                  
> check first
> 
> Out[3]= 1
> 
> In[4]:= G[-1]                                                                                                                                                                 
> check first
> 
> Out[4]= 0
> ```
> 
> Notice that in both cases, the pattern with the condition is evaluated first. If the condition is evaluated to `True`, the rule is applied. If not, the next rule in the list is applied.

Nothing is surprising or special to Mathematatica here, in contrast to how other programming language technology works. This is the way LHS usually works in an assignment.

In Python, for example:

```
F[5] = "foo"
```

The `F[5]` is treated differently than if it appeared on the RHS. On the LHS, you get the _address_ or _place_ to store the RHS rather than its value.

The only thing Mathematica does here is have more involved and complicated kinds of expressions on the left-hand side.  Logically, `Condition`, is tagged correctly as `HoldAll` as it should. (No surprise here).

It's just that on the Mathics implementation side, there has been fuzziness, vagueness, wrong "short-cut" hacks, and just plain wrong behavior in treating the left-hand side of assignments. No surprise here either.

But now is our chance to try to address this and remove the vagueness and fuzziness.

PR #1429, opened 2025-07-26
adjust SetEnvironment

mmatera:
This is a first example of implementation that works by chance due to a general misimplementation.
Also, now the result is compliant with WMA.

### DIFF
--- a/mathics/builtin/system.py
+++ b/mathics/builtin/system.py
@@ -603,7 +603,6 @@ class SetEnvironment(Builtin):
 
      Set two environment variables:
      S> SetEnvironment[{"FOO" -> "baz", "A" -> "B"}]
-      = SetEnvironment[{FOO -> baz, A -> B}]
 
      See that the environment variable has changed:
      S> GetEnvironment["FOO"]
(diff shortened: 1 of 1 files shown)


mathics/builtin/system.py
```diff
@@ -633,24 +633,36 @@ class SetEnvironment(Builtin):
...
+        for rule in rules.elements:
+            if not rule.has_form("System`Rule", 2):
+                return None
+            if not isinstance(rule.elements[0], String):
+                return None
+
         for rule in rules.elements:
             self.eval(rule, evaluation)
-        return None
+        return SymbolNull
```
rocky: If this is intended to be merged, then I suppose we should remove the duplicate return. 

If it is intended to be merged. LGTM.

mmatera: which one? I can merge the two if' s in lines 656 and 658, but I thought it was clearer in this way.

rocky: I couldn't figure out how to do this with the +/- marker so I did this as PR https://github.com/Mathics3/mathics-core/pull/1430

PR #1433, opened 2025-07-28
Processing options in patterts with Optional values and OptionsPattern together.

mmatera:
Digging on the issues with `Set*` I found some issues on how we store and apply pattern rules during evaluation. Here there is a bunch of cases of definitions that in the master branch passes the tests just because misimplementations in the way the rules are stored, sorted and applied. 
There are more changes to come, but these are easy. Most of them comes from the way in which `Optional` and `OptionsPattern` interact. The new rules are compatible with the WMA evaluation loop.

### DIFF
--- a/mathics/builtin/list/eol.py
+++ b/mathics/builtin/list/eol.py
@@ -198,25 +198,25 @@ class Cases(Builtin):
 
     summary_text = "list elements matching a pattern"
 
-    def eval(self, items, pattern, ls, evaluation, options):
-        "Cases[items_, pattern_, ls_:{1}, OptionsPattern[]]"
+    def eval(self, items, pattern, levelspec, evaluation, options):
+        "Cases[items_, pattern_, levelspec_:{1}, OptionsPattern[]]"
         if isinstance(items, Atom):
             return ListExpression()
 
-        if ls.has_form("Rule", 2):
-            if ls.elements[0].get_name() == "System`Heads":
-                heads = ls.elements[1] is SymbolTrue
-                ls = ListExpression(Integer1)
+        if levelspec.has_form("Rule", 2):
+            if levelspec.elements[0].get_name() == "System`Heads":
+                heads = levelspec.elements[1] is SymbolTrue
+                levelspec = ListExpression(Integer1)
             else:
-                evaluation.message("Position", "level", ls)
+                evaluation.message("Position", "level", levelspec)
                 return
         else:
             heads = self.get_option(options, "Heads", evaluation) is SymbolTrue
 
         try:
-            start, stop = python_levelspec(ls)
+            start, stop = python_levelspec(levelspec)
         except InvalidLevelspecError:
-            evaluation.message("Position", "level", ls)
+            evaluation.message("Position", "level", levelspec)
             return
 
         results = []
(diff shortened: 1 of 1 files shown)


rocky:
I looked at the PR, and from what I can tell, a new function `param_and_option_from_optional_place()` has been added. 

On first glance, I don't see anything wrong with it. But I would like more clarification of what is going on and where we are heading with this.

*Edit: the second PR clears this up a little, but only a little*

It looks at the last argument of what is presumably a left-hand-side expression. If that value is an option value, which it determines by looking at a list of options, it separates the option from the rest of the LHS and returns that separate from the rest of the LHS expression. 

How this translates into "improve some rules to be more resilient" is, to me, somewhat nebulous.

What problem(s) exactly are we trying to solve? Please give a simple, specific example. 

Small item "lhs, not "ls" is typically what is used to indicate the left-hand side of an assignment expression.

mmatera:
The problem that this PR solves is that the use of the combination Optional[...] and OptionsPattern only works here because of a bug in the evaluation loop. If I try to reproduce the behavior these patterns in WMA, I have the need to write something like the new function: an option in the position of an optional value is took by the interpreter as the optional value, and ruins the evaluation.

mathics/builtin/functional/apply_fns_to_lists.py
```diff
@@ -75,15 +76,14 @@ class Apply(InfixOperator):
...
     }
 
-    def eval_invalidlevel(self, f, expr, ls, evaluation, options={}):
-        "Apply[f_, expr_, ls_, OptionsPattern[Apply]]"
-
-        evaluation.message("Apply", "level", ls)
-
     def eval(self, f, expr, ls, evaluation, options={}):
-        """Apply[f_, expr_, Optional[Pattern[ls, _?LevelQ], {0}],
+        """Apply[f_, expr_, Optional[ls_, {0}],
```
mmatera: @rocky this is the place where this `ls` parameter happens. I propose this change because with a small change in the sequence of evaluations, these rules seem to fail, confusing the first option with the `Optional` parameter. This issue often happens when these two patterns are used together. However, with the changes in the last version of the other PRs, it seems I do not hit the issue.

mathics/builtin/functional/apply_fns_to_lists.py
```diff
@@ -415,19 +410,16 @@ class Scan(Builtin):
...
 
-    def eval_invalidlevel(self, f, expr, ls, evaluation, options={}):
-        "Scan[f_, expr_, ls_, OptionsPattern[Map]]"
-
-        evaluation.message("Map", "level", ls)
-
-    def eval_level(self, f, expr, ls, evaluation, options={}):
-        """Scan[f_, expr_, Optional[Pattern[ls, _?LevelQ], {1}],
+    def eval_level(self, f, expr, levelspec, evaluation, options={}):
+        """Scan[f_, expr_, Optional[levelspec_, {1}],
```
mmatera, line 414: `levelspec` must be the name of a symbol in Mathics, so cannot follow the snake name convention. `levelspecification` is too long,  and we already have used  `levelspec` in the `python_levelspec` function name.

rocky: Thanks - that's fine. `ls` is too vague and confusing &mdash; to prove it, I got confused.

mathics/builtin/functional/apply_fns_to_lists.py
rocky: Ok. Thanks for the explanation.

rocky:
LGTM

PR #1434, opened 2025-07-28
Improve get sort key

mmatera:
This is the second part, related to the way rules are ordered. Most of this PR is just to split a long chain of `if` statements, into specific methods for the different kind of patterns.

### DIFF
--- a/mathics/builtin/patterns/basic.py
+++ b/mathics/builtin/patterns/basic.py
@@ -10,6 +10,7 @@ from typing import Optional as OptionalType
 from mathics.core.builtin import PatternObject
 from mathics.core.evaluation import Evaluation
 from mathics.core.expression import Expression
+from mathics.core.pattern import ATOM_PATTERN_SORT_KEY
 from mathics.core.symbols import BaseElement
 
 # This tells documentation how to sort this module
@@ -101,6 +102,20 @@ class Blank(_Blank):
             else:
                 yield_func(vars_dict, None)
 
+    def get_sort_key(self, pattern_sort=True):
+        if pattern_sort:
+            return (
+                2,
+                11 if self.elements else 21,
+                1,
+                1,
+                0,
+                ATOM_PATTERN_SORT_KEY,
+                tuple(elem.get_sort_key(True) for elem in self.elements),
+                1,
+            )
+        return self.expr.get_sort_key()
+
 
 class BlankNullSequence(_Blank):
     """
@@ -143,6 +158,20 @@ class BlankNullSequence(_Blank):
 
     def get_match_count(self, vars_dict: OptionalType[dict] = None) -> tuple:
         return (0, None)
+
+    def get_sort_key(self, pattern_sort=True):
+        if pattern_sort:
+            return (
+                2,
+                13 if self.elements else 23,
+                1,
+                1,
+                0,
+                ATOM_PATTERN_SORT_KEY,
+                tuple(elem.get_sort_key(True) for elem in self.elements),
+                1,
+            )
+        return self.expr.get_sort_key()
 
 
 class BlankSequence(_Blank):
@@ -203,3 +232,17 @@ class BlankSequence(_Blank):
 
     def get_match_count(self, vars_dict: OptionalType[dict] = None) -> tuple:
         return (1, None)
+
+    def get_sort_key(self, pattern_sort=True):
+        if pattern_sort:
+            return (
+                2,
+                12 if self.elements else 22,
+                1,
+                1,
+                0,
+                ATOM_PATTERN_SORT_KEY,
+                tuple(elem.get_sort_key(True) for elem in self.elements),
+                1,
+            )
+        return self.expr.get_sort_key()
--- a/mathics/builtin/patterns/basic.py
+++ b/mathics/builtin/patterns/basic.py
@@ -111,6 +111,9 @@ class Blank(_Blank):
                 1,
                 0,
                 ATOM_PATTERN_SORT_KEY,
+                1
+                if any(element.undefined_sequence_length() for element in self.elements)
+                else 0,
                 tuple(elem.get_sort_key(True) for elem in self.elements),
                 1,
             )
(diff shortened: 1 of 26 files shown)


mmatera:
What is missing now is to modify the sort_keys for patterns,  in a way that a pattern of the form `F[___]` cames after a pattern of the form `F[_,_]`. This is important to ensure that more general rules come after the most specific rules.

mmatera:
What is missing now is to modify the sort_keys for patterns,  in a way that a pattern of the form `F[___]` came after a pattern of the form `F[_,_]`. This is important to ensure that more general rules come after the most specific rules.

rocky:
> This is the second part, related to the way rules are ordered. Most of this PR is just to split a long chain of `if` statements, into specific methods for the different kind of patterns.

So if I have this right, this doesn't change any existing behavior as far as we know and as far as the specific tests we have in place can determine. Right? 

I think this is an improvement, but it would be great to give more explanation in the docstring of "get_sort_key"  what the specific sort orders are trying to convey. 

I will mark elsewhere an example of one of the functions to be more specific. 

BTW, thanks for undertaking organizing this mess.

mathics/builtin/patterns/basic.py
```diff
@@ -101,6 +102,20 @@ def match(self, expression: BaseElement, pattern_context: dict):
             else:
                 yield_func(vars_dict, None)
 
+    def get_sort_key(self, pattern_sort=True):
```
rocky, line 105: I would be grateful for a docstring to describe in human terms what `(2, 11 | 21, 1, 1, 0, ...)` means.

And likewise in the other `get_sort_key()`  methods.

rocky: Thinking about this some more, Maybe we should do is define some Literal constants for `2`, `1`, `0` , `11`, `21`, etc. 

To give an analogous situation, consider operator precedence. 

We might have both `+` and `-`  have numeric precience 310 whiile `*` and `/` might have operator precedence 400. But as a literal value, 310 might be named `ADDITION_PRECEDENCE` while 400 might be the literal value for `MULTIPLICATION_PRECEDENCE`.

Looking at the code though, I don't know exactly what the `1`, `0`, `11`, mean.

mmatera:
> > This is the second part, related to the way rules are ordered. Most of this PR is just to split a long chain of `if` statements, into specific methods for the different kind of patterns.
> 
> So if I have this right, this doesn't change any existing behavior as far as we know and as far as the specific tests we have in place can determine. Right?
> 

Nop, until this point, everything works as before. This is way no tests are changed.

> I think this is an improvement, but it would be great to give more explanation in the docstring of "get_sort_key" what the specific sort orders are trying to convey.

I am working on documenting this now, which is probably the hardest part....

> 
> I will mark elsewhere an example of one of the functions to be more specific.
> 
> BTW, thanks for undertaking organizing this mess.

rocky:
> > > This is the second part, related to the way rules are ordered. Most of this PR is just to split a long chain of `if` statements, into specific methods for the different kind of patterns.
> > 
> > 
> > So if I have this right, this doesn't change any existing behavior as far as we know and as far as the specific tests we have in place can determine. Right?
> 
> Nop, until this point, everything works as before. This is way no tests are changed.
> 

Ok. Good. What I'd ask you to do outside of this PR then is to fix _one_ kind of problem as a sanity check to see that this is helpful. 

I guess it would be a PR based on this PR.

mathics/core/pattern.py
```diff
@@ -1199,3 +1218,51 @@ def per_name(yield_name: Callable, groups: Tuple, vars_dict: dict):
...
+    # Append (4,) to elements so that longer expressions have higher
+    # precedence
+    result = (
+        2,
+        0,
+        1,
+        1,
+        0,
+        pat.head.get_sort_key(True),
+        tuple(
```
mmatera, line 1291: TODO: to reproduce the WMA evaluation order, we need to add an element between the head and the elements that indicates if the pattern matches with a definite number of elements (leaves) or if it has `BlankSequence`-like elements. The goal is that `Expr[___]` comes after `Expr[_,_]`.

mathics/builtin/patterns/basic.py
mmatera: > Thinking about this some more, Maybe we should do is define some Literal constants for `2`, `1`, `0` , `11`, `21`, etc.
> 
> To give an analogous situation, consider operator precedence.
> 
Indeed, it is analogue to operator precedence, but with its own tricks...

> We might have both `+` and `-` have numeric precience 310 whiile `*` and `/` might have operator precedence 400. But as a literal value, 310 might be named `ADDITION_PRECEDENCE` while 400 might be the literal value for `MULTIPLICATION_PRECEDENCE`.
> 
> Looking at the code though, I don't know exactly what the `1`, `0`, `11`, mean.

At this point, I have two options to continue with this: one is using the current API, try to fix the few remaining issues with the sort keys to reproduce exactly the WMA behavior.
The other possibility is to reimplement the sort keys using bitmasks or another kind of structure. 
This other way would be more solid, and probably faster, but requires more changes over the current code. On the other hand, what each of the fields are is now documented in the code, and its explicit use is quite localized in a few modules. So, what do you think would be the best?

rocky: Creating `Literal` values that capture

```
 2,  # Expression
        0,  # Not blank or OptionValues
        1,  # Not pattern test
        1,  # Not a Pattern
        0,  # Not Optional
```

I don't think is too disruptive.  Something like 

```
from typing import Final, Literal

# Define an annotated constant integer Literal
SORT_ORDER_NOT_BLANK_OR_OPTION_VALUE: Final[Literal[1]] = 1
SORT_ORDER_NOT_PATTERN_TEST: Final[Literal[1]] = 1
SORT_ORDER_NOT_OPTIONAL: Final[Literal[0]] = 0
```

It is possible there could be another rewrite after this.  My main concern is the readability of something like `(2, 11 | 21, 1, 1, 0, ...)`

mmatera:
superseeded by #1447

mathics/builtin/patterns/basic.py
mmatera: #1447 is an step more on this direction

PR #1436, opened 2025-07-29
TraceEvaluation can filter rewrites or evaluations...

rocky:
New options booleans ShowRewrites and ShowEvaluation can be added to allow filtering TraceEvalation output to filter other either rewrites rules or evaluation expressions. Presumably you don't want to filter both.

### DIFF
--- a/mathics/core/definitions.py
+++ b/mathics/core/definitions.py
@@ -162,7 +162,7 @@ class Definitions:
         )
         self.inputfile = ""
 
-        # These are used by TraceEvaluation to indicate
+        # These are used by TraceEvaluation to```
         # whether what information to show.
         self.trace_evaluation = False
         self.trace_show_rewrite = False


mmatera (approved):
Interesting! This is better than the WMA debug... LGTM

mathics/core/definitions.py
```diff
@@ -162,7 +162,10 @@ def __init__(
         )
         self.inputfile = ""
 
+        # These are used by TraceEvaluation to indicate
+        # whether what information to show.
```
rocky, line 166:
```suggestion
        # to decide what information to show.
```

mathics/core/definitions.py
```diff
@@ -162,7 +162,10 @@ def __init__(
         )
         self.inputfile = ""
 
+        # These are used by TraceEvaluation to indicate
```
rocky:
```suggestion
        # These are used by TraceEvaluation to```

rocky:
> Interesting! This is better than the WMA debuger ... LGTM

While WMA will for a long while (if not always) be better than Mathics3 in speed and capabilities, one of the few areas where Mathics3 might be able to do better than WMA is in debugging and being more transparent about what goes on underneath.  

In this area, though, there is much, much more that can be done. I think it will happen, though, over time.

PR #1438, opened 2025-07-30
First cut at Interrupt handler...

rocky:
Also, `main.py` has been  moved to `__main__.py`  following Python conventions for main routines. This makes `python -m mathics` work.

GNU Readline history is enabled for "mathics" when it is available. It will share history files with mathicsscript.


mmatera:
> GNU Readline history is enabled for "mathics" when it is available. It will share history files with mathicsscript.

This is very welcome!

rocky:
While this isn't perfect, I think it's good enough to start off. 

In another commit, I'll deal with "trace" not being implemented. 

Another feature that might get added is another interrupt handler, such for SIGUS1, that shows what's getting run at any given time. In contrast to this handler it is non-interactive.

rocky:
Merging, since no negative comment. @mmatera @aravindh-krishnamoorthy - since the location of `mathics` changes (from `mathics.main` to `mathics.__main__`  make sure you do a `pip install -e .` after updating mater.

PR #1439, opened 2025-07-31
rule adjustments and sort_key for holdpattern

mmatera:


rocky:
@mmatera PR #1440 shows how to do most of this without adding `Pattern [Expression, ...]`. 

Your thoughts?

mathics/builtin/patterns/composite.py
```diff
@@ -173,6 +173,11 @@ def match(self, expression: Expression, pattern_context: dict):
         #     yield new_vars_dict, rest
         self.pattern.match(expression, pattern_context)
 
+    def get_sort_key(self, pattern_sort=True):
```
rocky, line 176: This change seems to be different than all of the other changes. Let's separate this part from the stuff that adds `"Pattern[expression, ... ` everywhere. 

In other words, instead of a PR for "rule adjustments and sort_key for holdpattern",  a PR for "sort_key for holdpattern, and "rule adjustments", two PRs.

The title "rule adjustments" my opinion, is less vaguely described as capturing the eval Expression inside eval methods that need them.

Thanks.

mmatera: Indeed, now this got decoupled of the other changes. Let's close this and reimplement it with your approach.

rocky: Thanks for your understanding, patience, and tolerance.

PR #1441, opened 2025-07-31
Using exceptions to handle conditions in the RHS

mmatera:
This is an alternative to #1428, using a specific exception to indicate that a rule fails to apply on an expression.


mathics/core/definitions.py
```diff
@@ -161,9 +161,6 @@ def __init__(
             "Global`",
         )
         self.inputfile = ""
-
```
rocky: These lines removed inadvertently?

mmatera: I think so....

rocky:
I agree this is improved from #1428. And again, I like the tests, but I may have more to say on this some other time.

Again, important gaps are being filled in and mistakes corrected.

I'd like to spend more time going over this in detail though. 

Thanks for the great work.

mathics/eval/assignments/assignment.py
```diff
@@ -1292,22 +1292,11 @@ def process_rhs_conditions(
     lhs: BaseElement, rhs: BaseElement, condition: Expression, evaluation: Evaluation
 ) -> Tuple[BaseElement, Optional[BaseElement]]:
     """
-    lhs = Condition[rhs, test] -> Condition[lhs, test]  = rhs
+    Add back the lhs conditions.
     """
     # To Handle `OptionValue` in `Condition`
     rulopc = build_rulopc(lhs.get_head())
     rhs_name = rhs.get_head_name()
```
rocky: `rhs_name`  is no longer used. Its only use is deleted below.

mathics/core/definitions.py
```diff
@@ -996,10 +993,26 @@ def insert_rule(values: List[BaseRule], rule: BaseRule) -> None:
 
     """
 
-    for index, existing in enumerate(values):
-        if existing.pattern.sameQ(rule.pattern):
-            del values[index]
-            break
+    def is_conditional(x):
```
rocky:
```suggestion
    def is_conditional(x) -> bool:
```

mathics/eval/assignments/assignment.py
```diff
@@ -1292,22 +1292,11 @@ def process_rhs_conditions(
     lhs: BaseElement, rhs: BaseElement, condition: Expression, evaluation: Evaluation
 ) -> Tuple[BaseElement, Optional[BaseElement]]:
     """
-    lhs = Condition[rhs, test] -> Condition[lhs, test]  = rhs
+    Add back the lhs conditions.
```
rocky, line 1295: The name "process_rhs_conditions" seems to me a misnomer. This function right now doesn't do _anything_ with the parameter, "rhs", except pass it back. 

So, as far as I can tell, the "rhs" parameter might be removed altogether. 

What this function does right now seems to be to _wrap_ `lhs` in a "Condition" expression, if parameter `condition` is an Expression.

mmatera: The idea is to finally remove this in the subsequent assignment refactor, but at this point, it is still entangled with other functions. I left it just as an intermediate step to finally remove it.

rocky: Ok.  Then it might be good to add this information as a comment, just in case it is inadvertently forgotten.

test/core/test_rules.py
```diff
@@ -51,6 +51,51 @@
 """
 
 
+@pytest.mark.parametrize(
```
rocky, line 54: Please move this to `test/builtins/test_assignment.py` or write code that stays within `mathics.core`. Thanks.

mathics/core/definitions.py
mmatera: I guess this is resolved now.

mathics/eval/assignments/assignment.py
mmatera: Now it is called `put_back_lhs_conditions` and has just the used parameters. I left it just to help merge the part related to assignments

test/core/test_rules.py
mmatera: These are tests for rules, not for assignment. But OK, they could be inside test.builtin.test_rules.py or something like that.

rocky: There is an important point regarding the organization of tests and the level of testing that isn't getting understood. 

If the `Set` Built-in functions are used in testing, that goes in the tests for the `Set`-class of commands. Even though your intent here is to do that to test some lower-level kind of behavior.  

If you want to write tests for `builtins` or core rules, fine.  _just stay in the abstraction level that those live in. Do not use the higher levels of abstraction_.

In other projects and code organizations, we could separate end-to-end tests from unit tests. In that scheme, in an end-to-end test (but not a unit test), it might be okay to use a higher-level routine to check its behavior to validate some lower-level concept. 

Our code is organized in layers. The core layer is independent of the built-in layer.  As such, core testing should be done using the core API only and not mix the Built-in API.

rocky:
The main thing I see missing from #1442 is changing variables with the name "ls" to "lhs".

mmatera:
> The main thing I see missing from #1442 is changing variables with the name "ls" to "lhs".

Again, the places where I found `ls` instead of `lhs` were variables meaning "level specification". But also, in this PR I could not find a case where this variable appears.

rocky:
> > The main thing I see missing from #1442 is changing variables with the name "ls" to "lhs".
> 
> Again, the places where I found `ls` instead of `lhs` were variables meaning "level specification". But also, in this PR I could not find a case where this variable appears.

Ok. Thanks for the information. Would you please then rename "ls" to "level_specificiation" to make this clear. Adding a type annotation would help here too.

PR #1440, opened 2025-07-31
Derive `Expression` inside Builtin `eval()` methods

rocky:
Introduce `get_eval_Expression()`
    
 `get_eval_Expression()` can be used in Mathics3 Builtin evaluation methods to find the Expression that caused the evaluation to get triggered.
    
It is often needed in error messages of "eval()" methods to report what `Expression` is in error.

PR #1442, opened 2025-07-31
Alternative assignment rework

mmatera:
This is an alternative, less disruptive implementation for the assignment refactor.


mathics/builtin/assignments/assignment.py
```diff
@@ -137,9 +137,9 @@ class Set(InfixOperator):
 
     summary_text = "assign a value"
 
-    def eval(self, lhs, rhs, evaluation):
-        "lhs_ = rhs_"
-
+    def eval(self, expr, evaluation):
+        "Pattern[expr, _ = _]"
```
rocky: See #1440 for how to do this without introducing `Pattern[expr, ...`

rocky:
I think this is much better.  

I really like that we are starting to add tests for this.  But more on that some other time. 

 I would like to spend some time going over this in more detail. 

This is good and important work that you are doing in these PRs. Although we might not get everything right on the first round, I'd like to understand what's going on and make sure we don't have to do stuff that causes problems should we decide to revise.

rocky:
> This is an alternative, less disruptive implementation for the assignment refactor.

The "alternative" PR for this is #1441, right? So only one of these PR's would go through. 

In what way is this "less disruptive" than #1441?  If anything, I am seeing more code changes.

mmatera:
No, this is the assignment refactor *based* on #1441 instead of #1428.  Is less disruptive because does not requires so many  changes in mathics.core.builtins

rocky:
> No, this is the assignment refactor _based_ on #1441 instead of #1428. Is less disruptive because does not requires so many changes in mathics.core.builtins

Ok, got it. I've suggested changes to #1441. Let's get them merged into master first. 

Do you want to make the changes, or should I make changes and put in a PR to go into #1441?

mathics/builtin/assignments/assignment.py
```diff
@@ -355,14 +353,21 @@ class UpSet(InfixOperator):
...
     summary_text = (
         "set value and associate the assignment with symbols that occur at level one"
     )
 
-    def eval(
-        self, lhs: BaseElement, rhs: BaseElement, evaluation: Evaluation
-    ) -> Optional[BaseElement]:
-        "lhs_ ^= rhs_"
+    def eval(self, expr: BaseElement, evaluation: Evaluation) -> Optional[BaseElement]:
+        "Pattern[expr, _ ^=_]"
```
rocky: I've been thinking about the use of `Pattern[expr]` to _create_ a placeholder for `expr` as opposed to using something like `Expression()` or `to_expression()` and I don't see using `Pattern[]` here as an improvement, even in those cases where `get_eval_Expression()` can't be used. (More on `get_eval_Expression()` below).

Previously, I mentioned the fact that now the docstring or pattern is a little more complicated and harder to understand the intent &mdash; capturing is put first and foremost, while everything else follows on the line.

It was claimed that this was somehow also _faster_. And here, I am not so sure...

First, upon loading, there is code to read the docstring and process it. This is _up-front_ work, whether or not the builtin is ever used. Second, on each evaluation, there is code to bind this variable. So here the question is about whether that work is less than the work done by `Expression` or `to_expression()`. Well, if this is done only to provide a more complete error message, then one expects that most of the time and in those cases where it matters most, doing lazy expression creation is also a time win.

Of course, this concern to me is a _secondary_ concern _after_ the misdirection aspect of adding `Patter[expr]` in a docstring describing the matching pattern, which causes a function to get invoked.

Now let me come back to `get_eval_Expression()` and how that is different from `Expression()`. In PR #1446, I made a pass over the Mathics3 Python code to convert `Expression()` and `to_expression()` into `get_eval_Expression()`. Here is what I learned from this.

One difference that can occur in the form of the expression. `get_eval_Expression()` shows the expression closer to how it was entered, whereas using `Expression()`, you can tweak this however you want to. But typically, the expression will reflect the reordering of parameters and canonicalization.

This kind of thing is most apparent in algebraic expressions.

The other "difference" is that Cython can have trouble introspecting the call stack. I guess this has to do with its "optimization". Personally, I think we should ditch Cython run over Mathics3 modules, unconditionally. As best as I can tell, it is not giving us any speed advantage nowadays. As for using it as a type checker, I think nowadays the other type checkers, like Mypy, are much better.

mathics/builtin/assignments/assignment.py
rocky: On further inspection, we don't use `expr` here at all. So the original seems better to me.

mmatera: Ok, let's return to the previous pattern.

mathics/builtin/assignments/assignment.py
mmatera: > I've been thinking about the use of `Pattern[expr]` to _create_ a placeholder for `expr` as opposed to using something like `Expression()` or `to_expression()` and I don't see using `Pattern[]` here as an improvement, even in those cases where `get_eval_Expression()` can't be used. (More on `get_eval_Expression()` below).
> 

OK, in any case, it was not the main goal of this PR, but I found the chance to check if this work and see if there was some consensus to change in that direction.

> Previously, I mentioned the fact that now the docstring or pattern is a little more complicated and harder to understand the intent — capturing is put first and foremost, while everything else follows on the line.
> 
> It was claimed that this was somehow also _faster_. And here, I am not so sure...
> 
> First, upon loading, there is code to read the docstring and process it. This is _up-front_ work, whether or not the builtin is ever used. 

OK, but this overhead happens at load time, and probably can be avoided if we compile and store the rules. 

> Second, on each evaluation, there is code to bind this variable. So here the question is about whether that work is less than the work done by `Expression` or `to_expression()`. Well, if this is done only to provide a more complete error message, then one expects that most of the time and in those cases where it matters most, doing lazy expression creation is also a time win.
> 
Actually, what would be faster would be something like `expr:Set[___]`, just because to match the pattern -when there are no other rules- would be faster than matching ` Set[x_,y_]`. In any case, this would impact performance only if we use several assignment statements in a loop. 


> Of course, this concern to me is a _secondary_ concern _after_ the misdirection aspect of adding `Patter[expr]` in a docstring describing the matching pattern, which causes a function to get invoked.
> 
> Now let me come back to `get_eval_Expression()` and how that is different from `Expression()`. In PR #1446, I made a pass over the Mathics3 Python code to convert `Expression()` and `to_expression()` into `get_eval_Expression()`. Here is what I learned from this.
> 
> One difference that can occur in the form of the expression. `get_eval_Expression()` shows the expression closer to how it was entered, whereas using `Expression()`, you can tweak this however you want to. But typically, the expression will reflect the reordering of parameters and canonicalization.
> 

An even simpler way to find the expression that is actually evaluated would be to store it in another attribute of the `Evaluation` object. This does not require any call to introspection functions or loops: just access to a Python variable. In any case, I think the proposal of using `get_eval_Expression()` is good in connection with the current development of the debugger API.

> This kind of thing is most apparent in algebraic expressions.
> 
> The other "difference" is that Cython can have trouble introspecting the call stack. I guess this has to do with its "optimization". Personally, I think we should ditch Cython run over Mathics3 modules, unconditionally. As best as I can tell, it is not giving us any speed advantage nowadays. As for using it as a type checker, I think nowadays the other type checkers, like Mypy, are much better.

This is another problem. In any case, this is a different problem. By now, since the goal is to improve `mathics.eval.assignment` module, I removed the changes in  `mathics.builtin` except for the ones required to pass the tests.

mmatera: > An even simpler way to find the expression that is actually evaluated would be to store it in another attribute of the `Evaluation` object. This does not require any call to introspection functions or loops: just access to a Python variable. In any case, I think the proposal of using `get_eval_Expression()` is good in connection with the current development of the debugger API.

See #1450 as an example of this approach

test/builtin/test_assignment.py
```diff
@@ -276,7 +268,7 @@ def test_set_and_clear(str_expr, str_expected, msg):
         (None, None, None),
         (
             (
-                "A[x_]:=B[x];B[x_]:=F[x_];F[x_]:=G[x];"
+                "A[x_]:=B[x];B[x_]:=F[x];F[x_]:=G[x];"
```
rocky: In this test, `SetDelayed` or `:=` is getting used. That is at a level _below_  the generic `mathics.builtin`. Either rewrite this to avoid using `SetDelayed` and the `Set`-like built-in functions, or move it under `test/builtin/assignments`. Thanks.

mathics/builtin/patterns/composite.py
```diff
@@ -173,6 +173,11 @@ def match(self, expression: Expression, pattern_context: dict):
         #     yield new_vars_dict, rest
         self.pattern.match(expression, pattern_context)
 
+    def get_sort_key(self, pattern_sort=True):
+        if not pattern_sort:
```
rocky: Instead of parameter `patten_sort`, why not define function `get_pattern_sort_key()`? 

It would be nice to add a docstring that explains that _how_ each sort is used, e.g. `get_sort_key` is used in ordering of expression elements. While `get_pattern_sort_key` is used to _prioritize_ which rule to select.

rocky:
I don't like the invented term "focus" to mean lhs without conditions, if that's what it is. Please change to something that conveys the meaning better. Thanks.

mmatera:
> something

target? main_expression?

rocky:
> > something
> 
> target? main_expression?

How about `lhs_without_condition` ?

mmatera:
> > > something
> > 
> > 
> > target? main_expression?
> 
> How about `lhs_without_condition` ?

without condition, holdpattern, exception,... stripped_lhs? bare_lhs?

rocky:
> > > > something
> > > 
> > > 
> > > target? main_expression?
> > 
> > 
> > How about `lhs_without_condition` ?
> 
> without condition, holdpattern, exception,... stripped_lhs? bare_lhs?

I don't understand. Is it correct to say this is the left-hand side with Condition expressions removed? 

If so, `lhs_without_condition` is reasonable. It has the "left-hand" part, "without" part and "condition" part in there and these are the important parts. If you want to use abbreviations for "condition", that is okay. lhs is already a common abbreviation that is used in the code. 

(rms in GNU Emacs once used the French "sans" for "without"; it is has a slightly culinary meaning in English as well, but Spanish "sin" would be kind of funny appropriate; but it would be confusing too since "sin" is an English word.)

mathics/builtin/patterns/composite.py
mmatera: See comment at https://github.com/Mathics3/mathics-core/pull/1447/files#r2252687143

mmatera:
> > > > > something
> > > > 
> > > > 
> > > > target? main_expression?
> > > 
> > > 
> > > How about `lhs_without_condition` ?
> > 
> > 
> > without condition, holdpattern, exception,... stripped_lhs? bare_lhs?
> 
> I don't understand. Is it correct to say this is the left-hand side with Condition expressions removed?
> 
> If so, `lhs_without_condition` is reasonable. It has the "left-hand" part, "without" part and "condition" part in there and these are the important parts. If you want to use abbreviations for "condition", that is okay. lhs is already a common abbreviation that is used in the code.
> 
> (rms in GNU Emacs once used the French "sans" for "without"; it is has a slightly culinary meaning in English as well, but Spanish "sin" would be kind of funny appropriate; but it would be confusing too since "sin" is an English word.)

`focus` was introduced by Angus in the old Mathics code, meaning that is the part of the expression that defines which definition is going to be modified. I guess focus is in the sense of *focal point* in optics. This does not mean exactly to remove the wrapping `Condition[...]`, but also other modifiers like `HoldPattern` or `Except`, or `Pattern[...]`.  
For example, if the lhs is
`expr:HoldPattern[Condition[F[Q[x_],y],x>0]]`
The assignment is done to a DownValue of `F`, or an `UpValue` of  `Q`, and not on `HoldPattern`, depending on the kind of assignment (Set,TagSet, or UpSet).

This is why I think that names like  `lhs_keine_condition`  are not good names.

rocky:
> > > > > > something
> > > > > 
> > > > > 
> > > > > target? main_expression?
> > > > 
> > > > 
> > > > How about `lhs_without_condition` ?
> > > 
> > > 
> > > without condition, holdpattern, exception,... stripped_lhs? bare_lhs?
> > 
> > 
> > I don't understand. Is it correct to say this is the left-hand side with Condition expressions removed?
> > If so, `lhs_without_condition` is reasonable. It has the "left-hand" part, "without" part and "condition" part in there and these are the important parts. If you want to use abbreviations for "condition", that is okay. lhs is already a common abbreviation that is used in the code.
> > (rms in GNU Emacs once used the French "sans" for "without"; it is has a slightly culinary meaning in English as well, but Spanish "sin" would be kind of funny appropriate; but it would be confusing too since "sin" is an English word.)
> 
> `focus` was introduced by Angus in the old Mathics code, meaning that is the part of the expression that defines which definition is going to be modified.


The common programming language term for this is: reference, and here, left-hand-side reference. 

>  I guess focus is in the sense of _focal point_ in optics.

The problem with "focus" as used here is that there are numerous ways this word could be interpreted and used as an analogy. Like starting with whether the meaning is used as a noun or a verb. 

In places where you want to appropriate a term from one domain to another, the term should have a very clear, specific, and unambiguous meaning.  For example, when Robert Tarjan took the term "amortized" from finance and applied it to algorithms, it is very clear what that means. 

 > This does not mean exactly to remove the wrapping `Condition[...]`, but also other modifiers like `HoldPattern` or `Except`, or `Pattern[...]`. For example, if the lhs is `expr:HoldPattern[Condition[F[Q[x_],y],x>0]]` The assignment is done to a DownValue of `F`, or an `UpValue` of `Q`, and not on `HoldPattern`, depending on the kind of assignment (Set,TagSet, or UpSet).
> 
> This is why I think that names like `lhs_keine_condition` are not good names.

Okay, fair enough. How about `lhs_reference` ?

test/builtin/test_assignment.py
mmatera: In this case, maybe is the same to use `=` and `:=`. But you mean to use the FullForm of the expressions?

mmatera:
> `lhs_reference`

`lhs_reference` sounds good.

rocky:
> > `lhs_reference`
> 
> `lhs_reference` sounds good.

Thanks for your patience and tolerance.

test/builtin/test_assignment.py
rocky: > In this case, maybe is the same to use `=` and `:=`. But you mean to use the FullForm of the expressions?

No. I guess I am not being clear here. I was using this as an example of the entire suite of tests that seems to be testing the `Set*` functions which appear in module `mathics.builtin.assignments`, not `mathics.builtin`.

But let me back up. What is the intention of this suite of tests?

As a principle, code in `test/builtin` should not use any code from the builtin modules under such as any of the `Set` functions.

mmatera:
> > > `lhs_reference`
> > 
> > 
> > `lhs_reference` sounds good.
> 
> Thanks for your patience and tolerance.

Sure. Thank you for your careful revision!

test/builtin/test_assignment.py
mmatera: @rocky, at this point, I have moved the tests in `test.builtin.test_assignment` to `test.builtin.assignment.test_assignment`, so this would address your comment, isn't it?

rocky: Yes, it does. Thanks!

mathics/core/definitions.py
```diff
@@ -161,9 +161,6 @@ def __init__(
             "Global`",
         )
         self.inputfile = ""
-
-        # These are used by TraceEvaluation to```
-        # whether what information to show.
         self.trace_evaluation = False
```
rocky, line 166:
```suggestion
        # TraceEvaluation uses these to 
        # decided what information to show
        self.trace_evaluation = False
```

test/builtin/assignments/test_assignment.py
```diff
@@ -1,20 +1,18 @@
 # -*- coding: utf-8 -*-
 """
 Unit tests for mathics.builtins.assignments.assignment
+
+Tests here looks for reinforce the compatibility of
```
rocky:
```suggestion
Tests here check the compatibility of
```

rocky (approved):
LGTM - but see suggested comment changes.

mmatera:
OK, let's go on with this, and iterate

PR #1443, opened 2025-07-31
Add get_eval_Expression()

rocky:
Introduce `get_eval_Expression()`

`get_eval_Expression()` can be used in Mathics3 Builtin evaluation methods to find the Expression that caused the evaluation to get triggered.

mathics.file_io.files module converted to use this function to test worthiness.

### DIFF
--- a/mathics/builtin/files_io/files.py
+++ b/mathics/builtin/files_io/files.py
@@ -49,6 +49,7 @@ from mathics.eval.files_io.read import (
     read_name_and_stream,
 )
 from mathics.eval.makeboxes import do_format, format_element
+from mathics.eval.stackframe import get_eval_Expression
 
 
 class Input_(Predefined):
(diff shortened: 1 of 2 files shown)


mathics/builtin/files_io/files.py
```diff
@@ -608,11 +609,10 @@ def do_format_output(expr, evaluation):
 
         return SymbolNull
 
-    def eval_default(self, exprs, filename, evaluation):
+    def eval_default(self, exprs, filename, evaluation: Evaluation):
         "Put[exprs___, filename_]"
-        expr = to_expression("Put", exprs, filename)
         evaluation.message("Put", "stream", filename)
-        return expr
+        return to_expression("Put", exprs, filename)
```
rocky, line 615: using get_eval_Expression() causes an infinite reevaluation loop here. I don't know why, but the problem doesn't have anything to do with `get_eval_Expression()`.

mmatera: It is probably related to the precedence order of the rules. I find something similar in the previous attempt to rewrite these rules.

mmatera: @rocky, since the rest works, just add this observation as a comment. The rest I think is very good. Merge when you feel it is ready.

PR #1445, opened 2025-08-01
Add sorting test

mmatera:
Just adding a test to check how expressions should be canonically sorted.


rocky:
The intent is good, even if we don't match WMA at the moment. I like that we will have more tests like this. 

If the tests use `check_evaluation`  and that mechanism, this is a test of the `Sort[]`  Builtin Function and should be moved to  `test.builtin.statistics.orderstats`. 

A test in `test.core` should not use Builtin Functions for testing, but instead stick inside the functions and lower-level API that stay within  `mathics.core`. 

Additionally, if rewritten as a core test using the API that this provides, there would be no question or need for using `scikit-image`.

mmatera:
I do not fully get why  `Sort` is inside `mathicsbuiltin.statistics.orderstats`, but OK...

rocky:
> I do not fully get why `Sort` is inside `mathicsbuiltin.statistics.orderstats`, but OK...

This came from https://reference.wolfram.com/language/guide/DescriptiveStatistics.html  but if you want to move this to another section, like https://reference.wolfram.com/language/guide/RearrangingAndRestructuringLists.html, that would be fine. 

But anyway, thanks for moving the test. Feel free to merge when you are satisfied.

PR #1447, opened 2025-08-02
Split key comparable

mmatera:
This reimplements #1434 by using constants, and with minimal changes in `mathics.builtin`.


mathics/core/keycomparable.py
```diff
@@ -0,0 +1,277 @@
...
+        raise NotImplementedError
+
+    def __eq__(self, other) -> bool:
+        return (
+            hasattr(other, "get_sort_key")
+            and self.get_sort_key() == other.get_sort_key()
+        )
+
+    def __gt__(self, other) -> bool:
+        return self.get_sort_key() > other.get_sort_key()
```
mmatera, line 87: @rocky, I glad you asked about this in https://github.com/Mathics3/mathics-core/pull/1442/files#r2251813566.

This is why we do not have two different methods for expression and pattern sort keys:

Expressions and Patterns are subclasses of this class. The methods on this class are called in different contexts, using the default value of the `get_sort_key`, depending on we are sorting rules or expressions.

When you call `sort(expr_list)` over a list of expressions `expr_list=[expr1, expr2,...]`, since the default  value of `pattern_sort`  is `False`, the list is ordered using the Expression criteria.

On the other hand, if we call  `sort(rule_list)` over a list of `Rule` or `Pattern` objects  `rule_list=[rule_1, rule_2,..]` since the default value of  `pattern_sort`  is `True` for these classes, the rules are sorted using the pattern sort keys. This is a clever but obfuscated way to achieve the right ordering in each context. Maybe there is a better way to do that, but I could not find it yet.

I planned to document this in some place if at the end of all this refactoring, I do not find a better way to implement the same behavior.

rocky: > @rocky, I glad you asked about this in https://github.com/Mathics3/mathics-core/pull/1442/files#r2251813566.
> 
> This is why we do not have two different methods for expression and pattern sort keys:
> 
> Expressions and Patterns are subclasses of this class. The methods on this class are called in different contexts, using the default value of the `get_sort_key`, depending on we are sorting rules or expressions.
> 
> When you call `sort(expr_list)` over a list of expressions `expr_list=[expr1, expr2,...]`, since the default value of `pattern_sort` is `False`, the list is ordered using the Expression criteria.
> 
> On the other hand, if we call `sort(rule_list)` over a list of `Rule` or `Pattern` objects `rule_list=[rule_1, rule_2,..]` since the default value of `pattern_sort` is `True` for these classes, the rules are sorted using the pattern sort keys. This is a clever but obfuscated way to achieve the right ordering in each context. Maybe there is a better way to do that, but I could not find it yet.
> 
> I planned to document this in some place if at the end of all this refactoring, I do not find a better way to implement the same behavior.

There is something really wrong with the design here.

Pattern prioritization and element sorting are fundamentally different things. So calling them the same thing or "unifying" they is going to lead to confusion and a massive explanation.

When `sort(rule_list)` is called, we know that we want pattern prioritization, so the key function can be set to pattern ordering can be set. Can't that also be done in other cases where comparison or sorting is done?

If you would like me to make a stab at changing the code for this, let me know.

mmatera: @rocky, if you are up for it, your changes are welcome. However, please consider looking at #1449 first and committing over it to avoid merge conflicts. Thanks!

rocky: Okay, I will start from #1449.  I don't know if I'll be successful, but at least I'll understand what's up. (I will be out most of Tuesday, though.)

mmatera:
Something still missing here is replacing uses of `get_sort_key` with the corresponding new properties. I will do that after finishing with this PR.

rocky:
LGTM, but please fix the mypy type mismatches. 

I have a slight preference for using the word "PRECEDENCE" when deciding which pattern goes first, but "SORT_KEY" is not wrong.

PR #1449, opened 2025-08-02
Using magic numbers instead of tuples in sort keys

mmatera:
This is the continuation of #1447, using integers instead of tuples for sort_keys. Names of the keys are debatable, but I think it starts to look more like how it should look.


test/core/test_rules.py
```diff
@@ -193,9 +193,9 @@ def test_default_optional_on_rules(str_expr, str_expected, msg):
         ("rule=A[a_.+B[b_.*x_]]->{a,b,x};", None, "Null", None),
         ("A[B[1]] /. rule", None, "{0, 1, 1}", None),
         ("A[B[x]] /. rule", None, "{0, 1, x}", None),
-        ("A[B[2*x]] /. rule", None, "{0, x, 2}", None),
+        ("A[B[2*x]] /. rule", None, "{0, 2, x}", None),
```
mmatera, line 195: Accidentally, a couple of tests now gives the same output I get  in WMA:
```
In[1]:= rule=A[a_.+B[b_.*x_]]->{a,b,x};                                         

In[2]:= A[B[2*x]] /. rule                                                       

Out[2]= {0, 2, x}

In[3]:= A[1+B[2*x]] /. rule                                                     

Out[3]= {1, 2, x}

```

test/core/test_rules.py
```diff
@@ -65,15 +65,15 @@
         (
             "SetAttributes[Q, {Orderless}];\
           {Q[a,1,b], Q[a,1,b]/.rule, Q[a, 1, b]/.ruled}",
-            "{Q[1, a, b], Q[a, 1, b], Q[a, 1, b]}",
-            "2. Set the attribute. Application is not affected.",
+            "{Q[1, a, b], True, Q[a, 1, b]}",
```
rocky, line 68: This test suite is in the wrong place. It should be under `test/builtin/patterns` or the tests rewritten.

We are using Built-in functions here. Compare with `test/core/test_pattern.py` which does not stray outside of `mathics.core`.

PR #1450, opened 2025-08-03
Store current expression in evaluation before rule apply function

mmatera:
Just an example of an alternative to make the Expression available in `eval_` method calls.


rocky:
I think this is a good idea.

rocky:
LGTM

PR #1451, opened 2025-08-05
Split get_sort_keys() into pattern_precedence and element_precedence

rocky:


rocky:
@mmatera I now recall that this is one of the big messes we've encountered in the past.  It complicates expression and box evaluation.  

I don't know how clean we'll get this. It may take several rounds. But let's start straightening this out.

PR #1455, opened 2025-08-11
remove some error checking-only eval methods and two small tweaks

rocky:
Remove two builtin eval-methods that existed only to issue some errors.

Instead, the tests which are parameter type checks, have been folded into the broader evaluation method for that builtin.

Also, add Makefile target `pytest-x` option to stop on first pytest failure. 
Add a ReduceList pytest.


mmatera (approved):
LGTM

PR #1456, opened 2025-08-11
Bye-bye get sort key

mmatera:
This completes the rework of the implementation of canonical and pattern precedence ordering.
The question now is whether we should keep the `get_sort_key` method with a deprecation warning, or just remove it.


mathics/core/definitions.py
```diff
@@ -1018,7 +1018,7 @@ def is_conditional(x) -> bool:
 
     # use insort_left to guarantee that if equal rules exist, newer rules will
     # get higher precedence by being inserted before them. see DownValues[].
-    bisect.insort_left(values, rule)
+    bisect.insort_left(values, rule, key=lambda x: x.pattern_precedence)
```
mmatera, line 1021: @rocky, the *half-smart* implementation seems to be required to support Python <=3.9, because `bisect.insort_key` does not support the parameter `key` in that version.

rocky: Yes, I saw that. I am also happy to drop support for Python 3.9 here. We'll be adding 3.14 soon when that comes out.

Your thoughts?

mathics/core/keycomparable.py
```diff
@@ -164,6 +164,64 @@ def __ne__(self, other) -> bool:
 )
 
 
+BLANK_WITH_PATTERN_PATTERN_SORT_KEY = (
```
rocky, line 219: It would be nice to add a comment containing an example pattern for each of these. This would provide someone with some clue as to the rationale behind the numbers as they change.

mathics/core/keycomparable.py
```diff
@@ -236,52 +216,50 @@ def __ne__(self, other) -> bool:
...
 # Blanks
-
 BLANK_WITH_PATTERN_PATTERN_SORT_KEY = (
     BASIC_EXPRESSION_PATTERN_SORT_KEY + PATTERN_SORT_KEY_BLANK_WITH_HEAD
-)
+)  # Blank[A]
 BLANK_GENERAL_PATTERN_SORT_KEY = (
     BASIC_EXPRESSION_PATTERN_SORT_KEY + PATTERN_SORT_KEY_BLANK_PURE
-)
+)  # Blank[]
```
mmatera: something like this?

rocky: Yes, and you could also give the operator form like `_`

rocky:
LGTM - merge when you are satisfied.

mathics/core/keycomparable.py
```diff
@@ -19,8 +19,12 @@ class KeyComparable:
     This class is not complete in of itself, it is intended to be
     mixed into other classes.
 
-    Each class should provide a `get_sort_key()` method which
+    Each class should provide a `element_order` property which
     is the primitive from which all other comparisons are based on.
+
+    The class also contains a `pattern_precedence` property that provides
+    the sort key used to order a list of rules according to the
+    precedence they have in the evaluation loop.
```
rocky, line 27: Um. Isn't it _more_ than just the evaluation loop? Isn't this also used by Builtin functions like MatchQ?

mmatera: `MatchQ` takes a single pattern, so there is nothing to sort.

rocky: Interesting. I looked at the WMA pattern-matching built-in functions, and indeed, I couldn't find any built-ins that work with a _collection_ of patterns. The usual kind of thing that I see is that there is the input parameters of a built-in function can be a list of expressions, which are matched against a _single_ pattern. 

So it looks like all this precedence mechanism is solely for deciding which rule or function to apply in expression evaluation.

mathics/core/keycomparable.py
```diff
@@ -218,23 +218,23 @@ def __ne__(self, other) -> bool:
...
+)  # BlankSequence[] (`pat__`)
 
 BLANKNULLSEQUENCE_WITH_PATTERN_PATTERN_SORT_KEY = (
     BASIC_EXPRESSION_PATTERN_SORT_KEY + PATTERN_SORT_KEY_BLANKNULLSEQUENCE_WITH_HEAD
-)  # BlankNullSequence[A]
+)  # BlankNullSequence[A] (`pat___A`)
 BLANKNULLSEQUENCE_GENERAL_PATTERN_SORT_KEY = (
     BASIC_EXPRESSION_PATTERN_SORT_KEY + PATTERN_SORT_KEY_BLANKNULLSEQUENCE_PURE
-)  # BlankNullSequence[]
+)  # BlankNullSequence[] (`pat___`)
```
rocky, line 237: Thank you very much!

rocky:
LGTM - merge when you are satisfied.

PR #1462, opened 2025-08-14
Assume @dataclasses exists and other small cleanups

rocky:
The dataclass module has existed since 3.7+

Reduce use of `get_{int,float}_value()`. 

Add Makefile targets `doctest-x`, `check-x` which stop on first error.


mmatera:
LGTM

PR #1466, opened 2025-08-16
Add argument mismatch checks and error messages

rocky:
Specifically to CubeRoot, Divide, Minus, Sqrt, and Subtract


mmatera (approved):
LGTM

rocky:
Thre are probably others, but we can add them in due time.

PR #1468, opened 2025-08-17
Even more error checking DRYness

rocky:
DRY error checking in `Round[]` and `Sign[]`.

### DIFF
--- a/mathics/builtin/numeric.py
+++ b/mathics/builtin/numeric.py
@@ -19,6 +19,7 @@ from mathics.core.atoms import (
     Complex,
     Integer,
     Integer0,
+    Integer1,
     IntegerM1,
     Number,
     Rational,
@@ -41,6 +42,7 @@ from mathics.core.symbols import (
     SymbolDivide,
     SymbolFalse,
     SymbolMachinePrecision,
+    SymbolPlus,
     SymbolTimes,
     SymbolTrue,
 )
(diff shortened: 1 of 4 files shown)


mathics/builtin/numeric.py
```diff
@@ -707,17 +708,28 @@ class Round(Builtin):
 
     attributes = A_LISTABLE | A_NUMERIC_FUNCTION | A_PROTECTED
 
+    # Set checking that the number of arguments required is one or two.
+    eval_error = Builtin.generic_argument_error
+    expected_args = (1, 2)
+
     rules = {
-        "Round[expr_?NumericQ]": "Round[Re[expr], 1] + I * Round[Im[expr], 1]",
```
rocky, line 711: This rule is called for handling a one-argument `Round[]` call which in part sets the default second argument of `Round[]` to 1.  

Looking at TraceEvaluation for, say `Round[5.5]`, we see a _lot_ of activity. It is stuff like this (all over the code historically) that I suspect is part of the reason we have been slow. 

I wouldn't object if it added _clarity_, but, in truth, it obfuscates. So the end result is that we historically had "clever" obfuscated code, which is _also_ slow code.

rocky: Note added after a test using Complex round failed. We keep some form of rewriting for rounding complex numbers. And in wolframscript, `Trace[6/(2 + 3I)]`, performing rewrites. 

However, the above code, which tests on `NumericQ` was rewriting in the simple and most common situations of rounding Reals of various sorts.  WMA does not apply any rules for these common occurrences. 

So while we don't _remove_ this, we at least _simplify_ it and narrow its scope.

Issue #1432, opened 2025-07-28
URL links with `$` in them are getting messed up


rocky:
## Description

WMA URLs in both Django and the TeX documents are getting messed up due to `$` substitution. 

## How to Reproduce and Output Given

In Django look at the URL for `$MachineName`.  I get `https://reference.wolfram.com/language/ref/MachineName.html`. 

Similarly,  in the LaTeX document this appears as 
`https://reference.wolfram.com/language/ref/\protect \TU\textdollar Machine.html`

## Expected behavior

For Django: `https://reference.wolfram.com/language/ref/$MachineName.html`. 
For LaTeX: `https://reference.wolfram.com/language/$Machine.html`

## Additional context

This was noticed while going over the documentation for a new release.

PR #1469, opened 2025-08-17
fix 1432

mmatera:
This PR fixes the issue #1432 by removing the escape character in front of dollar characters in URLs. The escape is necessary at the level of docstrings to avoid the confusion with the open character for a LaTeX equation. Once the expression is parsed, and we are inside a `<URL>` tag,  it is safe to remove the escape character.

### DIFF
--- a/mathics/builtin/system.py
+++ b/mathics/builtin/system.py
@@ -238,7 +238,7 @@ class Machine(Predefined):
 
 class MachineName(Predefined):
     """
-    <url>:WMA link:https://reference.wolfram.com/language/ref/MachineName.html</url>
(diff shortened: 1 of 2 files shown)


rocky:
Ok, but we need to find and fix all the other 40-50 other places.

rocky:
This works for the one case done here.  Let's merge, and I'll work on the others tomorrow.

mmatera:
If we agree in the approach, then it is just to adjust all the other cases.

rocky:
> If we agree in the approach, then it is just to adjust all the other cases.

It's not optimal, but the entire doc system has problems. #1470 Has the remaining changes. 

With this, we should be ready for release?

PR #1484, opened 2025-08-24
Improve traceprint

mmatera:
This PR changes `TracePrint` to show the result through the  `Print[]` interface. This allows us to show the result in the browser/CLI/docpipeline instead of the (raw) terminal.  

This is how it looks in the PDF documentation:
imagen

and this is how it looks in Mathics-Django help
imagen


rocky:
LGTM

PR #1485, opened 2025-08-25
More parameter checks

rocky:
Checks added to 'If', 'While', and 'Throw'. 

We start to handle 'argb' tag, "between x and y".

### DIFF
--- a/mathics/builtin/procedural.py
+++ b/mathics/builtin/procedural.py
@@ -380,12 +380,29 @@ class If(Builtin):
     You might use comments (inside '(*' and '*)') to make the branches of 'If'
     more readable:
     >> If[a, (*then*) b, (*else*) c];
+
+
+    Since one or more arugments to a boolean operation could be symbolic, it is\
+    possible that an 'If' cannot be evaluated. For example:
+
+    >> Clear[a, b]; If [a < b, a, b]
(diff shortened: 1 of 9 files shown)


mathics/builtin/procedural.py
```diff
@@ -377,15 +389,32 @@ class If(Builtin):
...
     """
 
-    summary_text = "if-then-else conditional expression"
     # This is the WR summary: "test if a condition is true, false, or
     # of unknown truth value"
     attributes = A_HOLD_REST | A_PROTECTED
+
+    # Set checking that the number of arguments required between 2 and 4.
+    eval_error = Builtin.generic_argument_error
+    expected_args = range(2, 5)
```
mmatera, line 416: Is it right to assign here an iterator instead a tuple?

rocky: > Is it right to assign here an iterator instead a tuple?

The `Union` in `mathics.core.builtin` includes `range`.

rocky: Clearly the types _have_ to be different to distinguish: "2 _or_ 5",  from "_between_ 2 _and_ 5"

mmatera: My question was about whether the object must be used more than once

PR #1486, opened 2025-08-26
Shorten TraceEvaluation

rocky:
Lines of the form:

Replacing: x -> x

don't add much, so remove them.

In the future, maybe we can have a "verbosity level" to include something like this. The only purpose I can see is to indicate that a rewrite rule was considered and did nothing. However, TraceEvaluations are typically very long and, in my experience, don't add clarity, just clutter.


mmatera:
LGTM

PR #1489, opened 2025-08-31
improve Information

mmatera:
This PR is one step towards improving the compatibility of the symbol `Information` between Mathics and Mathematica >=12.0.

Since the first implementation of the Builtin associated with this symbol (I guess, my first contribution to the project, circa 2013) and now, the way in which WMA implements this symbol has changed a lot.

Until WMA 11., `Information[symb_, LongForm->longform_]`  does not have an evaluation rule, just a format rule. The format rule printed some information about  `symb`, and return `Null` (like `Print`). The `LongForm` option controled if user defined rules where shown or not.

Since WMA 12., `Information[symb_, LongForm->longform_]` now is evaluated to an `InformationData[..]`, expression, which at the time, has format rules producing the output that we can see when we evaluate `?symb` or `?? symb`.

Since to have full compatibility with the current version requires many changes, I opted by starting with just provide a closer output to the new version, using the old "style". Then,  `InformationData` and related functions can be included.

### DIFF
--- a/mathics/builtin/atomic/symbols.py
+++ b/mathics/builtin/atomic/symbols.py
@@ -24,6 +24,7 @@ from mathics.core.attributes import (
 from mathics.core.builtin import Builtin, PrefixOperator, Test
 from mathics.core.convert.expression import to_mathics_list
 from mathics.core.convert.regex import to_regex
+from mathics.core.element import BaseElement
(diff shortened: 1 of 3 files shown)


rocky:
LGTM - thanks!

@mmatera: Merge when you are satisfied.

PR #1490, opened 2025-08-31
Find* now returns the best value found after a convergency error.

mmatera:
This PR improves the compatibility of `FindRoot`,`FindMinimum` and `FindMaximum`, by changing the behavior when the convergence is not achieved up to the requested tolerance. As in WMA, now the return value is the best candidate for the solution, even if the convergence fails.


rocky:
LGTM  - merge when satisfied.

PR #1491, opened 2025-08-31
Disable debug doctests

rocky:
Disable doctest info for Mathics3-Module-trepan. If it is installed, it will go into the debugger and halt testing.

This is somewhat temporary until better solutions exist. But we shouldn't allow the possibility of making a test go into a debugger.


mmatera (approved):
LGTM.

PR #1503, opened 2025-10-28
Revise Complex representation.

rocky:
Complex.value is now something that appears to be complex in Python. Complex.exact_value, however is an internal exact representation. It contains precision values and can be a Mathics3 Integer or Real, MachineReal number.

Property is_exact_value indicates whether a value is exact when converted to a Python complex number or not.

Some properties and tests are now cached.


rocky:
There is a bit of rot when I put in proper type checking. This needs more work reflective of the looseness and haphazardness of some of the current code.

mathics/core/atoms.py
```diff
@@ -131,7 +132,7 @@ def to_mpmath(self, precision: Optional[int] = None) -> mpmath.ctx_mp_python.mpf
                 return mpmath.mpf(self.value)
         return mpmath.mpf(self.value)
 
-    def to_python(self, *_, **kwargs):
+    def to_python(self, *_, **kwargs) -> Union[int, float, complex]:
```
mmatera: For mypy, `complex` contains `int` and `float`, so `Union` is not needed.

rocky:
This PR has problems.  Some of the smaller parts will be addressed in #1518

PR #1505, opened 2025-11-07
Revise ByteArray[]. Make it Atomic and conform to WMA more closely.

rocky:
This is work is in prepration for handling NumericArray which might be similar.

### DIFF
--- a/mathics/builtin/binary/bytearray.py
+++ b/mathics/builtin/binary/bytearray.py
@@ -71,7 +71,7 @@ class ByteArray_(Builtin):
         return String(f"ByteArray[<{len(baa.value)}>]")
 
     def eval_normal(self, baa, evaluation: Evaluation):
-        "System`Normal[ByteArray[baa_ByteArray]]"
+        "System`Normal[baa_ByteArray]"
         return to_mathics_list(*baa.value, elements_conversion_fn=Integer)
 
     def eval_list(self, values, evaluation) -> Optional[ByteArray]:


rocky:
@mmatera Bruce Lucas has been exploring what is needed to support Plotting, e.g. Plot3D, graphics and a jupyter-like interface with a performant `Manipulate[]`.  

A lot of details will follow later, but one clear thing is that having an implementation of NumericArray will be one piece that will move a lot of the code in the faster direction.

rocky:
Todo: 
- [x] Rename `ByteArrayAtom` to `ByteArray`.
- [x] Revise so `First[]` and `Last[]` work.
- [x] Split out sort order Strings from ByteArrays. These are not comparable without conversion which we don't want to do.
<strike>- [ ] DRY Sort eval routine, by adding a Rule for default behavior.</strike>

Later:

-  [ ] split  `mathics.atomic`  into modules `mathics.atomic.numeric`, `matics.atomic.array`, `mathics.atomic.string` and `mathics.atomic.symbol`.
- [ ] fill out ByteArray - releated functions.

Currently, I'm implementing `First[]` and `Last[]` in a non-optimal way. The code is expanding all elements and letting Python's `__getitem__` method pick out the requested element. 

Since ByteArray (and other Array's) have a way to do this _without_ expanding, a custom `__getitem__` function should be written. 

But this PR is getting large so I plan on doing that in a separate PR.

mathics/builtin/list/constructing.py
```diff
@@ -198,6 +198,8 @@ class Normal(Builtin):
     def eval_general(self, expr: Expression, evaluation: Evaluation):
         "Normal[expr_]"
         if isinstance(expr, Atom):
+            if hasattr(expr, "elements"):
```
bdlucas1: This is similar to but maybe not quite the same thing as elements in a non-Atomic Expression, so I wonder if it should have a different name, like "parts" maybe. For example, do we want code that is looking for elements in a general expression to see this - should Atoms appear, well, atomic except in some particular circumstances like this?

This could also alleviate my concern about performance if some code were to unwittingly look at .elements and thereby cause a performance issue.

mathics/builtin/list/eol.py
```diff
@@ -685,14 +694,22 @@ def eval(self, expr, evaluation: Evaluation, expression: Expression):
...
+                evaluation.message("First", "normal", Integer1, expression)
+                return
         expr_len = len(expr.elements)
         if expr_len == 0:
             evaluation.message("First", "nofirst", expr)
             return
+
+        if isinstance(expr, Atom):
+            # ByteArray or NumericArray, ...
+            return expr.elements[0]
```
bdlucas1: This same treatment could be given to Part.

mathics/builtin/list/constructing.py
```diff
@@ -198,6 +198,8 @@ class Normal(Builtin):
...
+                return ListExpression(*expr.elements)
```
bdlucas1: This needs to happen recursively for NumericArray - not sure this does that?

mathics/builtin/list/constructing.py
rocky: For ByteArray this is the right thing to do, so I'll change this to be more specific.

bdlucas1: To be clear, my question was not really about this specific line of code, but rather about whether the ByteArray Atom (or Atoms in general) ought to have a .elements property - would code that is looking for .elements be surprised to find an Atom that has .elements? I didn't seem any other examples of such Atoms, and naively it seems surprising to me that an Atom might have elements.

For example, is there code that will end up innocently looking at ByteArray.elements simply because .elements exists, and thereby cause a performance (or functional) problem?

So my question was instead of .elements, would it make sense for ByteArray and NumericArray to have some other propertly, e.g. .parts, that triggers the behavior in this line of code.

rocky: Yes, I sensed that from your other questions. I haven't figured out what my opinion is on this. 

"Elements" is the WMA term, and it is better than using "parts" which isn't the phrase that is used for this kind of thing, best as I can tell.

 Whether that needs an adjective to make sure the two fields do not have the same name - dunno. 

I understand your concern, but can see another argument regarding having to add lots what feels like arbitrary checks in the code if this _or_ that kind elementness.  

So in a sense, this is like predicting the future. 

Note that there might be a common function that works on elements that we might want to use.  And once function is called the distinction will no longer exist no matter what kind of collection it initially came from.

mmatera: Maybe this helps. The following code was evaluated in WMA:
```
In[1]:= x=ByteArray[{1,0,1,1,0}]                                                

Out[1]= ByteArray[<5>]

In[2]:= AtomQ[x]                                                                

Out[2]= True

In[3]:= x[[1]]                                                                  

Out[3]= 1

In[4]:= Head[x]                                                                 

Out[4]= ByteArray

In[5]:= x[[0]]

Out[6]= List
```

mathics/builtin/list/constructing.py
mmatera: This is what happens with `NumericArray` in WMA 12
```

In[1]:= x = NumericArray[Table[x*y, {x, 10.}, {y, 10.}], "Real64"]              

Out[1]= NumericArray[<10,10>, Real64]

In[2]:= Head[x]                                                                 

Out[2]= NumericArray

In[3]:= x[[0]]                                                                  

Out[3]= List

In[4]:= AtomQ[x]                                                                

Out[4]= True

In[5]:= x[[1]]                                                                  

Out[5]= NumericArray[<10>, Real64]

In[6]:= AtomQ[x[[1]]]                                                           

Out[6]= True

In[7]:= Head[x[[1]]]                                                            

Out[7]= NumericArray

In[8]:= x[[1]][[0]]                                                             

Out[8]= List
```

Notice that `Head` returns `NumericArray` both over the whole expression as the expression from a part. But `Part[x,0]` returns `List`, which suggest some kind of conversion when `Part` is used on these objects.

mmatera: Looking at the `TracePrint` output,
```
In[9]:= TracePrint[x[[0]], TraceInternal->True]                                
 x[[0]]
  Part
  x
  ByteArray[<4>]
  0
 ByteArray[<4>][[0]]
 List

Out[9]= List
```
it seems that the conversion is *implicit*: `Part` does not create a `List` expression from `ByteArray` (or `NumericArray`) but just access the elements, and return values as it would come from a `List` expression.

mathics/eval/list/eol.py
```diff
@@ -197,7 +197,7 @@ def parts_sequence_selector(pspec):
             raise MessageException("Part", "pspec", pspec)
 
     def select(inner):
-        if isinstance(inner, Atom):
+        if not hasattr(inner, "elements"):
```
mmatera, line 200: good catch!

rocky: In 961bade57 the code is now adjusted.

mathics/builtin/list/constructing.py
bdlucas1: > checks in the code if this or that kind elementness.

The supposition behind my question of course is that no code that currently looks at .elements wants to, nor should in the future, look at the new use of .elements in Atoms because Atoms are atomic, opaque. For example, these sub-parts don't and should not participate in pattern matching (WMA doc explicitly says that for NumericArray, and I think it's also true for ByteArray). To me this makes these things seem semantically rather different from elements.

In the doc I don't see any use of the term elements in conjunction with ByteArray or NumericArray. (In other parts of the doc I spotted both elements and args being used, the former for Lists and the latter for Expressions in general).

Anyway this is all the perspective of a relative beginner in both the language and the code base, so of course I'll defer to to your judgement.

test/builtin/test_testing_expressions.py
```diff
@@ -177,3 +175,49 @@ def test_matchq(str_expr, msgs, str_expected, fail_msg):
...
+        ),
+        ('Order["a", 1000]', "-1", "String comes after Integer"),
+        ("Order[0.9, 1]", "1", "Numeric less-than comparison between Real and Integer"),
+        (
+            "Order[1.2, 1]",
+            "-1",
+            "Numeric greater than comparison between Real and Integer",
+        ),
+        # FIXME: Intermittently fails - investigate why
+        # ("Order[F[2], A[2]]", "-1", "Function ordering in function name"),
```
mmatera: Try to clean first the Definitions object by adding a first "test" `(None, None, None),`

rocky: @mmatera it feels hoaky, but okay.  Using `Clear[A, F]` seems more explicit. But I'm not totally sure what you are suggesting. If you are up to it would use use the +/- to suggest a specific commit?

mathics/builtin/list/eol.py
rocky: @mmatera Your thoughts on renaming "elements" in ByteArray (and other Arrays in the future) to something like collection_elements?

mathics/builtin/list/constructing.py
bdlucas1: FWIW, Wolfram Cloud gives NumericArray instead of List for [[0]]:

image

test/builtin/test_testing_expressions.py
```diff
@@ -177,3 +175,49 @@ def test_matchq(str_expr, msgs, str_expected, fail_msg):
         failure_message=fail_msg,
         expected_messages=msgs,
     )
+
+
+@pytest.mark.parametrize(
+    ("str_expr", "str_expected", "assert_fail_msg"),
+    [
```
mmatera, line 182:
```suggestion
    [
            # Clean the definitions, because
            # a previous definition of `A` or `F` could make
            # the test to fail.
            (None, None, None),
```

mathics/builtin/list/constructing.py
rocky: Honestly, I  too am a beginner in the language part. And sadly too often I feel like I am a begginer in mysteries and rationale behind the code base. 

I think I'm okay with renaming "elements" in XXXArrays to array_elements, if that name is acceptable. And if not suggest something else, but I feel don't like inventing new names without reason, thought, and investigation which @bdlucas1  has done.

@mmatera - your opinion on this please?

test/builtin/test_testing_expressions.py
mmatera: @rocky, start parametrized tests with `(None, None, None),` is something that we have used along many other places, to ensure working on a clean `Definitions` object. `Clear[A, F]` could still leave some side effect definitions (like MakeBoxes definitions) that could spoil the test silently.

mathics/builtin/list/eol.py
mmatera: Regarding the rename,  in evaluation it does not seem a problem, since `Atom` (except by `Symbol`) do not have an `evaluate` method. But it is true that in pattern matching, the confusion could happen. So I guess the change makes sense.  Regarding the name, what about `items`?

rocky: I'm kind of worn out on this regarding the name, it can easily be changed in the figure. @bdlucas1 If "items" is acceptable, I'll change the code.

mathics/builtin/list/constructing.py
mmatera: https://github.com/Mathics3/mathics-core/pull/1505#discussion_r2506971630

mathics/builtin/list/eol.py
bdlucas1: .items is fine with me. And, since (assuming) this has never been used as a property name with atoms, I think if you feel comfortable with it you could use .items to trigger this behavior no matter the kind of atom (in preparation for NumericArray).

mathics/builtin/list/constructing.py
mmatera: I have check it in v 12.2.0. Probably they change this in more recent versions...

rocky:
@mmatera @bdlucas1 At this point, I think all of the concerns have been addressed, except possibly the one (or at least mine) where `Part[ByteArray[..]]` expands and then selects when it should not expand to select. 

That I'll address in another PR. This one however is getting large, so I'd like to merge and iterate, unless there are serious concerns remaing with this PR.

bdlucas1 (approved):
lgtm

mmatera:
Each of these atoms should define `Part` as UpValues` which are evaluated before the Downvalues

rocky:
> Each of these atoms should define `Part` as UpValues` which are evaluated before the Downvalues

@mmatera This is again something that I feel I can confidently do and get right. So please either suggest a commit +/-. Or we can commit this and start a new PR.  Your choice. Thanks.

mathics/builtin/list/eol.py
```diff
@@ -1172,19 +1200,18 @@ def eval(self, list, i, evaluation):
                 idx = idx.value
                 if idx == 0:
                     return SymbolByteArray
-                data = list.elements[0].value
-                lendata = len(data)
+                n = len(list.items)
                 if idx < 0:
-                    idx = data - idx
+                    idx = n - idx
```
mmatera, line 1205:
```suggestion
                    idx = n + idx
```

mmatera: All this block should be an UpValue. But leave this for a second round.

mathics/builtin/list/eol.py
```diff
@@ -1172,19 +1200,18 @@ def eval(self, list, i, evaluation):
...
                     idx = idx - 1
-                    if idx > lendata:
+                    if idx > n:
                         evaluation.message("Part", "partw", i, list)
                         return
-                return Integer(data[idx])
+                return list.items[idx]
             if idx is Symbol("System`All"):
                 return list
             # TODO: handling ranges and lists...
```
mmatera, line 1217: Maybe in this case we can just convert to list and use the List method.

mathics/builtin/list/eol.py
mmatera: Also, all this method should be inside a function in `mathics.eval.list.eol`

mathics/builtin/list/eol.py
mmatera, line 1217:
```suggestion
            # For the generic case, use
            list = to_mathics_list(*list.value, elements_conversion_fn=Integer)
            result = eval_Part([list], indices, evaluation)
            if result:
                try:
                    return ByteArray(bytearray([b.value for b in result.elements]))
                except TypeError:
                    return result
            return evaluation.current_expression
```

mmatera: and remove the next two lines...

mathics/builtin/binary/bytearray.py
```diff
@@ -35,44 +37,55 @@ class ByteArray(Builtin):
...
             evaluation.message("ByteArray", "lend", string)
-            return SymbolFailed
-        return Expression(SymbolByteArray, atom)
+            return None
+        return atom
 
-    def eval_to_str(self, baa, evaluation):
-        "ToString[ByteArray[baa_ByteArrayAtom]]"
+    def eval_to_str(self, baa, evaluation: Evaluation):
+        "ToString[ByteArray[baa_ByteArray]]"
```
mmatera:
```suggestion
        "ToString[baa_ByteArray]"
```

mathics/builtin/binary/bytearray.py
```diff
@@ -35,44 +37,55 @@ class ByteArray(Builtin):
...
         return String(f"ByteArray[<{len(baa.value)}>]")
 
-    def eval_normal(self, baa, evaluation):
-        "System`Normal[ByteArray[baa_ByteArrayAtom]]"
+    def eval_normal(self, baa, evaluation: Evaluation):
+        "System`Normal[ByteArray[baa_ByteArray]]"
```
mmatera:
```suggestion
        "System`Normal[baa_ByteArray]"
```

mathics/core/atoms.py
```diff
@@ -639,30 +640,45 @@ def to_sympy(self, *args, **kwargs):
...
+    # having these be distinct names may catch mistakes in coding
+    # where an expanded or Normal[]'d value is used when it should
+    # not be used.
+    _items: Optional[tuple] = None
+
+    class_head_name = "System`ByteArray"
+    hash: int
+
+    # We use __new__ here to ensure that two ByteArray's that have the same value
     # return the same object, and to set an object hash value.
```
bdlucas1, line 657: (Not a code review comment, just a general question maybe for discussion at a later time)

As far as I can see the code doesn't match this part of the comment: "ensure that two ByteArray's that have the same value return the same object", does it? In other instances of Atoms this is done, but not here unless I'm missing something.  And I'm wondering how much value doing that for ByteArrays would have anyway.

If the above is correct, could this more simply be `__init__`? Whenever I see `__new__` it gets my attention because I think something tricky is happening, and possibly dangerous or confusing (like for example the use of `__new__` in Builtin that returns an object of a completely unrelated class).

mathics/builtin/binary/bytearray.py
```diff
@@ -35,44 +37,55 @@ class ByteArray(Builtin):
...
 
-    def eval_to_str(self, baa, evaluation):
-        "ToString[ByteArray[baa_ByteArrayAtom]]"
+    def eval_to_str(self, baa, evaluation: Evaluation):
+        "ToString[ByteArray[baa_ByteArray]]"
         return String(f"ByteArray[<{len(baa.value)}>]")
 
-    def eval_normal(self, baa, evaluation):
-        "System`Normal[ByteArray[baa_ByteArrayAtom]]"
+    def eval_normal(self, baa, evaluation: Evaluation):
```
mmatera, line 73: These methods are not accesible until #1506 be fixed.

mathics/core/atoms.py
mmatera: Using `__new__`  is the standard way to implement *singleton* classes, i.e., classes that produce unique objects. We start doing this with symbols, and then we extend it to other atomic types. This helps to gain some performance in small evaluations, because reduces the need to recreate objects. On the other hand, if we want to deal with large numeric arrays, I guess we do not want to create a numeric object for each entry. So, indeed, we are goint to need go over this again.

rocky:
This is not the end of this topic, just the beginning...

Issue #1506, opened 2025-11-08
UpValues associated to Atoms are not taken into account in evaluations.


mmatera:
## Description

Differently than in WMA, Upvalues assigned to `Atom` types, like `Real` are not taken into account in evaluation

## How to Reproduce

```Mathics
In[1]:= Unprotect[Real]; Real/:F[x_Real]:=x;F[3.]
Out[1]= 3.
```

```
$mathics -c 'Unprotect[Real]; Real/:F[x_Real]:=x;F[3.]'  
```

## Output Given
Got `F[3.]` 

## Expected behavior

If the Upvalue were taken into account, the result -like in WMA- would be
`3.`

## Additional context

We would like that atoms like `ByteArray` use `UpValues` for defining rules associated to `Normal` or `Part`, to make the code more modular.

PR #1507, opened 2025-11-08
Using upvalues of atoms in evaluations.

mmatera:
In the `Expression.rewrite_apply_eval_step` method, in the upvalues lookup, to reduce the cost of evaluating expressions with many numerical elements, we use to skip atomic elements. This makes that upvalues of atomic elements get never evaluated (see  #1506). This PR fix by, instead of skipping atoms, just check beforehand if all the elements are of the same kind.


mathics/builtin/layout.py
```diff
@@ -60,7 +60,7 @@ class Format(Builtin):
 
     Raw objects cannot be formatted:
     >> Format[3] = "three";
-     : Cannot assign to raw object 3.
+     : Tag Integer in 3 is Protected.
```
mmatera, line 63: This matches better with WMA.

mathics/core/symbols.py
```diff
@@ -216,6 +216,13 @@ def get_head_name(self) -> "str":
     #        1/0
     #        return None if stop_on_error else {}
 
+    def get_lookup_name(self) -> str:
+        """
+        By default, atoms that are not symbols
+        have its class head_name as their lookup names.
```
rocky:
```suggestion
        have their class head_names as their lookup names.
```

rocky:
LGTM. See the very very small grammar change the docstring.

PR #1508, opened 2025-11-09
Rename element order values with _ELT_ORDER suffix

rocky:
My beef with the suffix `_SORT_KEY` is that it is vague (what is sorted, and why?).  Also, we use the `_SORT_KEY` suffix for the pattern selection precedence this has nothing to do with element ordering. 

The `_ELT_ORDER` suffix aims to make things clearer and less ambiguous.


mmatera:
LGTM

PR #1509, opened 2025-11-09
ByteArray access without item expansion in `Part[]`

rocky:
Add `__getitem(self, value)` to `ByteArray` so we can access a Python `bytearray` element without expanding the full Python `bytearray` value into a Mathics3 ListExpression. 

There might be other places in the Mathics3 where we might benefit from this. But this is a start.


mmatera:
LGTM

Issue #1511, opened 2025-11-10
Implement NumericArray


bdlucas1:
Currently functions like Plot3D generate Graphics3D objects that represent the mesh to be plotted as a List of Lists. For large meshes this is inefficient in space, and very inefficient in time to generate and consume for display. For example, generating and displaying a Plot3D with a useful degree of resolution takes many seconds, so is not feasible for interactive plotting with Manipulate.

A NumericArray implementation backed by NumPy arrays will greatly improve performance. Initial prototypes show orders of magnitude improvements in performance, allowing for example interactive plotting at useful resolution with Manipulate.

In addition I expect this could be useful for CSG operations on solids, where the boundary of the solid would be represented by a mesh stored as arrays.

I suggest a staged approach to implementation. Features to be implemented eventually:
* NumericArray atom
* NumericArray builtin
* Normal
* NumericArrayType, NumericArrayQ
* Equal, Length, ByteCount
* First, Last, Part, Take, Drop

mmatera:
It would be nice if `Normal`, instead of building a full nested `List` expression, would produce some kind of *proxy*, allowing to access the elements from the numpy array, to avoid conversions and creation of individual elements. This would be useful for linear algebra and data processing functions.

rocky:
> It would be nice if `Normal`, instead of building a full nested `List` expression, would produce some kind of _proxy_, allowing to access the elements from the numpy array, to avoid conversions and creation of individual elements. This would be useful for linear algebra and data processing functions.

I am certain we'll have a way for linear algebra and data processing functions to access numpy (or pandas, or whatever) arrays without having to expand. The details though I am not sure I can nail down. There are many ways to achieve this.

Also note that that we have `mathics.eval` functions that typically work off of lower level (i.e. not Mathics3 object) datatypes.

PR #1512, opened 2025-11-10
Numericarray atom

bdlucas1:
Initial feature for #1511: implement NumericArray atom.

This is sufficient to support the Plot3D/Graphics3D use case as those will only require NumericArrays generated and consumed in Python.


mathics/core/atoms.py
```diff
@@ -10,10 +10,16 @@
 import sympy
 from sympy.core import numbers as sympy_numbers
 
+try:  # pragma: no cover - optional dependency handled at runtime
```
mmatera: numpy is a dependence in sympy, so I guess it is safe to assume it is available.

mathics/core/atoms.py
```diff
@@ -1098,6 +1104,124 @@ def is_zero(self) -> bool:
...
+            raise ValueError(message)
+
+        # summary and hash
+        shape_string = "×".join(str(dim) for dim in self.value.shape) or "0"
+        self._summary_string = f"{self._type_name}, {shape_string}"
+        self._hash = None
+
+    # TODO: this is potentially expensive - what if we left it unimplemented? is hashing a numpy array reasonable?
+    # TODO: to make it less expensive only look at first 100 bytes - ok? needed?
+    def __hash__(self):
```
mmatera, line 1155: numpy arrays are mutable, so do not have a "hash". For our use here, it makes sense to use their `id` as Python objects.

mathics/core/atoms.py
```diff
@@ -1098,6 +1104,124 @@ def is_zero(self) -> bool:
...
+        if not self._hash:
+            self._hash = hash(
+                ("NumericArray", self.value.shape, self.value.tobytes()[:100])
```
mmatera:
```suggestion
                ("NumericArray", self.value.shape, id(self.value))
```

mathics/core/atoms.py
```diff
@@ -1098,6 +1104,124 @@ def is_zero(self) -> bool:
...
+        else:
+            return tuple(NumericArray(array) for array in self.value)
+
+    @property
+    def element_order(self) -> tuple:
+        return (
+            BASIC_ATOM_NUMERICARRAY_ELT_ORDER,
+            self.value.shape,
+            self.value.dtype,
+            self.value.tobytes(),
```
mmatera:
```suggestion
            id(self.value),
```

mathics/core/atoms.py
```diff
@@ -1098,6 +1104,124 @@ def is_zero(self) -> bool:
...
+            self.value, rhs.value
+        )
+
+    def to_sympy(self, **kwargs) -> None:
+        return None
+
+    # TODO: note that this returns a simple python list (of lists),
+    # not the numpy array - ok?
+    def to_python(self, *args, **kwargs) -> list:
+        return self.value.tolist()
```
mmatera, line 1208: why not just `self.value`?

mathics/core/convert/python.py
```diff
@@ -5,7 +5,12 @@
 
 from typing import Any
 
-from mathics.core.atoms import Complex, Integer, Rational, Real, String
+try:  # pragma: no cover - numpy is optional
```
mmatera: Again, we can assume that Numpy is available

mathics/core/atoms.py
bdlucas1: done

mathics/core/convert/python.py
bdlucas1: done

mathics/core/atoms.py
bdlucas1: done

mathics/core/atoms.py
bdlucas1: I wasn't sure what the contract of to_python is - will callers be surprised to find it return a numpy.ndarray and not a core Python builtin?

Just now I did some poking  around and found a [comment suggesting](https://github.com/Mathics3/mathics-core/blob/c6e5d283a5e52a3ce2cb5eec900356e6258c7c67/mathics/core/element.py#L329) it should be only a core Python type such as list.

However that comment also references some discussion suggesting the exact desired semantics may be unclear. And while numpy.ndarray is not a core Python builtin, it is kind of the next thing to it, being firmly entrenched in the Python ecosystem.

An advantage of returning numpy.ndarray is that converting to list has a performance cost, so code that innocently looks at to_python may unwittingly degrade performance.

@rocky do you have an opinion?

rocky: Some of this is loose, and can change based on what we find useful. However, to_python is intended for Python without any non-standard Python libraries. However note that to_python accepts keyword arguments, and the practice is that if you want to allow numpy, that is an option. Similar for mpmath, and sympy, although those have explciit "to_xxx" methods.

bdlucas1: I'm happy to go with whatever the consensus is.

mmatera: OK, I remember we have pass through the discussion about  what we should call "native" Python objects. So, let's keep lists here.

bdlucas1: OK. I left .tolist() and update comment.

mathics/core/atoms.py
```diff
@@ -1098,6 +1100,117 @@ def is_zero(self) -> bool:
...
+
+    def to_sympy(self, **kwargs) -> None:
+        return None
+
+    # TODO: this returns a list instead of np.ndarray in keeping with
+    # idea that to_python should return only "native" Python types.
+    # Keep an eye on this because there is a slight risk that code may
+    # naively call to_python and cause a performance issue due to
+    # the cost of converting to a nested list structure for a large array.
+    def to_python(self, *args, **kwargs) -> list:
```
rocky, line 1207: I should expand on the motivation and thinking here. 

It is okay to expand this method to accept a "allow_numpy" keyword and then adjust this based on that. 

We might also add a `to_numpy()` method in the future like we have to_sympy() methods. 

Note that  `to_sympy` also might allow "allow_numpy" (or "disallow_numpy") to adjust how it behaves.

My motivation for this was based on the desire to add translations to other variables. Just as Mathematica has a "ToTeX" thing (which is really ToLaTeX I think, there might be a ToPython (with options for what flavor of Python)

bdlucas1: Thanks for the perspective. I'm ok leaving it as-is for now - clients look at .value to get the numpy array - and thinking about what to do if in the future there is need for more numpy-amenable atoms. If so, personally to_numpy seems cleaner to me than allow_numpy - typing is simpler for example.

test/builtin/test_numericarray.py
```diff
@@ -0,0 +1,60 @@
...
+
+from mathics.core.atoms import Integer, NumericArray, String
+from mathics.core.convert.python import from_python
+
+#
+# Python API tests
+#
+
+
+def test_numericarray_atom_preserves_array_reference():
```
rocky: A small thing here. I believe these tests are really "mathics.core.atoms" tests rather than mathics.buitin tests. 

The core tests, test the core object which I think is what is happening here. This is a good thing and thanks for doing this kind of testing. 

The commented-out more heavyweight tests that check Mathics3 functions are however under mathics.builtin. 

Right now, I am not seeing this code filled out out, that's okay to defer for later.  That would be put under `mathics.builtin.binary` (binary data). 

And the uncommented version of this test would under `test.builtin.binary` rather than `test.builtin`.

bdlucas1: I'm not sure I'm following what specifically, if anything, you'd like me to do to enable the PR to be merged.

Personally, as you know, I'm not a big fan of having things split up and finely categorized. In this case for example personally I find it helpful to have all the tests related to NumericArray in one place. In general, it seems awkward for discoverability that support for NumericArray will be split up across by the time we're done maybe a dozen files. What is the purpose of separating out binary builtins and tests from other kinds of atoms - what makes an atom binary?

Anyway that's neither here nor there - just my personal bias - and I'll follow the established philosophy. Just let me know specifically what you'd like me to do.

rocky: We can discuss in (virtual) person soon.

mmatera (approved):
LGTM. The question of where the pytests should be can be deferred for later.

rocky:
> LGTM. The question of where the pytests should be can be deferred for later.

@mmatera Just discussed with Bruce. I have a PR in for the changes [here](https://github.com/bdlucas1/mathics-core/pull/11).

bdlucas1:
> PR in for the changes [here](https://github.com/bdlucas1/mathics-core/pull/11).

Thanks Rocky; merged.

rocky:
LGTM - @bdlucas1 good stuff here. Thanks for wading into the Stygian stables.

PR #1514, opened 2025-11-11
Plot structure tests

bdlucas1:
This is the first of a series of PRs aiming to improve plotting performance for Plot3D (initially):

* This PR improves Plot3D testing to look at the structure of the resulting Graphics3D expression, improving confidence that nothing is broken by subsequent refactoring.
* Refactor eval_Plot3D to separate the option processing stuff from the generation of the Graphics3D output in preparation for the subsequent step. This should cause no functional change, and the improved testing introduced by the previous PR will increase confidence that that is the case.
* Introduce optional (for now) generation of Graphics3D output using GraphicsComplex and NumericArray in Plot3D. This should provide greatly improved plotting performance for consumers who enable the option and are prepared to consume GraphicsComplex.
* Subsequent PRs may extend this to other types of plots, update internal (core) consumers of Graphics3D to understand GraphicsComplex, etc.

### DIFF
--- a/mathics/core/util.py
+++ b/mathics/core/util.py
@@ -117,20 +117,20 @@ def subranges(
             )
 
 
-def print_expression_tree(expr, indent="", marker = lambda expr: ""):
+def print_expression_tree(expr, indent="", marker=lambda expr: ""):
     """
     Print a Mathics Expression as an indented tree.
     Caller may supply a marker function that computes a marker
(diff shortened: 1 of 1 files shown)


rocky:
LGTM, and I am glad we are starting on a better testing regimine.

rocky:
@mmatera unless you have concern, let's merge this.

PR #1515, opened 2025-11-11
get ride of trailing _elements_fully_evaluated

mmatera:
Looking at the `elements_properties` stuff, I found this code, that now does not have any function, plus something that looks like a typo. @rocky, can confirm me that this is the case?


rocky:
@mmatera You are correct. There are typos/bugs there.

rocky:
LGTM

mmatera:
OK, let's merge and iterate. I found a few similar things in the way.

PR #1516, opened 2025-11-11
Correct conversion from Python bytes & bytearrays

rocky:
A small misfeature/Bug noticed by bdlucas1 in implementing NumericArray. Before the recent work on ByteArray, this was less wrong, but still a little wrong.


bdlucas1 (approved):
lgtm

PR #1517, opened 2025-11-12
Using `is_uniform` to avoid repeated checks

mmatera:
Now we have the `is_uniform` attribute, this PR uses it to make faster the match test with `SequenceBlank`  and `SequenceNullBlank`. Also, ensure that this property be computed and stored in list/expression creation operations (`Table`, `Sum`, `Product`).

* using is_uniform property in matching `SequenceBlank`  and `SequenceNullBlank` patterns
* `get_result` now accepts an optional attribute `is_uniform` to set this property in its result.
* `Table`, `Sum` and `Product` now set the element_property `is_uniform` to provide a faster evaluation of large lists.
* convert_expression_elements now handles is_uniform
* Start a workflow to benchmark some basic evaluations and pattern matching.

### DIFF
--- a/SYMBOLS_MANIFEST.txt
+++ b/SYMBOLS_MANIFEST.txt
@@ -244,6 +244,8 @@ System`Complement
 System`Complex
 System`ComplexExpand
 System`ComplexInfinity
+System`ComplexPlot
+System`ComplexPlot3D
 System`Complexes
 System`CompositeQ
 System`Composition
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -41,7 +41,7 @@ from mathics.core.builtin import (
     SympyObject,
     Test,
 )
-from mathics.core.convert.sympy import SympyExpression, from_sympy, sympy_symbol_prefix
+from mathics.core.convert.sympy import SympyExpression, from_sympy
 from mathics.core.element import BaseElement, ElementsProperties
 from mathics.core.evaluation import Evaluation
 from mathics.core.expression import Expression
@@ -62,6 +62,7 @@ from mathics.core.symbols import (
     SymbolPlus,
     SymbolTimes,
     SymbolTrue,
+    sympy_name,
 )
 from mathics.core.systemsymbols import (
     SymbolAnd,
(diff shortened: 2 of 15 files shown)


rocky:
LGTM, although I confess I don't see how this is going to help. If you want to give a specific example, I'd be grateful.

mmatera:
@rocky, suppose that you define a function
`F[vals__Real]=Sum[x,{x,vals}]`
and a list of values `A=Table[BesselJ[3,x],{x,0,10,1000}]`. Now, to evaluate
```
F@@A
```
first you need to check if the result of `Table[BesselJ[3,x],{x,0,10,1000}]` has all real entries. So, the `match` function (as is in master) has to check the head of each element in `A`. But, by construction, `A` has all their elements of the same type. So check the first is enough.
Same happens with the evaluation: since `A` is uniform, when you evaluate it, it is not necesary to check for upvalues on each element, you need just to look at the first one.

In this way,
`Table[A=Table[1.*i^2,{i,0,n}]; {n, Timing[MatchQ[A, {__Real}]]},{n,{2,10,100,10000}}]`
should show that the time required by `MatchQ` should be constant instead of linear in the size of `A`. 

However, for some reason, it seems we are missing the `element_properties` after the assigment, so the improvement is not shown.

rocky:
@mmatera Thanks for the explanaition.

mathics/eval/parts.py
```diff
@@ -179,8 +180,11 @@ def walk_levels(
                 cur_pos + [index + 1],
             )
             depth = max(element_depth + 1, depth)
-            elements.append(element)
-        new_expr = make_expression(head, *elements)
+            elements.append(new_element)
+            if new_element is not element:
+                elem_prop = None
+
+        new_expr = make_expression(head, *elements, elements_properties=elem_prop)
```
mmatera, line 187: Here is where the element_properties were lost. Now, for large blocks, I see a ~5% of improvement in performance in the folliwing operations:
```
In[1]:= a = Table[1./(1+i^2),{i,1000}];Timing[Plus@@a]
Out[1]= {0.00908414, 1.07567}

In[2]:= Timing[MatchQ[a, {__Real}]]
Out[2]= {0.00175423, True}

In[3]:= F[x__Real]:=1; Timing[F@@a]
Out[3]= {0.00755716, 1}
```

rocky:
LGTM. It would be nice to add a pytest timing test that tickles this feature so we might be able to detect regression. Thanks.

mmatera:
@rocky, now I adding some small benchmarks to compare the evaluation of expressions with and without the `uniform` property.



imagen

PR #1518, opened 2025-11-12
Try to lint atoms, start Order tests...

rocky:
Also, canonicalize color_directives in conversions, although this doesn't solve a problem I'll open an issue for.

### DIFF
--- a/mathics/builtin/colors/color_directives.py
+++ b/mathics/builtin/colors/color_directives.py
@@ -204,7 +204,12 @@ class _ColorObject(_GraphicsDirective, ImmutableValueMixin):
         return self.to_rgba()
 
     def to_expr(self):
-        return to_expression(self.get_name(), *self.components)
+        """Convert components to MachineReal consistently so that colors with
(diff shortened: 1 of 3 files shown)


rocky:
@mmatera for a while, I have wanted understand better and try to regularize our atom structure. (BaseElement is also on the list, but that is for some other time.)

Most of the stuff here, is going over mypy lint messages, starting to order the long list of methods more alphabetically. 

I note started an ordering testing, for keycomparable but there is way more to test and compare. 

I'll [open an issue I encountered, doing this](https://github.com/Mathics3/mathics-core/issues/1519) . Basically, Complex numbers are not comparable to Real numbers, but we don't have an ordering for this. But in trying to add this I am getting a weird problem in comparing RGBColors for Integer values versus the same Real values.

rocky:
@mmatera  I am glad this will get in before the work to address #1519 which will undoubtedly conflict with this (by proximity of code rather than something being fundamentally wrong that needs changing.)

PR #1520, opened 2025-11-13
Refactor plot3d

bdlucas1:
This PR substantially refactors Plot3D and DensityPlot (which share a
good bit of code), in preparation for adding a new vectorized
implementation of those two functions.  No function is changed or
added.

* The the flow from builtin to eval is cleaner - no more callbacks
  from eval to builtin. Options processing is done prior to calling
  eval_*. Plot3D and DensityPlot now have their own eval functions
  which share implementation code in eval rather than in builtin.

* A PlotOptions class has been introduced. This simplifies passing
  plotting options from builtin to eval. It may also be reusable for
  other plotting builtins.

* A new GraphicsGenerator class has been added. This consolidates and
  regularizes some scattered code for generating Graphics and
  Graphics3D expressions, which are rather complex in structure. This
  will be further enhanced to support GraphicsComplex in the PR that
  introduces vectorized plotting functions. Possibly could be retrofitted
  to other plotting functions.

* A test for DensityPlot was added, which was missed in the previous
  PR because I didn't realize how much code it shared with Plot3D.

This is a fairly big change, but I worked carefully, frequently
testing using the detailed tests for Plot3D and DensityPlot.  The
tests caught numerous errors as I worked, increasing confidence that
the warning light was hooked up and no functional changes have been
introduced.

Some quick timing tests show no change in performance as far as I can
see.

More can be done, particularly in the areas of error checking and
typing, but that shouldn't change the PR in any major way so hoping
to get some eyes on it in parallel.

### DIFF
--- a/mathics/eval/drawing/plot3d.py
+++ b/mathics/eval/drawing/plot3d.py
@@ -312,7 +312,7 @@ def compute_triangles(plot_options, evaluation):
 
         # add the mesh
         mesh_points = []
-        if mesh == SymbolFull:
+        if mesh is SymbolFull:
             for xi in range(plotpoints[0] + 1):
                 xval = xstart + xi / numx * (xstop - xstart)
                 mesh_row = []
@@ -368,7 +368,7 @@ def compute_triangles(plot_options, evaluation):
                 for mesh_line in mesh_points
                 if not any(x[2] is None for x in mesh_line)
             ]
-        elif mesh == SymbolAll:
+        elif mesh is SymbolAll:
             mesh_points = set()
             for t in triangles:
                 mesh_points.add((t[0], t[1]) if t[1] > t[0] else (t[1], t[0]))


mathics/builtin/drawing/plot.py
```diff
@@ -426,21 +520,25 @@ class _Plot3D(Builtin):
...
         options: dict,
     ):
-        """%(name)s[functions_, {x_Symbol, xstart_, xstop_},
-        {y_Symbol, ystart_, ystop_}, OptionsPattern[%(name)s]]"""
-        return eval_plot3d(
-            self, functions, x, xstart, xstop, y, ystart, ystop, evaluation, options
-        )
+        """%(name)s[args___, OptionsPattern[%(name)s]]"""
+
+        # TODO: test error for too many, too few, no args
```
rocky, line 552: FYI setting a class variable `expected_args` is the simple way we add limit checking to Builtin functions, that give the messages with `argx`, `argr`, `argrx`, `argt`, or `argtu` tags. `git grep -n expected_args` to see examples of this use.

mathics/builtin/drawing/plot.py
```diff
@@ -407,6 +405,102 @@ def process_function_and_options(
...
+                if limit is None:
+                    evaluation.message(expr.get_name(), "plln", limit_expr, range_expr)
+                    raise ValueError()
+                range.append(limit)
+            self.ranges.append(range)
+
+        # Mesh option
+        mesh_option = expr.get_option(options, "Mesh", evaluation)
+        mesh = mesh_option.to_python()
+        if mesh not in ["System`None", "System`Full", "System`All"]:
```
rocky: What is typically done here is not to use `to_python()`, but instead do checks against the symbols `[SymbolNone, SymbolAll,  Symbol("System``Full]`.  The first two Symbols are already defined in `mathics.atoms.systemsymbols`; `SymbolFull` should probably be added.

test/builtin/drawing/test_plot.py
```diff
@@ -325,3 +347,102 @@ def test_plot3d_nondefault():
...
+            Axes -> True,
+            AxesStyle -> {RGBColor[1,0,0],RGBColor[0,0,1]},
+            Background -> RGBColor[0,1,0],
+            ImageSize -> {200,200},
+            LabelStyle -> RGBColor[1,0,0],
+            PlotRange -> {0,1},
+            PlotRangePadding -> {1,2},
+            TicksStyle -> {RGBColor[0.5,0,0.5],GrayLevel[1]}
+        ]
+        """,
```
rocky, line 485: For what it's worth, when I run this in either the  Django or mathicsscript interfaces, I am getting a BoxExpressionError
````
...
 File "/src/external-vcs/github/bdlucas1/mathics-core/mathics/builtin/box/graphics.py", line 612, in _prepare_elements
    xmin, xmax, ymin, ymax, w, h, width, height = calc_dimensions(final_pass=False)
                                                  ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/src/external-vcs/github/bdlucas1/mathics-core/mathics/builtin/box/graphics.py", line 566, in calc_dimensions
    raise BoxExpressionError
```

rocky:
Much needed work. Thanks for undertaking. I will look at and see if I can answer the questions in the code in another pass, probably tomorrow.

mathics/builtin/drawing/plot.py
bdlucas1: Just to clarify - do you favor passing it to eval as a Symbol instead of a string?

test/builtin/drawing/test_plot.py
bdlucas1: Was that in master or this PR or both?

rocky: > Was that in master or this PR or both?

I just tried on master. I get the same BoxExpressionError.  The code we currently have is very fragile.

mathics/builtin/drawing/plot.py
rocky: Yes, Symbols are better to describe intent as well as being slightly faster.  

It is also possible one day that might be able to pickle everything in mathics.core.systemsymbols to make loading fast.  But that and autoload like GNU Emacs has is a different topic.

mathics/builtin/drawing/plot.py
```diff
@@ -407,6 +405,102 @@ def process_function_and_options(
...
+            max_depth = 15
+            evaluation.message(expr.get_name(), "invmaxrec", max_depth, 15)
+        else:
+            max_depth = 0
+            evaluation.message(expr.get_name(), "invmaxrec", max_depth, 15)
+        self.max_depth = max_depth
+
+        # ColorFunction and ColorFunctionScaling options
+        # This was pulled from construct_density_plot (now eval_DensityPlot).
+        # TODO: What does pop=True do? is it right?
```
rocky, line 504: Here's a guess. A dictionary method "pop" is used to remove an element from a dictionary. Since that's some sort of option, I think the idea is about whether this opiton is fully handled at this point and therefore remove it from the list of options, or keep it in the options dictionary for other routines to adjust behavior based on looking in the dictionary (as opposed to consulting a variable or some other state).

Just a guess though.

mathics/builtin/drawing/plot.py
```diff
@@ -407,6 +405,102 @@ def process_function_and_options(
...
+        # TODO: can we move some of the subsequent processing in eval_DensityPlot to here?
```
rocky, line 505: I realize this may be a rhetorical question, but I sometimes answer those. Yes, I hope so. The code organization here is currently a mess.

test/builtin/drawing/test_plot.py
```diff
@@ -325,3 +347,102 @@ def test_plot3d_nondefault():
...
+                        },
+                        {
+                            RGBColor[0.49621975000000007, 0.41002484999999994, 0.8144772499999999],
+                            RGBColor[0.49621975000000007, 0.41002484999999994, 0.8144772499999999],
+                            RGBColor[0.663226, 0.6872815, 0.9117649999999999]
+                        }
+                    }
+                ]
+            },
+            AspectRatio -> 1, (* TODO: is not passed through apparently - or is my misunderstanding? *)
```
rocky, line 475: Here is what I am seeing through experimentation. AspectRatio  _is_ handled properly for the simpler example: 

```mathematica
Graphics[
   Polygon[
      {{{0.0,0.0},{0.0,0.5},{0.5,0.0}}}
   ], 
   AspectRatio->0.5
]
```

The VertexColors option has not been added to Mathics3.

Also, when a list of polygons is given, I am not seeing anything, which could be a boxing roblem &mdash;  in neither instance am no I getting an error reported.

mathics/builtin/drawing/plot.py
bdlucas1: OK, thanks. Did that, and also added some other error checking for the range args.

test/builtin/drawing/test_plot.py
```diff
@@ -66,6 +66,11 @@ def test__listplot():
             "-Graphics3D-",
             None,
         ),
+        ("Plot3D[]", ("Plot3D called with 0 arguments; 3 arguments are expected.",), "Plot3D[]", None),
+        ("Plot3D[1, 2, 3]", ("Plot range 2 must be of the form {variable, min, max}, where max > min.",), "Plot3D[1, 2, 3]", None),
```
rocky: Kind of a minor thing, but the error I see in WL 13.2.0 for this expression and the next one  has a "pllim" tag instead of an "invrange" tag that says "Range specification 2 is not of the form {x, xmin, xmax}. 

It might be of interest in that it may give some insight as to how Mathematica is working.

mathics/eval/drawing/plot3d.py
```diff
@@ -306,7 +306,7 @@ def triangle(x1, y1, x2, y2, x3, y3, depth=0):
 
         # add the mesh
         mesh_points = []
-        if mesh == "System`Full":
+        if mesh == SymbolFull:
```
rocky: Actually, now that Symbols are used, all of the `==` tests can become `is` tests. (Symbols are unique).  Some linters will warn of this, I guess because "is" _is_ faster.

mathics/builtin/drawing/plot.py
bdlucas1: Agreed. Also, it's safer because typos are caught.

I think it would be good to encourage use of Symbols instead of strings. I don't know why the original coder might have used strings, but if I were writing the code I might have done the same thing because just typing a string is a lot easier than using a symbol, figuring out where it's defined (symbols or systemsymbols), and adjusting the import.

Here are some things I would find easier, in order of decreasing ease:

    # no import to adjust, no redundant Symbol, all symbols are in the same place
    import system.symbols as sym
    ... sym.Foo ...

    # no imports to adjust, all symbols are in the same place, redundant Symbol isn't that bad
    import system.symbols as sym
    ... sym.SymbolFoo ...

    # all symbols are in the same import, not split between symbols and systemsymbols
    from system.symbols import SymbolFoo
    ... SymbolFoo ...

In my view any of those would be an improvement. I understand though it might be a substantial change, and may not be judged worth it.

bdlucas1: Forgot to mention, I made the change.

test/builtin/drawing/test_plot.py
bdlucas1: OK. Fragility aside, I don't know that those are valid plots - the intent was just to check whether options are passed through correctly.

I have been thinking about how to do some end-to-end testing where valid plots are compared visually. In some ways those tests will be easier to do, though maybe harder to debug.  I think that's outside the scope of core though.

mathics/builtin/drawing/plot.py
bdlucas1: Makes sense as to what it does, but not sure why it's done only in this case. Anyway at some point I want to take a pass over the complete list of plotting options because currently only a subset are handled, and maybe as part of that I'll try to make handling more uniform. Getting options processing pulled out into its own thing was a first step.

mathics/builtin/drawing/plot.py
bdlucas1: Appreciate your response. I may look into that refactoring as part of the pass over all plotting options that I mentioned.

test/builtin/drawing/test_plot.py
bdlucas1: I took a quick look at it's error handling in for Plot3D ranges, and got some funky results for some of the mis-specified ranges. My conclusion is that their error handling is just as much an inconsistent mess as ours (and I think you said much the same in your email), so I think it's not worth putting excessive effort into emulating it in detail. For that matter, is there any reason to think they consider error messages a stable interface from version to version?

In this case I thought a simple generic error message would be low-effort and would suffice.

mathics/eval/drawing/plot3d.py
bdlucas1: thanks, done

mathics/builtin/drawing/plot.py
rocky: > Agreed. Also, it's safer because typos are caught.
> 
> I think it would be good to encourage use of Symbols instead of strings. I don't know why the original coder might have used strings, but if I were writing the code I might have done the same thing because just typing a string is a lot easier than using a symbol, figuring out where it's defined (symbols or systemsymbols), and adjusting the import.
> 
> Here are some things I would find easier, in order of decreasing ease:
> 
> ```
> # no import to adjust, no redundant Symbol, all symbols are in the same place
> import system.symbols as sym
> ... sym.Foo ...
> 
> # no imports to adjust, all symbols are in the same place, redundant Symbol isn't that bad
> import system.symbols as sym
> ... sym.SymbolFoo ...
> 
> # all symbols are in the same import, not split between symbols and systemsymbols
> from system.symbols import SymbolFoo
> ... SymbolFoo ...
> ```
> 
> In my view any of those would be an improvement. I understand though it might be a substantial change, and may not be judged worth it.

I am okay with the first or any of these. 

@mmatera - your opinion? 

I find it amusing to see "didn't think of that" roles reversed here.

An old Systems-Administration saying: 

Never ascribe to maliciousness that which can easily be explained by ignorance or thoughtlessness.

test/builtin/drawing/test_plot.py
rocky: > I took a quick look at it's error handling in for Plot3D ranges, and got some funky results for some of the mis-specified ranges. 

Great - thanks for doing this. 

> My conclusion is that their error handling is just as much an inconsistent mess as ours (and I think you said much the same in your email), so I think it's not worth putting excessive effort into emulating it in detail. For that matter, is there any reason to think they consider error messages a stable interface from version to version?

Dunno. @mmatera you're more of an expert here. 

> 
> In this case I thought a simple generic error message would be low-effort and would suffice.

I'm okay with this. As might be obvious, when one kind of at the bottom, one is not looking for _perfection_, just _improvement_.

mmatera: If it is ok for you, I will going over this during the weekend 
M

bdlucas1: Great, thanks!

mmatera: > I took a quick look at it's error handling in for Plot3D ranges, and got some funky results for some of the mis-specified ranges. My conclusion is that their error handling is just as much an inconsistent mess as ours (and I think you said much the same in your email), so I think it's not worth putting excessive effort into emulating it in detail. For that matter, is there any reason to think they consider error messages a stable interface from version to version?

Specifically for graphics, I think WMA neither have a consistent interface for error messages. This is probably because the kind of graphics and supported options have changed a lot from early versions. On the other hand, typical error messages (like the function you want to plot has non-numerical values) are more or less stable.

mmatera: > Kind of a minor thing, but the error I see in WL 13.2.0 for this expression and the next one has a "pllim" tag instead of an "invrange" tag that says "Range specification 2 is not of the form {x, xmin, xmax}.
> 
> It might be of interest in that it may give some insight as to how Mathematica is working.

These are examples in WMA 12.2

```
In[2]:= Plot3D[]                                                                

Plot3D::argrx: Plot3D called with 0 arguments; 3 arguments are expected.

Out[2]= Plot3D[]

In[3]:= Plot3D[1,2,3]                                                           

Plot3D::pllim: Range specification 2 is not of the form {x, xmin, xmax}.
```
`argx` is more or less stable since version WMA 6 at least,  but regarding the one associated with the range specification, I think have changed since this code was written.

mmatera: To check changes regarding legacy versions of WMA, I usually look at https://reference.wolfram.com/legacy/v6
See for instance https://reference.wolfram.com/legacy/v6/tutorial/Messages.html

However, very few information about error messages are described there...

rocky: @mmatera, thanks for investigating and giving insight.

PR #1521, opened 2025-11-15
Fix number comparisons vs expression comparisons

mmatera:
This PR improves the compatibility with the different  WMA sorting schemes for numbers and expressions.

### DIFF
--- a/test/builtin/test_file_operations.py
+++ b/test/builtin/test_file_operations.py
@@ -2,7 +2,7 @@
 """
 Unit tests for mathics.builtin.file_operations
 """
-
+import os
 import sys
 import time
 from test.helper import check_evaluation, evaluate
@@ -100,6 +100,10 @@ import pytest
             None,
         ),
     ],
+)
+@pytest.mark.skipif(
+    os.getenv("SANDBOX", False),
+    reason="Test doesn't work in a sandboxed environment with access to local files",
 )
 def test_private_doctests_file_properties(str_expr, msgs, str_expected, fail_msg):
     """file_opertions.file_properties"""


mathics/core/atoms.py
```diff
@@ -239,11 +239,14 @@ def __new__(cls, value) -> "Integer":
...
 
     def __eq__(self, other) -> bool:
-        return (
-            self._value == other.value
-            if isinstance(other, Integer)
-            else super().__eq__(other)
-        )
+        if isinstance(other, Integer):
+            return self._value == other.value
+        if isinstance(other, Number):
```
mmatera, line 244: Here, numbers are not treated as generic expressions in checking equality: This fixes the issue with `RGBColor[1,0,0]==RGBColor[1.,0,0]` keeping the right order in ```Sort[{1.,1,1.`3, 1.`9}]```

test/builtin/test_sort.py
```diff
@@ -6,6 +6,12 @@
 from mathics.core.symbols import Symbol, SymbolPlus, SymbolTimes
 
 
+def test_Sorting_Numbers():
```
rocky: A problem I have encountered with a test like this is that when it fails, it is very hard to figure out exactly which of the element-pair orderings is wrong. 

It is also hard to assess in a test like this whether we've covered all of the different pairs of elements.

I realize something like this is useful, and one could argue needed, as a _final_ sanity test that what users expect happens. 

But it should not be the _only_ test. Instead, please consider adding lower-level and more pinpointable unit tests under `test/core` somewhere. Thanks.

mathics/core/atoms.py
```diff
@@ -1019,6 +1087,17 @@ def __new__(cls, numerator, denominator=1) -> "Rational":
...
         return self
 
+    def __eq__(self, other) -> bool:
+        if isinstance(other, Rational):
+            return self.value.as_numer_denom() == other.value.as_numer_denom()
+        if isinstance(other, Integer):
+            return (other.value, 1)== self.value.as_numer_denom()
+        if isinstance(other, Number):
+            # For general numbers, rely on Real or Complex implementations.
+            return other.__eq__(self)
```
rocky, line 1069: I find it interesting that on line 1094, infix `==` is used, but here the `__eq__()` method is used. They are the same thing, right? 

The same thing happens at line 1038.

mmatera: I use it just to make explicit that we are calling a user defined method, instead of the default equality operator from Python.

test/builtin/test_sort.py
mmatera: I am thinking about that. The problem is that I am not fully sure that this implementation for the sort key for numbers is the best. What this test tries to reinforce is the final behaviour

mathics/core/atoms.py
rocky: But in the other places, Python's built-in equality operator is not used either. For Rational, it is a sympy.Float's `__eq__` and for Complex, I think we use Rational's `__eq__`.

test/builtin/test_sort.py
rocky: Ok, then please use a sequence of `Order[]` tests varying pairs of objects.

rocky:
@mmatera based on experimentation, element sort order also happens match how polynomials appear in many forms. Is this correct?

mmatera:
> @mmatera based on experimentation, element sort order also happens match how polynomials appear in many forms. Is this correct?

Yep, in `mathics.core.expression:893`, for expressions of the form `_Number var_^pow_Integer`, the element_order is treated as a *monomial*.

rocky:
LGTM - thanks for the explanation.

PR #1522, opened 2025-11-15
Vectorized plot3d

bdlucas1:
* Introduce a vectorized Plot3D in plot3d_vectorized.py. Since this
  generates Graphics3D containing GraphicsComplex with NumericArray,
  which no code will currently understand, for now it is enabled only
  if you set environment variable
  MATHICS3_USE_VECTORIZED_PLOT=yes. (TBD how we will eventually
  deploy.)

* Introduce a "compiler" in plot_compile.py that uses expr.to_sympy()
  followed by sympy.lambdify to generate a Python function that calls
  numpy to evaluate. This is a huge win because most Python functions
  can do vectorized operations on entire arrays. For now it's specific
  to plotting; TBD whether it might find other uses. See
  test_plot_compiler to get a sense of what function is currently
  supported.

* Slightly refactors GraphicsGenerator for simplicity.

Here are some timings comparing classic and vectorized Plot3D. The
visual comparison was done using a prototype front-end that
understands GraphicsComplex and NumericArray.
            
    Plot3D[Sin[x] Cos[y], {x,0,10}, {y,0,10}, PlotPoints->{p,p}, MaxRecursion->r]

                   p    r             eval_Plot3D
    classic        50   2 (default)   10083 ms      visible triangles on plot surface
    classic       100   0              3232 ms      less visible artifacts than above
    vectorized    200   n/a               5 ms      no visible artifacts

Plotting is now fast enough that useful interaction with Manipulate is
possible. I have a prototype of that and will submit a PR.

Other possible future work is to build out support for GraphicsComplex
and NumericArray in consumers of Graphics3D. It might be helpful to
have a "GraphicsParser" class that takes a Graphics3D expression and
turns it into Python data structures that are easy for consumers to
traverse.

### DIFF
--- a/mathics/core/util.py
+++ b/mathics/core/util.py
@@ -3,9 +3,7 @@
 Miscellaneous mathics.core utility functions.
 """
 
-import os
 import sys
-import time
 from itertools import chain
 from pathlib import PureWindowsPath
 from platform import python_implementation
--- a/test/builtin/drawing/test_plot_compile.py
+++ b/test/builtin/drawing/test_plot_compile.py
@@ -13,16 +13,15 @@ The tests that are uncommented pass; the commented ones need more work.
 """
 
 import inspect
+from test.helper import session
 
 import numpy as np
 
 from mathics.core.convert.python import from_python
 from mathics.core.expression import Expression
 from mathics.eval.drawing.plot_compile import plot_compile
-from mathics.session import MathicsSession
-from test.helper import session
 
-#session = MathicsSession()
+# session = MathicsSession()
 
 
 #
--- a/mathics/eval/drawing/plot3d.py
+++ b/mathics/eval/drawing/plot3d.py
@@ -21,8 +21,8 @@ from mathics.core.systemsymbols import (
     SymbolFunction,
     SymbolSlot,
 )
-from mathics.timing import Timer
 from mathics.eval.drawing.plot import compile_quiet_function
+from mathics.timing import Timer
 
 from .util import GraphicsGenerator
 


.gitignore
```diff
@@ -26,6 +26,7 @@ ChangeLog.rej
 Documents/
 Homepage/
 Test/
+!test/
```
bdlucas1: I added this because git was not seeing the new test file. Can be removed if it's a problem and test files can be manually added.

mathics/eval/drawing/plot_compile.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+# mappings = [dict(AdditionalMappings.__dict__), "numpy"]
+
+
+class CompileError(Exception):
+    pass
+
+
+def compile(evaluation, expr, names, debug=0):
```
rocky: `compile()` is a Python built-in function. Redefining it (especially when the uses of it use the unadorned name, i.e, `from .plot_compile import compile` can lead to confusion. Consider renaming to something other than the builtin function name.

(BTW, there are places in the code where we use "len" as a variable, which also violates this principle. This might have been done so that the variable in the function signature is the same name, "len", as appears in Mathematica documentation. But that doesn't happen here.)

One other side note, although I have more substantial thoughts right now, in PRs, the superficial ones tend to come first to dispose of the thought. The deeper ones often come last because they deserve more thought,  investigation, and effort to write down.

rocky:
> * Introduce a vectorized Plot3D in plot3d_vectorized.py. Since this
>   generates Graphics3D containing GraphicsComplex with NumericArray,
>   which no code will currently understand, for now it is enabled only
>   if you set environment variable
>   MATHICS_USE_VECTORIZED_PLOT=yes. (TBD how we will eventually
>   deploy.)
> * Introduce a "compiler" in plot_compile.py that uses expr.to_sympy()
>
> Here are some timings comparing classic and vectorized Plot3D. The visual comparison was done using a prototype front-end that understands GraphicsComplex and NumericArray.
> 
> ```
> Plot3D[Sin[x] Cos[y], {x,0,10}, {y,0,10}, PlotPoints->{p,p}, MaxRecursion->r]
> 
>                p    r             eval_Plot3D
> classic        50   2 (default)   10083 ms      visible triangles on plot surface
> classic       100   0              3232 ms      better than above
> vectorized    200   n/a               5 ms      no visible artifacts
> ```

By "better than above" you mean in terms of time, not visual appearance, right?

mathics/eval/drawing/plot_compile.py
bdlucas1: Agreed that its bad practice. In fact as I typed it emacs alerted me to the spoken-for nature of the identifier by colorizing it but I ignored it with I guess the thought in the back of my mind "meh, Rocky will never notice". Happy to keep you on your toes. Will fix.

bdlucas1:
> By "better than above" you mean in terms of time, not visual appearance, right?

No, I meant the visual appearance was better. I'm sure that the recursive subdivision thing has its moments, but I think for the most part its moments have passed, and on modern hardware you're better off just doing a regular computation on a larger grid.

One thing I need to think about: WMA handles piecewise-defined functions very nicely as separate disjoint meshes, whereas this code as it stands will join the meshes. I think it is possible to do the same efficiently with vectorized plot, but will need a little thought about how to accomplish.

mathics/builtin/drawing/plot.py
```diff
@@ -51,7 +52,18 @@
...
 )
-from mathics.eval.drawing.plot3d import eval_DensityPlot, eval_Plot3D
+
+# The vectorized plot function generates GraphicsComplex using NumericArray,
+# which no consumer will currently understand. So lets make it opt-in for now.
+# If it remains opt-in we'll probably want some combination of env variables,
+# Set option such as $UseVectorizedPlot, and maybe a non-standard Plot3D option.
+# For now an env variable is simplest.
+# TODO: work out exactly how to deploy.
+if os.getenv("MATHICS_USE_VECTORIZED_PLOT", False):
```
rocky: Another small thing. We've been moving from environment variable names that start MATHICS to MATHIC3, to distinguish what we have now from what came before. In fora, there still seems to be a little confusion. 

(Also appears in MATHICS_TIMING).

rocky:
> > By "better than above" you mean in terms of time, not visual appearance, right?
> 
> No, I meant the visual appearance was better. I'm sure that the recursive subdivision thing has its moments, but I think for the most part its moments have passed, and on modern hardware you're better off just doing a regular computation on a larger grid.
> 

Ok. Then a simple change would be to set the default to 0. That way, if someone wants poorer performance and visual appearence, it has to be requested specifically. ;-) 

> One thing I need to think about: WMA handles piecewise-defined functions very nicely as separate disjoint meshes, whereas this code as it stands will join the meshes. I think it is possible to do the same efficiently with vectorized plot, but will need a little thought about how to accomplish.

Ok.

mathics/builtin/drawing/plot.py
bdlucas1: Ok, will do.

(But as an end user that sounds a bit confusing to me - why do I need to remember in what version the env variable appeared?)

rocky: It is more that this is the MATHICS3 project and code,  not the MATHICS project which is in a different repository and probably can't be "archived". 

MATHICS3 is a fork and extension of MATHICS, but the closing of MATHICS is up to someone else.

rocky:
The pyodide failure looks weird to me. Tests that alter filesystems should be guarded by an environment variable with "sandbox" in the name which should not be run in pyoidide.

rocky:
I had a chance to try this, and it is again pretty cool. 

There is specialized timing stuff, which feels reasonable to me. Down the line, it might be made more general, but that can be down the line. 

@mmatera your thoughts?

rocky (approved):
LGTM - and thanks for going down this path.

mathics/eval/drawing/plot_compile.py
bdlucas1: Fixed.

mathics/builtin/drawing/plot.py
bdlucas1: Fixed.

bdlucas1:
> The pyodide failure 

I had noticed that and assumed the test was just flakey - no so? I can't see how it could be related to the changes in this branch though.

bdlucas1:
So the Pyiodide tests succeeded this time.

Depending on how it's configured, Pyodide might pull packages from the net at run time.  Any chance it's doing that and some CDN was down or flakey yesterday?

.gitignore
mmatera: `git add test/builtin/drawing/test_plot_compile.py` does not work?

mathics/core/util.py
```diff
@@ -141,3 +143,58 @@ def print_sympy_tree(expr, indent=""):
             print_sympy_tree(arg, indent + "    ")
     else:
         print(f"{indent}{expr.func.__name__}({str(expr)})")
+
+
+class Timer:
```
mmatera: I wonder whether this could be DRY with the core already available in  `mathics.timing`, or just put inside there.

.gitignore
bdlucas1: It does. But when it's new things like git status, and tools like GitHub desktop, don't see it, so you have to remember to manually add any new test files specifically with git add as you suggest.

I think (per Rocky) this is only an issue on MacOS where the Test/ exclusion also excludes test/ (because the fs is case-insensitive). On other platforms things in test/ are visible to status and GitHub desktop because the exclusion only applies to Test/, not test/.

Not a big deal; if you're not comfortable with it I'll remove the change.

mathics/eval/drawing/plot_compile.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+# mappings = [dict(AdditionalMappings.__dict__), "numpy"]
+
+
+class CompileError(Exception):
+    pass
+
+
+def plot_compile(evaluation, expr, names, debug=0):
```
mmatera, line 46: Is this about compile "plots" or functions? We already have code for "compiling" expressions (using LLVM for arithmetic expressions) in `mathics.compile` and `mathics.builtin.compilation`. Maybe this code could be there together with that too.

test/builtin/drawing/test_plot_compile.py
```diff
@@ -0,0 +1,418 @@
...
+    # dict(name="SquaresR", args=[0]),
+    # dict(name="Together", args=[0]),
+    # dict(name="Tr", args=[0]),
+    # dict(name="Undefined", args=[0]),
+    # dict(name="Underflow", args=[0]),
+    # dict(name="UnitStep", args=[0]),
+    # dict(name="Variables", args=[0]),
+]
+
+session = MathicsSession()
```
mmatera: Maybe use `test.helper.session` which is already available.

mathics/core/util.py
bdlucas1: I think it makes sense to move it there; will do.

The function it provides is a little different in that it maintains a stack of timing contexts and prints the times in nested "outline" form, which I find useful for getting an overall picture of where the time is going, so I'd prefer not to combine it with the others.

.gitignore
mmatera: The question is if by adding this code, any test I have inside this folder is automatically added to the repo. Usually, I have some tests that I prefer not to include in a branch until the functionality they handle is completely supported. But it is not a big deal.

mathics/eval/drawing/plot_compile.py
bdlucas1: It's about compiling functions, but currently is only used in the context of Plot3D. I think ultimately it could have broader use, but if it's ok with you I'd prefer to get a little experience with it and let it mature for a while in that context before deciding whether, and in what form, it's ready to use in other contexts, at which point we could rename and move it.

mmatera: I have problems to see the outputs in the MathicsDjango interface. I got the error `WARNING:root:System`GraphicsComplex is not a valid primitive or directive.` Do you have extra code in Mathics-Django to make it work?

bdlucas1: No, front-ends and/or core drawing code will need to be updated to understand GraphicsComplex before enabling it to take advantage of the faster plotting. I have a demo front-end that does understand it.

.gitignore
bdlucas1: Ah, got it. No problem, I've reverted the change.

mathics/core/util.py
bdlucas1: I moved it to mathics.timing.

test/builtin/drawing/test_plot_compile.py
bdlucas1: Done.

mmatera:
Another problem I found with this is that running `make doc` with `MATHICS3_USE_VECTOR_PLOT` show that some tests now are not passing. Just an example: `DensityPlot[Sin[x y], {x, -2, 2}, {y, -2, 2}, Mesh->Full]` raise the exception
```
  File "/tmp/mathics-core/mathics/core/rules.py", line 148, in yield_match
    new_expression = apply_fn(expression, vars, options, evaluation)
  File "/tmp/mathics-core/mathics/core/rules.py", line 413, in apply_function
    self.function(evaluation=evaluation, options=options, **vars_noctx)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/mathics-core/mathics/builtin/drawing/plot.py", line 576, in eval
    return self.do_eval(plot_options, evaluation, options)
           ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/mathics-core/mathics/builtin/drawing/plot.py", line 752, in do_eval
    graphics_expr = graphics.generate(options_to_rules(options, Graphics.options))
                    ^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'generate'

```

Also, I got some warnings:
```
<lambdifygenerated-6>:2: RuntimeWarning: invalid value encountered in divide
  return sin(x*y)/(x*y)
<lambdifygenerated-7>:2: RuntimeWarning: divide by zero encountered in log
  return log(x + y**2)
<lambdifygenerated-7>:2: RuntimeWarning: invalid value encountered in log
  return log(x + y**2)
```
Could be this related to the change in the Box interface for Plot3D and DensityPlot?

mmatera:
Same error can be produced by
```
MATHICS3_USE_VECTORIZED_PLOT=yes mathics -c 'Export["test.asy", DensityPlot[Sin[x y], {x, -2, 2}, {y, -2, 2}, Mesh->Full]]
'
```

mathics/eval/drawing/plot_compile.py
bdlucas1: Let me explain in a little more detail.

This project started with some conversations between Rocky and me about the best way to do performant plotting in the Mathics context. This resulted in a demo project, https://github.com/bdlucas1/math-plotting-demo. It uses prototype versions of the code that I'm now in the process of contributing to mathics. The demo includes a prototype Manipulate builtin that does interactive rendering of plots as you change parameters, made possible by the fast vectorized Plot3D.

If you're anxious to try it you're welcome to, but setup might be a bit finicky and I'm in the process of reconstituting the demo to use the development version of Mathics with the latest contributed code, so you might be better off waiting a few days for that.

The rendering in the demo front-ends is done by the front-end itself interpreting the Graphics3DBox output of Plot3D and driving Plotly and Dash. The highest performance will be obtained by feeding the numpy arrays that are the output of the vectorized Plot3D directly into high-performance rending libraries like Plotly in the front end.  Other rendering pipelines might be possible, performance TBD.

bdlucas1:
Right, DensityPlot is not yet implemented in plot3d_vectorized. In the spirit of merging early, I wanted see if we could get this in and then iterate. It's very much a WIP as you're observing.

There will also be a few other updates to testing needed to get all tests to pass using vectorized Plot3D - for example, the tests that look at the output of Plot3D. Getting all tests to pass with MATHICS3_USE_VECTORIZED_PLOT is a goal.

Regarding the warnings, thanks for pointing them out. Looks like those are coming from the sympy-compiled code, and need some investigation.

mmatera:
> Right, DensityPlot is not yet implemented in plot3d_vectorized. In the spirit of merging early, I wanted see if we could get this in and then iterate. It's very much a WIP as you're observing.
> 
> There will also be a few other updates to testing needed to get all tests to pass using vectorized Plot3D - for example, the tests that look at the output of Plot3D. Getting all tests to pass with MATHICS3_USE_VECTORIZED_PLOT is a goal.
> 
> Regarding the warnings, thanks for pointing them out. Looks like those are coming from the sympy-compiled code, and need some investigation.

Thanks for the explanation and all this work.  Since (keeping `MATHICS3_USE_VECTORIZED_PLOT` disabled) this PR does not seem to introduce bugs or loss of functionality,  I think it is OK to merge it as it is now and iterate.

mathics/eval/drawing/plot_compile.py
mmatera: Again, thanks for the explanations. I just wanted to know what would be important to review and check in this PR. In any case, the small reorganization of the code by itself is a good contribution to merging it.

bdlucas1:
Thanks, appreciate your review comments.

Actually, if you're interested in a quick preview, here's a Pyodide-based version of the demo that runs in your browser: https://bdlucas1.github.io/math-plotting-demo/lab/index.html. Double-click on demo.ipynb from the left column, then run the first cell, and give it a while to load the required packages. You should see a plot that you can vary interactively using two sliders. (If at any point something seems to break, try "Help | Clear browser" data to reset things.)

Keep in mind that interactive performance is not as good running under Pyodide as it is running natively. Also the demo just uses the default rendering parameters of Plotly regarding axes, color, aspect ratio; I will spend some time giving it a default appearance that more closely matches Mathematica.

PR #1523, opened 2025-11-17
Use max_recursion 0 in _Plot.

rocky:
[Recent discussion](https://github.com/Mathics3/mathics-core/pull/1522#issuecomment-3538930986) suggests that the default for max_recursion in plotting functions should be 0 for faster performance and better rendering. 

Also, use Symbols over character strings more often in Plot options routines.  (There is probably more that could be done here, but that's for some other time.)

### DIFF
--- a/mathics/builtin/drawing/plot.py
+++ b/mathics/builtin/drawing/plot.py
@@ -280,9 +280,15 @@ class _Plot(Builtin, ABC):
         # MaxRecursion Option
         max_recursion_limit = 15
         maxrecursion_option = self.get_option(options, "MaxRecursion", evaluation)
-        # Old default was 3. Bruce Lucas observes that
-        # 0 using more points is faster and gives better results.
-        maxrecursion = 0
+
+        # Investigate whether the maxrecursion value is optimal. Bruce
+        # Lucas observes that in some cases, using more points and
+        # decreasing recursion is faster and gives better results.
+        # Note that the tradeoff may be different for Plot versus
+        # Plot3D. Recursive subdivision in Plot3D is probably a lot
+        # harder.
+        maxrecursion = 3
+
         try:
             if maxrecursion_option is not SymbolAutomatic:
                 maxrecursion = maxrecursion_option.to_python()


bdlucas1:
> [Recent discussion](https://github.com/Mathics3/mathics-core/pull/1522#issuecomment-3538930986) suggests that the default for max_recursion in plotting functions should be 0 for faster performance and better rendering.

The better rendering in that case was achieved by turning off recursion, yes, but also increasing the number of points to compensate, but this PR doesn't change that, does it? So this will certainly improve performance, but will decrease quality. I'm not sure what value to set the number of points at to get equivalent quality, and what the comparative performance would be.

The tradeoff may also be different for this case, Plot, because I think recursive subdivision in Plot3D is probably a lot harder (guessing a bit because I haven't looked at the code for Plot, only for Plot3D).

In general, in my experiments I didn't find the performance of Plot as it stands to be problematic - Manipulate was able to get interactive performance from the current implementation of Plot. I've been focusing on Plot3D because it's a whole different story - you need a lot more data points to get a satisfactory rendering.

rocky:
> > [Recent discussion](https://github.com/Mathics3/mathics-core/pull/1522#issuecomment-3538930986) suggests that the default for max_recursion in plotting functions should be 0 for faster performance and better rendering.
> 
> The better rendering in that case was achieved by turning off recursion, yes, but also increasing the number of points to compensate, but this PR doesn't change that, does it? So this will certainly improve performance, but will decrease quality. I'm not sure what value to set the number of points at to get equivalent quality, and what the comparative performance would be.
> 
> The tradeoff may also be different for this case, Plot, because I think recursive subdivision in Plot3D is probably a lot harder (guessing a bit because I haven't looked at the code for Plot, only for Plot3D).
> 
> In general, in my experiments I didn't find the performance of Plot as it stands to be problematic - Manipulate was able to get interactive performance from the current implementation of Plot. I've been focusing on Plot3D because it's a whole different story - you need a lot more data points to get a satisfactory rendering.

Ok. Thanks for the clarification and additional detail. I've changed the default value back to 3 and tried to capture the wisdom here. However, if I got this wrong, please suggest alternate comments. 

So what remains are just the String to Symbol changes.

bdlucas1:
Comment looks good, and the other code improvements are worthwhile. LGTM.

PR #1524, opened 2025-11-18
Vectorized plot constant

bdlucas1:
Allow constant expression in vectorized Plot3D.

This change affects only the vectorized version;  added a test for constant expression to verify that it works already for Plot3D classic.


rocky:
LGTM

PR #1527, opened 2025-11-22
Add some benchmarks

mmatera:
The part of #1517 related to the benchmarks.


rocky:
LGTM.  Pulled it and tried it. Looks like someone needs to manually watch it.

mmatera:
> needs

The results of the benchmarks are stored for a while in the output of the GitHub workflows. We could also ask pytest to store the benchmarks in a file and use it to compare between releases or after a refactor. On the other hand, we have Mathics-Benchmarks. I do not know what is best for these kinds of things.

rocky:
>  I do not know what is best for these kinds of things.

I don't either. Possibly, with time and experience, we will know better.

PR #1530, opened 2025-11-24
New plot compile functions 2

bdlucas1:
* Added an optional keyword argument `raise_on_error` to `to_sympy` and `expression_to_sympy` that causes it to raise the specified exception in cases where it would otherwise silently have returned `None`. This is used by `plot_compile` to aid in identifying the cause of compilation failure, which will help to build out compilation functionality. That is the only place it is specified, and if it is not specified, behavior is as it was before.
* Added compilation test cases in `test_plot_compile.py` for `If`, `Switch`, and `Which`. These are defined in `procedural.py`, but they are not really procedural.
* Enabled `Mod`, `FractionalPart`, `IntegerPart`, `If`, `Which`, `Min`, and `Max` for compilation by sympy-enabling them: subclass from `SympyFunction` and implement `to_sympy`. Verified that this enables compilation using the test cases in `test_plot_compile.py`, and by doing some plots:

test-030

### DIFF
--- a/mathics/builtin/procedural.py
+++ b/mathics/builtin/procedural.py
@@ -316,7 +316,7 @@ class Do(IterationFunction):
     allow_loopcontrol = True
     summary_text = "evaluate an expression looping over a variable"
 
-    def get_result(self, _items):
+    def get_result(self, _items, is_uniform=False):
         return SymbolNull
 
 
--- a/SYMBOLS_MANIFEST.txt
+++ b/SYMBOLS_MANIFEST.txt
@@ -244,6 +244,8 @@ System`Complement
 System`Complex
 System`ComplexExpand
 System`ComplexInfinity
+System`ComplexPlot
+System`ComplexPlot3D
 System`Complexes
 System`CompositeQ
 System`Composition
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -41,8 +41,8 @@ from mathics.core.builtin import (
     SympyObject,
     Test,
 )
-from mathics.core.convert.sympy import SympyExpression, from_sympy, sympy_symbol_prefix
-from mathics.core.element import BaseElement
+from mathics.core.convert.sympy import SympyExpression, from_sympy
+from mathics.core.element import BaseElement, ElementsProperties
 from mathics.core.evaluation import Evaluation
 from mathics.core.expression import Expression
 from mathics.core.expression_predefined import (
@@ -62,6 +62,7 @@ from mathics.core.symbols import (
     SymbolPlus,
     SymbolTimes,
     SymbolTrue,
+    sympy_name,
 )
 from mathics.core.systemsymbols import (
     SymbolAnd,
@@ -382,7 +383,7 @@ language/ref/ConditionalExpression.html</url>
 
         sympy_cases = (
             (expr.to_sympy(**kwargs), sympy_cond),
-            (sympy.Symbol(sympy_symbol_prefix + "System`Undefined"), True),
+            (sympy.Symbol(sympy_name(SymbolUndefined)), True),
         )
         return sympy.Piecewise(*sympy_cases)
 
@@ -788,8 +789,12 @@ class Product(IterationFunction, SympyFunction, PrefixOperator):
     sympy_name = "Product"
     throw_iterb = False
 
-    def get_result(self, elements):
-        return Expression(SymbolTimes, *elements)
+    def get_result(self, elements, is_uniform=False):
+        return Expression(
+            SymbolTimes,
+            *elements,
+            elements_properties=ElementsProperties(is_uniform=is_uniform),
+        )
 
     def to_sympy(self, expr, **kwargs):
         if expr.has_form("Product", 2) and expr.elements[1].has_form("List", 3):
@@ -1025,8 +1030,12 @@ class Sum(IterationFunction, SympyFunction, PrefixOperator):
     # Do not throw warning message for symbolic iteration bounds
     throw_iterb = False
 
-    def get_result(self, elements) -> Expression:
-        return Expression(SymbolPlus, *elements)
+    def get_result(self, elements, is_uniform=False) -> Expression:
+        return Expression(
+            SymbolPlus,
+            *elements,
+            elements_properties=ElementsProperties(is_uniform=is_uniform),
+        )
 
     def to_sympy(self, expr, **kwargs) -> Optional[SympyExpression]:
         """
--- a/mathics/builtin/drawing/plot.py
+++ b/mathics/builtin/drawing/plot.py
@@ -38,6 +38,7 @@ from mathics.core.systemsymbols import (
     SymbolLog10,
     SymbolNone,
     SymbolRGBColor,
+    SymbolSequence,
     SymbolStyle,
 )
 from mathics.eval.drawing.charts import draw_bar_chart, eval_chart
@@ -53,6 +54,7 @@ from mathics.eval.drawing.plot import (
     get_plot_range,
     get_plot_range_option,
 )
+from mathics.eval.nevaluator import eval_N
 
 # The vectorized plot function generates GraphicsComplex using NumericArray,
 # which no consumer will currently understand. So lets make it opt-in for now.
@@ -62,11 +64,19 @@ from mathics.eval.drawing.plot import (
 # TODO: work out exactly how to deploy.
 use_vectorized_plot = os.getenv("MATHICS3_USE_VECTORIZED_PLOT", False)
 if use_vectorized_plot:
-    from mathics.eval.drawing.plot3d_vectorized import eval_DensityPlot, eval_Plot3D
+    from mathics.eval.drawing.plot3d_vectorized import (
+        eval_ComplexPlot,
+        eval_ComplexPlot3D,
+        eval_DensityPlot,
+        eval_Plot3D,
+    )
 else:
-    from mathics.eval.drawing.plot3d import eval_DensityPlot, eval_Plot3D
-
-from mathics.eval.nevaluator import eval_N
+    from mathics.eval.drawing.plot3d import (
+        eval_ComplexPlot,
+        eval_ComplexPlot3D,
+        eval_DensityPlot,
+        eval_Plot3D,
+    )
 
 # This tells documentation how to sort this module
 # Here we are also hiding "drawing" since this erroneously appears at the top level.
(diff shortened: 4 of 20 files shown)


bdlucas1:
I'll look into the test failure. I wonder why it would fail only in that one environment.

mmatera:
> I'll look into the test failure. I wonder why it would fail only in that one environment.

It does not fail in an environment, but in the workflow that checks the combinatorica package.

There, it produces this error:
`/opt/hostedtoolcache/Python/3.13.9/x64/lib/python3.13/site-packages/sympy/functions/elementary/piecewise.py:34: TypeError`

Can this be related to the changes in this PR?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,87 +0,0 @@
----
```
mmatera, line 1: Why are these files gone here?

mathics/eval/drawing/plot_compile.py
```diff
@@ -63,21 +63,8 @@ def plot_compile(evaluation, expr, names, debug=0):
         print("post-eval", expr)
 
     # Ask the expr Expression to generate a sympy expression and handle errors
-    try:
```
mmatera, line 66: I was looking again this code, and the behavior and results are very close to the one in `mathics.core.convert.functions.expression_to_callable_and_args`  and `expression_to_callable`. In `expression_to_callable`, we first try to use llvmlite to get a JIT compiled function, and if this fails, try with a Mathics-based  version of `lambdify`. The code in  `plot_compile` could be an intermediate step when llvm fails, but the expression can be evaluated with sympy.

mmatera: Also, I was playing with both implementations, to see what is faster. The following code uses both implementations, and also sympy `ufuncify`, which compile expressions to C. To make this last one work, I had to hack the from/to sympy code for Symbols, to replace the context separator (`) by a C compatible character.

```

from mathics.core.load_builtin import import_and_load_builtins
from mathics.core.convert.function import expression_to_callable_and_args
from mathics.eval.drawing.plot_compile import plot_compile
from sympy.utilities.autowrap import ufuncify
from sympy import Symbol as sympy_Symbol

from datetime import datetime

import_and_load_builtins()

session = MathicsSession(character_encoding="ASCII")



def do_test():
    session.evaluate("$Assumptions={Infinity>a,  a>0, a>-Infinity}")
    session.evaluate("F[x_,y_]:=(Print[x,y];x*y)")
    session.evaluate("G[u_]:=Sin[u]")
    session.evaluate("t=2.")
    expr = session.evaluate("expr=Integrate[G[F[a,x]],{x,0,b}]")
    vars = session.evaluate("{{a,_Real},{b,_Real}}").elements
    print(expr)
    print("as sympy", expr.to_sympy())


    xs = [.001*i for i in range(1000)]
    try:
        c1 = plot_compile(session.evaluation, expr,["_uGlobal_a","_uGlobal_b"], True)
        print("expr compiled:", c1)
        start = datetime.now()
        data = [c1(x,2) for x in xs]
        end = datetime.now()
        print(c1, data[-1], end-start)
    except Exception as e:
        print("Plot compile does not work...", e)

    try:
        symbols = [sympy_Symbol("_uGlobal_a"),sympy_Symbol("_uGlobal_b"),]
        cu =   ufuncify(symbols, expr.to_sympy())
        start = datetime.now()
        data = [cu(x,2) for x in xs]
        end = datetime.now()
        print(cu, data[-1], end-start)
    except Exception as e:
        print("Funcify does not work...", e)

        
    try:
        c2, c_args = expression_to_callable_and_args(expr, vars, session.evaluation)
        start = datetime.now()
        data = [c2(x,2) for x in xs]
        end = datetime.now()
        print(c2, data[-1], end-start)
    except Exception:
        print("expression to callable didn' t worlk")



if __name__=="__main__":
    do_test()
```


These are the results:

<function _lambdifygenerated at 0x74169eb43ec0> 1.415743153766258 0:00:00.000626
<ufunc 'wrapper_module_0'> 1.415743153766258 0:00:00.000950
<CFunctionType object at 0x74169e9fdb50> 1.4157431537662581 0:00:00.000443


LLVM version seems to be faster than the other versions.

bdlucas1: Thanks for the information and the code! That result makes sense for point evaluations, but from the perspective of plotting, what matters is applying a function to a large grid of points - for example, the 200x200 grids I've been using for Plot3D to get smooth-looking results. To that end I adapted your tests to evaluate that same function over a 200x200 numpy array:

```
from mathics.core.load_builtin import import_and_load_builtins
from mathics.core.convert.function import expression_to_callable_and_args
from mathics.eval.drawing.plot_compile import plot_compile
from mathics.core.symbols import Symbol, strip_context

from sympy.utilities.autowrap import ufuncify
import sympy
import numpy as np
import time

import_and_load_builtins()

session = MathicsSession(character_encoding="ASCII")

expr = "Times[Power[a, -1], Plus[1, Times[-1, Cos[Times[a, b]]]]]"
expr = session.evaluate(expr)
print(expr)

#a = np.full((200,200), 2.22 + 3.33j)
#b = np.full((200,200), 1.11 - 9.99j)
a = np.full((200,200), 1.33)
b = np.full((200,200), 2.44)

def compile_plot(expr):
    return plot_compile(session.evaluation, expr, ["a", "b"])

def compile_ufunc(expr):
    sympy_expr = expr.to_sympy()
    subs = {
        sym: sympy.Symbol(strip_context(str(sym)))
        for sym in sympy_expr.free_symbols
    }
    sympy_expr = sympy_expr.subs(subs)
    symbols = [sympy.Symbol("a"), sympy.Symbol("b")]
    fun = ufuncify(symbols, sympy_expr)
    return fun

def compile_llvm(expr):
    vars = session.evaluate("{{a,_Real}, {b,_Real}}").elements
    fun, args = expression_to_callable_and_args(expr, vars, session.evaluation)
    fun = np.vectorize(fun)
    return fun

def do_test(name, comp):
    try:
        for i in range(5):
            fun = comp(expr)
            start = time.time()
            fun(a, b)
            print(f"{name} {(time.time()-start)*1000:6.3f} ms")
    except Exception as e:
        print(f"{name}: {e}")

do_test("plot ", compile_plot)
do_test("ufunc", compile_ufunc)
do_test("llvm ", compile_llvm)
```
For operations on arrays vectorization (highly pipelined, often parallel) operation is a huge performance win. 

* plot_compile is inherently vectorized because it `lambdify` generates calls to native numpy functions, which are generally highly optimized and vectorized.
* ufunc is likewise inherently vectorized - its purpose is to compile functions that can efficiently operate on numpy arrays. But it's still user code so generally not as good as native numpy code.
* llvm does not seem to be vectorized, but its output can be vectorized by calling `numpy.vectorize`. However this is not as good functions that are especially compiled for vectorization like ufunc or native numpy functions.

Given the above, the following timings for the expression evaluated over a 200x200 numpy array aren't surprising:
```
plot   0.214 ms
plot   0.102 ms
plot   0.101 ms
plot   0.101 ms
plot   0.100 ms

ufunc  0.233 ms
ufunc  0.234 ms
ufunc  0.235 ms
ufunc  0.253 ms
ufunc  0.233 ms

llvm  15.853 ms
llvm  12.502 ms
llvm   9.758 ms
llvm   9.943 ms
llvm   9.849 ms
```
One caveat also is that ufunc and llvm don't seem to handle complex numbers.

So in any case it seems some combination of the above with fallbacks that may depend on the goal (fast point evaluation or fast vector evaluation) may give us the best combination of performance and coverage.

.github/ISSUE_TEMPLATE/bug_report.md
bdlucas1: Sorry about that, and thanks for the catch. I was experimenting and I thought I had restored to original state and I didn't catch it when I opened the pull request.

bdlucas1:
> Can this be related to the changes in this PR?

It may well be, and I have a suspicion as to the reason. I'm looking into running those tests.

mathics/eval/drawing/plot_compile.py
mmatera: @bdlucas1, thanks for the feedback! Indeed, lambdify is astonishingly fast!

bdlucas1:
> > Can this be related to the changes in this PR?
> 
> It may well be, and I have a suspicion as to the reason. I'm looking into running those tests.

The issue was that the new `If.to_sympy` was passing `sympy.Piecewise` an argument it didn't like under some conditions. Following the example of `ConditionalExpression.to_sympy` which likewise uses `sympy.Piecewise`, and looking into the code in `sympy.Piecewise` to see what it is expecting, I updated `If.to_sympy` to be more restrictive about what kinds of arguments it passes to `sympy.Piecewise`, and now the tests all pass.

 Here's the explanation I left in `If.to_sympy`:
```
            # sympy.Piecewise is is picky as to what type of conds it will accept,
            # allowing Boolean, Relational, and Symbol (as an honorary Boolean,
            # accompanied by a comment in the code that this isn't really correct).
            # Seems an attempt at early typechecking, but maybe too restrictive -
            # what about Exprs in general that might or might not evaluate to Boolean?
            # See similar code in mathics.builtin.arithmetic.ConditionalExpression.
```

mathics/builtin/procedural.py
```diff
@@ -443,6 +450,26 @@ def eval_with_false_and_other(self, condition, t, f, u, evaluation):
         else:
             return u.evaluate(evaluation)
 
+    def to_sympy(self, expr, **kwargs):
+        if len(expr.elements) == 3:
+            sympy_cond = expr.elements[0].to_sympy(**kwargs)
+            sympy_true = expr.elements[1].to_sympy(**kwargs)
```
mmatera: These conversions are not necessary if cond is not a valid condition. I'm would put this and the next line inside the conditional containing the successful  return.

mmatera: Apart from this detail, LGTM

bdlucas1:
@mmatera, any thoughts on this? I noticed you've been keeping it up-to-date with master, so I guess you're not ready to merge it yet?  If you have concerns and want more time to think about it that's absolutely fine from my perspective, as it's not holding up any of my work.

mathics/builtin/procedural.py
mmatera: @bdlucas1 This is my only comment on this. In any case, we can merge this and fix the order later.

PR #1531, opened 2025-11-25
Vectorized densityplot

bdlucas1:
Add support for vectorized DensityPlot. Default color palette for now; support for user-specified color palette TBD.


bdlucas1:
demo-220

test/timings/test_uniform_tables.py
bdlucas1, line 1: I guess these changes are a result of isort running on this file maybe when I upated from master, or something like that. Looks like it removed some unused imports. Let me know if this is a problem and I'll revert this change in the PR.

mmatera: This is just that I forgot to remove the imports in this minimal benchmark. Full tests are in #1517, so the imports will come back when we merge it.

mmatera (approved):
LGTM

Issue #1533, opened 2025-11-27
Sympy symbol names should not use '`' as context marks


mmatera:
Motivated by the good results obtained in using sympy lambdify for compiling expressions, I have been looking at the code that converts Mathics expression to sympy. One problem I found is that when I convert an expression to sympy, symbol and function names include the '`' character, which in Mathics is used to split contexts. In the sympy side, it looks OK until we try to produce source code (for example, when we try to use `ufuncify`). 
Ideally, context in sympy variables should be separated by a harmless character like `_`.

PR #1534, opened 2025-11-27
Improve sympy conversions

mmatera:
To start addressing #1533, this PR DRY the existing code, to avoid inconsistencies when the separator character be changed. 
Also, some constant names are changed to fullfil the Python conventions.

### DIFF
--- a/test/core/test_sympy_python_convert.py
+++ b/test/core/test_sympy_python_convert.py
@@ -182,16 +182,7 @@ class SympyConvert(unittest.TestCase):
             SymbolFunction,
             Expression(SymbolPlus, Integer1, Expression(SymbolSlot, Integer1)),
         )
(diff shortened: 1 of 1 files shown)


rocky:
LGTM - good observation. 

If I have this right, `_`, cannot be used in a WMA/Mathics3 Symbol name since `_`  is interpreted as an operator.  But it is a perfectly valid as a Python/SymPy name Right?

mmatera:
> LGTM - good observation.
> 
> If I have this right, `_`, cannot be used in a WMA/Mathics3 Symbol name since `_` is interpreted as an operator. But it is a perfectly valid as a Python/SymPy name Right?

Yep, this is why is a good candidate for replace ``` ` ``` in sympy.

PR #1535, opened 2025-11-30
Good bye sympy dummy attribute

mmatera:
This continues #1533. Now, let's remove the `sympy_dummy` attribute for symbols. When it is necessary (or useful) like in `Sum` or `Product` expressions, we can tell to `to_sympy` which symbols should be handled as sympy dummy symbols.

### DIFF
--- a/mathics/core/symbols.py
+++ b/mathics/core/symbols.py
@@ -342,7 +342,6 @@ class Symbol(Atom, NumericOperators, EvalMixin):
 
     name: str
     hash: int
-    sympy_dummy: Any
     _short_name: str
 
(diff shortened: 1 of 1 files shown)


mmatera:
@rocky, you have revised #1536 and #1534. This is just remove the `sympy_dummy` attribute from the `Symbol` class. Are you OK with this, too?

rocky:
> @rocky, you have revised #1536 and #1534. This is just remove the `sympy_dummy` attribute from the `Symbol` class. Are you OK with this, too?

Yes, thanks!

PR #1536, opened 2025-11-30
Underscore instead of grave tilde in sympy names as context separators

mmatera:
This continues #1535, by changing the way in which symbol names are translated to sympy.


mathics/core/convert/sympy.py
```diff
@@ -371,16 +370,16 @@ def old_from_sympy(expr) -> BaseElement:
...
                 else:
-                    name = name[len(SYMPY_SYMBOL_PREFIX) :]
+                    name = name[len(SYMPY_SYMBOL_PREFIX) :].replace("_", "`")
                 # Probably, this should be the value attribute
                 return Symbol(name)
             if is_Cn_expr(name):
                 return Expression(SymbolC, Integer(int(name[1:])))
             if name.startswith(SYMPY_SYMBOL_PREFIX):
-                name = name[len(SYMPY_SYMBOL_PREFIX) :]
+                name = name[len(SYMPY_SYMBOL_PREFIX) :].replace("_", "`")
```
rocky: This idiom appears about 3 times in this PR. Maybe it should be a little function?

mathics/core/symbols.py
```diff
@@ -754,7 +761,7 @@ def symbol_set(*symbols: Symbol) -> FrozenSet[Symbol]:
 
 def sympy_name(mathics_symbol: Symbol):
     """Convert a mathics symbol name into a sympy symbol name"""
-    return SYMPY_SYMBOL_PREFIX + mathics_symbol.name
+    return SYMPY_SYMBOL_PREFIX + mathics_symbol.name.replace("`", "_")
```
rocky, line 764: `SYMPY_SYMBOL_PREFIX + mathics.symbol.name.replace("``'", "_")` seems like an idiom. Maybe it could be turned into a function?

test/core/test_sympy_python_convert.py
```diff
@@ -49,10 +49,10 @@ def compare(self, mathics_expr, sympy_expr, **kwargs):
         self.compare_to_mathics(mathics_expr, sympy_expr)
 
     def testSymbol(self):
-        self.compare(Symbol("Global`x"), sympy.Symbol(f"{SYMPY_SYMBOL_PREFIX}Global`x"))
+        self.compare(Symbol("Global`x"), sympy.Symbol(f"{SYMPY_SYMBOL_PREFIX}Global_x"))
```
rocky: This too, feels like an idiom that is used a couple of times. Maybe turn into a function, too.

The thing about making functions like this that I appreciate is that the function name signals something, and a docstring is available to explain what's going on better.

rocky:
LGTM. Please consider DRYing the code.

mathics/core/symbols.py
mmatera: In this case, is already a function. What is not a function yet is the conversion in the opposite direction.

rocky (approved):
Thanks for the additional changes.

PR #1537, opened 2025-11-30
Convert results just if the condition is valid in If.to_sympy

mmatera:
Just a small detail about converting conditional expressions to sympy.


bdlucas1:
lgtm, thanks

PR #1538, opened 2025-11-30
Plot compile to lambdify compile

mmatera:
@bdlucas1, this is a proposal about where I think the lambdify code should be located. If this is not too disruptive, the next thing would be to use this code in the different places where we need compiled versions of WL expressions, like in table generations, numerical integration, and optimizers.


bdlucas1:
> If this is not too disruptive,
From that perspective I'm fine with this, if you think it's mature enough to be useful.

> use this code in the different places where we need compiled versions of WL expressions
I have no comment on this because I'm not familiar with those uses, other than to say I'm not sure how complete function coverage is for lambdify_compile - not really sure how exhaustive the list of tests is, and of those a lot fail, although those mostly tend to be obscure numerical functions.

One minor comment on the import in plot3d_vectorized.py which I will add, otherwise lgtm.

mathics/eval/drawing/plot3d_vectorized.py
```diff
@@ -8,12 +8,12 @@
 import numpy as np
 
 from mathics.builtin.colors.color_internals import convert_color
+from mathics.core.convert.lambdify import lambdify_compile as plot_compile
```
bdlucas1: A minor point - I think it would be better to import as lambdify_compile, and change the one use of it in this file, to get rid of the name plot_compile altogether.

mathics/core/atoms.py
```diff
@@ -628,7 +628,7 @@ def element_order(self) -> tuple:
         of an expression. The tuple is ultimately compared lexicographically.
         """
 
-        value = self._value
+        value: sympy.Float = self._value
```
rocky: Small thing. Generally, changing the type of  variables, here `value` is not a good practice, even in such as short span. How about using `v`,  `sympy_float` or something else.

rocky:
Just catching up on the discussion. LGTM. (One small optional suggestion though.)

mathics/core/atoms.py
mmatera:
```suggestion
        value = self._value
```

mmatera:
@bdlucas1  when I merge this with master, I got an error in the new `/home/mauricio/Projects/mathics-core/test/builtin/drawing/test_plot_detail.py`:
```
name = 'barchart-cls-def', str_expr = 'BarChart[{3,5,2,7}]', vec = False, opt = None, act_dir = '/tmp'

    def one_test(name, str_expr, vec, opt, act_dir="/tmp"):
        # update name and set use_vectorized_plot depending on
        # whether vectorized test
        if vec:
            name += "-vec"
            plot.use_vectorized_plot = vec
        else:
            name += "-cls"
    
        # update name and splice in options depending on
        # whether default or with-options test
        if opt:
            name += "-opt"
            str_expr = f"{str_expr[:-1]}, {opt}]"
        else:
            name += "-def"
    
        print(f"=== running {name} {str_expr}")
    
        try:
            # evaluate the expression to be tested
            expr = session.parse(str_expr)
            act_expr = expr.evaluate(session.evaluation)
            if len(session.evaluation.out):
                print("=== messages:")
                for message in session.evaluation.out:
                    print(message.text)
>           assert not session.evaluation.out, "no output messages expected"
E           AssertionError: no output messages expected
E           assert not [<mathics.core.evaluation.Message object at 0x7464cd741d30>]
E            +  where [<mathics.core.evaluation.Message object at 0x7464cd741d30>] = <mathics.core.evaluation.Evaluation object at 0x7464cf36a7a0>.out
E            +    where <mathics.core.evaluation.Evaluation object at 0x7464cf36a7a0> = <mathics.session.MathicsSession object at 0x7464dabeaa50>.evaluation

/home/mauricio/Projects/mathics-core/test/builtin/drawing/test_plot_detail.py:148: AssertionError
```
Any idea about what could produce this?

PR #1542, opened 2025-12-13
Add plot doctests

bdlucas1:
My understanding is that the doctests for plotting don't check the generated output and only look for the "-Graphics-" string, is that correct?

Here's a proposal for tests that compare the generated Graphics* output with expected output. The tests in `test_plot_detail.py` are now augmented with about 125 plot-related doctests, with corresponding expected reference output in `test_plot_detail_ref/doc-*`.

I hope this will give us a lot more confidence to refactor plotting code without breaking anything in the doc. These tests take about 10 seconds to run so having them separated out from the doctests will make development cycles faster.

Couple notes:
* I chose to store the test definitions in a yaml file `doc_tests.yaml` because with that number of tests a Python data structure is a bit unwieldy. The yaml file is leaner  and I think easier to peruse and edit.
* I obtained the 128 tests by scraping the output from running the current doctests and munging with sed, grep, etc.. Going forward maybe we can do something a little more robust - maybe from time-to-time run a python script tbd to extract them from the docstrings in the same way the existing doctests are extracted, compare to the existing list, and add new ones.


mmatera:
> My understanding is that the doctests for plotting don't check the generated output and only look for the "-Graphics-" string, is that correct?
> 

Actually, by testing the final textual output "-Graphics-", we verify that the entire evaluation of the plot expression has been completed without raising error messages. Then, when we run the docpipeline with a formatted output (html or latex) the graphics are actually generated.

There are also some pytests that check some basic aspects on converting graphics expressions into Asymptote and SVG expressions. This effort may go in that direction.

> I hope this will give us a lot more confidence to refactor plotting code without breaking anything in the doc. These tests take about 10 seconds to run so having them separated out from the doctests will make development cycles faster.

There are two typical ways to break documentation entries involving graphics. One is by building graphical expressions that can not be translated into box expressions. The other main possibility happens when the formatters that convert GraphicsBox expressions into SVG or Asymptote do not support some of the symbols or options in the GraphicsBox expressions. In any case, it is probably a good idea to check that after changes in the implementation of plot routines, the generated graphics do not change in an unexpected way.

bdlucas1:
> check that after changes in the implementation of plot routines, the generated graphics do not change in an unexpected way.

Yes, my particular concern is things like colors not being emitted correctly, plot range being wrong,  etc. The plot would still be generated without error, but the visual appearance would be wrong. Some of these details are fairly subtle both in interpretation of input and proper generation of output, so potentially easy to break with a refactoring.

bdlucas1:
After disabling a few tests on pyodide one-by-one due to differences there, I finally disabled all the new doctests because it didn't seem worthwhile to keep at it. Some differences:
* One test was a Plot3D with a `/ (x y)` and the plot range covered both the x and y axis, so I think this one was inherently numerically unstable, i.e. a small difference in computation could lead to a large difference in output.
* In once case the expression being plotted was `3` which was represented as an Integer type on all platforms but pyodide where it was a Real. Most likely a difference in what type numpy chose for an array given an input of `3`.
* In another case `Plot3D[Log[x + y^2], {x, -1, 1}, {y, -1, 1}]` aborted due to iteration limit. I think I experienced something similar in an unrelated plot involving Log. Possibly related to behavior of Log with negative inputs?

Looks like we run this same test with our doctests and it succeeds:
    
    2025-12-13T19:37:18.6203778Z 1447 ( 5): TEST Plot3D[Log[x + y^2], {x, -1, 1}, {y, -1, 1}]

Not sure why the behavior should differ. Looks like we have some earlier doctests that change the iteration limit - is that sticky?

bdlucas1:
To be clear, I meant I disabled the new tests on pyodide. They all pass on the other architectures, so I still think this is a worthwhile addition to our test suite.

rocky:
> After disabling a few tests on pyodide one-by-one due to differences there, I finally disabled all the new doctests because it didn't seem worthwhile to keep at it. Some differences:
> 
> * One test was a Plot3D with a `/ (x y)` and the plot range covered both the x and y axis, so I think this one was inherently numerically unstable, i.e. a small difference in computation could lead to a large difference in output.
> * In once case the expression being plotted was `3` which was represented as an Integer type on all platforms but pyodide where it was a Real. Most likely a difference in what type numpy chose for an array given an input of `3`.
> * In another case `Plot3D[Log[x + y^2], {x, -1, 1}, {y, -1, 1}]` aborted due to iteration limit. I think I experienced something similar in an unrelated plot involving Log. Possibly related to behavior of Log with negative inputs?
> 
> Looks like we run this same test with our doctests and it succeeds:
> 
> ```
> 2025-12-13T19:37:18.6203778Z 1447 ( 5): TEST Plot3D[Log[x + y^2], {x, -1, 1}, {y, -1, 1}]
> ```
> 
> Not sure why the behavior should differ. Looks like we have some earlier doctests that change the iteration limit - is that sticky?


Setting global variables is both "sticky" and icky. 

Now for a little history.  Originally, most of the tests were these doctests instead of having pytests or unit tests. That was just plain wrong.  The doctest is a wrong-headed homegrown approach that should not have existed. The long-term plan is to ditch this and the homegrown conversion to LaTeX, and instead go with some more conventional approach like sphinx.
 
In "doctests", the concern is more about documenting or describing behavior, and secondarily testing functionality.  That is why in my mind it's okay to report "there some sort of graphics here, and it's too difficult for me to try to check further than that".   For the "further than that" part, we should be doing _exactly_ the kind of thing you've been doing. Thanks for this.

rocky:
> * I obtained the 128 tests by scraping the output from running the current doctests and munging with sed, grep, etc.. Going forward maybe we can do something a little more robust - maybe from time-to-time run a python script tbd to extract them from the docstrings in the same way the existing doctests are extracted, compare to the existing list, and add new ones.

Sure. But given how things are going to change, this kind of thing is low on my list. The other stuff of working out how to test graphics more robustly and reasonably is more important.

rocky:
LGTM.

rocky:
> These tests take about 10 seconds to run so having them separated out from the doctests will make development cycles faster.

What we have typically been doing in such situations is to run the test _only_ when some environment variable is set. 

See, for example, the benchmarks test: https://github.com/Mathics3/mathics-core/blob/master/test/timings/test_uniform_tables.py#L17-L19

_It is important to keep the overall test running `make doctest` and `make pytest`  (which equals `make test`) not take too long._

Otherwise, people will tend to shy away from running tests before committing. And that is bad.

PR #1543, opened 2025-12-14
use `session.evaluate` instead of `session.parse(...).evaluate`

mmatera:
Just a small fix to the `test/builtin/drawing/test_plot_detail.py`: using `session.evaluate` ensures that trailing messages from previous tests are cleaned before the new evaluation.
This was the reason for the last failure in CI for #1538

### DIFF
--- a/test/builtin/drawing/test_plot_detail.py
+++ b/test/builtin/drawing/test_plot_detail.py
@@ -139,8 +139,7 @@ def one_test(name, str_expr, vec, opt, act_dir="/tmp"):
 
     try:
         # evaluate the expression to be tested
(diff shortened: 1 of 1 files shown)


bdlucas1:
Thanks for the fix.

Does this also clear out things like changes to iteration limit from previous tests, which I suspected might be the cause for one test failure?

mmatera:
> Thanks for the fix.
> 
> Does this also clear out things like changes to iteration limit from previous tests, which I suspected might be the cause for one test failure?

Apparently, it does. It seems that the message originates from a previous evaluation, where it was expected.

PR #1546, opened 2025-12-17
Use PyPI `Timed-Threads` for `TimeConstrained[]`

rocky:
Using newly released on PyPI Timed-Threads for `TimeConstrained[]` further running on Python 3.14+ packaging more easily.

### DIFF
--- a/.github/workflows/consistency-checks.yml
+++ b/.github/workflows/consistency-checks.yml
@@ -22,8 +22,6 @@ jobs:
       run: |
         sudo apt update -qq && sudo apt install llvm-dev remake
         python -m pip install --upgrade pip
-        # First, install our Timed-Threads
-        pip install Timed-Threads
         # We can comment out after next Mathics-Scanner release
         # python -m pip install -e git+https://github.com/Mathics3/mathics-scanner#egg=Mathics-Scanner[full]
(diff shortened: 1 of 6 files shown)


.github/workflows/consistency-checks.yml
```diff
@@ -22,6 +22,8 @@ jobs:
       run: |
         sudo apt update -qq && sudo apt install llvm-dev remake
         python -m pip install --upgrade pip
+        # First, install our Timed-Threads
```
rocky: @mmatera Now that we have a package that can be installed from PyPI should it be a required package (except when installing from pyodide and in that case dropped), or an optional one where we also install it here?

mmatera: I think it is OK to include it as a default package. Then maybe we can move the TimeConstrained tests to the tests in TimedThreads.

rocky:
@mmatera: now that Timed-Threads has been removed from explicit install in workflows CI, I think we are ready for a merge to master.  Right?

mmatera:
> @mmatera: now that Timed-Threads has been removed from explicit install in workflows CI, I think we are ready for a merge to master. Right?

sure. LGTM

PR #1547, opened 2025-12-18
Don't raise exception in `evaluate()` if parsing expr returns `None`

bdlucas1:
This can happen for example if you give it a string consisting of just a comment. I'm not sure if this behavior - returning None - is correct, but at least it's consistent with what parse does.

Also I suppose there should be a test for this, but not sure where the best place for it would be.

### DIFF
--- a/mathics/session.py
+++ b/mathics/session.py
@@ -19,6 +19,7 @@ from mathics_scanner.location import ContainerKind
 from mathics.core.definitions import Definitions
 from mathics.core.evaluation import Evaluation, Result
 from mathics.core.parser import MathicsSingleLineFeeder, parse
+from mathics.core.symbols import SymbolNull
 
 
 def autoload_files(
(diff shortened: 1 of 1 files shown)


mathics/session.py
```diff
@@ -150,7 +150,7 @@ def evaluate(self, str_expression, timeout=None, form=None):
         )
         if form is None:
             form = self.form
-        self.last_result = expr.evaluate(self.evaluation)
+        self.last_result = expr.evaluate(self.evaluation) if expr else None
```
mmatera: In WMA, an empty expression should result in `Null` (`SymbolNull`). I think that what we should do here instead is to check in line 147 if   `expr` is None, and return `SymbolNull`.

bdlucas1: Thanks. I've updated the PR accordingly.

rocky: > Thanks. I've updated the PR accordingly.

Thanks. LGTM. (And thanks @mmatera for spotting.)

mmatera: LGTM

mmatera: LGTM

PR #1548, opened 2025-12-21
Split out mathics.core.atoms

rocky:
Split mathics.core.atoms into:

* numeric atoms
* strings
* arrays

In the process, correct erroneous places where we were importing `Atom` from `mathics.core.atoms` (it resides in `mathics.core.symbols`).

Same thing was done one place for `Symbol` was imported from mathics.core.atoms.

Remove some unused imports

### DIFF
--- a/mathics/core/atoms/__init__.py
+++ b/mathics/core/atoms/__init__.py
@@ -2,7 +2,7 @@
 Mathics3 Atoms from which compound Mathics3 compound Expressions are made of.
 """
 from mathics.core.atoms.arrays import ByteArray, NumericArray
-from mathics.core.atoms.numeric import (
+from mathics.core.atoms.numerics import (
     COMPARE_PREC,
     FP_MANTISA_BINARY_DIGITS,
     MATHICS3_COMPLEX_I,
--- a/mathics/core/atoms/arrays.py
+++ b/mathics/core/atoms/arrays.py
@@ -7,7 +7,7 @@ from typing import Optional, Tuple, Union
 
 import numpy
 
-from mathics.core.atoms.numeric import Integer
+from mathics.core.atoms.numerics import Integer
 from mathics.core.atoms.strings import String
 from mathics.core.element import ImmutableValueMixin
 from mathics.core.keycomparable import (
--- a/mathics/core/atoms/numerics.py
+++ b/mathics/core/atoms/numerics.py
@@ -1,6 +1,8 @@
 """
 Numeric types: Number, Integer Real, MachineReal, PrecisionReal, Complex, Rational
 """
+# Note: Python warns of ambiguity numpy's module numpy.numerics if we name this file this numeric.py
+
 import math
 import re
 from functools import cache
--- a/mathics/core/atoms/strings.py
+++ b/mathics/core/atoms/strings.py
@@ -1,7 +1,7 @@
 """
 Mathics3 String
 """
-# Note: Python warns of ambiguity if we name this file this string.py
+# Note: Python warns of ambiguity Python's module string if we name this file this string.py
 
 import math
 


rocky:
@mmatera I wonder if we couldn't add more objects via numpy for `Tensors` and homogeneous arrays.

mmatera:
> @mmatera I wonder if we couldn't add more objects via numpy for `Tensors` and homogeneous arrays. 

From the WL side, all these objects are just List objects. I guess that the right way to do this would be to write a wrapper class for NDArray objects which provides the interface of a ListExpression. And then, add specific code for the expressions that build Lists, s.t. for numeric arrays, an instance of the wrapper it return. Then, we could reimplement the code in tensors, and linear algebra to take advantage of this representation.

rocky:
> > @mmatera I wonder if we couldn't add more objects via numpy for `Tensors` and homogeneous arrays.
> 
> From the WL side, all these objects are just List objects. I guess that the right way to do this would be to write a wrapper class for NDArray objects which provides the interface of a ListExpression. And then, add specific code for the expressions that build Lists, s.t. for numeric arrays, an instance of the wrapper it return. Then, we could reimplement the code in tensors, and linear algebra to take advantage of this representation.

Yes, I think so.

bdlucas1:
> > @mmatera I wonder if we couldn't add more objects via numpy for `Tensors` and homogeneous arrays.
> 
> From the WL side, all these objects are just List objects. I guess that the right way to do this would be to write a wrapper class for NDArray objects which provides the interface of a ListExpression. And then, add specific code for the expressions that build Lists, s.t. for numeric arrays, an instance of the wrapper it return. Then, we could reimplement the code in tensors, and linear algebra to take advantage of this representation.

In case it helps, I prototyped such a class in an early version of the vectorized plotting demo thinking to use it for Graphics output from vectorized plotting, but abandoned it when I discovered NumericArray. The primary difficulty I encountered was stamping out casual references to .elements that caused the list to be instantiated from the ndarray, killing performance (space and time). Many of these occurred even during instantiation of the List, some later, and no way to know when and where more such references might pop up, so I deemed it too fragile from a performance perspective. But maybe my approach was just not sophisticated enough and there's some way to avoid this.

In any case you can find it [here](https://github.com/bdlucas1/math-plotting-demo/blob/main/demo/list.py) if it helps.

rocky:
> > > @mmatera I wonder if we couldn't add more objects via numpy for `Tensors` and homogeneous arrays.
> > 
> > 
> > From the WL side, all these objects are just List objects. I guess that the right way to do this would be to write a wrapper class for NDArray objects which provides the interface of a ListExpression. And then, add specific code for the expressions that build Lists, s.t. for numeric arrays, an instance of the wrapper it return. Then, we could reimplement the code in tensors, and linear algebra to take advantage of this representation.
> 
> In case it helps, I prototyped such a class in an early version of the vectorized plotting demo thinking to use it for Graphics output from vectorized plotting, but abandoned it when I discovered NumericArray. The primary difficulty I encountered was stamping out casual references to .elements that caused the list to be instantiated from the ndarray, killing performance (space and time). Many of these occurred even during the instantiation of the List, some later, and no way to know when and where more such references might pop up, so I deemed it too fragile from a performance perspective. But maybe my approach was just not sophisticated enough and there's some way to avoid this.

When one starts off declaring an array kind of object, then for operations like `Times`, instead of converting the more efficient representation to the less-efficient but more general  Mathics3 ListExpression, `Times` would instead be modified to check to see for this special array-like list, and handle it in the more performant vector-like way. 

In a sense, wrapping an expression inside Plot does this declarative thing for the scope of just the expression (and uses numpy or sympy's version of `Times`.  If instead the variable used an explicit NumericArray or Tensor type, this information is stored inside a variable, which can have a larger scope than a single expression.

The thing compilation can give you that we don't do right now is type propagation, so something like a specific use of `Range`  can be specialized for a more performant way to feed into, say, numeric arrays. 

> 
> In any case you can find it [here](https://github.com/bdlucas1/math-plotting-demo/blob/main/demo/list.py) if it helps.

PR #1550, opened 2025-12-21
Split plots builtins by base class

bdlucas1:
This is a preliminary code rearrangement to make subsequent refactoring easier to do and easier to review. Plotting functions from plot.py that shared common base classes were pulled out into separate files by base class. Mostly a mechanical cut&paste operation, with a little manual cleanup to move variables defined at the top of the file.

### DIFF
--- a/mathics/builtin/drawing/plot_chart.py
+++ b/mathics/builtin/drawing/plot_chart.py
@@ -19,7 +19,7 @@ from mathics.eval.drawing.charts import draw_bar_chart, eval_chart
 from mathics.eval.nevaluator import eval_N
 
 # This tells documentation how to sort this module
-from .plot import sort_order # noqa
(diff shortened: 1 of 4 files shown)


bdlucas1:
Maybe that's where we'll end up, but at this point there is still some significant function in plot.py so creating a plot module would hide that significant function away in __init__.py which I think would be confusing (at least, it would confuse me). At this point I'd rather focus on what's going on in plot_plot.py, so maybe we can consider this a step along the way.

mmatera:
> Maybe that's where we'll end up, but at this point there is still some significant function in plot.py so creating a plot module would hide that significant function away in **init**.py which I think would be confusing (at least, it would confuse me). At this point I'd rather focus on what's going on in plot_plot.py, so maybe we can consider this a step along the way.

In that case, LGTM

PR #1551, opened 2025-12-22
Add a script entrypoint named `mathics3`.

hzhangxyz:
I'm always frustrated when starting mathics by uvx. I have to run `uvx --from mathics3 mathics` to start mathics.

This commit make user able to execute mathics with uvx directly like `uvx mathics3`.


rocky:
I am okay with this. @mmatera, your thoughts?

PR #1553, opened 2025-12-22
Emit only valid Graphics options from Plot

bdlucas1:
Currently Plot passes all options on to the generated Graphics object. This makes the tests more fragile than necessary because these options, not being valid Graphics options, are no-ops, and this is a bit of an impediment to refactoring.

* Only pass valid Graphics options on from Plot to the resulting Graphics
* Add the (non-standard) LogPlot option that we use to make logarithmic axes to list of valid Graphics options

I verified through my end-to-end tests in that render all the doctests (and more) using m3d and compare the resulting image to an expected image that there are no changes in the generated images, confirming that the options that were being passed are no-ops.


rocky:
LGTM

Issue #1554, opened 2025-12-23
ListPlot in non-vector mode is showing log scale


rocky:
## Description

Recent changes seem to have broken non-vector-mode tick marks in `ListPlot[]`

## How to Reproduce

```
mathicsscript --matplotlib -c 'ListPlot[{1, 1, 2}]'
```

## Output Given

<img width="596" height="476" alt="Image" src="https://github.com/user-attachments/assets/d142c812-6b2f-44bc-87cd-0de7b63a7443" />

## Expected behavior

No logplot scale, y axis goes from 0 to 2.

## Your Environment

Current Mathics core development environment 9.0.1.dev0 and same for mathicsscript
  mathics --version

## Workarounds

Use last release ;-)

bdlucas1:
See #1555 for analysis and fix.

We have a gap in testing, in that no (or maybe just a few?) tests check the generated SVG graphics. The tests in test_plot_detail.py can easily be enhanced to check that the SVG graphics for all the doctests aren't changed. If you'd like we can leave this issue open, assigned to me, to do that.

rocky:
Thanks for fixing. 

I will double-check the fix when I get a chance, sometime soon I hope.

bdlucas1:
I also submitted PR #1556, which is #1555 plus SVG tests. Please double-check one or the other, depending on whether you're ok with adding the SVG tests. I think they will help prevent mishaps like this in the future.

bdlucas1:
As noted on PR #1556, it's probably too fragile, so suggest just going with PR #1555 at this point.

rocky:
> As noted on PR [#1556](https://github.com/Mathics3/mathics-core/pull/1556), it's probably too fragile, so suggest just going with PR [#1555](https://github.com/Mathics3/mathics-core/pull/1555) at this point.

Ok.  As often happens here, there are good ideas, e. g., more robust testing or using compilation (or a more aggressive transformation) to speed up computation. But these ideas need iteration over time to get right. 

I haven't been following the testing process in detail, other than we're testing at the S-expression level (as opposed to a rendered SVG level), which is a good thing.

The obvious things that would seem to be needed are:

* some way to indicate tolerance on numeric values, and possibly
* checking counts of data items when things can vary widely  (e. g., all we careabout  are there are _n_ points, or _n_ points that range between, _x_ and _y_.)
[...]

PR #1555, opened 2025-12-23
Fix spurious log axes

bdlucas1:
Issue #1554 

PR #1553 restricted options passed from Plot to Graphics to only include valid Graphics options in order make the tests more robust. This required adding LogPlot to the list of allowable Graphics options, with a default of False. So now all Graphics objects have a LogPlot option, most with a value of False, whereas previously only log plots generated a LogPlot option in Graphics, with a value of True. However the SVG generation code in builtin/box/graphics.py did not check the value of the LogPlot option when deciding whether to generate log axes.


rocky:
I had a chance to verify the fix, and it addresses the problem I encountered. Thanks.

LGTM.

PR #1556, opened 2025-12-23
Fix spurious log axes plus add svg tests

bdlucas1:
This PR includes #1555, plus it also adds SVG tests: the doctest-based tests, in classic mode (using the non-vectorized plotting functions), in addition to checking the direct output of Plot*, also generate SVG files and check them.

I submitted this as PR separate from #1555 rather than update #1555 in case for any reason we decide not to add these tests we can still use #1555 to pick up the fix for issue #1554.

### DIFF
--- a/test/builtin/drawing/test_plot_detail.py
+++ b/test/builtin/drawing/test_plot_detail.py
@@ -73,9 +73,9 @@ except:
 from test.helper import session
 
 import mathics.builtin.drawing.plot as plot
+from mathics.core.expression import Expression
 from mathics.core.symbols import Symbol
 from mathics.core.util import print_expression_tree
-from mathics.core.expression import Expression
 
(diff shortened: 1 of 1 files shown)


bdlucas1:
OK, never mind. Looks like we're running into the same numerical precision issues where we see very slightly different results on different systems, so these are too fragile to run in our CI tests and I think a fix will be difficult.

I currently have a set of end-to-end tests that generate images and compare them using m3d that I run regularly as I do development. I will add these SVG tests to that suite. It's not ideal that they can't be run as CI tests, but I don't have a good solution for that.

You can close this PR unless you have some idea how these can be made into CI tests without inordinate effort.

bdlucas1:
Superceded by #1565 which does much the same thing but in such a way that the tests aren't sensitive to changes across platforms

PR #1560, opened 2025-12-25
Fix using bool where Symbols should used.

mmatera:
Looking at the code of MakeBoxes I found that in BoxExpression, some times we store options like Python objects instead of  Mathics expressions, which produce errors when we try to display them. This PR fixes that issue.


rocky:
LGTM - thanks!

PR #1558, opened 2025-12-24
Consolidate plot options processing into PlotOptions

bdlucas1:
Sorry about the size of this PR - I tried to find a way to do it more incrementally, but things were too tangled for me to figure out how to do that.

The main thrust of this PR is to consolidate options processing for plots in a common class, PlotOptions, eliminating as much duplicate code as possible. PlotOptions was created initially for and shared by the 3d plots that inherit from _Plot3D in plot_plots3d.py. With this PR it has been generalized and is also used by the plots in plot_plots.py that inherit from _Plot.

(With one exception: DiscretePlot is in plot_plots.py and inherits from _Plot, but it calls eval_ListPlot. TBD whether or how that can be untangled in a future PR).

I also tried to move the Python data structures in PlotOptions to use Symbols instead of strings. To facilitate this I added a keyword argument, preserve_symbols, to to_python. The effect is Symbols are not converted to strings if this arg is True.

All tests pass and I've also verified locally through automated testing that images generated from the Plots for the doctest tests (and more) both via Plotly (m3d) and via SVG are unchanged.

PR #1562, opened 2025-12-26
Consolidate plot options processing into PlotOptions (rebase)

mmatera:
Just rebase https://github.com/Mathics3/mathics-core/pull/1558


bdlucas1:
Looks like somehow this rebase missed the change to core/symbols.py that adds the preserve_symbols option to to_python.

PR #1566, opened 2025-12-27
Pane

mmatera:
This is a basic implementation of `Pane` and `PaneBox`, to be used in the rework of MakeBoxes. Right now, `PaneBox` has a visible effect just in LaTeX form.

### DIFF
--- a/mathics/builtin/box/layout.py
+++ b/mathics/builtin/box/layout.py
@@ -49,6 +49,8 @@ class BoxData(Builtin):
 
 class ButtonBox(BoxExpression):
     """
+
+    <url>:WMA link:https://reference.wolfram.com/language/ref/ButtonBox.html</url>
     <dl>
       <dt>'ButtonBox'[$boxes$]
       <dd> is a low-level box undocumented construct that represents a button \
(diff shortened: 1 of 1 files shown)


mathics/builtin/layout.py
```diff
@@ -211,6 +210,50 @@ class NonAssociative(Builtin):
     summary_text = "non-associative operator"
 
 
+class Pane(Builtin):
+    """
+    <url>:WMA link:https://reference.wolfram.com/language/ref/Pane.html</url>
+
+    <dl>
+      <dt>'Pane[$expr$, $width$]'
```
rocky: As best I can tell, we also support `Pane[$expr$]`. Right? If this is correct, I guess we should mention it here.

rocky: It looks like we also support `Pane[$expr$ {$width$, $height$}]`.

rocky:
First impressions: overall positive. I know this is going to require more PR iterations, and probably some discussions. Later today, I put in a review PR for just the documenation part.

mathics/builtin/box/layout.py
```diff
@@ -20,6 +20,7 @@
 from mathics.core.list import ListExpression
 from mathics.core.symbols import Symbol
 from mathics.core.systemsymbols import (
+    SymbolAutomatic,
```
rocky: This import isn't used. Will remove in upcoming PR.

rocky:
LGTM

PR #1567, opened 2025-12-27
Fix complex cast warnings

mmatera:
Just some small adjustments to avoid these warnings.


bdlucas1:
@mmatera thanks for looking at this. I had just noticed those the other day (are they new or was I blind?) and was wanting to look into them. I think the fixes are probably right but I have a small doubt I'd like to confirm. If you don't mind holding off a bit I'll take a look later today or tomorrow.

rocky:
LGTM. Had noticed as well.

PR #1569, opened 2025-12-27
Plot image tests workflow

mmatera:
This should do the work

### DIFF
--- a/test/builtin/drawing/test_plot_detail.py
+++ b/test/builtin/drawing/test_plot_detail.py
@@ -369,8 +369,7 @@ if __name__ == "__main__":
     parser.add_argument("--update", action="store_true", help="update reference files")
     args = parser.parse_args()
     UPDATE_MODE = args.update
-    if not os.environ.get("MATHICS_PLOT_DETAILED_TESTS", False) and not UPDATE_MODE:
-        exit(0)
+
     try:
         test_all()
     except AssertionError:


.github/workflows/plot-tests.yml
```diff
@@ -1,31 +1,41 @@
...
 
-  - name: Install OS dependencies
-    run: |
-      sudo apt-get update -qq
-      sudo apt-get install -qq liblapack-dev llvm-dev tesseract-ocr remake
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: '**'
```
bdlucas1, line 7: Sorry, not familiar with workflows - does this mean it runs on pull requests on any branch? Just checking that the intention is in general this should run on pull requests.

.github/workflows/plot-tests.yml
```diff
@@ -1,31 +1,41 @@
-  - name: Set up Python 3.11
-    uses: actions/setup-python@v5
-    with:
-      python-version: '3.11'
+name: Mathics3 (ubuntu)
```
bdlucas1: Is this the name that shows up on the front page of the PR where it shows the tests that are running? If so it might be nice to have a distinctive name so one can confirm at a glance that the plot tests have finished (successfully, or not)

test/builtin/drawing/test_plot_detail.py
```diff
@@ -363,8 +368,9 @@ def make_ref_files():
     parser = argparse.ArgumentParser(description="manage plot tests")
     parser.add_argument("--update", action="store_true", help="update reference files")
     args = parser.parse_args()
-    update_mode = args.update
-
+    UPDATE_MODE = args.update
+    if not os.environ.get("MATHICS_PLOT_DETAILED_TESTS", False) and not UPDATE_MODE:
```
bdlucas1: Why would this be conditional? If you arrive here you've run something like python -m test.builtin.drawing.test_plot_detail, and that can't have any meaning other than that you explicitly want to run those tests.

.github/workflows/plot-tests.yml
mmatera: I copy and paste from other workflow, and I forgot changing the name. Thanks for noticing it!

test/builtin/drawing/test_plot_detail.py
mmatera: I just followed the pattern we use in other tests. What it does is to prevent the tests from being run on the main test workflows.

test/builtin/drawing/test_plot_detail.py
```diff
@@ -147,7 +148,7 @@
 
 # determines action to take if actual and reference files differ:
 # either raise assertion error, or update reference file
-update_mode = False
+UPDATE_MODE = False
```
mmatera, line 151: Just to make the linter happy, global variables should be in uppercase.

Makefile
```diff
@@ -127,6 +128,9 @@ clean: clean-cython clean-cache
 	rm -f mathics/data/op-tables || true; \
 	rm -rf build || true
 
+plot-detailed-tests:
```
bdlucas1, line 131: I'm not sure how other developers work, but my current habit is to make pytest-x, doctest-x, and the Combinatoria thing before submitting a pull request, in order to run the most comprehensive set of tests that I know of. I think now I will need to add plot-detailed-tests to that, is that right?

I'm fine doing that, but I wonder if it would make sense to have a target that runs those four (and maybe others I don't know about?) if they want to run a comprehensive set of tests, and to encourage then to do so. My personal preference is to be as comprehensive as is reasonable before submitting a PR.

.github/workflows/plot-tests.yml
mmatera: The workflow is executed on each PR.

test/builtin/drawing/test_plot_detail.py
bdlucas1: This particular entry point is not on a main test workflow though - it's a CLI specific to this file. The main test workflows are via the test_all function itself, which you've marked with skipif make this test conditional. Or am I missing something fundamental?

test/builtin/drawing/test_plot_detail.py
bdlucas1: Makes sense. Thanks, for the explanation,  I did wonder a bit.

Makefile
bdlucas1: Above comment isn't a big deal to me though, just a thought for your consideration.

mmatera: In general, I run `make check`, which runs all the tests. I can include `plot-detaild-tests` there too.  What I did was to put `plot-detailed-tests` in `make check-x`, so you can run all the tests sequentially or in parallel (`make -j 4 check-x`).

test/builtin/drawing/test_plot_detail.py
mmatera: In the script, there are more global variables in lowercase, but I would like to correct them in another round.
Also, I would like to use `pytest.mark.parametrize` to decompose  `test_all` into several smaller tests, to see the progress bar.

Makefile
bdlucas1: That sounds good, thanks.

bdlucas1:
lgtm, and thanks again for stepping in.

PR #1571, opened 2025-12-28
Fix Alphabet documentation entry

mmatera:
Just fix a missing escape character, which makes the LaTeX documentation compile.


rocky:
LGTM - thanks.

PR #1565, opened 2025-12-27
Plotting SVG tests

bdlucas1:
These tests generate SVG files from from existing tests described in the `*.yaml` files, output them in distilled, limited-precision form, and then diff them against reference files.

(You will also see some activity on this branch aimed at similarly generating PNG files that has been abandoned. The sticking point there is that font rendering seems to be different on different platforms. Possibly those tests can be revived by either solving that problem or maybe allowing a tolerance, but I think the SVG tests may be a better option anyway.)

A README file has been added explaining how the tests work and also giving some options for running the tests during development that I found useful.

A bug in boxes_to_svg was fixed.

PR #1574, opened 2025-12-28
Vectorized 2d plot for Plot, LogPlot, PolarPlot, ParametricPlot

bdlucas1:
This PR implements vectorized plotting for subclasses of _Plot:  Plot, LogPlot, PolarPlot, ParametricPlot, when enabled. In many cases current performance of 2d plot functions is ok, but for some use cases this PR provides a noticeable improvement. For example in the attached image of the Riemann zeta function showing zeros on the critical line

* classic plot took 250 ms to plot 57 points, a little sluggish for interactive use, and the low resolution doesn't quite resolve all the zeros - the cusps don't all quite reach the 0 line. A more complex plot with more curves would be slower still.
 
* vectorized plot took 3 ms to plot 1000 points. The higher resolution resolves the zeros a bit better.

image


bdlucas1:
Once #1565 settles I will add some test cases to this PR.

bdlucas1:
The test failures in Plot Detailed Tests after merging #1565 are due to some new tests that were introduced by that branch but whose reference files were computed before vectorized plotting functions were available, so the plots were computed by the non-vectorized plot functions in a few cases. This is the change we see comparing the new reference files to the old in GitHub desktop - the points formerly represented by Lists of Lists are now represented by NumericArrays:

image

But the diff also highlights that the PlotPoints parameter being passed to the Graphics isn't quite right:

image

It shouldn't be a string, it should be a Symbol. I will rework that plus add some new tests to this PR. (I think Automatic is correct though, but will think that through as well).

So I'm happy to see that the effort we've put into these tests is paying off.

rocky:
LGTM.


> image

One difference I see between this and WMA plots is that the labels here are tilted 90 degrees and go up the y axis whereas in WM the y axis lables go horizontal rather than vertical.  See for example: https://wolfram.com/xid/0obfm-gzj8f6

PR #1575, opened 2025-12-29
Linting with ruff

mmatera:
This PR has a bunch of small changes suggested by ruff. Additionally, it identified some tests that were inactive due to miscoding.   

Most of the changes are removing unused variables and unused imports, and change assigments from lambda functions by function definitions.


rocky:
LGTM - sure, let's do it! 

Looked over all of the changes, and this looks good. (And I've used Ruff for linting in other projects, so I can vouch for its usability.)

PR #1577, opened 2025-12-29
split 'test_all' into a parametrized test

mmatera:
This PR splits the large and silent `test_all` into a `pytest.mark.parametrized` test, which shows a progress bar. I would like to do the same with `test_yaml`.


test/builtin/drawing/test_plot_detail.py
```diff
@@ -369,7 +369,54 @@ def yaml_tests(fn, vec):
...
+@pytest.mark.skipif(
+    pyodide is not None,
+    reason="Does not work in Pyodide",
+)
+@pytest.mark.parametrize(("vec"), [vec for vec in (True, False)])
+def test_yaml(vec):
+    """
+    Do yaml tests
+    """
+    yaml_tests("doc_tests.yaml", vec=False)
```
bdlucas1: This misses the second .yaml file (and I expect there may be more - for example I had in mind to split out the nondoc-* tests from doctests.yaml into their own .yaml file).

bdlucas1: This illustrates a maintenance problem going forward having this path in addition to the do_test_all/CLI path

mmatera:
```suggestion
    yaml_tests("doc_tests.yaml", vec=vec)
```

mmatera: Thanks for catching the but. 
 I tried in this round to make the proposal as non-disruptive as possible. But the idea would be
* Put the list of parameters (tests) in a generator function.
* rewrite `do_tests_all` as a call to `test_one_test` inside a loop using the same generator.

mmatera: > This misses the second .yaml file (and I expect there may be more - for example I had in mind to split out the nondoc-* tests from doctests.yaml into their own .yaml file).
I would prefer to avoid adding large bunch of files, which must be updated each time there is a slight change in the implementation. Again, I think the right way to do this is to use a generator function that reads from a single yaml file.

bdlucas1: > * Put the list of parameters (tests) in a generator function.
> * rewrite `do_tests_all` as a call to `test_one_test` inside a loop using the same generator.

That sounds like a workable solution. I could take on doing that in a followup PR at my next good context switch point, after #1565 and #1574 have settled. 

>I would prefer to avoid adding large bunch of files, which must be updated each time there is a slight change in the implementation.

I agree, and I don't expect to create a large number of .yaml files. I think it will be helpful to have a small number of separate ones reflecting different gross characteristics e.g. doc/nondoc, vec=True/False. (Regarding the latter, I have in mind to add a DEFAULTS entry to the yaml file to set vec=True/False default for the whole file, rather than embedding that in the code).

> Again, I think the right way to do this is to use a generator function that reads from a single yaml file.

Agreed with the generator idea, but that can be easily applied to a small number of yaml files.

bdlucas1: Also, regarding the non-yaml tests, that I was thinking to replace that with a yaml file that does the same tests, with the options and other variations explicitly in the yaml file instead of implicit in the code. That code I think is a bit convoluted and hard to follow anyway, and doing this will simplify the code so that there's only one mechanism, the yaml files. What do you think?

mmatera: Fully agree: the best would be to have some generator functions that read the corresponding `yaml` files and feed the parametrized tests and `do_test_all`.

PR #1578, opened 2025-12-29
Plot image tests yaml

bdlucas1:
@mmatera since I'm waiting on Rocky for his further comments on the PR I went ahead and implemented the things we discussed in your PR (which I have already merged): with this PR all tests are now driven by yaml files, the yaml tests are fed one-by-one into pytest by a generator, and the CLI is driven by the same generator.

If you don't mind could you take a look before I merge this into #1565?

### DIFF
--- a/test/builtin/drawing/doc_tests.yaml
+++ b/test/builtin/drawing/doc_tests.yaml
@@ -1,10 +1,11 @@
 #
 # These tests were automatically picked up from doc 2025-12-13
 # The doc-xxx numbers are simply sequential numbers in alphebatized list as of that date
-# The doc-xxx labels correspond to test output so shouldn't be changed
-# New tests can just be added at end of list, disregarding alphabetical order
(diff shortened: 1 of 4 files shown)


bdlucas1:
@mmatera I think the test failures have nothing to do with this PR, do you agree? They seem be related to a time-constrained test that is aborting. I've seen that test failing frequently, but not always. I wonder if we need to skip it, or maybe give it more time.

mmatera:
> @mmatera I think the test failures have nothing to do with this PR, do you agree? They seem be related to a time-constrained test that is aborting. I've seen that test failing frequently, but not always. I wonder if we need to skip it, or maybe give it more time.

Indeed, it is a random related to our current implementation of `TimeConstrained` and related symbols.  Probably we should put these tests apart, but on the other hand, it is easy to recognize them and run it again.

mmatera:
Nice! LGTM

bdlucas1:
Thanks for suggesting this approach

PR #1580, opened 2025-12-30
Fix format numbers

mmatera:
This is another step in making the formatting more compatible with WMA. This time, in adjusting how numbers are formatted in different forms and representations. 

I tried to stick to the form in which WMA does it, with a detail: Except in `OutputForm`, PrecisionReal numbers in WMA are shown with all the decimal places included in the internal representation. For example,
```
In[1]:= N[Pi,4]//InputForm                                                      

Out[1]//InputForm= 3.1415926535897932385`4.
```
In Mathics, because of our internal representation, ```N[Pi,4]``` results in a number with exactly 5 decimal digits  (`3.1416`4`) 

Apart from this difference,  the references for the tests were collected from WMA.

In order to make this work, some bugs and misimplementations were fixed:
* `RowBox.elements` now stores a `ListExpression` instead of the `RobBox.items` directly.
* `String.make_boxes` uses the internal value `String._value` instead of the proxy `String.value`.
* A more sophisticated analysis is used to determine when parenthesize the index in a `SuperscriptBox` in its `text` representation.

### DIFF
--- a/test/builtin/test_makeboxes.py
+++ b/test/builtin/test_makeboxes.py
@@ -82,17 +82,17 @@ def test_makeboxes_real(str_expr, str_expected, msg):
 @pytest.mark.parametrize(
     ("str_expr", "str_expected", "msg"),
     [
-        (r"MakeBoxes[1.4`]", r"1.4`", "StandardForm always shows a prec mark."),
+        (r"MakeBoxes[1.4`]", r"1.4`", "StandardForm always shows a precision mark."),
         (r"MakeBoxes[OutputForm[1.4]]", r"1.4", "OutputForm, no mark for MachineReal"),
         (
             r"MakeBoxes[3.142`3]",
             r"3.142`3",
-            "StandadForm with Prec real shows all the stored digits, and prec",
+            "StandardForm with PrecisionReal shows all the stored digits, and precision",
         ),
         (
             r"MakeBoxes[OutputForm[3.142`3]]",
             r"3.14",
-            "OutputForm trims digits up to prec.",
+            "OutputForm trims digits up to precision.",
         ),
         (r"MakeBoxes[1.5`20]", r"1.5`20.", None),
         (r"MakeBoxes[1.4`20]", r"1.4`20.", None),


rocky:
Actually, this PR is in some ways is more helpful than #1581 because it is easier to see what output is changed. Changing _both_ the number format along with the testing mechanism obscures making it easier for someone to actually figure out what got improved.

mmatera:
OK, let`s reinstate this, and put #1581 as a continuation.

mathics/eval/makeboxes/numberform.py
```diff
@@ -251,20 +266,31 @@ def split_string(s, start, step):
...
-        )
-    else:
-        return method(String(s), String(base), String(pexp), options)
+    boxed_s = _boxed_string(s, number_as_text=SymbolFalse)
+    if pexp == "":
+        return boxed_s
+
+    boxed_base = _boxed_string(base, number_as_text=SymbolFalse)
+    boxed_pexp = _boxed_string(pexp, number_as_text=SymbolFalse)
+    print("parts:", [boxed_s, boxed_base, boxed_pexp], method)
```
mmatera:
```suggestion

```

mmatera:
Mental note: add a warning in the precommit script when a `print` statement is added in a PR...

test/builtin/test_makeboxes.py
```diff
@@ -79,8 +82,18 @@ def test_makeboxes_real(str_expr, str_expected, msg):
 @pytest.mark.parametrize(
     ("str_expr", "str_expected", "msg"),
     [
-        (r"MakeBoxes[1.4]", r"1.4`", None),
-        (r"MakeBoxes[1.4`]", r"1.4`", None),
+        (r"MakeBoxes[1.4`]", r"1.4`", "StandardForm always shows a prec mark."),
```
rocky: See comments in #1581 about using "prec" versus being more explicit with the word "precedence".

But since I put in a change that, I suppose it is simpler and less work to deal with that in #1581

mmatera: I change all the occurences on this file.

rocky:
LGTM - thanks for the attention to detail and follow-up on the suggestions.

PR #1582, opened 2025-12-30
Wma format comparison

mmatera:
This PR adds a WL script to do the same tests that `test_format.py` does. WMA 12 does not support `YAML` files, so I had to translate the reference file to json. The output is also included.

To run the tests in WMA, we can open `test.m` in the Notebook interface, or run in the command line

```
wolframscript -f test.m
```
or
```
cat test.m | math
```


mmatera:
It is a pity that mathics cannot read the `json` file well, so we could compare the outputs in both CLI.

mmatera:
I noticed some small inconsistencies between the CLI and the notebook interface. For example:

image

image

We see that the precision mark is shown always in CLI, but not in the Notebook interface.

rocky:
> I noticed some small inconsistencies between the CLI and the notebook interface. For example:
> 
> image image
> We see that the precision mark is shown always in CLI, but not in the Notebook interface.

In the back of my head, I've always had the thought that this inconsistency might be due to the same person who had inconsistencies in this code base.

It's probably not worth the effort to mimic both behaviors based on a setting.

rocky:
LGTM

mmatera:
> > I noticed some small inconsistencies between the CLI and the notebook interface. For example:
> > image image
> > We see that the precision mark is shown always in CLI, but not in the Notebook interface.
> 
> In the back of my head, I've always had the thought that this inconsistency might be due to the same person who had inconsistencies in this code base.
> 
> It's probably not worth the effort to mimic both behaviors based on a setting.

That' s the point. The idea is to be consistent enough to be compatible with the libraries that expects a certain format of the Boxed expressions. There are also some aspects that we can probably do better, like pretty print, or LaTeX processing.

PR #1583, opened 2025-12-30
use JSON library to import JSON files

mmatera:
This improves the support for reading JSON files needed in #1582.

### DIFF
--- a/mathics/builtin/fileformats/jsonformat.py
+++ b/mathics/builtin/fileformats/jsonformat.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+
+"""
+JSON
+
+Basic implementation for an JSON importer.
+"""
+
+import json
+
+from mathics.core.atoms import String
+from mathics.core.builtin import Builtin
+from mathics.core.convert.python import from_python
+from mathics.core.expression import Evaluation
+
+
+class JSONImport(Builtin):
+    """
+    ## <url>:native internal:</url>
+
+    <dl>
+      <dt>'JSON`Import`JSONImport["file"]'
+      <dd>parses "string" as a JSON file, and returns the data as a nested
+          list of rules.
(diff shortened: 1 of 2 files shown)


mathics/builtin/fileformats/jsonformat.py
```diff
@@ -0,0 +1,42 @@
...
+          list of rules.
+    </dl>
+
+    """
+
+    summary_text = "import elements from json"
+    context = "JSON`Import`"
+    messages = {"dec": "Decoding Error at `1`"}
+
+    def eval(self, filename, evaluation: Evaluation):
```
rocky, line 33: This (and the other eval routines) should be moved under `mathics.eval`

mmatera: Eventually I will do that, but this is just to help #1582

rocky: Ok. You are on the honor system to make good on this. I am concerned that a number of the things that get put off don't get done. 

LGTM.

mmatera: In concrete, the former implementation of the JSON imported does not work. I want to be able to import JSON to share lists of tests between WMA and the pytest system of Mathics, to narrow the differences in formatting, and maybe later, on other aspects. 
In particular, the JSON implementation I wrote is quite small. A refactor of the importers would requiere to review the code of the HTML and XML importers, which are much more complex. In order to finish something, I need to focus on one problem...

bdlucas1: > Ok. You are on the honor system to make good on this. I am concerned that a number of the things that get put off don't get done.

TODO/FIXME comments, Issues, Kanban boards, PMs, stand-ups, all-hands, ..., tell me when you want me to stop ...

PR #1532, opened 2025-11-26
ComplexPlot, ComplexPlot3D

bdlucas1:
image
image

PR #1584, opened 2025-12-31
More vectorized plotting tests

bdlucas1:
No code changes, just tests


rocky:
LGTM. One thing I noticed though is that ContourPlot documentation doesn't have a single example:

image


Maybe one of the nicer test examples can be added to the user-facing documentation? (Even better would be to add some helpful text around the example.

bdlucas1:
> ContourPlot documentation doesn't have a single example:

@mmatera in #1532 you had expressed interest in doing that for ComplexPlot and ComplexPlot3D - does that apply to ContourPlot also?

In any case I think that's an effort for a separate PR. I opened #1586 so it won't get forgotten.

rocky:
> > ContourPlot documentation doesn't have a single example:
> 
> @mmatera in #1532 you had expressed interest in doing that for ComplexPlot and ComplexPlot3D - does that apply to ContourPlot also?

Yes, thanks.

> 
> In any case I think that's an effort for a separate PR. I opened #1586 so it won't get forgotten.

Thanks.

PR #1585, opened 2025-12-31
Two options for suppressing numpy warnings

bdlucas1:
You may have noticed warnings like the following being produced during the vectorized plotting tests:

```
test/builtin/drawing/test_plot_detail.py::test_yaml[parms246]
  <lambdifygenerated-8>:2: RuntimeWarning: invalid value encountered in sqrt

test/builtin/drawing/test_plot_detail.py::test_yaml[parms253]
  <lambdifygenerated-6>:2: RuntimeWarning: divide by zero encountered in divide

test/builtin/drawing/test_plot_detail.py::test_yaml[parms257]
  <lambdifygenerated-8>:2: RuntimeWarning: divide by zero encountered in log
```

These come from numpy, and are intended to help developers directly working with numpy to do numerical computation that there might be a problem in their code. Our situation is a little different - end users are not directly interacting with numpy. Such errors might for example occur when a user does `Plot[1 / x, {x, 0, 1}]`. In this case floating point infinity is a completely reasonable value and is handled appropriately during rendering.

Three possible solutions:

- Capture the warnings and turn them into session messages. I mention this for completeness, but I don't think it is the right solution - as mentioned these are reasonable and useful values. As far as I can tell WMA doesn't produce such messages in plotting functions.
- Globally suppress numpy warnings. While I know for certain that disabling during plotting is reasonable, I would posit that in any context where numpy is being used infinity, nan, etc. are reasonable and useful values, but I'm no certain of that.
- More conservatively, suppress the warnings only in vectorized plotting code.

The PR contains code for both of the last two solutions for concreteness, but only one of them is needed.

### DIFF
--- a/mathics/builtin/numpy_utils/with_numpy.py
+++ b/mathics/builtin/numpy_utils/with_numpy.py
@@ -14,12 +14,6 @@ import numpy
 from mathics.core.list import ListExpression
 
 #
-# Suppress numpy warnings for things like divide by zero
-#
-
-numpy.seterr(all="ignore")
-
-#
 # INTERNAL FUNCTIONS
 #
 


mathics/builtin/numpy_utils/with_numpy.py
```diff
@@ -13,6 +13,12 @@
 
 from mathics.core.list import ListExpression
 
+#
+# Suppress numpy warnings for things like divide by zero
+#
+
+numpy.seterr(all="ignore")
```
mmatera: What about `if not DEBUG:...` ?

rocky:
> You may have noticed warnings like the following being produced during the vectorized plotting tests:
> 
> ```
> test/builtin/drawing/test_plot_detail.py::test_yaml[parms246]
>   <lambdifygenerated-8>:2: RuntimeWarning: invalid value encountered in sqrt
> 
> test/builtin/drawing/test_plot_detail.py::test_yaml[parms253]
>   <lambdifygenerated-6>:2: RuntimeWarning: divide by zero encountered in divide
> 
> test/builtin/drawing/test_plot_detail.py::test_yaml[parms257]
>   <lambdifygenerated-8>:2: RuntimeWarning: divide by zero encountered in log
> ```
> 

Actually, I hadn't noticed. Thanks for pointing this out, paying _attention_ to it, and coming up with remedies. Attention to detail is greatly appreciated!

> These come from numpy, and are intended to help developers directly working with numpy to do numerical computation that there might be a problem in their code. Our situation is a little different - end users are not directly interacting with numpy. Such errors might for example occur when a user does `Plot[1 / x, {x, 0, 1}]`. In this case floating point infinity is a completely reasonable value and is handled appropriately during rendering.
> 
> Three possible solutions:
> 
> * Capture the warnings and turn them into session messages. I mention this for completeness, but I don't think it is the right solution - as mentioned these are reasonable and useful values. As far as I can tell WMA doesn't produce such messages in plotting functions.

I agree, we shouldn't turn them into session messages for the reasons you so clearly indicate.

> * Globally suppress numpy warnings. While I know for certain that disabling during plotting is reasonable, I would posit that in any context where numpy is being used infinity, nan, etc. are reasonable and useful values, but I'm no certain of that.
> * More conservatively, suppress the warnings only in vectorized plotting code.
> 
> The PR contains code for both of the last two solutions for concreteness, but only one of them is needed.

Let's remove the global divide by zero suppression, but keep the code to disable this in vectorized plotting.

Thanks.

mathics/builtin/numpy_utils/with_numpy.py
```diff
@@ -13,12 +13,6 @@
 
 from mathics.core.list import ListExpression
 
-#
-# Suppress numpy warnings for things like divide by zero
-#
-
-numpy.seterr(all="ignore")
```
rocky, line 20: Thanks.

rocky: LGTM

mmatera:
LGTM

PR #1590, opened 2026-01-01
Implement ParametricPlot3D (vectorized)

bdlucas1:
vec-parametricplot3d-torus-vec
vec-parametricplot3d-rings-vec
vec-parametricplot3d-trefoil-vec
vec-parametricplot3d-multi-vec

### DIFF
--- a/mathics/builtin/drawing/plot.py
+++ b/mathics/builtin/drawing/plot.py
@@ -481,7 +481,9 @@ class PlotOptions:
         self.exclusions = exclusions
 
         # Mesh option (returns Symbol)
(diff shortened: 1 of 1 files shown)


bdlucas1:
vec-parametricplot-moebius-vec

rocky:
@bdlucas1 At this point, when you add new builtins, you should be doing this properly.  That means adding an example and looking to see if Wikipedia has a name for the built-in. If so, then add a link to the builtin. 

The first time or so, I suppose ignorance and not reading and following docs is understandable. But at this point, please do better.

Thanks.

Issue #1589, opened 2025-12-31
Support GraphicsComplex and NumericArray for SVG generation


bdlucas1:
Currently the SVG generation routines don't understand how to handle GraphicsComplex and NumericArray used in Graphics* by the new vectorized plotting routines for performance. Implementing this would benefit front-ends that don't want to do their own rendering, and would allow examples to be included in the documentation.

@rocky I think you said at one point you would be interested in doing this?

rocky:
> Currently the SVG generation routines don't understand how to handle GraphicsComplex and NumericArray used in Graphics* by the new vectorized plotting routines for performance. Implementing this would benefit front-ends that don't want to do their own rendering, and would allow examples to be included in the documentation.
> 
> [@rocky](https://github.com/rocky) I think you said at one point you would be interested in doing this?

I'm happy to let others do this. Also, I am probably not going to get to this for a while. 

My current thought was to use [drawsvg2](https://pypi.org/project/drawsvg/).  And do this as an optional Python Module outside of Mathics-core. It gets installed though if one runs `pip install -e .[full]`

Here are the benefits of doing this: 

1. We force a rendering API which is desirable
2. We can remove the SVG rendering from mathics core. 
3. We provide support for NumericArray and GraphicsComplex

Issue #1586, opened 2025-12-31
Review doc and add examples for new vectorized plotting functions


bdlucas1:
ContourPlot, ComplexPlot, ComplexPlot3D, ParametricPlot3D, SphericalPlot3D, ContourPlot3D

See test/builtin/drawing/vec_tests.yaml for possible examples.

bdlucas1:
I think this is dependent on #1589, unless it's ok to have examples in the doc with no visual output.

bdlucas1:
## ContourPlot

```
vec-contourplot-waves:
    expr: ContourPlot[x - y^3, {x, -2, 2}, {y, -1, 1}, AspectRatio->Automatic]
```

```
vec-contourplot-two-equations:
    expr: ContourPlot[{Re[Sin[x + I y]] == 5, Im[Sin[x + I y]] == 0}, {x, -10, 10}, {y, -10, 10}]
```

<img width="40%" alt="Image" src="https://github.com/user-attachments/assets/cb392ce5-9707-48b6-be73-30bd8004fe58" />
<img width="40%" alt="Image" src="https://github.com/user-attachments/assets/33383c04-3459-4410-8e39-a3bcc9a0f3de" />


## ComplexPlot

```
vec-complexplot-poles-zeros:
    expr: ComplexPlot[(z^2 + 1)/(z^2 - 1), {z, -2 - 2 I, 2 + 2 I}]
```

<img width="40%" alt="Image" src="https://github.com/user-attachments/assets/144a2eb5-a55c-4d26-9fd7-a43a66434fcc" />

The color pinwheels show zeros and poles of the function, with the pinwheels going opposite directions around poles compared to zeros.

## ComplexPlot3D

```
vec-complexplot3d-poles-zeros:
    expr: ComplexPlot3D[(z^2 + 1)/(z^2 - 1), {z, -2 - 2 I, 2 + 2 I}, PlotRange-> {0,4}]
```

<img width="40%" alt="Image" src="https://github.com/user-attachments/assets/e5f450c3-452a-4741-b536-8f9636ed1d27" />

Color pinwheels show poles and zeros as for ComplexPlot, with the addition of showing the magnitude in the third axis.

## ParametricPlot3D

```
vec-parametricplot3d-trefoil:
    expr: ParametricPlot3D[{Sin[t] + 2 Sin[2 t], Cos[t] - 2 Cos[2 t], -Sin[3 t]}, {t, 0, 2 Pi}]

[...]

PR #1592, opened 2026-01-03
Implement ContourPlot3D

bdlucas1:
vec-contourplot3d-sphere-vec

vec-contourplot3d-hyperboloids-vec

vec-contourplot3d-sphere-multi-vec

vec-contourplot3d-mesh-vec


rocky (changes requested):
Please follow the existing documenting conventions, where at least one user-facing example is given. Examples with commentary would be nice. And also add a link to wikipedia for the concept at the beginning when that is applicable as it often is here.   Thanks.

mathics/builtin/drawing/plot_plot3d.py
```diff
@@ -237,6 +239,40 @@ class ContourPlot(_Plot3D):
     graphics_class = Graphics
 
 
+class ContourPlot3D(_Plot3D):
+    """
+    <url>:WMA link: https://reference.wolfram.com/language/ref/ContourPlot3D.html</url>
+    <url>:Isosurface: https://en.wikipedia.org/wiki/Isosurface</usr>
```
rocky:
```suggestion
    <url>:Isosurface: https://en.wikipedia.org/wiki/Isosurface</url> (
    <url>:WMA link: https://reference.wolfram.com/language/ref/ContourPlot3D.html</url>)
```

We put the Wiki link first. Also that's `</url>` not `</usr>`.

mathics/builtin/drawing/plot_plot3d.py
```diff
@@ -237,6 +239,40 @@ class ContourPlot(_Plot3D):
...
+    <dl>
+      <dt>'ContourPlot3D'[$f(x,y,z)$, {$x$, $x_{min}$, $x_{max}$}, {$y$, $y_{min}$, $y_{max}$, {$y$, $y_{min}$, $y_{max}$}]
+      <dd>creates a three-dimensional contour plot of $f(x,y,z)$ over the specified region on $x$, $y$, and $z$.
+
+      >> ContourPlot3D[x ^ 2 + y ^ 2 - z ^ 2, {x, -1, 1}, {y, -1, 1}, {z, -1, 1}]
```
rocky: Examples don't belong inside `<dl>` lists. Look again at the docs to see how this is typically done.

If it is not fixed in a day or so, I'll put in a PR to fix this.

bdlucas1: Fixed

rocky (approved):
Ideally, I'd like to understand the code in more detail. Don't have time right now, though. Possibly some other time. 

Overall, this is cool stuff.  I am concerned about (inadvertently) making other existing stuff non-functional.

mathics/eval/drawing/plot3d_vectorized.py
```diff
@@ -222,6 +223,33 @@ def emit(graphics, i, xyzs, _, quads):
     return make_surfaces(plot_options, evaluation, dim=2, is_complex=False, emit=emit)
 
 
+def equations_to_contours(plot_options):
+    """
+    For contour plots, convert functions of the form f==g to f-g,
+    and adjust contours ot [0] and disable background
```
rocky: ot? is to? Or something else?

bdlucas1:
>  I am concerned about (inadvertently) making other existing stuff non-functional.

I am too; that's why I put a lot of effort into putting tests in place for the existing plotting function to reduce that risk. What I think those tests tell us is that every example in the doc still generates the same output, both Graphics* and SVG. And in addition to testing at that level, I have tests in m3d where I generate PNGs from the SVGs, and from the Graphics* (via m3d rendering)  and compare them automatically for changes and manually for mismatches.

mathics/eval/drawing/plot3d_vectorized.py
bdlucas1: Fixed

rocky:
> > I am concerned about (inadvertently) making other existing stuff non-functional.
> 
> I am too; that's why I put a lot of effort into putting tests in place for the existing plotting function to reduce that risk. What I think those tests tell us is that every example in the doc still generates the same output, both Graphics* and SVG. And in addition to testing at that level, I have tests in m3d where I generate PNGs from the SVGs, and from the Graphics* (via m3d rendering) and compare them automatically for changes and manually for mismatches.

The expanded testing is appreciated, although it probably will take me a little while to learn and get used to. 

It's not the stuff that we currently test that is a worry, it is the stuff that we _don't_ test. Like documentation. I think @mmatera has some tests here, but stuff like adding a like to a relevant wiki page and making regular the formatting is not something that is easily done.

bdlucas1:
> it is the stuff that we _don't_ test. Like documentation. I think @mmatera has some tests here, but stuff like adding a like to a relevant wiki page and making regular the formatting is not something that is easily done.

That is why I have been hesitant to touch the docstrings for builtins etc., and was happy with the offer of help from @mmatera on the first PR I submitted without examples to add those later. I thought process we were on was to accumulate them on #1586 to be done later in a batch (whether by me at the end of the day, or by someone more experienced with the arcane system).   From my perspective it would have been far more efficient, and would likely have produced a better result, to have done all those in a batch in a single PR. Spending the morning struggling with an arcane markup in three inter-dependent PRs and dealing with things like Windows charsets was not how I had envisioned spending my Saturday morning.

mmatera:
> > it is the stuff that we _don't_ test. Like documentation. I think @mmatera has some tests here, but stuff like adding a like to a relevant wiki page and making regular the formatting is not something that is easily done.
> 
> That is why I have been hesitant to touch the docstrings for builtins etc., and was happy with the offer of help from @mmatera on the first PR I submitted without examples to add those later. I thought process we were on was to accumulate them on #1586 to be done later in a batch (whether by me at the end of the day, or by someone more experienced with the arcane system). From my perspective it would have been far more efficient, and would likely have produced a better result, to have done all those in a batch in a single PR. Spending the morning struggling with an arcane markup in three inter-dependent PRs and dealing with things like Windows charsets was not how I had envisioned spending my Saturday morning.

@bdlucas1, now I am working on another part of the code. When I finish (at least a part) of what I am doing, the next is to go over the new symbols to add the missing documentation.

mmatera:
For my part, I am happy if @bdlucas1 completes the implementation of these graphics routines and tests. Adding (basic) docs is for me something that is not too hard to do before the release.

rocky:
> > it is the stuff that we _don't_ test. Like documentation. I think @mmatera has some tests here, but stuff like adding a like to a relevant wiki page and making regular the formatting is not something that is easily done.
> 
> That is why I have been hesitant to touch the docstrings for builtins etc., and was happy with the offer of help from @mmatera on the first PR I submitted without examples to add those later. I thought process we were on was to accumulate them on #1586 to be done later in a batch (whether by me at the end of the day, or by someone more experienced with the arcane system).

If you want to add builtins, you should be taking more responsibility for understanding how builtins are written and following the conventions we use. 

The first or first couple of builtins, it is understandable that something might slip through. 


> From my perspective it would have been far more efficient, and would likely have produced a better result, to have done all those in a batch in a single PR. Spending the morning struggling with an arcane markup in three inter-dependent PRs and dealing with things like Windows charsets was not how I had envisioned spending my Saturday morning.

From my perspective, things like this slip until it becomes a mess. Right before release, there is this huge chore of trying to get things regularized and sometimes unbroken. 

If I recall correctly, the more important part was to get a rendering API and architecture in place.

bdlucas1:
> @bdlucas1, now I am working on another part of the code. When I finish (at least a part) of what I am doing, the next is to go over the new symbols to add the missing documentation.

Thanks @mmatera.  I've added some suggested examples to #1586 in case it helps. Let me know if there's anything else I can do to help.

mmatera:
I started with the documentation. The problem I find is that symbols like `ComplexPlot` are loaded, but cannot be evaluated if the variable `use_vectorized_plot` is set to `False`. Shall we set this variable to `True` by default? Or load the symbols just if it the variable is set to `True`?

rocky:
> I started with the documentation. The problem I find is that symbols like `ComplexPlot` are loaded, but cannot be evaluated if the variable `use_vectorized_plot` is set to `False`. Shall we set this variable to `True` by default? Or load the symbols just if it the variable is set to `True`?

This was exactly the kind of thing I wanted to detect and avoid. 

From the project point of view, it is better to figure this out earlier than later.

bdlucas1:
> The problem I find is that symbols like ComplexPlot are loaded, but cannot be evaluated if the variable use_vectorized_plot is set to False. Shall we set this variable to True by default?

The status of these feature is that they generate Graphics* output that contains GraphicsComplex and NumericArray, which will not be understood by most current consumers (aside from the experimental front-end m3d, which I use as a proof-of-concept and test vehicle). That was the reason for my reference in #1586 to a dependency on #1589, which is one way of enabling current consumers to be able to consume the output of the vectorized plotting functions. 

> Or load the symbols just if the variable is set to True?

The intention of the variable is that it be dynamic, e.g. so we can run both vectorized and non-vectorized plotting tests. I haven't looked, but I suspect that's the reason for the test failures in your PR. Having this be dynamic has also been a great help in development work (my initial implementation was static and it proved to be awkward).

In the examples that I added this morning, I just showed the unevaluated expression as the output, which is what the doctests want to see. This of course isn't helpful to a user.

I'm thinking that given the incompleteness (from an end-to-end system) point of view of the vectorized plotting features, it might be reasonable to leave them undocumented until we have a more complete picture. Or in the alternative document them but clearly mark them as "preview" or "incomplete" or "coming soon".

rocky:
> I'm thinking that given the incompleteness (from an end-to-end system)

The incompleteness was recently created. It's going to have to be fixed before we can release. And it would be better to fix it now. 

If you don't want to fix the incompleteness caused by these recent features, that's okay. I'll do it when I can. 

But then hold off on doing more, please.

rocky:
>  I'm thinking that given the incompleteness (from an end-to-end system)

One other thing regarding incompleteness. The reality, right now, is that we probably have to provide fallback ability to plot for plots where there are scoped variables and procedural-type statements (until we get compilation under control, and that's a big topic).

PR #1595, opened 2026-01-03
Tweak section header title

### LINKED ISSUE #1587: Incorrect description in plot_plot3d.py
The user facing description in plot_plot3d.py isn't quite right:

> Plots with Points Having 3 Items.

"Items" is a bit vague but I can't think of an interpretation under which it is completely correct:
 
* ContourPlot of equations of the form f(x,y)==g(x,y) have 2 items
* If you consider each complex number to be 2 items then
    * ComplexPlot3D has 4 items (one input complex number, one output complex number displayed as magnitude and phase)
* If you consider each complex number to be single item then
    * ComplexPlot3D has 2 items
    * ComplexPlot has 1.5 items (one input complex number, one output scalar which is the phase of the output complex number)

I can't think of a good user-facing summary for that module. That module is useful from a code organization perspective (it groups classes that share the _Plot3D base class) but not from a user-facing documentation perspective. I think the underlying problem is that a single mechanism is used for both organizing code and for organizing the document structure. That is reasonable when docstrings are used to document the code itself, but maybe not so much when the docstrings are documenting something else.

rocky:
Tweak a section header title, and add a missing space in the doc.  Fixes #1587

### DIFF
--- a/mathics/builtin/drawing/plot_plot3d.py
+++ b/mathics/builtin/drawing/plot_plot3d.py
@@ -1,13 +1,13 @@
 """
 Custom Plots
 
-A Density plot is kind of <url>:heat map:https://en.wikipedia.org/wiki/Heat_map</url> that represents magnitude or individual values as a color.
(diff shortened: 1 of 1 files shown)


mmatera (approved):
LGTM

PR #1596, opened 2026-01-03
Improve `FullForm` compatibility

mmatera:
Another small step that requires changing many lines. With this change, all the tests for `FullForm` in `makeboxes_tests-WMA.yaml` pass in Mathics3.
Changes are:

* FullForm does not add spaces after the separator ",".  Instead, the space is added in the functions that convert RowBoxes into text and latex strings.
* Precision in `PrecisionReal` numbers have a decimal point.
* trailing zeros in the mantisa of a number in `StandardForm`, `TraditionalForm`, `InputForm` and `FullForm` are removed.
* Negative numbers in `StandardForm`, `TraditionalForm` and  `FullForm` split as `RowBox[{"-",absolutevalue}]`
* adjust the previos tests.


mathics/builtin/box/layout.py
```diff
@@ -100,7 +100,17 @@ def init(self, num, den, **options):
         self.box_options = options
 
     def to_expression(self):
-        return Expression(SymbolFractionBox, self.num, self.den)
+        elems = (
+            self.num,
+            self.den,
+        )
+        return Expression(
```
rocky: This idiom: 

``` 
   return Expression(<symbol> *(...)) 
```

appears a number of places. It should be a function with a docstring...

which brings up the other question. Do you know what's up with turning a BoxExpression into an Expression when inside some other BoxExpression? 

This is the thing that would be explained inside the docstring of the helper function.

mathics/builtin/numbers/exp.py
```diff
@@ -172,8 +172,8 @@ class Exp(MPMathFunction):
      = E
     >> Exp[10.0]
      = 22026.5
-    >> Exp[x] //FullForm
-     = Power[E, x]
+    >> Exp[x] // FullForm
+     = Power[E,x]
```
rocky: I don't understand. Please educate me.

The space in `// FullForm` is not necessary, as far as I can tell. So what was it added? 

Also in wolframscript I do see a space after the `,`, e.g. `Power[E, x]`.

mathics/builtin/box/layout.py
rocky: Oh. I think I get a little bit of this. `to_expression()` purpose is deliberately to turn a BoxExpression back into an Expression.  Okay. That kind of thing would go into the function name and docstring. `boxexpression_to_expression()`.

mathics/builtin/compress.py
```diff
@@ -23,7 +23,7 @@ class Compress(Builtin):
     </dl>
 
     >> Compress[N[Pi, 10]]
-     = eJwz1jM0MTS1NDIzNTZLMDQAABh8Ay0=
+     = eJwz1jM0MTS1NDIzNTZLMDTQAwAb1wNb
```
rocky: Let's make this `...` since it keeps changing.

mathics/eval/makeboxes/numberform.py
```diff
@@ -147,11 +147,11 @@ def NumberForm_to_String(
         if digits_after_decimal_point is None:
             is_int = True
     elif isinstance(value, Real):
-        prec = dps(value.get_precision())
+        precision = dps(value.get_precision())
```
rocky, line 150: `prec` -> `precision` is geatly appreciated - Thanks!

test/format/format_tests.yaml
```diff
@@ -103,28 +103,28 @@
...
+    System`TraditionalForm: "-1.`\xD710^6"
 '-4':
   latex:
     System`InputForm: '-4'
     System`OutputForm: '-4'
     System`StandardForm: '-4'
     System`TraditionalForm: '-4'
   mathml:
-    System`InputForm: <mtext>-4</mtext>
+    System`InputForm: <mn>-4</mn>
```
rocky, line 124: Good to see here and other places that these are now getting expressed as  `<mn>` insteads of `<mtext>`.

mathics/builtin/numbers/exp.py
mmatera: This is just to make the style of the tests more uniform, and then make easier to find the pattern `// FullForm`.

mathics/builtin/compress.py
mmatera: OK

mathics/builtin/numbers/exp.py
rocky: Ok. Thanks explanation on what's up with the space before `FullForm`. But what about the space in the output that mismatches WolframScript?

mmatera: Good catch! The space had to be added back in the functions that convert rowboxes to strings. Now it is done.

mathics/builtin/box/expression.py
```diff
@@ -208,7 +210,7 @@ def tex_block(self, tex, only_subsup=False):
     def to_expression(self) -> Expression:
         # FIXME: All classes should store their symbol name.
```
rocky:
```suggestion
```

Based on the code, this has already been done, right? So this comment is no longer relevant.

mathics/builtin/box/expression.py
```diff
@@ -208,7 +210,7 @@ def tex_block(self, tex, only_subsup=False):
...
         # So there should be a self.head.
```
rocky:
```suggestion
```

Part of the comment above. See above.

rocky:
LGTM - see small removal of what I guess is an obsolete comment.  Somehow, I think in a later round, more stuff can be DRY'd or revised.

PR #1598, opened 2026-01-04
remove default options in boxes

mmatera:
This PR slightly modifies `options_to_rules` in a way that allows to remove trivial options (i.e., options that are set to their default value). Here I do it for Boxes, but probably we want to do the same with other symbols that have a lot of options, like  `Integrate` or `Solve`.

### DIFF
--- a/mathics/builtin/options.py
+++ b/mathics/builtin/options.py
@@ -10,6 +10,7 @@ the default behavior that function. Default options can be queried or set.
 :WMA link:
 https://reference.wolfram.com/language/guide/OptionsManagement.html</url>
 """
+from typing import Callable, Optional
 
 from mathics.builtin.image.base import Image
 from mathics.core.atoms import Integer1, String
@@ -17,6 +18,7 @@ from mathics.core.builtin import Builtin, Predefined, Test, get_option
 from mathics.core.evaluation import Evaluation
 from mathics.core.expression import Expression
 from mathics.core.list import ListExpression
+from mathics.core.parser import parse_builtin_rule
 from mathics.core.symbols import Symbol, SymbolList, ensure_context, strip_context
 from mathics.core.systemsymbols import SymbolDefault, SymbolRule, SymbolRuleDelayed
 from mathics.eval.patterns import Matcher, get_default_value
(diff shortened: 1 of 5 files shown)


mathics/builtin/options.py
```diff
@@ -570,12 +572,44 @@ def eval(self, symbol: Symbol, options: Expression, evaluation: Evaluation):
...
 
 
-def options_to_rules(options, filter=None):
+def filter_non_default_values(builtin):
+    """
+    Return a filter function that removes those
+    options which have associated their default values.
+    """
+    builtin_options = builtin.options
+    builtin_options = {
```
mmatera, line 581: `filter_non_default_values` produce a function that stores the expressions (Mathics3 elements) representing the options. Maybe we should store this in the builtins (in the contribute phase) or directly to store the filter function.

mathics/builtin/options.py
```diff
@@ -570,12 +572,44 @@ def eval(self, symbol: Symbol, options: Expression, evaluation: Evaluation):
...
+        if name not in builtin_options:
+            return True
+        if value.sameQ(builtin_options[name]):
+            return False
+        return True
+
+    return filter
+
+
+def filter_from_iterable(elems):
```
mmatera, line 597: This function is used in Plot. However, I think `filter_non_default_values` would also be a better choice in that case.

rocky:
LGTM - not sure of the motivation. But okay.

mmatera:
The motivation is what I stated before: when an expression is reconstructed, and the expression involved options, we only want to include the options which give non-default information. In particular, when an expression is converted into boxes, it store the values of all the options, which are used then for modify the way the textual form is generated. But we can also want to convert the box expressions objects back to normal expressions. Then, the reconstructed expressions should contain just the non default options. That the reason I need this at this point.

rocky:
> The motivation is what I stated before: when an expression is reconstructed, and the expression involved options, we only want to include the options which give non-default information. In particular, when an expression is converted into boxes, it store the values of all the options, which are used then for modify the way the textual form is generated. But we can also want to convert the box expressions objects back to normal expressions. Then, the reconstructed expressions should contain just the non default options. That the reason I need this at this point.

Ok. Thanks for the (re)explanation.

PR #1599, opened 2026-01-06
Misc tweaks...

rocky:
Start giving meaning to some of the magic precedence values Start a form directory for handing OutputForm, InputForm, etc. (Rendering/Format, and Forms are different)

### DIFF
--- a/mathics/builtin/atomic/strings.py
+++ b/mathics/builtin/atomic/strings.py
@@ -897,7 +897,7 @@ class ToString(Builtin):
         return self.eval_form(value, SymbolOutputForm, evaluation, options)
 
     def eval_form(self, expr, form, evaluation: Evaluation, options: dict):
-        "ToString[expr_, form_, OptionsPattern[ToString]]"
+        "ToString[expr_, form_Symbol, OptionsPattern[ToString]]"
         encoding = options["System`CharacterEncoding"]
         return eval_ToString(expr, form, encoding.value, evaluation)
 
--- a/mathics/builtin/box/layout.py
+++ b/mathics/builtin/box/layout.py
@@ -14,20 +14,12 @@ from mathics.builtin.options import filter_non_default_values, options_to_rules
 from mathics.core.atoms import String
 from mathics.core.attributes import A_HOLD_ALL_COMPLETE, A_PROTECTED, A_READ_PROTECTED
 from mathics.core.builtin import Builtin
-from mathics.core.element import BaseElement, BoxElementMixin
+from mathics.core.element import BaseElement, BoxElementMixin, EvalMixin
 from mathics.core.evaluation import Evaluation
 from mathics.core.exceptions import BoxConstructError
 from mathics.core.expression import Expression
 from mathics.core.list import ListExpression
 from mathics.core.symbols import Symbol
-from mathics.core.systemsymbols import (
-    SymbolFractionBox,
-    SymbolRowBox,
-    SymbolSqrtBox,
-    SymbolSubscriptBox,
-    SymbolSubsuperscriptBox,
-    SymbolSuperscriptBox,
-)
 from mathics.eval.makeboxes import to_boxes
 
 # This tells documentation how to sort this module
@@ -105,8 +97,8 @@ class FractionBox(BoxExpression):
     def elements(self):
         if self._elements is None:
             self._elements = elements_to_expressions(
+                self,
                 (
-                    self,
                     self.num,
                     self.den,
                 ),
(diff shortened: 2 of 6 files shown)


mathics/eval/makeboxes/makeboxes.py
```diff
@@ -142,7 +142,7 @@ def eval_makeboxes_outputform(expr, evaluation, form, **kwargs):
     Build a 2D representation of the expression using only keyboard characters.
     """
     from mathics.builtin.box.layout import PaneBox
-    from mathics.format.outputform import expression_to_outputform_text
+    from mathics.form.outputform import expression_to_outputform_text
```
mmatera, line 145: Maybe the right place for `outputform.py` would be inside `eval/makeboxes`, like `inputform.py`. `mathics.format` contains modules to convert (Boxed) expressions into strings in different representations (plane text, latex, mathml). OutputForm (and InputForm) does something similar, but on (norma) expressions.

rocky: At some level, it does not matter that much other than "Form" handling and rendering are different concepts and levels; so `mathics.format` is definitely the wrong place.

Personally, I think of `mathics.eval` having code to implement the primitives of built-in functions. 

Of course, there are primitives for boxing. So, sure, this fits into this category. However...

Boxing is considered its own phase. And while it is a kind of evaluation too, it is not the kind of primitive used for addition and multiplication, which must be implemented in Python, SymPy, etc. 

In theory, Boxing could be written purely as Mathics3 or WMA code.  And there is enough of it that, in my opinion, we'd be better off just putting that in its own place. 

But again, other than it _not_ being part of `mathics.format`, I don't have a strong feeling like this. Aesthetically, I think it is just better put in `mathics.form`.

mmatera: > At some level, it does not matter that much other than "Form" handling and rendering are different concepts and levels; so `mathics.format` is definitely the wrong place.
> 
At this point, I agree.

> Personally, I think of `mathics.eval` having code to implement the primitives of built-in functions.
> 
> Of course, there are primitives for boxing. So, sure, this fits into this category. However...
> 
> Boxing is considered its own phase. And while it is a kind of evaluation too, it is not the kind of primitive used for addition and multiplication, which must be implemented in Python, SymPy, etc.
> 

OK, but we have a specific submodule for Box evaluations

> In theory, Boxing could be written purely as Mathics3 or WMA code. And there is enough of it that, in my opinion, we'd be better off just putting that in its own place.

If in principle that is true, it is also true that we could implement in WL most of the code inside "eval". However, would be unnecesarily complicated and very slow. 

> 
> But again, other than it _not_ being part of `mathics.format`, I don't have a strong feeling like this. Aesthetically, I think it is just better put in `mathics.form`.

I do not have strong feelings about this either, just an argument: from the point of view of code organization, it appears that we would have a lot of space for the different steps of formatting. On the other hand, the complexity inside would not be comparable with the one in `core`, `builtin` or `eval`.

rocky: > > At some level, it does not matter that much other than "Form" handling and rendering are different concepts and levels; so `mathics.format` is definitely the wrong place.
> 
> At this point, I agree.
> 
> > Personally, I think of `mathics.eval` having code to implement the primitives of built-in functions.
> > Of course, there are primitives for boxing. So, sure, this fits into this category. However...
> > Boxing is considered its own phase. And while it is a kind of evaluation too, it is not the kind of primitive used for addition and multiplication, which must be implemented in Python, SymPy, etc.
> 
> OK, but we have a specific submodule for Box evaluations
> 

Personally, I'd like the boxing and form stuff put somewhere else, since I think of eval as more of the kind of stuff that a compiler would work on.


> > In theory, Boxing could be written purely as Mathics3 or WMA code. And there is enough of it that, in my opinion, we'd be better off just putting that in its own place.
> 
> If in principle that is true, it is also true that we could implement in WL most of the code inside "eval". However, would be unnecesarily complicated and very slow.
> 

So, again this is the kind of thing, in my view of the code organization that should distinguish eval from other parts.  Stuff that really has to be done in Python (Sympy, numpy, etc.) is definitely in eval. Stuff that could easily be done in WMA, is somewhere else; especially if it is a well-defined "phase".  There may be gray areas in the middle. 

Analogy: think of "eval" as "physics". But there is "chemistry" and "organic chemistry" as more like "boxing" and "forms". While this _could_ be put under the category of "physics", typically we don't do it in that way

> > But again, other than it _not_ being part of `mathics.format`, I don't have a strong feeling like this. Aesthetically, I think it is just better put in `mathics.form`.
> 
> I do not have strong feelings about this either, just an argument: from the point of view of code organization, it appears that we would have a lot of space for the different steps of formatting. On the other hand, the complexity inside would not be comparable with the one in `core`, `builtin` or `eval`.

If you don't have a strong feeling, would you mind catering again to my idiosyncracies, and let's put this in either "mathics.form" or "mathics.box" or have two modules?

mmatera: @rocky, OK, let's do as you proposed here and see how this evolve.

PR #1601, opened 2026-01-07
InputForm

mmatera:
This reimplements InputForm

### DIFF
--- a/mathics/builtin/forms/output.py
+++ b/mathics/builtin/forms/output.py
@@ -130,7 +130,7 @@ class FullForm(FormBaseClass):
     """
 
     in_outputforms = True
-    in_printforms = True
+    in_printforms = False
     summary_text = "underlying M-Expression representation"
 
     def eval_makeboxes(self, expr, fmt, evaluation):
@@ -262,7 +262,7 @@ class _NumberForm(Builtin):
 
     default_ExponentFunction = None
     default_NumberFormat = None
-
+    in_outputforms = True
     messages = {
         "npad": "Value for option NumberPadding -> `1` should be a string or a pair of strings.",
         "dblk": "Value for option DigitBlock should be a positive integer, Infinity, or a pair of positive integers.",


rocky:
Thanks for doing this! I will try to look at in detail later today.

mmatera:
PRs #1602 - #1605 contains the part of this PR which are not directly related, and can be analyzed separately,

rocky:
> PRs #1602 - #1605 contains the part of this PR which are not directly related, and can be analyzed separately,

I appreciate your separating PRs into smaller separable ones. Thanks.

mathics/form/inputform.py
```diff
@@ -0,0 +1,455 @@
+"""
+This module builts the string associated to the InputForm.
+
+`InputForm` produces a textual output suitable for being parsed and directly
```
rocky: The comment here should probably go into the user manual (and then removed or condensed) here.

mathics/form/inputform.py
```diff
@@ -0,0 +1,455 @@
...
+        format_expr = format_expr.elements[0]
+
+    lookup_name: str = format_expr.get_head().get_lookup_name()
+
+    try:
+        result = EXPR_TO_INPUTFORM_TEXT_MAP[lookup_name](
+            format_expr, evaluation, **kwargs
+        )
+        return result
+    except _WrongFormattedExpression as e:
```
rocky: Here and in the following except clause, `e` is not used. Either use or remove.

mathics/form/inputform.py
rocky: Change that - the user doc for "InputForm"?

mmatera: Let's see first if this is clear enough. Then we can rephrase the part that makes sense at the user level in the docstring of `InputForm`.

rocky: I was in the process of trying to understand what's up here. I'll put in a PR to merge tomorrow.

There were other small things I saw along this vein.

rocky:
LGTM.   

So I guess we can move on to OutputForm.

PR #1602, opened 2026-01-07
fix ToString rule

mmatera:
This PR fixes the rule for `ToString`, to avoid that options like `CharacterEncoding` be taken as the `Form` parameter.


rocky:
LGTM

PR #1603, opened 2026-01-07
remaining adjusts for text render functions and boxed objects

mmatera:
This PR contains some remaining adjusts on the routines that built strings from Boxes, after the change in the `elements` property.


rocky:
LGTM

PR #1604, opened 2026-01-07
Times is 'Left' associative

mmatera:
Like `Plus`,  `Times` is  `Left '- associative. This PR fixes its format function accordingly.

### DIFF
--- a/mathics/builtin/arithfns/basic.py
+++ b/mathics/builtin/arithfns/basic.py
@@ -686,11 +686,11 @@ class Times(InfixOperator, SympyFunction):
         positive = [Expression(SymbolHoldForm, item) for item in positive]
         negative = [Expression(SymbolHoldForm, item) for item in negative]
         if positive:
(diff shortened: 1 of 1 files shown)


rocky:
LGTM

mathics/builtin/arithfns/basic.py
```diff
@@ -686,11 +686,11 @@ def inverse(item):
         positive = [Expression(SymbolHoldForm, item) for item in positive]
         negative = [Expression(SymbolHoldForm, item) for item in negative]
         if positive:
-            positive = create_infix(positive, op, 400, "None")
+            positive = create_infix(positive, op, 400, "Left")
```
rocky, line 689: We should get rid of the magic number 400.

mmatera: This and other format functions in this module are going to be moved and reworked later.  This is the minimal change needed to make `InputForm` work.

PR #1607, opened 2026-01-07
Fix evaluation of InterpretationBox with options

mmatera:
Another round of fixes. These are necessary for  `format_test.m`  work with `InputForm`


mathics/builtin/box/layout.py
```diff
@@ -233,9 +237,13 @@ def elements(self):
             )
         return self._elements
 
-    def eval_create(self, reprs, expr, evaluation):
-        """InterpretationBox[reprs_, expr_]"""
-        return InterpretationBox(reprs, expr)
+    def eval_create(self, reprs, expr, evaluation, options):
+        """InterpretationBox[reprs_, expr_, OptionsPattern[]]"""
+        if isinstance(reprs, EvalMixin):
+            reprs = reprs.evaluate(evaluation)
```
rocky, line 252: What are examples of this code getting used?

mmatera: the exampe is in `test.format.format_test.m`

rocky: That's generic test code. Where in the YAML though?

mmatera: Yep, the YAML file contains the examples

rocky: Yes, I figured that, but anything _specific_.

mmatera:
```
    InputForm:
      expect: InterpretationBox[StyleBox["a - b", ShowStringCharacters -> True, NumberMarks-> True], InputForm[a - b], Editable -> True, AutoDelete -> True]
      expr: MakeBoxes[a-b//InputForm]
```
Each time MakeBoxes is applied to an expression of the form  `InputForm[...]` an `InterpretationBox` object is created. Now, to the output of `MakeBoxes[a-b//InputForm]` with the *expected* result `InterpretationBox[StyleBox["a - b", ShowStringCharacters -> True, NumberMarks-> True], InputForm[a - b], Editable -> True, AutoDelete -> True]`, this second expression must be evaluated. In the version we have in master, this expression is not evaluated because of the options. Then, the comparison between the result of the test and the expected expression fails.

mmatera: When the evaluation is successful, the final expression shows the options in lexicographical order.

rocky: Thanks for the information. This is sufficiently complicated and subtle that a comment should be added. 

I've been trying to think of one, but I am not sure I fully understand. 

Is the idea that reprs is either some sort of evaluatable expression  (or is it just some of StyleBox)? 

If it is in this former category, then we need to evaluate it to turn it into a box which will process and remove the options?

mmatera: This rule converts an `Expression` object into a `InterpretationBox` object.  Expression does not have `boxes_to_*` methods, and `InterpretationBox` does not have an `evaluate` method. In order for the conversion happens, the rule must match with the expression. But the expression we have is of the form `InterpretationBox[boxed, expr, opt1_, opt2_,...]`.
Also, `InterpretationBox` has a HoldAll attribute, so if the first element is not a BoxElementMixin, we need to evaluate it first.

rocky: Ok. Let's add this as a comment. I can do that tomorrow, along with the other small doc changes. 

There is a lot of code in this PR that feels mysterious and bottom-up, as opposed to there beings some higher-level principle that can be used to derive the details coded. 

Maybe later we will have a simpler model for expressing InputForm and how that relates `IntrpretationBox`. IF that happens, it might be that it fits so well that we don't need to make a test like this, or the test feels more obvious.

mmatera: This mechanism of conversions has been around for a while, but probably requires better documentation. The main idea is that `MakeBoxes[]` does not produce `Expression` objects, but `BoxElementMixin` objects. `*Box` expressions should then be evaluated to these kind of objects. The problem that this PR tries to fix is that this conversion fails for `InterpretationBox` expressions having optional parameters.

mathics/builtin/box/layout.py
```diff
@@ -239,6 +243,11 @@ def elements(self):
 
     def eval_create(self, reprs, expr, evaluation, options):
         """InterpretationBox[reprs_, expr_, OptionsPattern[]]"""
+        # If the first element is not a litera, this
```
rocky: litera -> literal

(I am not seeing the +/- commit button, or I'd suggest a commit.)

Thanks for the additional comments and clarification. This helps a lot.

mathics/builtin/box/layout.py
```diff
@@ -233,17 +241,26 @@ def elements(self):
             )
         return self._elements
 
-    def eval_create(self, reprs, expr, evaluation):
-        """InterpretationBox[reprs_, expr_]"""
-        return InterpretationBox(reprs, expr)
+    def eval_create(self, reprs, expr, evaluation, options):
+        """InterpretationBox[reprs_, expr_, OptionsPattern[]]"""
+        # If the first element is not a litera, this
```
mmatera:
```suggestion
        # If the first element is not a literal, this
```

rocky:
LGTM. Many thanks for sticking with this.

PR #1610, opened 2026-01-08
`InputForm` code tweaks

rocky:


mathics/builtin/forms/output.py
```diff
@@ -186,23 +186,48 @@ def eval_mathml(self, expr, evaluation) -> Expression:
...
+     >> Format[Foo[x], InputForm] := Bar
+
+     >> Foo[x] // InputForm
+      = Bar
+
+    In contrast, 'FullForm' output is not altered via 'Format' assignment :
+     >> Format[Foo[x], InputForm] := Baz
+
+     >> Foo[x] // FullForm
+      = Foo[x]
```
rocky, line 230: @mmatera you also provided an example showing how `InputForm` is oblivious to `MakeBoxes`.   I would like to add an example here, but the example that was given also combines an example showing the behavior of `Format` assignment. 

Could you come up with a _simple_ example (showing one concept, not two combined) that shows this?

mmatera: Do you mean this example?
```
In[1]:=MakeBoxes[InputForm[G[x_]],f_]:=RowBox[{"--mb if G--", MakeBoxes[InputForm[x],f]}]

(*the formatter evaluates MakeBoxes[OutputForm[Inputform[G[t]]], StandardForm]*)
In[2]:= InputForm[G[t]]                                                                                                                                                       

Out[2]//InputForm= G[t]

(*the formatter evaluates MakeBoxes[Inputform[M[G[t]]], StandardForm]*)
In[3]:=InputForm[M[G[t]]]//StandardForm 
Out[3]//StandardForm= M[G[t]]
```

However,

```
(*the formatter evaluates MakeBoxes[Inputform[G[t]], StandardForm]*)
In[4]:= InputForm[G[t]]//StandardForm                                                                                                                                         

Out[4]//StandardForm= --mb if G--t
```
This happens because the formatter applies the MakeBoxes rule, and it happens to match the full expression.

rocky: Can this example be simplified? Is there a way to do this without combining the differences that appear as a result of `InputForm` (or `FullForm`) appearing on the left-hand side of a `SetDelayed`, which is a second difference?

mmatera: I thought of these examples as a way to show the internal behavior. Without setting a specific rule that fails to apply, they do not make any sense.

mmatera: I think that the discussion about the combination of InputForm and MakeBoxes is more suitable for the development guide. Here is a little bit out of scope.

rocky: Ok. Then no example is needed here. Thanks.

PR #1611, opened 2026-01-09
Reorganize modules and doc for Forms

rocky:
This organization I think, makes more sense and separates data-specific wrapping modules, like `MatrixForm`, from general-purpose forms like `StandardForm`, which appears in `$PrintForm` at startup.

Our Mathics3-specific forms have been isolated.

Text has been gone over.


mmatera (approved):
LGTM

PR #1620, opened 2026-01-12
Fix MathMLForm for empty operators

mmatera:
"<mo></mo>" cannot be parsed in Mathics-Django browser interface, so avoid converting `""` into `<mo></mo>`.


rocky:
LGTM

PR #1621, opened 2026-01-12
Typo in documentation

mmatera:
This typo spoils the compilation of the LaTeX documentation. Also fix some broken references.

### DIFF
--- a/mathics/builtin/box/layout.py
+++ b/mathics/builtin/box/layout.py
@@ -124,8 +124,8 @@ class FormBox(BoxExpression):
     def init(self, *elems, **kwargs):
         self.box_options = kwargs
         self.form = elems[1]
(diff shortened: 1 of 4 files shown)


rocky:
Ok. But with the backslash, it appears in Django.  So something additional is needed somewhere.

mmatera:
I reformulate the changes following the pattern used in https://github.com/Mathics3/mathics-core/blob/501b9c612777a555642d35719a5b9bb1feba6371/mathics/builtin/atomic/strings.py#L243 and other similar uses. Maybe we should adjust Mathics-Django eventually.

rocky:
Ok. https://github.com/Mathics3/mathics-django/issues/235 was added to note the breakage caused by the change in `$` handling here. 

LGTM

PR #1628, opened 2026-01-17
More on WMA Form compatibility

mmatera:
This PR we go with another step in make Mathics format routines closer to WMA:

* For compatibility with WMA, TeXForm and MathMLForm are now based on TraditionalForm.
* Start to reduce MakeBoxes rules with a non-boxform symbol as its second argument.
* When LaTeX or XML documentation is generated with  `docpipeline.py`, output is wrapped in `StandardForm` to make those expressions look closer to the WMA interface.


rocky:
Something I don't understand.  In WMA: 

```mathematica
In[5]:= Integrate[f[x],x] //TeXForm                                              

Out[5]//TeXForm= \int f(x) \, dx
```

In mathics3: 

```mathematica
In[1]:= Integrate[f[x],x] //TeXForm
Out[1]//TeXForm= \int f\left(x\right) \, dx
```

I suppose one could argue that `\left( ... \right)` is an improvement for TeX/LateX formatting

In Django:

```
Integrate[f[x],x] //TeXForm
    \int f\left[x\right] \, dx
```

Note the square brackets still. And doing the above, substituting `//MathMLForm`, I also see square brackets? 

@mmatera What's going on?

mmatera:
> Something I don't understand. In WMA:
> 
> ```mathematica
> In[5]:= Integrate[f[x],x] //TeXForm                                              
> 
> Out[5]//TeXForm= \int f(x) \, dx
> ```
> 
> In mathics3:
> 
> ```mathematica
> In[1]:= Integrate[f[x],x] //TeXForm
> Out[1]//TeXForm= \int f\left(x\right) \, dx
> ```
> 
> I suppose one could argue that `\left( ... \right)` is an improvement for TeX/LateX formatting
> 

In WMA, `\left` and `\right` are used just when the inner expression contains elements like `\frac{...}{...}` or `..^...` which are taller than the brackets. For example:
```
In[1]:= F[Q[R[a]]]//TeXForm                                                   
Out[1]//TeXForm= F(Q(R(a)))

In[2]:= F[Q[x]/R[a]]//TeXForm                                                   
Out[2]//TeXForm= F\left(\frac{Q(x)}{R(a)}\right)
```
In the first case, we have a nexted expression, but does not include "2D" text. On the other hand, adding `\left` and  `\right` does not make a difference.

> In Django:
> 
> ```
> Integrate[f[x],x] //TeXForm
>     \int f\left[x\right] \, dx
> ```
> 

Not sure why, but it probably ask for `StandardForm` by default. I have to check. In any case, it seems to be the right behavior, if we want to mimic the WMA notebook interface.

> Note the square brackets still. And doing the above, substituting `//MathMLForm`, I also see square brackets?
> 
> @mmatera What's going on?

mmatera:
Indeed, Mathics-Django uses StandardForm,  https://github.com/Mathics3/mathics-django/blob/b24f3710db8af868c9b8298476420cc15748ecfa/mathics_django/web/format.py#L103C1-L103C84

as I suggested here too https://github.com/Mathics3/mathics-core/pull/1628/changes#diff-7b7617ca0d7cccdfd745672321d56473f683fc875ffb27bd95e97042fb116e4bR272

rocky:
Ok. Thanks for the investigation/explanation. LGTM. 

Please let's start to address some of the older outstanding issues before on. Thanks again.

mmatera:
> Ok. Thanks for the investigation/explanation. LGTM.
> 
> Please let's start to address some of the older outstanding issues before on. Thanks again.

On each round I do with the refactor of Format/Boxes I find some of these older issues. The ones that are related I try to handle, but my plan is to at least finish with the basic framework of formatting this month. Then I can focus on other issues.

rocky:
> > Ok. Thanks for the investigation/explanation. LGTM.
> > Please let's start to address some of the older outstanding issues before on. Thanks again.
> 
> On each round I do with the refactor of Format/Boxes I find some of these older issues. The ones that are related I try to handle, but my plan is to at least finish with the basic framework of formatting this month. Then I can focus on other issues.

Ok. When adding or finding `format_xxx` routine in `mathics.builtin` that is to be changed, I ask that you put the bulk of the code in  `mathics.form` and in the builtin just add a call to this new routine. I do not believe this will take much effort, and it makes it easier to go over the form aspect overall.

PR #1631, opened 2026-01-18
Format module reorganization

mmatera:
Here there is a proposal of reorganization of the code associated to format. I move the code of modules `mathics.eval.makeboxes`, `mathics.form` and `mathics.format` inside of a module `mathics.format`.
The logic of the organization is explaind in the __init__.py file of `mathics.format`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,7 @@ MATHICS3_MODULE_OPTION ?= --load-module pymathics.graph,pymathics.natlang
    djangotest \
    gstest \
    latexdoc \
+   mypy \
    plot-detailed-tests\
    pytest \
    pytest-x \
@@ -127,6 +128,9 @@ clean: clean-cython clean-cache
 	rm -f factorials || true; \
 	rm -f mathics/data/op-tables || true; \
 	rm -rf build || true
+
+mypy:
+	mypy --install-types --ignore-missing-imports --non-interactive mathics
 
 plot-detailed-tests:
 	MATHICS_CHARACTER_ENCODING="ASCII" MATHICS_PLOT_DETAILED_TESTS="1" $(PYTHON) -m pytest -x $(PYTEST_OPTIONS) test/builtin/drawing/test_plot_detail.py
--- a/mathics/builtin/forms/data.py
+++ b/mathics/builtin/forms/data.py
@@ -96,7 +96,10 @@ class BaseForm(FormBaseClass):
     def eval_makeboxes(self, expr, n, f, evaluation: Evaluation):
         """MakeBoxes[BaseForm[expr_, n_],
         f:StandardForm|TraditionalForm|OutputForm]"""
-        return eval_baseform(expr, n, f, evaluation)
+        try:
+            return eval_baseform(expr, n, f, evaluation)
+        except ValueError:
+            return None
 
 
 class _NumberForm(Builtin):
--- a/mathics/builtin/list/predicates.py
+++ b/mathics/builtin/list/predicates.py
@@ -2,11 +2,11 @@
 Predicates on Lists
 """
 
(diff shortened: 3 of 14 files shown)


rocky:
Overall, I think this is a good idea. If I understand this correctly, `mathics.form` disappears and is put under `mathics.format`, which is now subdivided into: 
* formvalues
* forms
* makeboxes
* export

I can suggest small changes to the docstrings for grammar, but let's first get the bigger terminology. 

"export" feels to me like the wrong word. Here is what dictionary.com says about export: 

> to ship (commodities) to other countries or places for sale, exchange, etc.
> to send or transmit (ideas, institutions, etc.) to another place, especially to another country.
> Computers. to save (documents, data, etc.) in a format usable by another software program.

In contrast, for render: 

> to translate into another language.
>  to represent; depict, as in painting
> Digital Technology.
> to use the processing power of computer hardware and software to synthesize (the components of an image or animation) in a final graphic output.

I give both the computer usage and non-computer usage to give a feel for the broader sense of the word.

The other term I think would be better to change "makeboxes" to "box". "MakeBoxes" is a WMA built-in function.  The general concept is "boxing", and the verb for that is "box". 

To regularize things, we have been using singular rather than plural, so "form" and "formvalue" would be better.  But the problem with the term "formvalue" is that it, too, is jargon that most people won't know because the built-in function `FormValue` is not documented in WMA. I think `form_rule` would be better, which conveys more simply what this does. 

BTW I am seeing `asy_numpy.py` not moved. 

@mmatera your thoughts?

mmatera:
> Overall, I think this is a good idea. If I understand this correctly, `mathics.form` disappears and is put under `mathics.format`, which is now subdivided into:
> 
>     * formvalues
> 
>     * forms
> 
>     * makeboxes
> 
>     * export
> 
> 
> I can suggest small changes to the docstrings for grammar, but let's first get the bigger terminology.
> 
> "export" feels to me like the wrong word. Here is what dictionary.com says about export:
> 
> > to ship (commodities) to other countries or places for sale, exchange, etc.
> > to send or transmit (ideas, institutions, etc.) to another place, especially to another country.
> > Computers. to save (documents, data, etc.) in a format usable by another software program.
> 
> In contrast, for render:
> 
> > to translate into another language.
> > to represent; depict, as in painting
> > Digital Technology.
> > to use the processing power of computer hardware and software to synthesize (the components of an image or animation) in a final graphic output.
> 
> I give both the computer usage and non-computer usage to give a feel for the broader sense of the word.
> 
OK, so let's use `render`

> The other term I think would be better to change "makeboxes" to "box". "MakeBoxes" is a WMA built-in function. The general concept is "boxing", and the verb for that is "box".
> 
> To regularize things, we have been using singular rather than plural, so "form" and "formvalue" would be better. But the problem with the term "formvalue" is that it, too, is jargon that most people won't know because the built-in function `FormValue` is not documented in WMA. I think `form_rule` would be better, which conveys more simply what this does.
> 
> BTW I am seeing `asy_numpy.py` not moved.

where is this file?

> 
> @mmatera your thoughts?

@rocky, I think you are in a better position to make these choices and corrections. Feel free to make the adjustments in the names and docstrings.

rocky:
> > BTW I am seeing `asy_numpy.py` not moved.
> 
> where is this file?

Ooops. This is probably something I was working on a while ago.  

> 
> > @mmatera your thoughts?
> 
> @rocky, I think you are in a better position to make these choices and corrections. Feel free to make the adjustments in the names and docstrings.

Thanks. I will put in a PR to this PR in a day or so.

rocky:
Not sure why we are getting errors now. I don't have any further stuff. LGTM after we understand what's up with the errors.

mmatera:
The errors are in the mypy tests. I am going over this later.

PR #1634, opened 2026-01-19
More tweaks in InputForm

mmatera:
* Handle  the `Optional` function/operator in InputForm expressions
* take `Precedence` into account for `Pre/PostFix`
* Add more annotations.
* Take the chance to for moving `format_plus` and `format_times` implementation to `mathics.format.form_rule`

### DIFF
--- a/mathics/builtin/patterns/defaults.py
+++ b/mathics/builtin/patterns/defaults.py
@@ -36,7 +36,7 @@ class Optional(InfixOperator, PatternObject):
     >> f[a]
      = {a, 1}
 
-    Note that '$symb$ : $pattern$' represents a 'Pattern' object. However, there is no
+    Note that '$symb$ : $pattern$' represents a 'Pattern' object. However, there is no \
     disambiguity, since $symb$ has to be a symbol in this case.
 
     >> x:_ // FullForm
@@ -46,7 +46,7 @@ class Optional(InfixOperator, PatternObject):
     >> x:_+y_:d // FullForm
      = Pattern[x, Plus[Blank[], Optional[Pattern[y, Blank[]], d]]]
 
-    's_.' is equivalent to 'Optional[s_]' and represents an optional parameter which, if omitted,
+    's_.' is equivalent to 'Optional[s_]' and represents an optional parameter which, if omitted, \
     gets its value from 'Default'.
     >> FullForm[s_.]
      = Optional[Pattern[s, Blank[]]]


mathics/builtin/atomic/symbols.py
```diff
@@ -266,19 +269,20 @@ class Definition(Builtin):
      .
      . N[r, MachinePrecision] = 3.5
      .
-     . Format[args___, MathMLForm] = Infix[{args}, "~"]
+     . Format[r[args___], MathMLForm] = Infix[{args}, "~"]
```
rocky: What does `r[args___]` do? 

Let's pick up "~" from a table.

rocky:
> * Handle `Optional`

In the user docs, we should mention and give an example

mathics/builtin/atomic/symbols.py
mmatera: > What does `r[args___]` do?
> 
This says that there is a format rule for `r[args__]` associated to `MathMLForm`. In master, the argument of `Format` is formatted in `OutputForm`, so we get something meaninless.

> Let's pick up "~" from a table.

But this is the reference result of a Doctest, so we want to write explicitly. What would make more sense is to format `Infix[{elem_}, "~"]` (i.e., with a single element) as `Infix[{elem}]`.

rocky: > > What does `r[args___]` do?
> 
> This says that there is a format rule for `r[args__]` associated to `MathMLForm`. In master, the argument of `Format` is formatted in `OutputForm`, so we get something meaninless.

Thanks for the enlightenment. 
> 
> > Let's pick up "~" from a table.
> 
> But this is the reference result of a Doctest, so we want to write explicitly. What would make more sense is to format `Infix[{elem_}, "~"]` (i.e., with a single element) as `Infix[{elem}]`.

Yes, you are correct on both counts.

mmatera: Correction: here, "~" is not the default value of `Infix`, but something arbitrary used for the example. It could be any other symbol or string.

rocky: <strike>To make things clearer, let's use a different symbol then. </strike> I see you've done that already. Thanks!

mathics/builtin/atomic/symbols.py
```diff
@@ -244,7 +239,7 @@ class Definition(Builtin):
 
     Definition of a rather evolved (though meaningless) symbol:
     >> Attributes[r] := {Orderless}
-    >> Format[r[args___]] := Infix[{args}, "~"]
+    >> Format[r[args___]] := Infix[{args}, "#"]
```
rocky, line 242: Thanks!

mathics/builtin/atomic/symbols.py
mmatera: I use `#` now

mathics/builtin/atomic/symbols.py
```diff
@@ -73,18 +78,17 @@ def format_rule(
...
         if isinstance(rule, Rule):
-            r = rhs(
+            lhs_pat = Expression(SymbolInputForm, lhs(rule.pattern.expr))
+            repl_expr = rhs(
                 rule.replace.replace_vars(
                     {"System`Definition": Expression(SymbolHoldForm, SymbolDefinition)}
                 )
             )
-            r = Expression(SymbolInputForm, r)
+            repl_expr = Expression(SymbolInputForm, repl_expr)
```
rocky, line 87: Thanks!

mathics/format/form/inputform.py
```diff
@@ -169,7 +169,7 @@ def _infix_expression_to_inputform_text(
     result = parenthesize(precedence, operand, result, parenthesized)
 
     # Process the rest of operands
-    parenthesized = (SymbolLeft, SymbolNonAssociative)
+    parenthesized = group in (SymbolLeft, SymbolNonAssociative)
```
rocky: This smells fishy. Normally, you shouldn't have to list specific symbols; it's all done by comparing precedence values. 

Let's comment to this effect, and we'll come back and address this some other time.

mathics/format/form/outputform.py
```diff
@@ -379,7 +379,7 @@ def _infix_outputform_text(expr: Expression, evaluation: Evaluation, **kwargs) -
     result = parenthesize(precedence, operand, result, parenthesized)
 
     # Process the rest of operands
-    parenthesized = (SymbolLeft, SymbolNonAssociative)
+    parenthesized = group in (SymbolLeft, SymbolNonAssociative)
```
rocky: See above.

mmatera: The idea is to check if `group` belongs to one of these lists. Maybe should be a `set` and maybe it should be in constants., like 
```
FIRST_PARENTHESIZED={SymbolRight, SymbolNonAssociative}
REST_PARENTHESIZED={SymbolLeft, SymbolNonAssociative}
```

mathics/format/form/inputform.py
mmatera: The reason to compare this is the need to handle the case of operations with the same precedence, or cases where the operands have signs. See what happened with `-1;-1;-1//HoldForm` when `parenthesized` has a wrong value.

rocky: > The reason to compare this is the need to handle the case of operations with the same precedence, 

When operators have the same precedence, then an associative property is used to indicate a specific association order, should it be different from the normal way. For example, if I want to indicate "(a + (b + c))" where "+" is left associative. 

> or cases where the operands have signs. See what happened with `-1;-1;-1//HoldForm` when `parenthesized` has a wrong value.

What happened? I don't have enough information to know what you mean here.

rocky: At any rate, I think we should add a comment as to why normal precedence and association rules aren't used, possibly marking it for revision later.

mmatera: > 
> What happened? I don't have enough information to know what you mean here.

Depending on the group value, `-1;;-1;;-1` would be parenthesized as `(-1);;(-1);;(-1)`, for example. In any case, it seems that we are not treating the `Group` parameter in the same way as in WMA. Probably we are going to pass through this again. In any case, it goes beyond the goal of this PR.

mathics/format/form/util.py
```diff
@@ -50,6 +50,20 @@ class _WrongFormattedExpression(Exception):
 PRECEDENCE_POWER: Final[int] = PRECEDENCES.get("Power", 590)
 
 
+# These constants are used to decide if two operands with the
```
rocky, line 56:
```suggestion
# TODO: (rocky) See if we can accomplish parenthesization using precedence and association values as is common for this kind of thing.
# 
# These constants are used to decide if two operands with the
```

rocky:
> > * Handle `Optional`
> 
> In the user docs, we should mention and give an example

@mmatera can you come up with a good example?

mmatera:
> > > * Handle `Optional`
> > 
> > 
> > In the user docs, we should mention and give an example
> 
> @mmatera can you come up with a good example?

In this case, I am not sure what would be a good example. This is just a minimal, simple example:

https://github.com/Mathics3/mathics-core/pull/1634/changes#diff-2c7428121ef76f0d9412d29b2ed8098ca90ff657482cfa0d9fbb4c3b99efbb2eR59

mathics/builtin/patterns/defaults.py
rocky:
```suggestion
    Note that '$symb$ : $pattern$' represents a 'Pattern' object. However, there is no \
```

rocky:
```suggestion
    's_.' is equivalent to 'Optional[s_]' and represents an optional parameter which, if omitted, \
```

rocky:
LGTM - the docs need going over. I will do that in a separate PR.

 (And it looks like we are not tagging the Optional operator properly in mathics-scanner, I will put in a PR for that too.)

Thanks for your patience.

mmatera:
> LGTM - the docs need going over. I will do that in a separate PR.
> 
> (And it looks like we are not tagging the Optional operator properly in mathics-scanner, I will put in a PR for that too.)
> 

Why?

Notice that as in WMA, `A_.` is a valid entry, but `A__.` is taken as an incomplete line. This is why I add the parentheses in InputForm. 

> Thanks for your patience.

rocky:
@mmatera If I have it correct, if we merge this in before the revised OutputForm, then we can just merge OutputForm into master. Is that acceptable?

For OutputForm, tomorrow, I will try to go over and understand the OutputForm YAML. I think the code part is fine, just want to double check.

mmatera:
> For OutputForm, tomorrow, I will try to go over and understand the OutputForm YAML. I think the code part is fine, just want to double check.

I just split this into InputForm and OutputForm to facilitate the review. So, as you prefer.

rocky:
> > LGTM - the docs need going over. I will do that in a separate PR.
> > (And it looks like we are not tagging the Optional operator properly in mathics-scanner, I will put in a PR for that too.)
> 
> Why?

I assume the "why" question is about tagging Optional. (Because for the docs, the section on Optional is very confusing)

Right now, in the named character tables, we list "Option", when it should be Option _al_  . It should also have the "operator" field filled in.

> 
> Notice that as in WMA, `A_.` is a valid entry, but `A__.` is taken as an incomplete line. This is why I add the parentheses in InputForm.

I don't understand what this is in reference to.

mmatera:
> > Why?
> 
> I assume the "why" question is about tagging Optional. (Because for the docs, the section on Optional is very confusing)
> 
Yep

> Right now, in the named character tables, we list "Option", when it should be Option _al_ . It should also have the "operator" field filled in.
> 

OK, but this goes beyond what I am trying to do here. I just propose these changes to keep the tests working, and as close as possible to WMA. But most of this is about formatting.

> > Notice that as in WMA, `A_.` is a valid entry, but `A__.` is taken as an incomplete line. This is why I add the parentheses in InputForm.
> 
> I don't understand what this is in reference to.

I guess you was talking about parsing. But again, it is another discussion. What I tried to do here was just to get from `InputForm` something that can be parsed again in Mathics3 or WMA.

rocky:
> I guess you was talking about parsing. But again, it is another discussion. 

Yep, PR created over in mathics-scanner. 

> What I tried to do here was just to get from `InputForm` something that can be parsed again in Mathics3 or WMA.

Ok. Then let's merge and iterate.

PR #1570, opened 2025-12-28
OutputForm

mmatera:
Now we have `PaneBox` and `InterpretationBox`, this PR implements a version of `OutputForm` that is more compatible with the one in WMA.

PR #1635, opened 2026-01-19
OutputForm rebase

mmatera:
This PR rebase #1570, including just the changes required to use the new implementation of  `OutputForm`.
* Remove rules of the form `MakeBoxes[_, OutputForm]`
* Implement  `eval_makeboxes_outputform`
* use `eval_makeboxes_outputform` in `MakeBoxes[OutputForm[_],_]`
* Modify `format_element` in a way that use `form` as the second argument of `eval_makeboxes` if `form` is a BoxForm.
* Update `format_test.yaml` according to the new results, closer to WMA.

### DIFF
--- a/mathics/builtin/list/eol.py
+++ b/mathics/builtin/list/eol.py
@@ -1172,6 +1172,7 @@ class Part(Builtin):
 
         i = i.get_sequence()
         list = Expression(SymbolMakeBoxes, list, f).evaluate(evaluation)
+        # FIXME: pick up values LeftDoubleBracket and RightDoubleBracket from named-characters.yaml
         open, close = "\u301a", "\u301b"
         indices = list_boxes(i, f, evaluation, open, close)
         result = RowBox(list, *indices)


mathics/builtin/forms/print.py
```diff
@@ -165,7 +170,7 @@ class MathMLForm(FormBaseClass):
     summary_text = "format expression as MathML commands"
 
     def eval_mathml(self, expr, evaluation) -> Expression:
-        "MakeBoxes[MathMLForm[expr_], (OutputForm|StandardForm|TraditionalForm)]"
+        "MakeBoxes[MathMLForm[expr_], Alternatives[StandardForm,TraditionalForm]]"
```
rocky: What's the difference between using `|` vs `Alternatives[]`?

mathics/doc/documentation/1-Manual.mdoc
```diff
@@ -934,12 +942,12 @@ For instance, you can override 'MakeBoxes' to format lists in a different way:
...
    = {1, 2, 3}
 
-  #> {1, 2, 3} // TeXForm
+  >> {1, 2, 3} // TeXForm
    = \left[1 2 3\right]
 
 However, this will not be accepted as input to \Mathics anymore:
   >> [1 2 3]
-  : Expression cannot begin with "[1 2 3]" (line 1 of "<test-Formatting Output-28>").
+  : Expression cannot begin with "[1 2 3]" (line 1 of "<test-Formatting Output-29>").
```
rocky: Can we use the more general (and probably more robust) "..." instead of 28/29 here?

mathics/builtin/forms/print.py
mmatera: I have seen problems confusing `Alternatives` with `Or`, but now I am not able to reproduce it. I can revert it.

rocky: No need to change. I was just curious.

mathics/doc/documentation/1-Manual.mdoc
mmatera: Docpipeline works if I  replace the whole line with `...`,  but not just a part.

mmatera:
```suggestion
  : ...
```

mathics/builtin/list/eol.py
```diff
@@ -1168,14 +1168,11 @@ class Part(Builtin):
...
-        f:StandardForm|TraditionalForm|OutputForm|InputForm]"""
+        (f:StandardForm|TraditionalForm)]"""
 
         i = i.get_sequence()
         list = Expression(SymbolMakeBoxes, list, f).evaluate(evaluation)
-        if f.get_name() in ("System`OutputForm", "System`InputForm"):
-            open, close = "[[", "]]"
-        else:
-            open, close = "\u301a", "\u301b"
+        open, close = "\u301a", "\u301b"
```
rocky, line 1176:
```suggestion
        # FIXME: pick up values LeftDoubleBracket and RightDoubleBracket from named-characters.yaml
        open, close = "\u301a", "\u301b"
```

mathics/doc/documentation/1-Manual.mdoc
rocky: Then let's replace the entire line. This is very fragile, and we use other mechanisms for testing output.

test/format/format_tests.yaml
```diff
@@ -314,22 +300,20 @@ Association[a -> x, b -> y, c -> Association[d -> t, Association[e -> u]]]:
     System`StandardForm: <|a->x, b->y, c-><|d->t, e->u|>|>
     System`TraditionalForm: <|a->x, b->y, c-><|d->t, e->u|>|>
 Complex[1.09*^12, 3.]:
+  msg: Complex number
   latex:
     System`InputForm: 1.09*{}^{\wedge}12 + 3.*I
-    System`OutputForm: 1.09\times 10^{12}\text{ + }3. I
+    System`OutputForm: '\text{1.09×10${}^{\wedge}$12 + 3. I}'
```
rocky: In Wolframscript I am seeing `\times` used instead of `×` which I think is more correct.  This also occurs many other places in this file.

test/format/format_tests.yaml
```diff
@@ -87,25 +87,25 @@
   msg: very small negative real number (<10^-6)
   latex:
     System`InputForm:  -1.*{}^{\wedge}-6
-    System`OutputForm: -1.\times 10^{-6}
+    System`OutputForm: '\text{-1.×10${}^{\wedge}$-6}'
```
rocky: In WolframScript, I am seeing `grave` and * used: 

```mathematica
-\text{1.$\grave{ }$*${}^{\wedge}$-6}
```

test/format/format_tests.yaml
rocky: Regarding the recent CI failures, I think if we use plain "*", then TeX and LaTeX will convert this to a nice character operator automatically. And I think I have seen this done in Wolframscript output for TeXForm.

rocky:
Overall, there is a big improvement. I am seeing a lot of discrepancies in `format_tests.yaml` between wolframscript and what we produce. 

Let's discuss how to handle.

test/format/format_tests.yaml
mmatera: When I build the table, I found some problems with the TeXForm and MathMLForm output produced by WMA: sometimes it just shows errors and fails to produce an output, or the produced output does not compile well in LaTeX or render properly in the browser. The partial solution I found was to pick the current output from Mathics, and  generate native WMA tests to compare and improve the cases that make sense to improve.

mmatera: Regarding this particular character, I just committed a patch to produce the right LaTeX command.

test/format/format_tests.yaml
mmatera: This is the output I get in WolframScript for this test case:
```
In[1]:= -1. 10^6 //InputForm//TeXForm                                           

Out[1]//TeXForm= \text{-1.*${}^{\wedge}$6}

In[2]:= -1. 10^6 //OutputForm//TeXForm                                          

Out[2]//TeXForm= \text{      6$\backslash $n-1. 10}

In[3]:= -1. 10^6 //StandardForm//TeXForm                                        

Out[3]//TeXForm= -1.\times 10^6

In[4]:= -1. 10^6 //TraditionalForm//TeXForm                                     

Out[4]//TeXForm= -1.\times 10^6
```

The main difference in `OutputForm` comes from the pretty-print-like output, which we currently do not support. 
`\grave{ }` is used to mark the tick character, used in `StandardForm` for 'Real' numbers:
```
In[1]:= -1. 10^6 //StandardForm
Out[1]//StandardForm= -1.`*^6
```
However, as can be seen in the example, it is not used in TeXForm.

rocky:
This is a bit messy, and I think it would have been better to have focused on something smaller, like getting TeXForm working better, before combining that with OutputForm work.  However, I think we should commit this and iterate.

mmatera:
> This is a bit messy, and I think it would have been better to have focused on something smaller, like getting TeXForm working better, before combining that with OutputForm work. However, I think we should commit this and iterate.

Again, the goal is to establish a framework that works at least as well as our last version, but that allows to make formatting more compatible with WMA. This is why I tryied to avoid fixing some details in our TeXForm implementation, which in comparison with WMA is not too bad...

mathics/builtin/forms/print.py
mmatera: I have changed all the occurences of `Alternatives` in favor of using `|`.

mathics/doc/documentation/1-Manual.mdoc
mmatera: Done.

test/format/format_tests.yaml
mmatera: I put this in another PR. We can see to load these equivalences from the tables there.

mmatera:
> Overall, there is a big improvement. I am seeing a lot of discrepancies in `format_tests.yaml` between wolframscript and what we produce.
> 
> Let's discuss how to handle.

I start to put more on the true output of WS in `test/format/format_test-WMA.yaml`. See #1639. The content `format_tests.yaml` should try to be similar or in some cases *better* than what we see in `test/format/format_test-WMA.yaml`.

rocky:
> > This is a bit messy, and I think it would have been better to have focused on something smaller, like getting TeXForm working better, before combining that with OutputForm work. However, I think we should commit this and iterate.
> 
> Again, the goal is to establish a framework that works at least as well as our last version, but that allows to make formatting more compatible with WMA. This is why I tryied to avoid fixing some details in our TeXForm implementation, which in comparison with WMA is not too bad...

The situation though, is that I think some concepts like understanding precedence is missing, and instead we have precedence more hard-coded, which will probably mean there are omissions.  And we have hard-coded characters in there.   So I expect there will be another pass at this. 

But this is a good start. LGTM.

mmatera:
> > > This is a bit messy, and I think it would have been better to have focused on something smaller, like getting TeXForm working better, before combining that with OutputForm work. However, I think we should commit this and iterate.
> > 
> > 
> > Again, the goal is to establish a framework that works at least as well as our last version, but that allows to make formatting more compatible with WMA. This is why I tryied to avoid fixing some details in our TeXForm implementation, which in comparison with WMA is not too bad...
> 
> The situation though, is that I think some concepts like understanding precedence is missing, and instead we have precedence more hard-coded, which will probably mean there are omissions. And we have hard-coded characters in there. So I expect there will be another pass at this.
> 
Another pass is coming to handle these aspects. 
> But this is a good start. LGTM.
Thanks for the large review!

rocky:
> > Overall, there is a big improvement. I am seeing a lot of discrepancies in `format_tests.yaml` between wolframscript and what we produce.
> > Let's discuss how to handle.
> 
> I start to put more on the true output of WS in `test/format/format_test-WMA.yaml`. See #1639. The content `format_tests.yaml` should try to be similar or in some cases _better_ than what we see in `test/format/format_test-WMA.yaml`.

Better is good.

PR #1636, opened 2026-01-20
Go over Optional doc

rocky:


mathics/builtin/patterns/defaults.py
```diff
@@ -19,48 +19,58 @@
...
      = {a, 1}
 
-    Note that '$symb$ : $pattern$' represents a 'Pattern' object. However, there is no \
-    disambiguity, since $symb$ has to be a symbol in this case.
-
-    >> x:_ // FullForm
-     = Pattern[x, Blank[]]
-    >> _:d // FullForm
-     = Optional[Blank[], d]
-    >> x:_+y_:d // FullForm
```
rocky, line 46: This example does not seem motivated.

mathics/builtin/patterns/defaults.py
```diff
@@ -19,48 +19,58 @@
...
-     = Pattern[x, Plus[Blank[], Optional[Pattern[y, Blank[]], d]]]
-
-    's_.' is equivalent to 'Optional[s_]' and represents an optional parameter which, if omitted, \
-    gets its value from 'Default'.
-    >> FullForm[s_.]
-     = Optional[Pattern[s, Blank[]]]
-
-    'InputForm' shows it in its 'Infix' or 'Postfix' form depending on the \
-    number of parameters:
```
rocky, line 55: InputForm, by default, uses infix for equations.  I do not understand from the example where the number of parameters is coming into play. And I do not see where postifx notation is selected.

mmatera: There is no Infix form with just one argument...

mathics/builtin/patterns/defaults.py
mmatera: I guess the motivation is to show the order of interpretation: the second colon is interpreted first, to build the inner  `Optional` expression, and then the first one is interpreted as a named Pattern.

rocky: Ok. I'll add this example back, adding this explanation.

mathics/builtin/patterns/defaults.py
rocky: Technically, this degenerate case is defined on Infix:

```
In[2]:= Infix[{1}, "+"]
Out[2]= 1
```

But the larger and more important point is: what does this Infix and Postfix stuff have to do with the examples?

mathics/builtin/patterns/defaults.py
rocky: > I guess the motivation is to show the order of interpretation: the second colon is interpreted first, to build the inner `Optional` expression, and then the first one is interpreted as a named Pattern.

Hmm... It looks like we may have the wrong Precedence indicated for Pattern. We give Pattern value 670 while it is 150 in WMA. In both WMA and Mathics3, though,  Optional is 140.  So the behavior is natural.<strike>in WMA while in Mathics3, it goes contrary to what is specified.</strike>

Is this something we specifically hacked around? <strike>and would get wrong were we to follow the usual precedence rules?</strike>

*Edit* what we have would give the same behavior. But, I wonder if there is a reason not to use the same value 150.

rocky: I looked at the precedence issue, and it is tagged with the right value in mathics_scanner, but Pattern is not tagged as an operator in mathics-core, and if that's changed, other things break.  670 is the magic number that's given when we can't find a precedence value tagged.

Sigh.

mathics/builtin/patterns/defaults.py
mmatera: Notice also that `Optional` uses a different character in the case of one and two parameters. I am not completely sure that can be handled in the same way that other "Infix" operators. I just used the terms Infix/Postfix for the position of the elements and the symbol, but I do not know if the analogy goes beyound that.

rocky:
@mmatera thoughts on this PR?

PR #1639, opened 2026-01-20
Update the yaml examples of MathML and TeX Form outputs

mmatera:
This PR contains some update on the database of examples collected from WolframScript. Some examples were removed because WMA do not support producing a MathMLForm or TeXForm  output for certain expressions, including some very simple ones (numbers). In this sense, our implementation is stronger than the one in WMA.


rocky:
LGTM. One small suggestion in the YAML file at some point. It would help readability if there were a space of some sort between each example.  This makes it clearer whether the last line, I think it is "msg" belongs to the group above as opposed to the group below.

rocky:
LGTM

PR #1644, opened 2026-01-24
Use MathicsScanner tables for render LaTeX

mmatera:
This should work when https://github.com/Mathics3/mathics-scanner/pull/144#issue-3849291168 gets merged.


rocky:
LGTM. I think a lot more could be done, but this is a step in the right direction.

PR #1642, opened 2026-01-21
Makeboxes overhault

mmatera:
This PR does the largest step so far in making the formatting process in Mathics compatible with the one in WMA. 
The main change is in the sequence of formatting. Now, `MakeBoxes` rules are not `Downvalues`, of the `MakeBoxes` symbol, but are stored as `FormatValues` of the corresponding symbols. Rules in `MakeBoxes` are now restricted to call the  `format_element` function, and return a Box expression that represents its input. Hence, in loading definitions, `MakeBoxes` is not a special symbol anymore.
Also, the default implementation for formatting basic elements like symbols, expressions, and lists does not pass through the evaluation process until explicit rules are set by the user.

PR #1646, opened 2026-01-24
Implement Format MakeBoxes and Format[...]//OutputForm

mmatera:
This PR implements the formatting of `Format[...]` expressions. I implement it in a way that works independently from #1642, but still work with it.


mathics/builtin/layout.py
```diff
@@ -62,12 +62,56 @@ class Format(Builtin):
...
+
+    Notice that differently from WMA, 'Format' expressions are not \
+    formatted in 'InputForm':
+    >> Format[{a->Integrate[F[x], x]}, StandardForm]
+     = ...
+    >> Format[{a->Integrate[F[x], x]}, StandardForm] //InputForm
+     = Format[{a -> Integrate[F[x], x]}, StandardForm]
+
+    This choice is more consistent with the meaning of 'InputForm' \
+    in the sense it gives the text required to reproduce the expression.
```
rocky: I don't understand. Above is written:

> 'Format' expressions are not formatted in 'InputForm': ...

So, how does _supplying_ an InputForm make things more consistent with InputForm?

mmatera: No, what I said is that if you want to see the format rule, you need to use a format that does not apply the format rule to the expression. Suppose you define a rule:
```
In[1]:= Format[F[x__]]:=HoldForm[x+F]                                           
```

And then ask for the FormatValues:
```
In[2]:= FormatValues[F]                                                         

FormatType::ftype: Value of option FormatType -> FormatType is not valid.
Out[2]= {HoldPattern[MakeBoxes[(x__) + F, FormatType_]] :> x + F, 
 
>    HoldPattern[(x__) + F] :> x + F}
```
Then, the rule is also formatted, because the format applies to any form. In WMA, also, if you apply `InputForm`

```
In[3]:= FormatValues[F] // InputForm                                            


FormatType::ftype: Value of option FormatType -> FormatType is not valid.
Out[2]//InputForm= 
{HoldPattern[MakeBoxes[F[x__], FormatType_]] :> x + F, 
 HoldPattern[F[x__]] :> HoldForm[x + F]}
```
the format is applyied to the rule. However, this is not something that you can enter in the command line to produce the same rule. This is why I say that WMA is inconsistent. With this PR, Mathics3 has the right behavior, giving something that we can enter to generate the rule.

rocky: > No, what I said is that if you want to see the format rule, you need to use a format that does not apply the format rule to the expression. Suppose you define a rule:
> 
> ```
> In[1]:= Format[F[x__]]:=HoldForm[x+F]                                           
> ```
> 
> And then ask for the FormatValues:
> 
> ```
> In[2]:= FormatValues[F]                                                         
> 
> FormatType::ftype: Value of option FormatType -> FormatType is not valid.

In WMA, I do not see a `FormatType::ftype` error.  What gives here?

mmatera: In master, there is no error because `Format[...]` does not have an associated MakeBoxes rule. In this branch I got
```
In[1]:= Format[F[x__]]:=HoldForm[x+F] 

In[2]:= FormatValues[F]      
FormatType::ftype: Value of option FormatType -> OutputForm is not valid.
Out[2]= {HoldPattern[<math display="block"><mrow><mtext>x__</mtext> <mo>+</mo> <mi>F</mi></mrow></math>] ⧴ x + F, HoldPattern[x__ + F] ⧴ x + F, HoldPattern[x__+F] ⧴ x + F, HoldPattern[\text{x\_\_}+F] ⧴ x + F, HoldPattern[x__+F] ⧴ x + F}
```

rocky: I think giving an error here is not correct since WMA doesn't. Right?

mmatera: WMA does show the error, but is a slighly different error.
```
FormatType::ftype: Value of option FormatType -> FormatType is not valid.
Out[2]= {HoldPattern[MakeBoxes[(x__) + F, FormatType_]] :> x + F, 
 
>    HoldPattern[(x__) + F] :> x + F}

```

mmatera: The difference comes from the way in which WMA implements the assignment to `Format` with a single argument:
```
In[3]:= FormatValues[F] //FullForm                                              

Out[3]//FullForm= 
 List[RuleDelayed[HoldPattern[MakeBoxes[F[Pattern[x, BlankSequence[]]], 
 Pattern[FormatType, Blank[]]]], 
 Format[HoldForm[Plus[x, F]], FormatType]], 
 RuleDelayed[HoldPattern[Format[F[Pattern[x, BlankSequence[]]]]],   HoldForm[Plus[x, F]]]]
```
We just still are not there yet.

rocky: Is it a simple matter to just add this in Mathics3? 

If not, would you open an issue to cover this for later? (You understand this far better than I).

Otherwise, LGTM.

mmatera: There is a difference on how WMA and Mathics3 handles this:
`Format[expr_]:=rhs_ `
In Mathics3, a format with an empty "Form" in the formats dictionary of the definition object. WMA defines a MakeBoxes rule and the empty rule:
```
{HoldPattern[MakeBoxes["p", FormatType_]] :> Format["p", FormatType],
  HoldPattern[Format["p"]] :> "p"}
``` 
Eventually, I am going to do this change, but I want to finish with the adjustments of the previous PR to go on with it.

PR #1648, opened 2026-01-25
Update doc for `FormBox`

rocky:
Go over `FormBox` user documentation and example.

### DIFF
--- a/mathics/builtin/makeboxes.py
+++ b/mathics/builtin/makeboxes.py
@@ -3,9 +3,10 @@
 Low-level Format definitions
 """
 
+
 from mathics.core.attributes import A_HOLD_ALL_COMPLETE, A_READ_PROTECTED
 from mathics.core.builtin import Builtin, Predefined
-from mathics.format.box import format_element
+from mathics.format.box import eval_generic_makeboxes, format_element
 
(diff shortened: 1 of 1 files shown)


mathics/builtin/box/layout.py
```diff
@@ -86,21 +86,41 @@ def is_constant_list(list):
...
+    r"""<url>
     :WMA link:
     https://reference.wolfram.com/language/ref/FormBox.html</url>
 
     <dl>
-      <dt>'FormBox[boxes, form]'
-      <dd> is a low-level box construct that displays as \
-    boxes and keep information about the form used to generate \
-    the box representation.
+      <dt>'FormBox[$boxes$, $form$]'
```
mmatera, line 94: This does not seem to be the goal of this Box expression. According to the documentation, FormBox gives a way to `ToExpression` to determine how the Box expression must be interpreted (parsed?) to recover the original expression, from which the Box expression comes from. 
Is like `InterpretationBox`, but instead of storing the original expression, just keep the information of what form the boxed expression comes from.
Honestly, I do not fully understand why this structure is used instead of TagForm, but I introduce it for compatibility with WMA.

mmatera:
I rebase in #1653, following the right order of the changes history

PR #1651, opened 2026-01-25
Makeboxes overhault

mmatera:
This is what follows from #1650 and #1652. Lets review it and merge before going over these ones.

The remaining changes inside this PR are not independent of each other.

PR #1652, opened 2026-01-25
Another independent, small part of #1651

mmatera:
* `eval_makeboxes_matrix` -> `eval_makeboxes` in `MatrixForm` to avoid hide the rule.
* fix escape characters in formatting Strings.
* Improve the way in which differences between expected and results are shown when assertions fail in `test.helper`.


rocky:
LGTM

PR #1653, opened 2026-01-25
Rebase #1648

mmatera:
@rocky, I rebase here your proposal in #1648, and reproduce my comment.

### DIFF
--- a/mathics/builtin/box/layout.py
+++ b/mathics/builtin/box/layout.py
@@ -91,14 +91,32 @@ class FormBox(BoxExpression):
     https://reference.wolfram.com/language/ref/FormBox.html</url>
 
     <dl>
-      <dt>'FormBox[$boxes$, $form$]'
-      <dd> is a low-level boxing construct that wraps $boxes$ and $form$ into a box.
+      <dt>'FormBox'[$boxes$, $form$]
+      <dd> is a low-level boxing construct that wraps $boxes$ and $form$ into a box. \
+      'form' must be one of the forms in '$BoxForms' list.
     </dl>
 
     ## No examples because our implementation and understanding of the concept
     ## may be lacking. See https://github.com/Mathics3/mathics-core/pull/1653
     ## for the sordid discussion.
     """
+
+    # FormBox provides a way to tell the interpreter in `ToExpression`
+    # how to interpret the 'boxes' expression to reconstruct
+    # an expression. For example, the box expression
+    # `RowBox[{"Sin", "(","Pi", ")"}]`
+    # is interpreted in `StandardForm` as `Times[Sin, Pi]`.
+    # However, if it is enclosed in `FormBox[..., TraditionalForm]`
+    # it is interpreted as `Sin[Pi]`.
+    #
+    # It also has effect in how the WMA notebook interface renders
+    # the box expression: variables in `TraditionalForm` are shown
+    # in italics, while in other forms are shown in regular a regular
+    # font.
+    # On the other hand, the form does not have any effect on
+    # `ToString`, `MathMLForm` and `TeXForm`, so at the render level,
+    # we can not notice any difference in the currently available
+    # Mathics3 frontends.
 
     attributes = A_PROTECTED | A_READ_PROTECTED
     summary_text = "wrap boxes with an association to a particular form"


mathics/builtin/box/layout.py
```diff
@@ -86,21 +86,41 @@ def is_constant_list(list):
...
     :WMA link:
     https://reference.wolfram.com/language/ref/FormBox.html</url>
 
     <dl>
-      <dt>'FormBox[boxes, form]'
-      <dd> is a low-level box construct that displays as \
-    boxes and keep information about the form used to generate \
-    the box representation.
+      <dt>'FormBox[$boxes$, $form$]'
+      <dd> is a low-level boxing construct that wraps $boxes$ and $form$ into a box. \
```
mmatera, line 95: This does not seem to be the goal of this Box expression. According to the documentation, FormBox gives a way to ToExpression to determine how the Box expression must be interpreted (parsed?) to recover the original expression, from which the Box expression comes from.
Is like InterpretationBox, but instead of storing the original expression, just keep the information of what form the boxed expression comes from.
Honestly, I do not fully understand why this structure is used instead of TagForm, but I introduce it for compatibility with WMA. The only difference with `TagForm` is that the second argument is a the head of a function, and `FormBox` is a symbol.

mathics/builtin/box/layout.py
```diff
@@ -86,21 +86,41 @@ def is_constant_list(list):
...
+      even though the surrounding form may be different.
     </dl>
+
+
+    We can use 'FormBox' to specify the way a particular \
+    box context renders, independent of the encompassing form it might be inside.
+
+    Consider the following 'SuperscriptBox' nested inside 'TraditionalForm'
+
+    >> TraditionalForm[FormBox[SuperscriptBox["x", "2"], InputForm]]
```
mmatera: This is what I get in WMA:
```
In[1]:= TraditionalForm[FormBox[SuperscriptBox["x", "2"], InputForm]]           

Out[1]//TraditionalForm= FormBox[SuperscriptBox[x, 2], InputForm]
```
The first observation is that `TraditionalForm` is not affecting the format. Let's compare with
```
In[5]:= TraditionalForm[F[SuperscriptBox["x", "2"], InputForm]]                 

Out[5]//TraditionalForm= F(SuperscriptBox[x, 2], InputForm)
```
or 

```
In[5]:= TraditionalForm[FormBox[SuperscriptBox["x", "2"], NotAForm]]                 

Out[5]//TraditionalForm= FormBox[SuperscriptBox[x, 2], NotAForm]
```
I guess that the reason is WMA does not try to format (internal) `FormBox` Box-expressions.

mathics/builtin/box/layout.py
rocky: > This does not seem to be the goal of this Box expression. 

What part is not true? 

FormBox wraps "boxes" and "form" into a box, right?  And it is a lower-level boxing construct, i.e., something that is purely about the boxing aspect for notebook-like use.


> According to the documentation, FormBox gives a way to ToExpression to determine how the Box expression must be interpreted (parsed?) to recover the original expression, from which the Box expression comes from.

What is said is that "FormBox affects the interpretation [which I take to mean specifically rendering] of its contents" and the example used involves "ToExpression".  It does not mean that it _only_ involves "ToExpression", but rather that this is the effect in that particular example. 


> Is like InterpretationBox, but instead of storing the original expression, just keep the information of what form the boxed expression comes from. 

The reason the information box is not changed, is so that the outside formatting has access to the box and the form. 

> Honestly, I do not fully understand why this structure is used instead of TagForm, but I introduce it for compatibility with WMA.

TagBox's second parameter is a "tag" which can be anything, whereas in FormBox's second box is a Form which is not anything, but has specific rendering rules. The form is what is used to do the rendering inside that particular box, independent of the enclosing form. 

So, in other words, if you have some text that overall is to be formatted according to one form's principles, but within that you want some particular subpart to go by some other Form rules, use FormBox. 

You might be able to use TagBox as well, but you'd probably have to define how the tag changes formatting. This has already been done when a form is used.


Here is another example. 

The following example creates a superscript (like x**2) but tells Mathics3 to treat it as TraditionalForm (mathematical notation), instead of the default StandardForm (Mathics3 coding notation). 

```mathematica
DisplayForm[
  FormBox[
    SuperscriptBox["x", "2"], 
    TraditionalForm
  ]
]
```

SuperscriptBox["x", "2"] is the raw "box" data representing.  This wrapper tells the Front End, "interpret these boxes specifically as TraditionalForm" even though it is enclosed inside DisplayForm[...].

In TraditionalForm, single-letter variables like "x" are automatically shown in italics, whereas in StandardForm, they are typically plain text. By changing TraditionalForm to StandardForm in the code above, you can see how FormBox directly controls the visual style of the same underlying data.

One example I can think of from days of TeX typseting of your is that you have a language that goes left to right, but inside that you quote a language that goes right-to-left. If you have to break the quoted line that is too long, you want to use right-to-left breaking and hyphenation.

mathics/builtin/box/layout.py
rocky: > The first observation is that TraditionalForm is not affecting the format. 

More precisely, it is not changing what goes on inside FormBox. However, outside of that, it does. To keep things short, I didn't include stuff outside. 

> I guess that the reason is WMA does not try to format (internal) FormBox Box-expressions.

That's a *consequence* of the bigger picture described, but it is not the bigger idea.

mathics/builtin/box/layout.py
mmatera: > > This does not seem to be the goal of this Box expression.
> 
> What part is not true?
> 

The only part I can say *is not true* is the result of the tests, in WMA. What I said is that I do not understand the concept of this symbol to give a deeper description than the one I gave. 

> FormBox wraps "boxes" and "form" into a box, right? And it is a lower-level boxing construct, i.e., something that is purely about the boxing aspect for notebook-like use.
> 
Indeed. 
> 
> What is said is that "FormBox affects the interpretation [which I take to mean specifically rendering] of its contents" and the example used involves "ToExpression". It does not mean that it _only_ involves "ToExpression", but rather that this is the effect in that particular example.
> 
Yes, and apart from `ToExpression` is seems to affect the frontend. 
> 
> The reason the information box is not changed, is so that the outside formatting has access to the box and the form.
> 
OK, but what I see is that the second argument does not affect how the expression is rendered, in the CLI or in the notebook interface.

Image

> TagBox's second parameter is a "tag" which can be anything, whereas in FormBox's second box is a Form which is not anything, but has specific rendering rules. The form is what is used to do the rendering inside that particular box, independent of the enclosing form.
> 

As far as I could see, the second argument only affects the result of `ToExpression`, but not how the expression is rendered. For example, 
```
In[1]:= FormBox[SuperscriptBox["x", "2"], NotAForm[2]]//DisplayForm                                                                                                          

                       2
Out[1]//DisplayForm= x
```
on the other hand, unter `ToExpression` there are differences. In the first place, `ToExpression` does not accept `FormBox` if its second argument is not in `$BoxForms`:
```
In[1]:= FormBox[SuperscriptBox["x", "2"], InputForm]//ToExpression                                                                                                            

MakeExpression::boxfmt: InputForm in MakeExpression[SuperscriptBox[x, 2], InputForm] is not a box formatting type. A box formatting type is any member of $BoxForms.

Syntax::sntxi: Incomplete expression; more input is needed .

ToExpression::esntx: Could not parse FormBox[SuperscriptBox[x, 2], InputForm] as input.

Out[1]= $Failed
```
while in `TagBox` can be anything:
```
In[2]:= TagBox[SuperscriptBox["x", "2"], NotAForm]//ToExpression                                                                                                              
                  2
Out[2]= NotAForm[x ]
```
and in this case, the second argument wraps the result.

If the second argument is in $BoxForms, the result is the same:

```
In[3]:= FormBox[SuperscriptBox["x", "2"], TraditionalForm]//ToExpression                                                                                                      

         2
Out[3]= x

In[4]:= TagBox[SuperscriptBox["x", "2"], TraditionalForm]//ToExpression                                                                                                       

         2
Out[4]= x
```

But in `TagBox`, if the second argument is in `$PrintForms`, the expression is not wrapped either:
```
In[5]:= TagBox[SuperscriptBox["x", "2"], InputForm]//ToExpression                                                                                                             

         2
Out[5]= x
```

> So, in other words, if you have some text that overall is to be formatted according to one form's principles, but within that you want some particular subpart to go by some other Form rules, use FormBox.
> 

This is not something that I am seeing happen

> You might be able to use TagBox as well, but you'd probably have to define how the tag changes formatting. This has already been done when a form is used.
> 
> Here is another example.
> 
> The following example creates a superscript (like x**2) but tells Mathics3 to treat it as TraditionalForm (mathematical notation), instead of the default StandardForm (Mathics3 coding notation). 
> 
> DisplayForm[
>   FormBox[
>     SuperscriptBox["x", "2"], 
>     TraditionalForm
>   ]
> ]
> 
> SuperscriptBox["x", "2"] is the raw "box" data representing. This wrapper tells the Front End, "interpret these boxes specifically as TraditionalForm" even though it is enclosed inside DisplayForm[...].
> 

This is what happen with this expression in the front end:

Image

Notice that even if the second argument is anything, the result is *almost* the same.

> In TraditionalForm, single-letter variables like "x" are automatically shown in italics, whereas in StandardForm, they are typically plain text. By changing TraditionalForm to StandardForm in the code above, you can see how FormBox directly controls the visual style of the same underlying data.
> 
OK, this is something that seems to happen, that I missed the first time: if the second argument is `TraditionalForm`, then it uses italics for the variables, while other symbols don't. 


> One example I can think of from days of TeX typseting of your is that you have a language that goes left to right, but inside that you quote a language that goes right-to-left. If you have to break the quoted line that is too long, you want to use right-to-left breaking and hyphenation.

OK, but again is not something that I am seeing here.

mathics/builtin/box/layout.py
```diff
@@ -86,21 +86,41 @@ def is_constant_list(list):
...
+    the way 'TraditionalForm' handles 'SuperscriptBox':
+
+    >> InputForm[FormBox[SuperscriptBox["x", "2"], TraditionalForm]]
+     = FormBox[SuperscriptBox["x", "2"], TraditionalForm]
+
     """
 
     attributes = A_PROTECTED | A_READ_PROTECTED
-    summary_text = "box with an associated form"
+    summary_text = "wrap boxes with an association to a particular form"
```
mmatera, line 122: This statement is fine

mathics/builtin/box/layout.py
rocky: The topics are getting too long, so let's break them down into smaller parts.

> OK, but what I see is that the second argument does not affect how the expression is rendered, in the CLI or in the notebook interface.

I don't have access to a notebook interface. But in wolframscript I get the output shown in the examples. 

Again: 

```Mathematica
In[4]:= TraditionalForm[FormBox[SuperscriptBox["x", "2"], InputForm]]                                                         

Out[4]//TraditionalForm=   FormBox[SuperscriptBox[x, 2], InputForm]

In[5]:= InputForm[FormBox[SuperscriptBox["x", "2"], TraditionalForm]]                                                         

Out[5]//InputForm= 
FormBox[SuperscriptBox["x", "2"], TraditionalForm]
```
The difference between the two in the Outs is whether "x", and "2" are surrounded in quotes.  In Inputform you wouldn't put quotes around x, while in traditional form, I guess you would. Is that correct?

rocky: > OK, but what I see is that the second argument does not affect how the expression is rendered, in the CLI or in the notebook interface.
> 
> Image

In the CLI, yes, but I guess not in that notebook.

mmatera: In `InputForm`, you put the quotes because that's how you enter a string. In `TraditionalForm`, strings are not enclosed in quotes. This is the standard behavior for any expression.

mmatera: > In the CLI, yes, but I guess not in that notebook.

The only difference I see in the notebook interface is the second argument, but not how the first argument is formatted. Am I missing something?

mathics/builtin/box/layout.py
```diff
@@ -86,21 +86,41 @@ def is_constant_list(list):
...
+      <dt>'FormBox[$boxes$, $form$]'
+      <dd> is a low-level boxing construct that wraps $boxes$ and $form$ into a box. \
+      It indicates to outer encompassing Forms, that $boxes$ \
+      should be rendered using rules of $form$, \
+      even though the surrounding form may be different.
     </dl>
+
+
+    We can use 'FormBox' to specify the way a particular \
+    box context renders, independent of the encompassing form it might be inside.
```
mmatera: If this were true, and I am still not convinced, the example should be a pure box expression -an expression made just of *Box symbols, strings and lists- , like
RowBox[{"F","[",  FormBox[SuperscriptBox["x","2"], TraditionalForm] ,"]"}]

followed by //DisplayForm to force the interpreter to render.

mathics/builtin/box/layout.py
rocky: > > So, in other words, if you have some text that overall is to be formatted according to one form's principles, but within that you want some particular subpart to go by some other Form rules, use FormBox.
> 
> This is not something that I am seeing happen

Well, you do see that the formatting of "x" has changed, italicized or not. And that varies depending on the Form, right?  Is that consistent with what is said? 

I guess what you are observing is that in other parts of the rendering, the form rendering is not respected.

If so, we modify this say that some parts, like variable rendering or strings, are changed. 

> 
> This is what happen with this expression in the front end:
> 
> Image
> Notice that even if the second argument is anything, the result is _almost_ the same.

When there are errors, that's a whole different situation. To start, I think we just stick to the behavior when the input is valid.

mmatera: We are talking about this one:

Image

In this case, I do not see any difference, apart from the symbol in the second parameter.

mmatera: I do see a difference here:

Image

When we ask to render the expression, if the `FormBox` has `TraditionalForm` as its second parameter, the inner variable  `x` is in italics.

mmatera: But it affects how the inner variables are shown, not the outer ones;

Image

mmatera: Notice also that unter `TeXForm` and `MathMLForm`, the tag has no effect:
```
In[1]:= DisplayForm[FormBox[SuperscriptBox["x", "2"], InputForm]] // TeXForm    

Out[1]//TeXForm= x^2

In[2]:= DisplayForm[FormBox[SuperscriptBox["x", "2"], TraditionalForm]] // TeXForm                                                                                            

Out[2]//TeXForm= x^2

```


```
In[3]:= DisplayForm[FormBox[SuperscriptBox["x", "2"], InputForm]] // MathMLForm 

Out[3]//MathMLForm= 
   <math>
    <semantics>
     <msup>
      <mi>x</mi>
      <mn>2</mn>
     </msup>
     <annotation
      encoding='Mathematica'>TagBox[FormBox[SuperscriptBox[&quot;x&quot;,
      &quot;2&quot;], InputForm], DisplayForm]</annotation>
    </semantics>
   </math>

In[4]:= DisplayForm[FormBox[SuperscriptBox["x", "2"], TraditionalForm]] // MathMLForm                                                                                         

Out[4]//MathMLForm= <math>
                     <semantics>
                      <msup>
                       <mi>x</mi>
                       <mn>2</mn>
                      </msup>
                      <annotation encoding='Mathematica'>TagBox[FormBox[SuperscriptBox[&quot;x&quot;, &quot;2&quot;], TraditionalForm], DisplayForm]</annotation>
                     </semantics>
                    </math>

```

mmatera: > When there are errors, that's a whole different situation. To start, I think we just stick to the behavior when the input is valid.

I look at what happens with invalid inputs to see when the input affects the result.

mathics/builtin/box/layout.py
rocky: Looking at https://reference.wolfram.com/language/ref/FormBox.html?q=FormBox

Under basic examples, I see: 

> Create a cell with TraditionalForm style settings applied
> ... 
> Create a cell with StandardForm style settings applied:

This is more or less what I was trying to indicate, substituting "style settings applied" for "rendering".

However, I think the safest thing is to scale back on this description and go with more observable stuff.  I will update this PR soon.

mathics/builtin/box/layout.py
mmatera: Notice that in rendering (DisplayForm), the WMA kernel does not bother to show a warning when the second argument is not what is expected. It just neglects it.

mathics/builtin/box/layout.py
mmatera: Maybe we should leave our findings in this discussion as a large comment after the docstring. Now I think I understand better why these two structures were introduced. However, they would make more sense when we start to do a more serious/sophisticated  implementation of `ToExpression`.

mathics/builtin/box/layout.py
```diff
@@ -92,31 +92,22 @@ class FormBox(BoxExpression):
...
-
-    Compare with the following where we reverse this. The outside form is 'InputForm', \
-    but dictate inside the 'FormBox' that we want the 'SuperscriptBox' portion to get rendered \
-    the way 'TraditionalForm' handles 'SuperscriptBox':
+    Compare with the following where we reverse 'TraditionalForm' and 'InputForm':
 
     >> InputForm[FormBox[SuperscriptBox["x", "2"], TraditionalForm]]
      = FormBox[SuperscriptBox["x", "2"], TraditionalForm]
 
+    The difference is in how the form style settings are applied to the variable $x$.
```
mmatera: The difference here comes from the outer head `InputForm` or `TraditionalForm`, not from the second parameter of `SuperscriptBox`. 

Differences coming from that parameter are not visible in any of our front ends, because all of them are based on pure text, MathML and TeX forms, which are not affected by that parameter.

mathics/builtin/box/layout.py
rocky: I will add a comment with a link to this PR. 

I don't think we've gotten to the bottom of this, and this discussion is pretty rambly.

mathics/builtin/box/layout.py
rocky: Ok. Then I guess we have to remove the example for now. I note that one of the outputs compared to WMA is wrong anyway. See https://github.com/Mathics3/mathics-core/issues/1654

mmatera: I put in the main branch a draft of a comment with the discoveries we made here.

mathics/builtin/box/layout.py
```diff
@@ -86,18 +86,19 @@ def is_constant_list(list):
...
-    the box representation. 'form' must be one of the forms \
-    in '$BoxForms' list.
+      <dt>'FormBox'[$boxes$, $form$]
+      <dd> is a low-level boxing construct that wraps $boxes$ and $form$ into a box. \
+      'form' must be one of the forms in '$BoxForms' list.
     </dl>
+
+    ## No examples because our implementation and understanding of the concept
+    ## may be lacking. See https://github.com/Mathics3/mathics-core/pull/1653
+    ## for the sordid discussion.
```
mmatera, line 101: jajaja

mathics/builtin/box/layout.py
mmatera: I agree, we are far from the bottom of this discussion. And I think it is necessarily rambly because we do not have yet the full context where the discussion makes sense: this kind of box expressions are in the interface with other topics like `ToExpression` and the notebook front end.

PR #1658, opened 2026-01-26
More on escape characters

mmatera:
Some missing characters ASCII in LaTeX conversions. Together with #1621, the documentation is built without critical errors.
Unicode tables does not load these conversions, because the original characters are ASCII. Since it is a very small set, it seems OK to be hard-coded.


rocky:
I suspect there is a more elegant way to do this, but I don't want to spend the time on it right now. 

This is an improvement on what we have. 

LGTM

PR #1661, opened 2026-01-29
MathML tweaks

mmatera:
This PR makes some tweaks on how mathml output is render. 
* adds line breaks following the WMA usage.
* Handle more carefully `InterpretationBox` to work with `InputForm` and `OutputForm`.
* Remove `OutputForm[s_String]` format rule.
* update `format-tests-WMA.yaml` with more tests compared against WS.
* Update  `format-tests.yaml` accordingly.

### DIFF
--- a/mathics/format/render/mathml.py
+++ b/mathics/format/render/mathml.py
@@ -345,10 +345,6 @@ def string(s: String, **options) -> str:
             # Mathics-Django:
             if text == "":
                 return ""
-            if text == named_characters["DifferentialD"]:
-                return render(
-                    '<mo form="prefix" lspace="0.2em" rspace="0">%s</mo>', text
-                )
             if text == named_characters["InvisibleTimes"]:
                 return render(
                     '<mo form="prefix" lspace="0" rspace="0.2em">%s</mo>', text


mathics/format/render/mathml.py
```diff
@@ -53,15 +54,15 @@ def encode_mathml(text: str) -> str:
...
-    "\u00d7",
-    "\u2032",
-    "\u2032\u2032",
-    " ",
-    "\u2062",
-    "\u222b",
-    "\u2146",
+    "\u301a",  # [[
+    "\u301b",  # ]]
+    "\u00d7",  # \[Times]
```
rocky: Why can't this be picked up from character tables?

mmatera: Good question. I have to check if we are not doing already that.

mathics/format/render/mathml.py
```diff
@@ -53,15 +54,15 @@ def encode_mathml(text: str) -> str:
...
-    "\u301a",
-    "\u301b",
-    "\u00d7",
-    "\u2032",
-    "\u2032\u2032",
-    " ",
-    "\u2062",
-    "\u222b",
-    "\u2146",
+    "\u301a",  # [[
```
rocky: Recently, RightDoubleBracket and LeftDoubleBracket were tagged as operators. I suspect that these can now be picked up from the character tables.

mathics/format/render/mathml.py
```diff
@@ -172,7 +188,7 @@ def fractionbox(self, **options) -> str:
     _options = self.box_options.copy()
     _options.update(options)
     options = _options
-    return "<mfrac>%s %s</mfrac>" % (
+    return "<mfrac>\n%s\n %s\n</mfrac>" % (
```
rocky: Is there a typographic difference here, or is this simply to make human comprehension easier when looking at the text?

mmatera: It is just human readability, and helps to compare with WMA output, which split the output in lines. Also, WMA uses indentation for the different levels, which we are not supporting right now. For example
```
In[1]:= a^b/c//MathMLForm                                                       

Out[1]//MathMLForm= <math>
                     <mfrac>
                      <msup>
                       <mi>a</mi>
                       <mi>b</mi>
                      </msup>
                      <mi>c</mi>
                     </mfrac>
                    </math>
```
while in Mathics3

```
In[1]:= a^b/c//MathMLForm
Out[1]//MathMLForm= <mfrac>
                    <msup>
                    <mi>a</mi> 
                    <mi>b</mi>
                    </msup>
                     <mi>c</mi>
                    </mfrac>
```

mmatera: Having the output split in lines, we can add the indentation later, or remove it from the WMA output for doing the comparison.

rocky: > It is just human readability, and helps to compare with WMA output, which split the output in lines. Also, WMA uses indentation for the different levels, which we are not supporting right now. For example
> 
> ```
> In[1]:= a^b/c//MathMLForm                                                       
> 
> Out[1]//MathMLForm= <math>
>                      <mfrac>
>                       <msup>
>                        <mi>a</mi>
>                        <mi>b</mi>
>                       </msup>
>                       <mi>c</mi>
>                      </mfrac>
>                     </math>
> ```
> 
> while in Mathics3
> 
> ```
> In[1]:= a^b/c//MathMLForm
> Out[1]//MathMLForm= <mfrac>
>                     <msup>
>                     <mi>a</mi> 
>                     <mi>b</mi>
>                     </msup>
>                      <mi>c</mi>
>                     </mfrac>
> ```

Ok. Got it. I was not opposed to it, but just wondered about the intention. (BTW, adding a comment to this effect will help others from wondering or thinking about this.)

mmatera: The question is where to leave the comment. Maybe in the head of the file?

rocky: > The question is where to leave the comment. Maybe in the head of the file?

In the docstring for the module is fine. This is something that users and Mathics3 Python developers might want to know, and it is general idea.

rocky: > Having the output split in lines, we can add the indentation later, or remove it from the WMA output for doing the comparison.

This is probably another indication that the problem is probably not being approached from the standpoint of box expression information. Indentation level is just a parameter on a recursive call or a box property. If it is getting done in WMA, I suspect that is a reflection of the fact that it approaches this in the more traditional way.

mathics/format/render/mathml.py
```diff
@@ -102,29 +124,47 @@ def render(format, string):
...
                 )
             return render("<mo>%s</mo>", text)
         elif is_symbol_name(text):
             return render("<mi>%s</mi>", text)
         else:
-            outtext = ""
-            for line in text.split("\n"):
-                outtext += render("<mtext>%s</mtext>", line)
-            return outtext
+            return "".join(
```
rocky: This is where getting the proper indent level would be done.

mathics/format/render/mathml.py
mmatera: I did realize that indentation is something that can be done on the fly. The question is if we want to do it. I had my doubts about adding the line breaks, because -differently from WMA graphics interface- we use mathml form to render output. So, adding complexity and length to the output makes harder to the browser show the output.

On the other hand, adding the line breaks makes easier the comparison between Mathics3 and WMA results. So, I proposed an intermediate format.

rocky: > I did realize that indentation is something that can be done on the fly. The question is if we want to do it. I had my doubts about adding the line breaks, because -differently from WMA graphics interface- we use mathml form to render output. So, adding complexity and length to the output makes harder to the browser show the output.
> 
> On the other hand, adding the line breaks makes easier the comparison between Mathics3 and WMA results. So, I proposed an intermediate format.

It's okay to discuss approaches before coding and PR's.

mmatera: OK, so, thoughts?

rocky: BoxExpression should have an integer "nesting_level" attribute, and that value times the string value of an indent is what prefaces the tags after a "\n". 

```
INDENT_SPACES = "  " # or "\t" or whatever
(f"\n{self.nesting_level * INDENT_SPACES}).join(...)
```

mmatera: OK, but the first thing is that we want to indent the mathml code. A way to achieve that is to use an argument for the render functions tracking the indentation level of the context. If we agree that it a good idea, I can implement it.

rocky: If it is set in a `BoxExpresssion` object, MathML rendering can pick up this attribute using its `self` parameter.

mathics/format/render/mathml.py
```diff
@@ -295,31 +345,32 @@ def render(format, string):
             # Mathics-Django:
             if text == "":
                 return ""
-            if text == "\u2146":
+            if text == "\u2146":  # DifferentialD
```
```suggestion
            if text == named_characters["DifferentialD"]
```

mathics/format/render/mathml.py
```diff
@@ -295,31 +345,32 @@ def render(format, string):
...
                 return render(
                     '<mo form="prefix" lspace="0.2em" rspace="0">%s</mo>', text
                 )
-            if text == "\u2062":
+            if text == "\u2062":  # InvisibleTimes
```
```suggestion
            if text == named_characters["InvisibleTimes"]
```

rocky:
@mmatera What the test expects is ⅆ  (U2146), but what the named character encoded is 𝑑  (U0001D451). 

Notes in the YAML  say that the latter is what we want. 

So do we adjust the test expectation or change to use 2146, which I think we decided looks worse?

mathics/format/render/mathml.py
```diff
@@ -295,31 +345,32 @@ def render(format, string):
             # Mathics-Django:
             if text == "":
                 return ""
-            if text == "\u2146":
+            if text == named_characters["DifferentialD"]:
                 return render(
                     '<mo form="prefix" lspace="0.2em" rspace="0">%s</mo>', text
                 )
```
```suggestion
```

This code does not follow the comment about how this turns into invisible space. When this was `\u2146` that might have been true, but we use for DifferentialD the value U0001D451

mmatera:
Superseed by #1665

mmatera:
> @mmatera What the test expects is ⅆ  (U2146),

I would keep this one (ⅆ). Let's adjust it in another round.

 but what the named character encoded is 𝑑  (U0001D451). 
> 
> Notes in the YAML  say that the latter is what we want. 
> 
> So do we adjust the test expectation or change to use 2146, which I think we decided looks worse? 
> 
>

rocky:
> > @mmatera What the test expects is ⅆ  (U2146),
> 
> I would keep this one (ⅆ). Let's adjust it in another round.

We don't have a named character for this. How does this get into the output then?

mmatera:
> > > @mmatera What the test expects is ⅆ  (U2146),
> > 
> > 
> > I would keep this one (ⅆ). Let's adjust it in another round.
> 
> We don't have a named character for this. How does this get into the output then?

Maybe '\[DifferentialD]' must be associated to 'ⅆ' in Mathics-Scanner.

rocky:
> > > > @mmatera What the test expects is ⅆ  (U2146),
> > > 
> > > 
> > > I would keep this one (ⅆ). Let's adjust it in another round.
> > 
> > 
> > We don't have a named character for this. How does this get into the output then?
> 

As I wrote before: 
> Notes in the YAML say that the latter is what we want.

mmatera:
In that case, OK, let's update the renders to use the `U0001D451`  character.

rocky:
> In that case, OK, let's update the renders to use the `U0001D451` character.

That's already done automatically. That's why we got the test failure.

PR #1662, opened 2026-01-29
Fix error when integrate a non-sympy integrand.

mmatera:
If an expression involves 'String' elements, it fails to be converted to sympy. This produced an error in the implementation of `Integrate`, due to the wrong order in the cases considered. Then, evaluating the expression
```
Integrate[F[x,"p"],{x,a,b}]
```
produced an error. This PR fixes that issue.


rocky:
LGTM

PR #1663, opened 2026-01-29
Improve TeXForm

mmatera:
* Improve conversion from RowBox of function call and parenthesized form to LaTeX : 
   - Use `\left` `\right` when the embraced expression does not contain `^`, `_` or `\`. 
   - Improve the formatting of some bracket characters, like '<|' and '|>'.
* fix typo in documentation.

### DIFF
--- a/mathics/format/render/latex.py
+++ b/mathics/format/render/latex.py
@@ -7,7 +7,7 @@ we may make use of via the mathics-scanner tables.
 LaTeX formatting is usually initiated in Mathics via TeXForm[].
 
 TeXForm in WMA is slightly vague or misleading since the output is
-typically LaTeX rather than Plain TeX. In Mathics, we also assume AMS
+typically LaTeX rather than Plain TeX. In Mathics3, we also assume AMS
 LaTeX or more specifically that we the additional AMS Mathematical
 Symbols exist.
 """
@@ -37,6 +37,7 @@ from mathics.core.convert.op import (
     UNICODE_TO_AMSLATEX,
     UNICODE_TO_LATEX,
     get_latex_operator,
+    named_characters,
 )
 from mathics.core.exceptions import BoxConstructError
 from mathics.core.formatter import (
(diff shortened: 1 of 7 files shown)


mathics/format/render/latex.py
```diff
@@ -376,6 +438,45 @@ def rowbox(self, **options) -> str:
...
+        items[-1],
+    )
+    try:
+        bracket_data = BRACKET_INFO[key]
+    except KeyError:
+        return None
+
+    contain = rowbox_sequence(items[1:-1], **options) if len(items) > 2 else ""
+
+    if any(c in contain for c in ("\\", "^", "_")):
```
rocky: This is extremely fragile and myopic.  Is there a more general principle underlying this? 

I think the situation is that boxes should be noting whether their bounding box is strictly one-dimensional linear or can expand into the two-dimensional space. And that is what dictates whether `\left` and `\right' are needed.

mmatera: One way to do it better would be to add an attribute to each `BoxExpression` giving information about if the expression is *one-dimensional*.  I didn´t do that because
1) most of the cases are properly covered with this
2) the cases which are not covered (expressions with escaped operators) still give expressions that compile in LaTeX
3) It is not too hard to change the condition to something more robust, but it requires more changes here and there. 
For the last point I can do a proposal on the top of this, to make it easy to review.

rocky: > One way to do it better would be to add an attribute to each `BoxExpression` giving information about if the expression is _one-dimensional_. 

That would be fine.

> I didn´t do that because...

The problem is that this represents thinking about the situation in fundamentally the wrong way. And this has to be undone later.   Better, in my opinion, is not to start down the hacky path. We've seen that before. 

Keeping this around as a draft so that the output changes can be compared against is a good idea, though.

mmatera: #1664 can be used instead of checking for characters in the encoded inner LaTeX expression.

mmatera: This should be fixed now...

rocky:
LGTM. 

I suspect there are more opportunities for the use of named characters; the parenthesis routine can be more table-driven,  but that can come after more iterations. 

As is, this is a big improvement over what came before.

PR #1664, opened 2026-01-29
is_multiline property

mmatera:
This property can be used in #1663  instead of looking for specific characters in the LaTeX representation


rocky:
LGTM. 

Spoiler alert - the more general form that will come up is bounding box information.

mmatera:
> LGTM.
> 
> Spoiler alert - the more general form that will come up is bounding box information.

Bounding box information is generated at render time, because it depends on the target form.

PR #1666, opened 2026-01-30
$BoxForms and ParentForm

mmatera:
This PR reimplements and extends $BoxForms to reproduce the behavior in WMA. An example about how it works is presented in its Docstring documention.
Changes:
* Handle `RowBox` expressions without elements (fix bug)
* Add `boxforms` as an attribute of the `Definitions` object.
* `PrintForms` and `OutputForms` constants in `mathics.core.definitions` were renamed to `PRINT_FORMS` and `OUTPUT_FORMS` following Python Standards. 
* `BOX_FORMS` was moved from `mathics.format.boxes.makeboxes` to `mathics.core.definitions`.
* `mathics.format.render` is not loaded from the initialization of `Definitions`, but from `import_and_load_builtins`.
* In `FormBaseClass` a new attribute `in_boxforms` was added to mark that Forms which are in `$BoxForms`. `OutputForm` was marked as `in_printforms`.
* `$BoxForms` was moved from `mathics.builtin.makeboxes` to `mathics.builtin.forms.variables`. Symbol `ParentForm` was added.
*

### DIFF
--- a/.github/workflows/pyodide.yml
+++ b/.github/workflows/pyodide.yml
@@ -55,7 +55,7 @@ jobs:
           pip install "setuptools>=70.0.0" PyYAML click packaging pytest
 
           # We can comment out after next Mathics-Scanner release
-          python -m pip install -e git+https://github.com/Mathics3/mathics-scanner#egg=Mathics-Scanner
+          python -m pip install --no-build-isolation -e git+https://github.com/Mathics3/mathics-scanner#egg=Mathics-Scanner
 
           pip install --no-build-isolation -e .
           make mathics/data/op-tables.json mathics/data/operator-tables.json


rocky:
LGTM - this is all good stuff and foundational.  

Something I noticed in trying the example is that we get a ParentForm error even when the formatting does not require it. 

In particular: 

```mathematica
MakeBoxes[x_Integer, MyBoxForm] := StringJoin[Table["o",{x}]]
AppendTo[$BoxForms, MyBoxForm]
MyBoxForm[3]
```
gives the message about ParentForm not being set, even though `MyBoxForm[3]` should not need that. I wonder if this is something that might be improved in the code. Does WMA mimic this weird behavior?

mmatera:
> LGTM - this is all good stuff and foundational.
> 
> Something I noticed in trying the example is that we get a ParentForm error even when the formatting does not require it.
> 
> In particular:
> 
> ```mathematica
> MakeBoxes[x_Integer, MyBoxForm] := StringJoin[Table["o",{x}]]
> AppendTo[$BoxForms, MyBoxForm]
> MyBoxForm[3]
> ```
> 
> gives the message about ParentForm not being set, even though `MyBoxForm[3]` should need that. I wonder if this is something that might be improved in the code. Does WMA mimic this weird behavior?

Indeed, this is the behavior in WMA. I could not find a good motivation for it, so we can avoid this behavior, by moving the line that makes the check at the point where the ParentForm is required.

rocky:
> > LGTM - this is all good stuff and foundational.
> > Something I noticed in trying the example is that we get a ParentForm error even when the formatting does not require it.
> > In particular:
> > ```mathematica
> > MakeBoxes[x_Integer, MyBoxForm] := StringJoin[Table["o",{x}]]
> > AppendTo[$BoxForms, MyBoxForm]
> > MyBoxForm[3]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > gives the message about ParentForm not being set, even though `MyBoxForm[3]` should need that. I wonder if this is something that might be improved in the code. Does WMA mimic this weird behavior?
> 
> Indeed, this is the behavior in WMA. 

Ok. Thanks for checking, and for the information. Let's keep things WMA compatible.

mathics/builtin/forms/variables.py
```diff
@@ -5,13 +5,147 @@
...
+
+    attributes = A_LOCKED | A_PROTECTED
+    name = "$OutputForms"
+    summary_text = "the list of output forms"
+
+    def evaluate(self, evaluation):
+        return ListExpression(*evaluation.definitions.outputforms)
+
+
+class ParentForm(Builtin):
```
rocky, line 131: Just noticed that ParentForm is not a variable, so it should not be here. Probably should be moved to "Low-level Format Definitions".

mmatera: It is not a variable, but it is strongly related to $BoxForms. I thought it makes sense to have it in the same chapter.

mmatera: It is not a variable, but it is strongly related to $BoxForms. I thought it makes sense to have it in the same chapter.

rocky: So is MakeBoxes. And ParentForm is strongly related to MakeBox. And both are low-level formatting primitives.

PR #1667, opened 2026-01-30
Security/sandbox system commands rebased

rocky:
Revise #1660 Disallow various shell escape and OS/environment queries in a sandboxed environment. Rename SANDBOX variable MATHICS3_SANDBOX

### DIFF
--- a/mathics/builtin/system.py
+++ b/mathics/builtin/system.py
@@ -333,7 +333,7 @@ class Machine(Predefined):
 
 
 class MachineName(Predefined):
-    r"""
+    """
     <url>:WMA link:https://reference.wolfram.com/language/ref/\\$MachineName.html</url>
 
     <dl>
@@ -890,10 +890,8 @@ class UserName(Predefined):
     name = "$UserName"
     summary_text = "get login name of the user that invoked the current session"
 
+    @not_in_sandboxed_environment
     def evaluate(self, evaluation: Evaluation) -> String:
-        if not settings.ENABLE_SYSTEM_COMMANDS:
-            evaluation.message("$UserName", "dis")
-            return SymbolFailed
         try:
             user = os.getlogin()
         except Exception:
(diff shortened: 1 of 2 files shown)


rocky:
@RinZ27 Here's a revised PR. Let me know if this meets your concerns. 

In the previous PR, some commands did not report $Failed, but instead returned empty lists or some sort of value of the type that was supposed to be returned. I think it is better to report $Failed, since errors are caught earlier and do not cascade too much beyond the point where things start to go wrong.

RinZ27:
Thanks for the cleanup, @rocky. The decorator approach is much cleaner and avoids the boilerplate I was adding to each function. I also agree that returning `$Failed` is more idiomatic for blocked system calls than an empty list—it makes the failure explicit. Everything looks solid on my end.

rocky:
> Thanks for the cleanup, @rocky. 

Thanks for noting the problem, working on it, and keeping with it. (Side note: some of the commits and mistakes feel like the kinds of responses I get when I use various AI tools. These things sometimes make these weird, wacky changes tacitly.)


> The decorator approach is much cleaner and avoids the boilerplate I was adding to each function. I also agree that returning `$Failed` is more idiomatic for blocked system calls than an empty list—it makes the failure explicit. Everything looks solid on my end.

Thanks.

We could use help on the Django side to fix this. I'd appreciate it if you would put in PRs for the adjustments. Also, we probably need to alter mathicsscript so that it rejects `!` in Sandboxed mode. 

Lots to do!

rocky:
BTW, I suspect this won't be the last of this, and that there's more to do. Expect more iterations.

RinZ27:
@rocky I'm up for the next steps. I'll dive into `mathics-django` soon to align the sandbox logic there. Restricting `!` in `mathicsscript` when sandboxed is also a great point—I'll work on a PR for that as well.

PR #1669, opened 2026-01-31
In CI Use newer URL form for pip git

rocky:
Tread water: the more things change, the more we have to change to stay just to keep to the level of function we had.

### DIFF
--- a/.github/workflows/pyodide.yml
+++ b/.github/workflows/pyodide.yml
@@ -55,7 +55,7 @@ jobs:
           pip install "setuptools>=70.0.0" PyYAML click packaging pytest
 
           # We can comment out after next Mathics-Scanner release
-          python -m pip install -e git+https://github.com/Mathics3/mathics-scanner#egg=Mathics-Scanner
+          python -m pip install --no-build-isolation -e git+https://github.com/Mathics3/mathics-scanner#egg=Mathics-Scanner
 
           pip install --no-build-isolation -e .
           make mathics/data/op-tables.json mathics/data/operator-tables.json


.github/workflows/pyodide.yml
```diff
@@ -55,7 +55,7 @@ jobs:
           pip install "setuptools>=70.0.0" PyYAML click packaging pytest
 
           # We can comment out after next Mathics-Scanner release
-          python -m pip install --no-build-isolation -e git+https://github.com/Mathics3/mathics-scanner#egg=Mathics-Scanner
+          python -m pip install -e "Mathics-Scanner[full] @ git+https://github.com/Mathics3/mathics-scanner"
```
mmatera: `--no-build-isolation` is no longer necessary?

rocky: > `--no-build-isolation` is no longer necessary?

Does not seem to work. I think it fails the way it used to here: not being able to find PyYAML.

https://github.com/Mathics3/mathics-core/actions/runs/21546476644/job/62088574615

I suspect what's going on is that the recent change to workflows forced Pyodide to not work, and possibly Pyodide folks are busy trying to fix things on their end.   I am seeing a lot of "pyodide not available" messages.

But if you have any suggestions, feel free to commit on top and just continue. I really don't care how it's fixed, just that it be fixed. And if we have to remove this from CI (temporarily), I am ok with that too.

mmatera:
CI for Pyodide steel seems to work in master. Maybe we can just restore it from there by now.

mmatera:
I merged this in #1666 and then restored the Pyodide workflow from master. If it works, we can do here the same.

mmatera:
> I merged this in #1666 and then restored the Pyodide workflow from master. If it works, we can do here the same.

I confirm that it works. So, to make it properly, lets revert here the Pyodide workflow, merge this and then #1666

rocky:
> > I merged this in #1666 and then restored the Pyodide workflow from master. If it works, we can do here the same.
> 
> I confirm that it works. So, to make it properly, lets revert here the Pyodide workflow, merge this and then #1666

Ok. Thanks for the investigation/information. Looks like it will work now. When done, we should do as you say, merge to master this, and then merge to master BoxForms/ParentForm. 

I suspect part of the "working" though is somewhat temporary, and there will be pyodide breakage like there was for non-pyoidide workflows recently. But we deal with that when it comes up.

PR #1672, opened 2026-02-02
Add tests for BoxExpression

mmatera:
This PR includes some tests involving the conversion between BoxExpressions and Expressions. Also, fix an issue in `Expression.sameQ` that prevents a `BoxExpression` from coming from an expression to be properly compared against the original expression.


rocky:
LGTM

PR #1673, opened 2026-02-03
Don't hardcode unicode-symbols

rocky:
Don't hardcode unicode-symbols. Also, op-tables.json -> character-tables.json since that's what it is.

### DIFF
--- a/mathics/format/render/latex.py
+++ b/mathics/format/render/latex.py
@@ -359,9 +359,9 @@ def superscriptbox(self, **options):
 
     sup_string = self.superindex.get_string_value()
     # Handle derivatives
-    if sup_string == named_characters["LeftDoubleBracket"]:
+    if sup_string == named_characters["Prime"]:
         return "%s'" % tex1
-    if sup_string == named_characters["LeftDoubleBracket"] * 2:
+    if sup_string == named_characters["Prime"] * 2:
         return "%s''" % tex1
     base = self.tex_block(tex1, True)
     superidx_to_tex = lookup_conversion_method(self.superindex, "latex")


mathics/format/render/latex.py
```diff
@@ -358,9 +359,9 @@ def superscriptbox(self, **options):
 
     sup_string = self.superindex.get_string_value()
     # Handle derivatives
-    if sup_string == "\u2032":
+    if sup_string == named_characters["LeftDoubleBracket"]:
```
mmatera: This is not `"\[LeftDoubleBracket]"` but ` "\[Prime]"` (the derivative, see comment above)

mathics/format/render/latex.py
```diff
@@ -79,26 +80,26 @@
         "latex_closing_large": r"\right]",
     },
     (
-        String("\u301A"),
-        String("\u301B"),
+        String(named_characters["LeftDoubleBracket"]),
```
mmatera, line 83: This makes a lot of sense, thanks!

mathics/format/render/latex.py
```diff
@@ -358,9 +359,9 @@ def superscriptbox(self, **options):
 
     sup_string = self.superindex.get_string_value()
     # Handle derivatives
-    if sup_string == "\u2032":
+    if sup_string == named_characters["LeftDoubleBracket"]:
         return "%s'" % tex1
-    if sup_string == "\u2032\u2032":
+    if sup_string == named_characters["LeftDoubleBracket"] * 2:
```
mmatera: This is This is not `"\[LeftDoubleBracket]\[LeftDoubleBracket]"` but ` "\[Prime]\[Prime]"` (the derivative, see comment above)

mmatera (approved):
Besides the two wrong names, LGTM

mathics/format/render/latex.py
rocky: @mmatera, thanks for the correction.  Is this correct now?

mmatera: It is, thanks!

PR #1674, opened 2026-02-03
Add Box type annotations to render routines.

rocky:
* Add Box type annotations to render routines
* Remove "_" from non-private classes. Go over render docstrings.
*   Fix type errors found by mypy:
      - Initialize BoxExpression.boxes.
      - Initialize GraphicsBox.boxwidth.
      - Initialize GraphicsBox.boxheight

### DIFF
--- a/mathics/builtin/box/graphics.py
+++ b/mathics/builtin/box/graphics.py
@@ -4,7 +4,7 @@ Boxing Symbols for 2D Graphics
 """
 from abc import ABC
 from math import atan2, cos, degrees, pi, sin
-from typing import Any, Dict, Optional, Tuple
+from typing import Any, Dict, Final, List, Optional, Tuple
 
 from mathics.builtin.box.expression import BoxExpression
 from mathics.builtin.colors.color_directives import (
@@ -51,6 +51,12 @@ class GraphicsElementBox(BoxExpression, ABC):
         self.style = style
         self.opacity = opacity
         self.is_completely_visible = False  # True for axis elements
+
+
+# GraphicsElementBox Builtin class that should not get added as a definition,
+# and therefore not added to to external documentation.
+
+DOES_NOT_ADD_BUILTIN_DEFINITION: Final[List[BoxExpression]] = [GraphicsElementBox]
 
 
 class _Polyline(GraphicsElementBox):
--- a/mathics/builtin/forms/base.py
+++ b/mathics/builtin/forms/base.py
@@ -1,3 +1,5 @@
+from typing import Final, List
+
 import mathics.core.definitions as definitions
 from mathics.core.builtin import Builtin
 from mathics.core.symbols import Symbol
@@ -44,8 +46,7 @@ class FormBaseClass(Builtin):
         return instance
 
 
-# FormBaseClass is a public Builtin class that
-# should not get added as a definition (and therefore not added to
-# to external documentation.
+# FormBaseClass is a Builtin class that should not get added as a
+# definition, and therefore not added to to external documentation.
 
-DOES_NOT_ADD_BUILTIN_DEFINITION = [FormBaseClass]
+DOES_NOT_ADD_BUILTIN_DEFINITION: Final[List[Builtin]] = [FormBaseClass]


rocky:
@mmatera: you asked for changes in #1671 to be split out. 

Here, we just go over type annotations.  This aspect was, in fact, a larger problem than just MathML. But it should come first because, in order have a more informed discussion of MathML, we need a better foundation.

The changes here seem larger than what's really going on: renaming "self" to "box" and adding specific type definitions. (And in doing so, already I am seeing a type failure.)

mmatera:
I am seing now some type errors in master too, with the latest version of mypy (1.19.1). Some of the errors here should be fixed by adding an assertion / condition to narrow the type of the argument. But I guess that in all the render routines, the type of the argument `box` should be `BoxElementMixin`, and then narrow them inside each function.

mmatera:
BTW, thanks for the work of splitting this in parts.

mathics/builtin/box/graphics.py
```diff
@@ -43,7 +43,7 @@
 SymbolRegularPolygonBox = Symbol("RegularPolygonBox")
 
 
-class _GraphicsElementBox(BoxExpression, ABC):
+class GraphicsElementBox(BoxExpression, ABC):
```
mmatera, line 46: I think the underscore was to avoid considering this class as a Builtin to be shown in the documentation.

mmatera: (In WMA this builtin does not exist)

rocky: > I think the underscore was to avoid considering this class as a Builtin to be shown in the documentation.

The _entire_ module is marked no-doc. So this should no longer be a reason for having an underscore. In Python, an underscore typically indicates a private class, and this is not private. If it had an underscore previously, that was the wrong way to solve the problem. But at any rate this problem does not exist.

rocky: > (In WMA this builtin does not exist)

I don't understand what this is getting at. We have many classes in Mathics3 and some built-in function classes that do not exist in WMA.

rocky:
> I am seing now some type errors in master too, with the latest version of mypy (1.19.1). Some of the errors here should be fixed by adding an assertion / condition to narrow the type of the argument. 

Ok. 

> But I guess that in all the render routines, the type of the argument `box` should be `BoxElementMixin`, and then narrow them inside each function.

I don't think "all", but just the top-level generic ones.

mathics/builtin/box/graphics.py
mmatera: > > I think the underscore was to avoid considering this class as a Builtin to be shown in the documentation.
> 
> The _entire_ module is marked no-doc. So this should no longer be a reason for having an underscore. In Python, an underscore typically indicates a private class, and this is not private. If it had an underscore previously, that was the wrong way to solve the problem. But at any rate this problem does not exist.



It should not be loaded as a Built-in symbol. Check what happens in the CLI when you entry

`?BoxGraphicsElement`

The underscore prevents that this class contribute with a definition. But this is a symptom that this module does not belongs to `mathics.builtin.box` module.

mmatera: This is what I get in this branch:

```
In[1]:= ? GraphicsElementBox
Out[1]= box representation for a graphics element
        
        Attributes[GraphicsElementBox] = {Protected, ReadProtected}

```

mmatera:
> I don't think "all", but just the top-level generic ones.

I mean, all the functions which are registered.

mathics/builtin/box/graphics.py
rocky: > It should not be loaded as a Built-in symbol. Check what happens in the CLI when you entry
> 
> `?BoxGraphicsElement`

I think you mean `?GraphicsElementBox`, right? 

> 
> The underscore prevents that this class contribute with a definition. But this is a symptom that this module does not belongs here.

If it gets renamed with the word Box first, I don't think that will get added as a definition, right? I think that's how BoxExpression works, right?

rocky:
> > I don't think "all", but just the top-level generic ones.
> 
> I mean, all the functions which are registered.

This is not the situation for RowBox, and several other boxes like that. `def rowbox(box: RowBox..)`  should always get passed a RowBox.

mathics/builtin/box/graphics.py
mmatera: > > It should not be loaded as a Built-in symbol. Check what happens in the CLI when you entry
> > `?BoxGraphicsElement`
> 
> I think you mean `?GraphicsElementBox`, right?
> 
yep
> > The underscore prevents that this class contribute with a definition. But this is a symptom that this module does not belongs here.
> 
> If it gets renamed with the word Box first, I don't think that will get added as a definition, right? I think that's how BoxExpression works, right?

We was very clever if we did that...

mmatera: As it is, it is loaded. But OK, it is not terrible. We can adjust it later.

rocky: I just looked at the code. What is done to define list `DOES_NOT_ADD_BUILTIN_DEFINITION = [...]` in the module. 

See https://github.com/Mathics3/mathics-core/blob/master/mathics/builtin/forms/base.py#L51

And that is far better than overloading the meaning of underscore at the beginning of a list. I'll add this in there now.

rocky:
> LGTM. Merge when you feel it is ready.

Great! Tomorrow, I'll look over the separation of box attributes, which belong to the box object, from boxing options, which are passed via the `**options` parameter.

After that, we can then start a discussion on the code that needs to compute box information and where and how that should work.

PR #1676, opened 2026-02-04
box attribute versus Box option separation for MathMLForm

rocky:
Box attribute versus Box option separation for MathMLForm.

### DIFF
--- a/mathics/format/render/mathml.py
+++ b/mathics/format/render/mathml.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 """
-Mathics3 Graphics3D box rendering to MathML strings.
+Mathics3 Box rendering to MathML strings.
 
 MathML rendering is usually initiated via MathMLForm[].
 """


rocky:
@mmatera The test_format errors I am seeing are all a result of the difference between whether the `<ms>` tag or the `<mtext>` tag is to be used.  That seems to be determined by the value of `ShowStringCharacters`. 

Where and how this gets set for those particular tests is a mystery to me.  I don't see how the changes make here is changing this value.

Also odd is that we seem to be wrapping MathMLForm inside InputForm in `test/format/test_format.py`. 

The same kinds of changes need to be done to other render files like `text.py` and `latex.py`. If you are up for doing these, great. If not, I will do them tomorrow.

mmatera:
@rocky, the tests reinforce the behavior which makes that quotes appear in Mathics-Django when they should. If you want to do changes, please check that Mathics-Django produce the right output. For example,
imagen

Remember that ```"<ms>x</ms>"``` add quotes to the output, while ```"<mtext>x</mtext>"``` doesn' t.

mmatera:
With these changes, it seems that  information from the outer boxes is not being propagated to the inner boxes. Suppose you have something like
```
StyleBox[RowBox[{"F","[","x","]"}],  ShowStringCharacters->False]
```
`ShowStringCharacters` is not a property of `RowBox`, but must be propagated to each String contained in the RowBox. This is why I think you cannot list all the possible "attributes" explicitly. Also, I still think it is not a good idea to attach these attributes to the objects in the tree structure of BoxExpressions. The alternative is to do something like HTMElements in javascript: all the elements have all the possible CSS attributes, most of them inherited from the top level elements, and many of them not used in rendering the particular object (Ej:  elements has a `justify-content` attribute, used when the element contains text, not an image).

rocky:
> With these changes, it seems that information from the outer boxes is not being propagated to the inner boxes. Suppose you have something like
> 
>```
> StyleBox[RowBox[{"F","[","x","]"}],  ShowStringCharacters->False]
> ```

[Option] information from the outer boxes does propagate to the inner boxes. That's what the `**options` parameter is supposed to do. There may be a bug somewhere, though. I'll check on this later, and that may be what's up with the discrepancy with `ShowStringCharacters.`

> Also, I still think it is not a good idea to attach these attributes to the objects in the tree structure of BoxExpressions.

It's always assuring to learn that your opinion is in agreement with folks who have thought about and studied this, like Donald Knuth (one of the inventors of the attribute grammar idea).  :-)  No slouch, that Donald Knuth!

rocky:
> ShowStringCharacters is not a property of RowBox, but must be propagated to each String contained in the RowBox.

Ok, if that's not happening, as it appears to be the case, then this is a bug. 

> This is why I think you cannot list all the possible "attributes" explicitly

This part doesn't follow. I'll track down what's going wrong here and fix it. With this, hopefully, you'll understand the pattern better.

mathics/format/render/mathml.py
```diff
@@ -2,7 +2,7 @@
 """
 Mathics3 Graphics3D box rendering to MathML strings.
```
mmatera: This line is outdated. Now this module renders all the Box Expressions.

mmatera:
LGTM. @rocky, merge when you think it is ready.

rocky:
I think we merge and then tomorrow do text and possibly latex together, rather than lump this here. 

This will cause merge conflicts in the MathML PRs, but we may as well do this in little steps, so it is clear what's going on and to catch any little issues that might arise.

mmatera:
> This will cause merge conflicts in the MathML PRs, but we may as well do this in little steps, so it is clear what's going on and to catch any little issues that might arise.

I just fixed all these merge conflicts.

PR #1677, opened 2026-02-05
Revise `render/latex.py`  and `render/text.py` for box attributes

rocky:
Revise render/latex.py and render/text.py for box attributes

### DIFF
--- a/mathics/format/render/latex.py
+++ b/mathics/format/render/latex.py
@@ -284,6 +284,7 @@ add_conversion_fn(PaneBox, pane_box)
 
 
 def fractionbox(box: FractionBox, **options) -> str:
+    # Note: values set in `options` take precedence over `box_options`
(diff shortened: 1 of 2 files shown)


rocky:
These changes are now pretty straightforward. 

However, in making the changes, I see that there is a bit of duplication across the form rendering that can be eliminated and simplified. 

We could add this to this PR, but I'm thinking another smallish PR for DRYing the code.

Also in the grand scheme of things, there is removing hard-coding characters, and related to that, whether some of the code that does this can't be directed more by table properties from YAML files. 

@mmatera: your thoughts?

mmatera:
As you prefer. Any of these changes should be too disturbing to the existing prs

rocky:
> As you prefer. Any of these changes should be too disturbing to the existing prs

We will merge an interate then.

PR #1679, opened 2026-02-05
First round of DRYing common render code.

rocky:
Actually, before DRYing, I ran into a little issue here that the `boxes` field always only contains a single box, and calling lookup routines for "boxes" is confusing or misleading. 

So before doing the DRYing, going in small steps, I think it's best to get this out of the way.

### DIFF
--- a/mathics/builtin/box/expression.py
+++ b/mathics/builtin/box/expression.py
@@ -83,7 +83,7 @@ class BoxExpression(BuiltinElement, BoxElementMixin):
 
     def __init(self, *args, **kwargs):
         super().__init(args, kwargs)
-        self.boxes = []
+        self.inner_box = None
 
     def do_format(self, evaluation, format):
         return self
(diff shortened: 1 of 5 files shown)


rocky:
@mmatera: your thoughts?

mmatera:
> @mmatera: your thoughts?

If I get this right, this is just renaming. As far as I can see, I think is OK. LGTM

rocky:
> > @mmatera: your thoughts?
> 
> If I get this right, this is just renaming. As far as I can see, I think is OK. LGTM

There are also some additional slight changes to `mathics.core.formatter` that will assist in DRYing code. And there's a redundant hard-coded character translation table removed. (The original copy should be removed or reduced as well, but that's for another time.

mmatera:
Just a detail: these methods are called in Mathics Django and several of the Mathics3 modules. We need to adjust them accordingly.

rocky:
> Just a detail: these methods are called in Mathics Django and several of the Mathics3 modules. We need to adjust them accordingly.

Noted.

PR #1681, opened 2026-02-05
Start to DRY render code

rocky:
Start to DRY some of the generic box conversion routines, where all that's done is copying a box's `box_option` field into its child options and recursing those child fields. 

Sort methods in text and LaTeX. 

rendering for text, and LaTeX was done. MathML still needs a look-over, and another round might be needed for text and LaTeX. But this PR is big enough already. (overall 20, lines were removed).


rocky:
Thanks. 

A heads up for the next PR. To increase meaningful diffs in the MathML branches, the functions should be sorted by name first. (It is a pity this wasn't done earlier.)

PR #1683, opened 2026-02-06
Remove `boxes_to` deprecation and duplicate TeX mapping tables.

rocky:
* DRY a little more. 
* Remove deprecation message since this doesn't seem to be deprecated. 
* Remove unused TeX-format code that has been moved to the TeX renderer.

### DIFF
--- a/mathics/core/element.py
+++ b/mathics/core/element.py
@@ -417,13 +417,10 @@ class BoxElementMixin(ImmutableValueMixin):
         return box_to_format(self, format, **options)
 
     def boxes_to_mathml(self, **options) -> str:
-        """For compatibility deprecated"""
         return self.box_to_format("mathml", **options)
 
     def boxes_to_tex(self, **options) -> str:
-        """For compatibility deprecated"""
         return self.box_to_format("latex", **options)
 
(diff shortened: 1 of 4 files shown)


rocky:
@mmatera After this PR, we finally move on to removing hard-coded characters. 

Then ... finally ... we can get to MathMLForm indenting (and probably other forms after that).  Keep the faith.

mmatera:
Ok. I am trying to keep updated the other branches.

rocky:
> Ok. I am trying to keep updated the other branches.

One technique that I have used in the past is to just keep a diff around from a working base. (`git diff > tmp/d.diff`).

Then, instead of merging. Apply the diff off of the new base. This is sort of like what a `git rebase` does. However, the other main benefit is that I can use it to understand the essential changes. So starting from scratch isn't so bad. Or the diff can be edited. (which I guess is sort of like `git rebase -i`).

rocky:
> Ok. I am trying to keep updated the other branches.

For some of the PRs, like the pretty-print PR, you'll find that starting from scratch, after first embracing the general idea of working in a syntax-directed tree-attribute-oriented fashion, will lead to better results.

PR #1684, opened 2026-02-07
Remove hard coded latex

rocky:
Try to use YAML table information for named characters more often.

### DIFF
--- a/mathics/builtin/box/compilation.py
+++ b/mathics/builtin/box/compilation.py
@@ -23,6 +23,9 @@ class CompiledCodeBox(BoxExpression):
         self._elements = args
         self.box_options = kwargs
 
+    # FIXME "boxes_to..." should be renamed "box_to..."
+    # since there is only one box element
+
     def boxes_to_text(self, elements=None, **options):
         if elements is None:
             elements = self.elements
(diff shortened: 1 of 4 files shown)


mathics/builtin/box/compilation.py
```diff
@@ -23,6 +23,9 @@ def init(self, *args, **kwargs):
         self._elements = args
         self.box_options = kwargs
 
+    # FIXME "boxes_to..." should be renamed "box_to..."
+    # since there is only one box element
+
```
mmatera: Since the attribute belongs to a BoxMixin, `to_text` (`to_format`) should be enough, isn't it?

rocky:
@mmatera with this. I think we're ready now for MathML with indent or the other PR without the indent. 

I think I now have a good sense of where the render code.  Overall, things have greatly improved, and I think we would be in a good position now to add external Render modules for Matplotlib or drawsvg2. 

Still, I have a gut feeling that there is a bit of room for improvement in code organization, but hopefully it will come over time. 

Previously, I was a bit afraid of adding the next level of complexity that is required by 2D-type boxes and renderers. (I still am a little bit). But I think we are in a better position for adding that now.

PR #1680, opened 2026-02-05
Change Symbol order to have uppercase names before lowercase ones.

hrueter:
Having uppercase names before lowercase ones is the canonical order in WL as defined by the sort function.
https://reference.wolfram.com/language/ref/Sort.html

PR #1685, opened 2026-02-07
Alphabetic canonical order

mmatera:
@hrueter, this is a proposal for the reimplementation #1680, for taking into account the WMA alphabetic ordering. Thanks for noticing this.


hrueter:
Thanks, this fix seems to work as expected.

rocky:
LGTM

rocky:
@hreuter: Thanks for reporting drafting a fix and checking @mmatera: Thanks
for finishing

PR #1686, opened 2026-02-07
Mypy findings fixes

mmatera:
These are some fixes for issues found by mypy 1.17.1.


rocky:
LGTM. That reminds me, we should also update the version of black we use. It caused problems in a recent PR.

PR #1688, opened 2026-02-07
Update black

mmatera:
This PR updates the version of `black` used to normalize the base code.


rocky:
LGTM

Note that we should also update other Mathics3 repositories, which also use(d) 23.12.1, to be in sync with this repository.

PR #1689, opened 2026-02-08
fix box3d render

mmatera:
This fix the issue with mathics server

### DIFF
--- a/mathics/core/formatter.py
+++ b/mathics/core/formatter.py
@@ -91,8 +91,6 @@ def lookup_method(self, format: str) -> Callable:
     if getattr(BoxElementMixin, f"to_{format}") is box_to_method:
         box_to_method = None
     if box_to_method:
-        print("using", box_to_method)
-
         # The elements is not used anywhere.
         def ret_fn(box, elements=None, **opts):
             assert elements is None, "elements parameter is not used anymore."


mathics/core/formatter.py
```diff
@@ -91,9 +91,12 @@ def lookup_method(self, format: str) -> Callable:
     if getattr(BoxElementMixin, f"to_{format}") is box_to_method:
         box_to_method = None
     if box_to_method:
+        print("using", box_to_method)
```
rocky: This should not be here.

mmatera:
```suggestion

```

rocky:
LGTM

rocky:
BTW I am seeing for

```
MathMLForm[Integrate[F[x],{x,a,b}]]
```
image

So it looks like there is a bug in MathMLForm.

mmatera:
> BTW I am seeing for
> 
> ```
> MathMLForm[Integrate[F[x],{x,a,b}]]
> ```
> 
> image
> 
> So it looks like there is a bug in MathMLForm.

I find this issue too: the problem is that we are not able to escape HTML/MATHML code when we pass a string. The same happends if we try to print a string with that form:
imagen

but I am not sure if it is a problem of `to_mathml`, from the mathics-Django formatter function, or the `js` code.

PR #1691, opened 2026-02-12
More args checking

rocky:
Argument testing on more built-in functions.
    
Also, split out test_intfns.py into its module component


rocky:
There is a lot more to do here. But this is all I have time for right now, and it is not trivial.

mmatera (approved):
LGTM

PR #1694, opened 2026-02-12
Handle RuleDelayed in Map over Association expressions

mmatera:
One slight change over #1692. Also, I add a comment about differences with WMA output in the case that the `Association` expression does not have the form `Association[__(Rule|RuleDelayed)]`.

### DIFF
--- a/mathics/builtin/intfns/combinatorial.py
+++ b/mathics/builtin/intfns/combinatorial.py
@@ -117,6 +117,9 @@ class _BooleanDissimilarity(Builtin, ABC):
             counts[(a << 1) + b] += 1
         return self._compute(len(py_u), *counts)
 
+    eval_error = Builtin.generic_argument_error
+    expected_args = 2
+
 
 class _NoBoolVector(Exception):
     pass
@@ -217,8 +220,6 @@ class DiceDissimilarity(_BooleanDissimilarity):
      = 1 / 2
     """
 
-    eval_error = Builtin.generic_argument_error
-    expected_args = 2
     summary_text = "Dice dissimilarity"
 
     def _compute(self, n, c_ff, c_ft, c_tf, c_tt) -> Expression:
@@ -325,6 +326,8 @@ class LucasL(SympyFunction):
     """
 
     attributes = A_LISTABLE | A_NUMERIC_FUNCTION | A_PROTECTED | A_READ_PROTECTED
+    eval_error = Builtin.generic_argument_error
+    expected_args = (1, 2)
 
     sympy_name = "lucas"
     summary_text = "get a Lucas number or polynomial"
@@ -439,6 +442,8 @@ class PolygonalNumber(Builtin):
     """
 
     attributes = A_LISTABLE | A_NUMERIC_FUNCTION | A_PROTECTED | A_READ_PROTECTED
+    eval_error = Builtin.generic_argument_error
+    expected_args = (1, 2)
     rules = {
         "PolygonalNumber[n_Integer]": "PolygonalNumber[3, n]",
         "PolygonalNumber[r_Integer, n_Integer]": "(1/2) n (n (r - 2) - r + 4)",
@@ -592,6 +597,8 @@ class Subsets(Builtin):
         "Subsets[list_ , Pattern[n,_List|All|DirectedInfinity[1]], spec_]": "Take[Subsets[list, n], spec]",
     }
 
+    eval_error = Builtin.generic_argument_error
+    expected_args = range(1, 4)
     summary_text = "list all the subsets"
 
     def eval_list(self, list, evaluation):
--- a/test/builtin/intfns/test_combinatorial.py
+++ b/test/builtin/intfns/test_combinatorial.py
@@ -36,6 +36,28 @@ import pytest
             ["EulerE called with 0 arguments; 1 or 2 arguments are expected."],
             "Euler argument number error",
         ),
+        (
+            "JaccardDissimilarity[]",
+            ["JaccardDissimilarity called with 0 arguments; 2 arguments are expected."],
+            "JaccardDissimilarity argument number error",
+        ),
+        (
+            "LucasL[]",
+            ["LucasL called with 0 arguments; 1 or 2 arguments are expected."],
+            "LucasL argument number error",
+        ),
+        (
+            "PolygonalNumber[]",
+            ["PolygonalNumber called with 0 arguments; 1 or 2 arguments are expected."],
+            "PolygonalNumber argument number error",
+        ),
+        (
+            "Subsets[]",
+            [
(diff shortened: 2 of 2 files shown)


rocky:
LGTM. Thanks for the generalization and clarification.

Li-Xiang-Ideal:
btw now `MapAt` maps on values for `Association[__Rule]` only:
https://github.com/Mathics3/mathics-core/blob/a60d4f791562e98c913f3691b39c904347e7dda5/mathics/eval/functional/apply_fns_to_lists.py#L41-L46

mathics/builtin/functional/apply_fns_to_lists.py
```diff
@@ -192,7 +192,7 @@ def callback(level):
             # Fixing this would require a different implementation of this eval_ method.
             #
 
-            if is_association and level.has_form(("Rule", "RuleDeyaled"), 2):
+            if is_association and level.has_form(("Rule", "RuleDelayed"), 2):
```
rocky, line 194: This mistake wasn't caught in the existing tests. So, probably a test needs to be added here for the new features added.

mmatera: The question is, what should we test? For the full behavior, I have in mind a bunch of tests, that in the current implementation would not pass. The example in the comment is one of them.
To match with the WMA behavior, this method should be rewritten, and probably it would be also be interesting to define a specific class (like `ListExpression`) to store valid `Association` expressions. None of this would be too complicated, but right now I do not have the time to handle it.
What I could do is to list the tests here, or just add the tests to a pytest module, and mark them as xfail.

rocky: > The question is, what should we test? 

So you are saying that nothing in this PR right now benefits a user?  There is promise, though, that one day it will? 

> For the full behavior, I have in mind a bunch of tests, that in the current implementation would not pass. The example in the comment is one of them. To match with the WMA behavior, this method should be rewritten, and probably it would be also be interesting to define a specific class (like `ListExpression`) to store valid `Association` expressions. None of this would be too complicated, but right now I do not have the time to handle it. What I could do is to list the tests here, or just add the tests to a pytest module, and mark them as xfail.

Yes, it is better to add xfail or commented-out tests now while the issue is fresh in your mind rather than do it later. Thanks.

mmatera: > > The question is, what should we test?
> 
> So you are saying that nothing in this PR right now benefits a user? There is promise, though, that one day it will?
> 
> > For the full behavior, I have in mind a bunch of tests, that in the current implementation would not pass. The example in the comment is one of them. To match with the WMA behavior, this method should be rewritten, and probably it would be also be interesting to define a specific class (like `ListExpression`) to store valid `Association` expressions. None of this would be too complicated, but right now I do not have the time to handle it. What I could do is to list the tests here, or just add the tests to a pytest module, and mark them as xfail.
> 
> Yes, it is better to add xfail or commented-out tests now while the issue is fresh in your mind rather than do it later. Thanks.

This PR improves handling of some typical cases, but to cover the most general situation, we need to reimplement this.

test/builtin/list/test_association.py
```diff
@@ -193,3 +193,75 @@ def test_associations_private_doctests(
...
+            "Map[F, Q[a->1, b:>Association[p->3,q->4]],{2}]",
+            None,
+            "Q[F[a]->F[1], F[b]:>F[Association[p->3, q->4]]]",
+            "Acting on a nested association, the inner association is treated as normal.",
+        ),
+        (
+            "Map[F, Association[a->1, b:>Association[p->3,q->4]], {0}]",
+            None,
+            "F[Association[a->1, b:>Association[p->3, q->4]]]",
+            "Special behavior happends at the first level.",
```
rocky: happends -> occurs

test/builtin/list/test_association.py
```diff
@@ -193,3 +193,75 @@ def test_associations_private_doctests(
...
+            "Map[F, Association[a->1, b:>Association[p->3,q->4]]]",
+            None,
+            "Association[a->F[1], b:>F[Association[p->3, q->4]]]",
+            "Acting on a nested association, the inner association is treated as normal.",
+        ),
+        (
+            "Map[F, Association[a->1, b:>Association[p->3,q->4]], {1}]",
+            None,
+            "Association[a->F[1], b:>F[Association[p->3, q->4]]]",
+            "Special behavior happends at the first level.",
```
rocky: happends -> occurs

test/builtin/list/test_association.py
```diff
@@ -193,3 +193,75 @@ def test_associations_private_doctests(
...
+            "Map[F, Association[a->1, b:>Association[p->3,q->4]], {2}]",
+            None,
+            "Association[a->1, b:>Association[F[p->3],F[q->4]]]",
+            "Special behavior happends at the first level.",
+        ),
+        (
+            "Map[F, Association[a->1, b:>Q[p->3, q->4]], {2}]",
+            None,
+            "Association[a->1, b:>Q[F[p->3],F[q->4]]]",
+            "Special behavior happends at the first level.",
```
rocky: happends -> occurs

test/builtin/list/test_association.py
mmatera: Thanks!

mmatera:
> btw now `MapAt` maps on values for `Association[__Rule]` only:
> 
> https://github.com/Mathics3/mathics-core/blob/a60d4f791562e98c913f3691b39c904347e7dda5/mathics/eval/functional/apply_fns_to_lists.py#L41-L46

thanks for the observation!

mathics/builtin/functional/apply_fns_to_lists.py
rocky: > > > For the full behavior, I have in mind a bunch of tests, that in the current implementation would not pass. The example in the comment is one of them. To match with the WMA behavior, this method should be rewritten, and probably it would be also be interesting to define a specific class (like `ListExpression`) to store valid `Association` expressions. None of this would be too complicated, but right now I do not have the time to handle it. What I could do is to list the tests here, or just add the tests to a pytest module, and mark them as xfail.
> > 
> > 
> > Yes, it is better to add xfail or commented-out tests now while the issue is fresh in your mind rather than do it later. Thanks.
> 
> This PR improves handling of some typical cases, but to cover the most general situation, we need to reimplement this.

It occurs to me that it might be good to open an issue with this information, so we don't forget after this PR is merged. Thanks.

mmatera: I just opened a related issue #1696, with the basic content of this discussion.

rocky:
Let's merge. If there are other changes that can be done in another PR.

PR #1695, opened 2026-02-12
Arg checking for `mathics.builtin.intfns.combinatorial`

rocky:
Finish argument checking for built-in functions in `mathics.builtin.intfns.combinatorial`.

### DIFF
--- a/mathics/builtin/intfns/combinatorial.py
+++ b/mathics/builtin/intfns/combinatorial.py
@@ -117,6 +117,9 @@ class _BooleanDissimilarity(Builtin, ABC):
             counts[(a << 1) + b] += 1
         return self._compute(len(py_u), *counts)
 
+    eval_error = Builtin.generic_argument_error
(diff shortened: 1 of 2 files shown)


mathics/builtin/intfns/combinatorial.py
```diff
@@ -325,6 +326,8 @@ class LucasL(SympyFunction):
     """
 
     attributes = A_LISTABLE | A_NUMERIC_FUNCTION | A_PROTECTED | A_READ_PROTECTED
+    eval_error = Builtin.generic_argument_error
```
rocky, line 329: Side note. I considered adding `eval_error` in the class this is in, here `SymPyFunction`.

However, for that to work, because (I think) there are default values, we would have to correct all functions in the class to have their proper number of arguments listed. 

So DRYing this is left until all functions have been revised to include their argument check values.

mmatera:
LGTM

PR #1700, opened 2026-02-20
Add unicode box characters

### LINKED ISSUE #1622: Escape sequences in string parsing
## Description

Looking at #1619, I found a slight difference on how WMA and Mathics parse strings with escaped characters.

## How to Reproduce

In CLI:

```
In[1]:="\\\` " //InputForm
```

## Output Given

In Mathics3
```
Out[1]:="\\\\` " //InputForm
```

## Expected behavior

In WMA,
```
Out[1]:="\\\` " //InputForm
```
The issue manifests when using `StringForm`, because we cannot put a backslash before a placeholder:
```
In[1]:=StringForm["C:\\`1`", "root"]
```
In WMA, we get
```
Out[1]:= C:\root
```
but in Mathics we get an error:
```
StringForm::sfq: Unmatched backquote in C:\`1`.
Out[1]= C:\`1`
```

## Workarounds

The workaroung is to put the backslash in a parameter:
```
In[2]:=StringForm["C:`2``1`", "root", "\\"]
Out[2]= C:\root
```

In other contexts, it is enough to avoid escape the backquote character.

```
In[1]:= "\\`"
```

## Priority

Low

## Additional information

This is how WMA and Mathics process similar cases:

WMA
```
In[1]:=  cases = {"\\", "\ ", "\\ ", "\\\ ",                                    
  "`", "\` ", "\\` ", "\\\` "}                                                  

Out[1]= {\,  , \ , \ , `, \` , \` , \\` }

In[2]:= Table[{case, case // InputForm, StringLength[case],                     
   StringTake[case, {1}] }, {case, cases}] // TableForm                         

Out[2]//TableForm= \      "\\"      1   \

                          " "       1    

                   \      "\\ "     2   \

                   \      "\\ "     2   \

                   `      "`"       1   `

                   \`     "\` "     2   \`

                   \`     "\\` "    3   \

                   \\`    "\\\` "   3   \

```

Mathics
```
In[1]:=  cases = {"\\", "\ ", "\\ ", "\\\ ",
          "`", "\` ", "\\` ", "\\\` "}
Out[1]= {\,  , \ , \ , `, \` , \` , \\` }

In[2]:= Table[{case, case // InputForm, StringLength[case], 
           StringTake[case, {1}] }, {case, cases}] // TableForm
Out[2]//TableForm= \      "\\"       1   \
                   
                          " "        1    
                   
                   \      "\\ "      2   \
                   
                   \      "\\ "      2   \
                   
                   `      "`"        1   `
                   
                   \`     "\\` "     3   \
                   
                   \`     "\\` "     3   \
                   
                   \\`    "\\\\` "   4   \

```

rocky:
Handle boxing escape characters inside a string, in built-in functions `Characters`, `StringLength`, `StringTake`, and `ToCharacterCode`.

Tracks API changes in MathicsScanner, <strike>so must be merged not before https://github.com/Mathics3/mathics-scanner/pull/152</strike>.

Fixes #1622

### DIFF
--- a/mathics/format/form/outputform.py
+++ b/mathics/format/form/outputform.py
@@ -865,8 +865,15 @@ def stringform_render_output_form(
     curr_indx = 0
     strform_str = safe_backquotes(replace_box_unicode_with_ascii(strform.value))
     parts = strform_str.split("`")
-    # Rocky: This looks like a hack to me: is it needed?
+
+    # Rocky: This looks like a hack to me. It is used right now
+    # to use allow 'Backquote' to be to escaped with a backslash:
+    # >> StringForm["`` is Global\\`a", a]
+    #  = a is Global`a
+    # There is probably needs to be another change to the scanner, and/or
+    # there is something deeper going on and a change to StringForm.
     parts = [part.replace("\\[RawBackquote]", "`") for part in parts]
+
     result = [parts[0]]
     if len(parts) <= 1:
         return result[0]
--- a/mathics/eval/strings.py
+++ b/mathics/eval/strings.py
@@ -169,10 +169,6 @@ def eval_StringForm_MakeBoxes(strform: String, items, form, evaluation: Evaluati
     strform_str = safe_backquotes(replace_box_unicode_with_ascii(strform.value))
 
     parts = strform_str.split("`")
-
-    # Rocky: This looks like a hack to me: is it needed?
-    parts = [part.replace(r"\[RawBackquote]", "`") for part in parts]
-
     result = [String(parts[0])]
     if len(parts) <= 1:
         return result[0]


mathics/format/form/outputform.py
```diff
@@ -861,8 +863,9 @@ def stringform_render_output_form(
     items = [render_output_form(item, evaluation, **kwargs) for item in items]
 
     curr_indx = 0
-    strform_str = safe_backquotes(strform.value)
+    strform_str = safe_backquotes(replace_box_unicode_with_ascii(strform.value))
     parts = strform_str.split("`")
+    # Rocky: This looks like a hack to me: is it needed?
```
mmatera: Indeed, it is a hack, to allow escape from the backquote. Maybe with the last changes in the scanner, it is no longer needed.

test/builtin/test_strings.py
```diff
@@ -219,30 +219,25 @@
...
-        (
-            'StringTake["abc", {0, 0}]',
-            ('Cannot take positions 0 through 0 in "abc".',),
-            "StringTake[abc, {0, 0}]",
-            None,
-        ),
+        # These tests are commented out due to the bug reported in issue #906
+        # Octal and hexadecimal notation works alone, but fails
+        # as a part of another expression. For example,
+        # F[\.78\.79\.7A]   or "\.78\.79\.7A" produces a syntax error in Mathics.
```
mmatera: <strike>I do not see this fail in master, and the previous version of Mathics-Scanner. Maybe this is a new bug?</strike> Now all these tests passes without problem.

test/builtin/test_strings.py
```diff
@@ -219,30 +219,25 @@
...
-        ('StringTake["abcd", 0] // InputForm', None, '""', None),
-        ('StringTake["abcd", {3, 2}] // InputForm', None, '""', None),
-        ('StringTake["", {1, 0}] // InputForm', None, '""', None),
-        (
-            'StringTake["abc", {0, 0}]',
-            ('Cannot take positions 0 through 0 in "abc".',),
-            "StringTake[abc, {0, 0}]",
-            None,
-        ),
+        # These tests are commented out due to the bug reported in issue #906
```
mmatera: I checked with this version, and the commented tests seem to pass without problems

test/builtin/test_strings.py
rocky: Thanks for checking. Tests reinstated now.

mathics/format/form/outputform.py
rocky: Looks like we still need in order to make: 

```
    >> StringForm["`` is Global\\`a", a]
     = a is Global`a
```

work. It might require a further change in the scanner. Or maybe there is something deeper here; a better change to `StringForm` is needed?

Either way, it's something that requires more thought, and since this stuff is already too large for a single change, I'd like to get this down and come back to this.

test/builtin/test_strings.py
rocky: Thanks for investigating and reporting back. Reinstated now.

mathics/format/form/outputform.py
mmatera: Thanks for check this. We can look again in another round

rocky: > Thanks for check this. We can look again in another round

Teamwork!

rocky:
I'd like to mention something related to this change. 

In the past, there was a thought that _all_ operators had to be represented by Unicode symbols.  (This kind of thing was last done circa tokenized Basic in the mid 1970s.)

This is not that. Here, we are doing this _only_ for Box Expression "operators"  (where grouping is also treated as an "operator"). And even then, *only inside strings*. 

To my mind, the thing that justifies this is the behavior of `StringTake`, `Characters`, `StringLength`, and `ToCharacters.` Especially `ToCharacters`, where the specific Unicode values are explicit.

Exactly *why* it is that these functions work in this behavior is not readily apparent to me.  I am sure there was _some_ reason for this. But still, it feels like more of a *hack* than a fundamental principle. 

And even here, WMA was, I think, careful to restrict this Unicode character idea to the private-use section of Unicode.

PR #1705, opened 2026-03-02
Add Graphics symbols and Graphics Option symbols

rocky:
Copy Graphics symbols (currently only found in `mathics-m3d/m3d/sym.py` into core.


mmatera (approved):
LGTM

PR #1707, opened 2026-03-03
Get named-character JSON from scanner rather than create a local copy.

rocky:
Get named-character JSON from scanner rather than create a local copy.


mmatera:
LGTM

PR #1709, opened 2026-03-03
Combine How to reproduce and output given sections

rocky:
Combine "How to Reproduce" and "Output given" sections


mmatera:
Makes sense! LGTM

PR #1710, opened 2026-03-03
Revise Polynomial functions like `Expand` and `Apart`

rocky:
* Handle inner recursive function closure of `Expand` and `ExpandAll` properly.
* Break out trig/hypertrig and relational functions.
* Separate eval code of `Apart` add sympy tracing, and have it handle relational operators


mmatera:
LGTM

Issue #1655, opened 2026-01-25
Builtin `Begin` not adding variable in context and `End` leaves traces of variables defined by `Begin[]`


rocky:
## Description

`Begin` is not adding context, and `End` might not be removing variables added. 

## How to Reproduce and Output Given

```mathematica
In[1]:= ??f
Out[1]= Global`f

In[2]:= Begin["MyContext`"]
Out[2]= "MyContext`"

In[3]:= f[x_] := x^2

In[4]:= ??f
Out[4]= Global`f  (* Wrong. Should be MyContext` *)       
        f[x_] = x^2

In[5]:= End[]
Out[5]= "MyContext`"

In[6]:= ??f
Out[6]= Global`f (* Wrong. Should have been removed *)
        
        f[MyContext`x_] = MyContext`x^2
```

## Additional context

Noticed while trying to understand why certain Mathematica packages are failing.

mmatera:
The problem seems to be in `Information`: it defines the symbol, which hides the `MyContext` definition. In WMA:
```
In[1]:= ?? f                                                                    

Out[1]= Missing[UnknownSymbol, f]

In[2]:= Begin["MyContext`"]                                                     

Out[2]= MyContext`

In[3]:= f[x_]=2 x                                                               

Out[3]= 2 x

In[4]:= End[]                                                                   

Out[4]= MyContext`

In[5]:= ?? f                                                                    

Out[5]= Missing[UnknownSymbol, f]

In[6]:= ?? MyContext`f                                                          

Out[6]= MyContext`f
        MyContext`f[MyContext`x_] = 2*MyContext`x
         

In[7]:= f[z_]:=3                                                                

In[8]:= ?? f                                                                    

Out[8]= Global`f
        f[z_] := 3
```
I am planning to fix this over the weekend.

mmatera:
@rocky, if I am right, what is reported here is not an issue of the `Begin`/`BeginPackage` implementations, but of how `?? ...` is parsed as an expression.

rocky:
> [@rocky](https://github.com/rocky), if I am right, what is reported here is not an issue of the `Begin`/`BeginPackage` implementations, but of how `?? ...` is parsed as an expression.

@mmatera `??` parsing is not the essential problem reported here, but instead a side effect of how I was conveying the problem. 

```mathematica
In[1]:= ?? f                                                                    
Out[4]= Global`f
```

[...]

Issue #1713, opened 2026-03-07
`? symb` and  `?? symb` should be parsed as `Information["symb"]` and `Information["symb", LongForm->True]` respectively


mmatera:
## Description

This issue is related to #1655. In WMA, 
```
In[1]:= ? A//FullForm                                                                                                                                                         

Out[1]//FullForm= Missing["UnknownSymbol", "A"]

In[2]:= ?? A//FullForm                                                                                                                                                        

Out[2]//FullForm= Missing["UnknownSymbol", "A"]
```
while in Mathics3
```
In[1]:= ?A//FullForm
Out[1]//FullForm= Information[A, Rule[LongForm, False]]

In[2]:= ?? A//FullForm
Out[2]//FullForm= Information[A, Rule[LongForm, True]]
```
As a consequence, in WMA  `?? A`  does not create a definition associated to the symbol `A`, `A` is not evaluated:
```In[1]:= ?? A                                                                    

Out[1]= Missing[UnknownSymbol, A]

In[2]:= ?? A                                                                    

Out[2]= Missing[UnknownSymbol, A]
```

In Mathics3, on the other hand,
```
In[1]:= ?? A
Out[1]= Global`A
```
This is because in the evaluation of `Information[A, LongForm->True]`, a definition for `A` is created.

## Additional context

This behavior is behind the issue #1655: When information for a symbol is request, the symbol is created in the current context, and can hide the symbol created inside a  `Begin[]...End[]` or `BeginPackage[]...EndPackage[]`, which is what is reported in that issue.

mmatera:
Looking at this example:
```
In[21]:= s ??   ad@`sa$sd asdj*21    +   a  //Hold//FullForm                                                                                                                            

Out[21]//FullForm= Hold[Times[s, Information["ad@`sa$sd asdj*21", Rule[LongForm, True]], Plus[a]]]

```

It seems that after `?` or `??`, a sequence that matches the pattern r"[A-Za-z0-9$@`* ]+" (or letter-like characters like `\[Alpha]`)  must be parsed as a string, and used as the first element of  `Information[]`.

rocky:
> Looking at this example:
> 
> ```
> In[21]:= s ??   ad@`sa$sd asdj*21    +   a  //Hold//FullForm                                                                                                                            
> 
> Out[21]//FullForm= Hold[Times[s, Information["ad@`sa$sd asdj*21", Rule[LongForm, True]], Plus[a]]]
> ```
> 
[...]

PR #1714, opened 2026-03-07
`Information` builtin improvements. Partial fix for #1713

mmatera:
This PR implements a partial fix for #1713. 
* Now, `?? symb` is parsed as `Information["symb", LongForm->True]`
* `?? "Plu*"` gives the list of symbols matching the pattern, as expected.

Still, this can not hanble
```
?? Plu*
``` 
which goes for another round.

### DIFF
--- a/mathics/core/parser/parser.py
+++ b/mathics/core/parser/parser.py
@@ -176,6 +176,14 @@ class Parser:
         """
         Parse the single top-level or "start" expression.
         This is called right after doing parse setup.
+
+        In scanning tokens, the scanning mode might get altered on
+        seeing specific tokens inside the tokenizer.
+
+        In particular, in:
+           << symbol
+        the tokenizer in tokenizing "<<" changes the scanning mode for how
+        symbol gets parsed.
         """
         result = []
         while self.next().tag != "END":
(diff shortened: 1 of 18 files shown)


mmatera:
Also, this should fix #1655.

rocky:
This is conceptually not right. We should have a parse routine for parsing a string pattern as described in [Names](https://reference.wolfram.com/language/ref/Names.html) and mentioned in discussion #1713

mmatera:
> parse

I didn´t touch the part that processes this in `Information`. <strike>But since this seems to give the expected behavior, either we are using the same function used in `Names`, or we can DRY it.</strike>

Both methods uses `definitions.get_matching_names(symbol_pat)` to convert a pattern string into a list of symbol names.

mathics/core/parser/parser.py
```diff
@@ -1151,8 +1151,19 @@ def p_Increment(self, _: Token) -> Node:
...
         q = prefix_operators["Information"]
+
+        # This is not completely right:
+        # this token should consume any alphanumeric character
+        # sequence which could match with a symbol, but also character-like,
+        # spaces, `@`, `$` or `*`.
+        #
+        # See issue #1713
+
         child = self.parse_expr(q)
```
rocky: We shouldn't be calling `self.parse_expr(q)`.  Instead, we are looking for something that matches a pattern as found in the string form of Names.

rocky:
@mmatera I will put in a PR on top of this to try to handle a `Names` pattern in a couple of days. Although sometimes it is good to handle a small piece and then do more in another PR, here, I am concerned about the two ideas being intertwined. So I'd like to see what happens when we flesh out more of the full job. 

If it turns out to be too much of a burden, then we can fall back and commit this piece and then commit the next piece.

mathics/core/parser/parser.py
mmatera: @rocky, This is in the line of the comment I put just above. Thanks for looking at it.

test/core/parser/test_parser.py
```diff
@@ -634,12 +634,12 @@ def testInequality(self):
...
-        self.check("??a", "Information[a, LongForm -> True]")
-        self.check("a ?? b", "a Information[b, LongForm -> True]")
+        self.check("??a", 'Information["a", LongForm -> True]')
+        self.check("a ?? b", 'a Information["b", LongForm -> True]')
         self.check("a ?? + b", 'Times[a, Missing["UnknownSymbol", Plus[b]]]')
-        self.check("a + ?? b", "a + Information[b, LongForm -> True]")
-        self.check("??a + b", "Information[a, LongForm -> True] + b")
-        self.check("??a * b", "Information[a, Rule[LongForm, True]]*b")
+        self.check("a + ?? b", 'a + Information["b", LongForm -> True]')
+        self.check("??a + b", 'Information["a", LongForm -> True] + b')
```
rocky: A test like this is weird. First, it makes no sense. Second, this is not what I see in WMA. Instead, I see: 

```
In[1]:= ??a + b                                                         

Out[1]= b Missing[UnknownSymbol, a]

```

rocky:
Closing by merging #1715 into master directly. #1715 was based on#1714 initially, so there should be no loss.

PR #1715, opened 2026-03-08
Handle Names wildcards in the prefix form of `??` and `?` ('Information')

rocky:
Basic support for name wildcards (`*` and `@`) as the operand of the prefix operator form of `Information` (`??` and `?`).

Also, correct some bugs in detecting and handling name wildcards.

### DIFF
--- a/mathics/builtin/atomic/symbols.py
+++ b/mathics/builtin/atomic/symbols.py
@@ -651,6 +651,8 @@ class Symbol_(Builtin):
     """
 
     attributes = A_LOCKED | A_PROTECTED
+    eval_error = Builtin.generic_argument_error
+    expected_args = 1
 
     messages = {
         "symname": (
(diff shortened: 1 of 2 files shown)


mathics/core/parser/parser.py
```diff
@@ -572,12 +580,52 @@ def parse_expr(self, precedence: int) -> Optional[Node]:
...
         """Parse a "p_"-tagged expression.
         "p_" tags include prefix operators, left-bracketed expressions
         and tokens that can be identified by some prefix, like a number
         or a string.
+
+        In scanning tokens, the scanning mode might get altered on
+        seeing specific tokens inside the tokenizer.
+
+        In particular, in:
+           ?? symbol
```
mmatera: I bet this is not really necessary: the tokenizer knows that `?` with a previous argument should be handled as  `Question`, and that after `Question`, it comes a String or a  NamePattern string, as well as how to segmentate both cases.

rocky: > I bet this is not really necessary: the tokenizer knows that `?` with a previous argument should be handled as `Question`, and that after `Question`, it comes a String or a NamePattern string, as well as how to segmentate both cases.

You are correct. This was a remnant of an earlier version of the code before adding `parse_name_pattern`.

The comment has been removed. Thanks for catching.

PR #1717, opened 2026-03-10
Missing from a String should not produce a Symbol definition.

mmatera:
This PR completes what was missing to fix the wrong behavior of `Information` reported in #1655, and refined in #1713. 
What was still wrong in the current master branch is that when a String is passed to `Information`, and the string is not a name pattern, the standard routine for `Symbol` is used, creating the definition. As a result, in the example shown in https://github.com/Mathics3/mathics-core/issues/1655#issuecomment-4024136105 , the first line creates the definition that hides the effect of the `Begin/End` block.


rocky:
LGTM

PR #1719, opened 2026-03-10
Begin package with two parms

mmatera:
This PR implements the support for `BeginPackage` with its second argument.


rocky:
LGTM and thanks!

PR #1721, opened 2026-03-11
Revise Quotient[] builtin

rocky:
Revise `Quotient` so it handles the 3-argument form, lists, and numbers other than Integers.

(Noticed in trying to run some standard WMA packages.)

### DIFF
--- a/mathics/builtin/intfns/divlike.py
+++ b/mathics/builtin/intfns/divlike.py
@@ -314,10 +314,9 @@ class Quotient(Builtin):
       'Round[(m-d)/n]'.
     </dl>
 
-    ## Plot showing the step-like 'Floor' behavior of 'Quotient':
-    ##
-    ## >> DiscretePlot[Quotient[n, 5], {n, 30}]
-    ##  = -Graphics-
+    Plot showing the step-like 'Floor' behavior of 'Quotient':
+     >> DiscretePlot[Quotient[n, 5], {n, 30}]
+      = -Graphics-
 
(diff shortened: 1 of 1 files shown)


mathics/builtin/intfns/divlike.py
```diff
@@ -294,34 +300,82 @@ def eval(self, a: Integer, b: Integer, m: Integer, evaluation: Evaluation):
...
-      <dd>computes the integer quotient of $m$ and $n$.
+      <dd>computes the integer quotient of $m$ and $n$. For non-complex numbers, this \
+      equivalent to 'Floor[m/n]'. When a complex number is involved, it is 'Round[m/n]'.
+      <dt>'Quotient[m, n, d]'
+      <dd>computes the integer quotient of $m-d$ and $n$. For non-complex numbers, this \
+      is equivalent to 'Floor[(m-d)/n]'. When a complex number is involved, it is \
+      'Round[(m-d)/n]'.
     </dl>
 
+    ## Plot showing the step-like 'Floor' behavior of 'Quotient':
```
rocky: There is a bad bug that recently popped up involving the interaction of args checking and things that wrap arguments in a sequence, like DiscretePlot, which uses compile_function_quiet.

mmatera: @rocky, would you put an issue with this example, to look at it later?

rocky: Sure. I'm not sure exactly what's up, but this seems like a nasty bug.

Issue #1283, opened 2025-01-10
Better Error checking on Inequality builtin functions `<`, `Between`


rocky:
`Between` should check that there are two items in a range.  
Comparison in `<` and others should give an error if compared to a complex value. 

In general, the error checking of `mathics.core.testing_expressions.equality_inequality` should be improved.

PR #1724, opened 2026-03-11
Fix expected args checking bug

### LINKED ISSUE #1722: Generic error argument getting called in symbolic evaluations?
## Description

I am seeing argument errors when trying to evaluate built-in functions that use `generic_argument_error` in a symbolic context.

```mathematica

In[1]:= DiscretePlot[Quotient[n, 5], {n, 30}]
Quotient::argt: Quotient called with 2 arguments; 2 or 3 arguments are expected.
Quotient::argt: Quotient called with 2 arguments; 2 or 3 arguments are expected.
Out[1]= -Graphics-
```

A simpler example may be using `N[Quotient[n, 5]`.

## Expected behavior

No error. Interestingly, the result produced is fine. We just get these erroneous argument check messages.

## Workarounds

Ignore the messages. The result is correct in any event.

## Priority

While it does not stop evaluations from occurring, the bug seems pervasive, annoying, and confusing to someone not familiar with the bug (which is almost everyone).

So this should be addressed soonish.

## Additional context

Noticed in trying standard Mathematica packages and `DiscretePlot[Quotient[n, 5], {n, 30}]`.

rocky:
Fix expected args checking bug. This routine is triggered by (implicit) function pattern match. However, it can get called when the arguments are correct. So we need to test for this. 

Fixes #1722

### DIFF
--- a/mathics/core/builtin.py
+++ b/mathics/core/builtin.py
@@ -163,7 +163,7 @@ class Builtin:
              ...
     ```
 
-    the options are stored as a dictionary in the last parameter. For
+    The options are stored as a dictionary in the last parameter. For
     example, if the rule is applied to ``F[x, Method->Automatic]`` the
     expression is replaced by the output of ``eval_with_options(x,
     evaluation, {"System`Method": Symbol("Automatic")})
@@ -181,7 +181,7 @@ class Builtin:
 
     Notice that for creating a Builtin, we must pass to the
     constructor the option ``expression=False``. Otherwise, an
-    Expression object is created, with the ``Symbol`` associated to
+    Expression object is created, with the ``Symbol`` associated with
     the definition as the ``Head``.  For example,
 
     ```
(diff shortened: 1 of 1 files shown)


rocky:
This bug is sufficiently blocking other issues #1283 and #1725 that I'm merging in.

mathics/core/builtin.py
```diff
@@ -445,6 +445,12 @@ def contextify_form_name(f):
     # arguments expected) It assumes each builtin defines
     # "expected_args" for the correct number of arguments to give.
     # See class mathics.builtin.arithfns.basic.Sqrt for how to set up.
+    #
+    # The way this gets called is weird. It is pattern-matched
+    # along with other function call patterns. As a result, it might
+    # get called when there is no error. Therefore, we need to
+    # determine whether the expected arguments count matches the actual
+    # argument counts of the arguments provided.
```
mmatera, line 453: Something that I would not check yet is whether this issue with the fallback rules that show messages is related to a wrong pattern order.

rocky: I don't understand enough about what the "right" order is. If the order is wrong, that kind of thing needs isolation to show what's up.  It's not something I feel confident in undertaking right now.

PR #1729, opened 2026-03-14
First cut at ExpandNumerator

rocky:
A first cut at adding `ExpandNumerator`.

### DIFF
--- a/mathics/builtin/numbers/algebra.py
+++ b/mathics/builtin/numbers/algebra.py
@@ -765,6 +765,34 @@ class ExpandDenominator(_Expand):
         return expand_polynomial(expr, False, True, **kwargs)
 
 
+class ExpandNumerator(_Expand):
+    """
+    <url>:WMA link:
+    https://reference.wolfram.com/language/ref/ExpandNumerator.html</url>
+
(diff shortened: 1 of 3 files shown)


mathics/builtin/numbers/algebra.py
```diff
@@ -765,6 +765,34 @@ def eval(self, expr, evaluation: Evaluation, options: dict):
...
+    <url>:WMA link:
+    https://reference.wolfram.com/language/ref/ExpandNumerator.html</url>
+
+    <dl>
+      <dt>'ExpandNumerator'[$expr$]
+      <dd>expands out products and powers appearing in the numerator in $expr$.
+    </dl>
+
+    >> ExpandNumerator[ ( x - 1 ) ( x - 2 ) / x ]
+     = -3 + 2 / x + x
```
axkr, line 779: WMA returns `(2 - 3*x + x^2)/x `

rocky: I don't see what we have wrong, just different.  Still, I think I will be removing this function. For now, it is far more trouble than it is worth.

PR #1743, opened 2026-03-17
Improve handling  `N[x_Integer]` for large `x`.

mmatera:
This PR fixes #1741

PR #1744, opened 2026-03-17
Use mpmath when we need a large mantissa

### LINKED ISSUE #1741: Implement `MachinePrecision` option  for large numbers that fall outside of Python's builtin `float` mantissa
## Description

`N[integer, MachinePrecision]` is not implemented.

Currently, we are forcing `N` for a large `Integer` argument (doesn't fit into a `float` mantissa) to be PrecisionReal.

In the source code that implements the evaluation function,  remove lines [94-95 of `mathics.core.eval.nevaluator`](https://github.com/Mathics3/mathics-core/blob/master/mathics/eval/nevaluator.py#L94-L95): 

```
if isinstance(expr, Number):
        return expr.round(d)
````

Now use `N` with `MachinePrecision` specified: 

```mathematica

In[1]:= N[265613988875874769338781322035779626829233452653394495974574961739092490901302182994384699044001, MachinePrecision]
Out[1]= 265613988875874769338781322035779626829233452653394495974574961739092490901302182994384699044001
```

I believe this is showing that numeric conversion to `MachinePrecision`  has not been implemented.

## Expected behavior

```
In[2]:= N[3^200, MachinePrecision]                                                                        
Out[2]= 2.65614 10^95

```

## Additional context

Related to issue #1740

rocky:
Fixes #1741

Thanks to @mmatera for his guidance and information on this.

### DIFF
--- a/mathics/core/atoms/numerics.py
+++ b/mathics/core/atoms/numerics.py
@@ -310,19 +310,30 @@ class Integer(Number[int]):
         return self._value == 0
 
     def round(self, d: Optional[int] = None) -> Union["MachineReal", "PrecisionReal"]:
-        """Produce a Real approximation value of ``Integer`` with
-        decimal precision ``d``.  If ``d`` is ``None`` we force the
-        mantissa to fit the entire integer value. This value is
-        created from a Python ``float`` if the number fits in a float
-        mantissa. Otherwise, we use a mpmath.mpf value with
-        (mpmath.mp.dps) adjusted to hold the integer.  In either case
-        a ``MachineReal`` value is returned.
+        """Produce a Real approximation rounding value of ``Integer`` with
+        decimal precision ``d``.
 
-        If ``d`` is given, we return a ``PrecisionReal`` value from a
-        sympy.Float.
+        If ``d`` is ``None`` we force the mantissa to fit the entire
+        integer value, provided it is less than magical number
+        1024. 1024 is a common internal Mathematica implementation limit where
+        it switches from using MachineReal to PrecisionReal.
+
+        If a decimal precision ``d`` is not None, then we convert to
+        a PrecisionReal using that value d.
+
+        When ``d`` is ``None`` but the mantissa does not fit into a
+        Python float, we implement the value as a mpmath.mpf value.
         """
         if d is None:
-            return MachineReal(self.value)
+            d = self.value.bit_length()
+            # Many WMA implementations seem to change behavior of the integer
+            # represetation that have more than 1024 digits. In theory this is
+            # number can vary depending on hardware characteristics.
+            # In practice, a reasonable
+            if d <= 1024:
+                return MachineReal(self.value)
+            else:
+                d = 16  # MACHINE_PRECISION_VALUE rounded up
 
         return PrecisionReal(sympy.Float(self.value, d))
 
@@ -424,7 +435,7 @@ class Real(Number[T]):
         update(b"System`Real>" + str(self.to_sympy().n(_prec)).encode("utf8"))
 
 
-# This has to come before PrecisionReal which uses MachineReal.
+# This has to come before PrecisionReal, which uses MachineReal.
 # FIXME: rocky: float is not right. It should be Union[float, mpmath.mpf]
 # but I don't understand how to get the type annotation system to handle his.
 class MachineReal(Real[float]):
@@ -432,6 +443,8 @@ class MachineReal(Real[float]):
     Machine precision real number.
 
     Stored internally as a Python float or a mpmath.mpf
+
+    Precision for these numbers is `MachinePrecision`.
     """
 
     # Dictionary of MachineReal constant values defined so far.
--- a/test/builtin/test_numeric.py
+++ b/test/builtin/test_numeric.py
@@ -80,6 +80,24 @@ def test_realvalued():
             "Numeric converts a large integer to a MachineReal without losing precision",
         ),
         (
+            "N[2^1023]",
+            None,
+            "8.98847×10^307",
+            "Numeric display digits for value under 2^1024 is DefaultPrintDisplay value 6",
+        ),
+        (
+            "N[2^1024]",
+            None,
+            "1.797693134862316×10^308",
+            "Numeric display digits for value on or orver= 2^1024 is $MachineDigits",
+        ),
+        (
+            "N[3^200]",
+            None,
+            "2.65614×10^95",
+            "Numeric converts a large integer to a MachineReal without losing precision",
+        ),
+        (
             "p=N[Pi,100]",
             None,
             "3.141592653589793238462643383279502884197169399375105820974944592307816406286208998628034825342117068",


test/builtin/test_numeric.py
```diff
@@ -73,6 +73,12 @@ def test_realvalued():
 @pytest.mark.parametrize(
     ("str_expr", "msgs", "str_expected", "fail_msg"),
     [
+        (
+            "N[3^200]",
+            None,
+            "2.65614×10^95",
```
rocky, line 97: Heads up!  I note here is that we are testing using the Unicode × even though tests are supposed to use ASCII encoding. I suspect that correcting ToString to do the right thing is going to cause a lot of changes in (wrong) test expectations.

mmatera: Yep. But step by step.

mathics/core/atoms/numerics.py
```diff
@@ -308,18 +310,25 @@ def is_zero(self) -> bool:
...
         if d is None:
             d = self.value.bit_length()
+
             if d <= FP_MANTISA_BINARY_DIGITS:
                 return MachineReal(float(self.value))
-            else:
-                d = MACHINE_PRECISION_VALUE
+
+            with mpmath.workdps(d):
+                return MachineReal(mpmath.mpf(self.value))
```
mmatera: The problem with this approach is that we are lying about working with "MachineReal".

rocky: > The problem with this approach is that we are lying about working with "MachineReal".

I don't understand. Please enlighten me.

mmatera: @rocky, please compare with #1743 and let's take what makes more sense of each one.

mmatera: > > The problem with this approach is that we are lying about working with "MachineReal".
> 
> I don't understand. Please enlighten me.

MachineReal is supposed to be a wrapper for native `float`. With this approach, you are hiding a mpmath number in a MachineReal to make it look like a native `float`. My proposal, which gives the same results that you get in WMA, is to try to convert to float, and if it fails, produce a PrecisionReal.

rocky: I have looked at #1743. This PR addresses the expectation that `N[2^300]` gives the same results as WMA. #1743 does not. So it does not fix the issue, but instead asserts incompatibility with WMA when none is needed here.

If the issue is that we aren't "rounding" when, and the expectation is that we should, Then we should create another function and arrange `N[2^300]` to call this implementation. 

Rounding an integer _is_ the same as the integer, so in that sense, I don't have a problem with the method called "round".

mmatera: Now, try with N[3^647]

rocky: Ok - good. 

With this branch I get: 

```
General::ovfl: Overflow occurred in computation.
```

with #1743 I get:

```
4.98255158407003×10^308
```

Both are different from WMA: 

```
4.982551584070027 10^308
```

rocky: From the above,  what I learn is that we should correct our implementation of  `MachineReal` so that it is either a float or a mpmath mpf.

I will be doing that now.

mmatera:
@rocky, please check if this PR reproduces the WMA behavior for this: 
```
>> {SameQ[N[3^646]//Precision, MachinePrecision],SameQ[N[3^647]//Precision, MachinePrecision]}
  = {True, False}
```

rocky:
> @rocky, please check if this PR reproduces the WMA behavior for this:
> 
> ```
> >> {SameQ[N[3^646]//Precision, MachinePrecision],SameQ[N[3^647]//Precision, MachinePrecision]}
>   = {True, False}
> ```

If you can explain why the precision changes when we go from 3^636 to 3^647, I am sure we can mimic that behavior in the code. Right now, though, I don't understand the rationale behind this, and one can't (or shouldn't try) to code what one does not understand.

mmatera:
The rationale is this one: `N[x]` , equivalent to  `N[x, MachinePrecision]` tries to convert `x` into an approximate number represented by a MachineReal number. A MachineReal number is a number that can be handled by the CPU/FPU in a "native" (fast) way.
When this conversion is not possible, WMA produces a `PrecisionReal` number, with a precision (almost) equivalent to the precision of a floating point number.
The change of behavior (both in formatting as in the result of `Precision`) corresponds to this change in the argument, from a number that can be approximately represented by a floating point (float) number, and a number that must be treated as a (non-native) mpmath number. You can also see the difference if you print both numbers: in WMA 
```
In[2]:= {N[3^646], N[3^647]}                                                                                                                                                  

                   308                      308
Out[2]= {1.66085 10   , 4.982551584070027 10   }

```

mmatera:
Also,
```
class MachineReal(Real[float]):
    """
    Machine precision real number.

    Stored internally as a python float.
    """
```

(not `mpmath.mpf`)

rocky:
> Also,
> 
> ```
> class MachineReal(Real[float]):
>     """
>     Machine precision real number.
> 
>     Stored internally as a python float.
>     """
> ```
> 
> (not `mpmath.mpf`)

Note that this gets changed in this PR: https://github.com/Mathics3/mathics-core/pull/1744/changes#diff-4962cb3725acd09a6aa5b0033a5d789873396cbc7f785559d214cd93e5f3f9f7R434

I will write more in response to what you inform in the rationale.

rocky:
> The rationale is this one: `N[x]` , equivalent to `N[x, MachinePrecision]` tries to convert `x` into an approximate number represented by a MachineReal number. A MachineReal number is a number that can be handled by the CPU/FPU in a "native" (fast) way. 

Ok. But be aware that CPU "native" and "fast" are extremely subjective in the context of Python, and in the context of alternate implementations of Mathematica that attempt to conform to Mathematica.

A Python `int` is not the same thing as a native CPU register. The `int` and `float` "objects" have some sort of wrapping in , when you look at the lower-level implementation, much in the same way as our Integer wraps the _Python_ int and pre-this patch `MachineReal` wraps `float`.  I would find it hard to believe that WMA `MachineReal` wraps a CPU float, because 1989, since `N[3^200]` has given a mantissa, according to this explanation, that does not fit in an IEEE double float.

Also, note that Python's underlying `int` representation automatically grows as the integer does. 


> When this conversion is not possible, WMA produces a `PrecisionReal` number, with a precision (almost) equivalent to the precision of a floating point number. 

So, what we are trying to understand to match what the likely WMA behavior is; when in WMA, this conversion is not possible.

I understand in _theory_ that it can vary from machine to machine. However, we should shoot for most _likely_ values. These are the kinds of values you find in a laptop, in the cloud, on a workstation, etc.  

I suspect that on most implementations to cutover point mathissa value will be larger than the IEEE double standard. I would not even be surprised if that were not also the case on a mobile phone implementation as well.

> The change of behavior (both in formatting and in the result of `Precision`) corresponds to this change in the argument, from a number that can be approximately represented by a floating point (float) number, and a number that must be treated as a (non-native) mpmath number. You can also see the difference if you print both numbers: in WMA
> 
> ```
> In[2]:= {N[3^646], N[3^647]}                                                                                                                                                  
> 
>                    308                      308
> Out[2]= {1.66085 10   , 4.982551584070027 10   }
> ```

Yes, I noticed that. In the back of my mind, I was asking if this is a result of some WMA mantissa overflow, exactly what is that overflow value? If we know that, we can easily adjust the Mathics3 code to match that.

And in the end, we can document all of this behavior.

rocky:
> @rocky, please check if this PR reproduces the WMA behavior for this:
> 
> ```
> >> {SameQ[N[3^646]//Precision, MachinePrecision],SameQ[N[3^647]//Precision, MachinePrecision]}
>   = {True, False}
> ```

With latest commit we track this. Not surprisingly, the breakout was when the mantissa had size 1024.

mmatera:
> 
> Ok. But be aware that CPU "native" and "fast" are extremely subjective in the context of Python, and in the context of alternate implementations of Mathematica that attempt to conform to Mathematica.
> 
> A Python `int` is not the same thing as a native CPU register. The `int` and `float` "objects" have some sort of wrapping in , when you look at the lower-level implementation, much in the same way as our Integer wraps the _Python_ int and pre-this patch `MachineReal` wraps `float`. I would find it hard to believe that WMA `MachineReal` wraps a CPU float, because 1989, since `N[3^200]` has given results that, according to this explanation, do not fit in an IEEE double float .
> 
3^200 ~ 2.65614e95 and 95 fits in a 11-bits exponent, so its approximate value fits in a IEEE double float. Remember that `N[x, MachinePrecision]` gives a float number that *approximates* the value of `x`.



> Also, note that Python's underlying `int` representation automatically grows as the integer does.
> 
OK, but there is no `MachineInteger` type or something like that in WMA. 

> > When this conversion is not possible, WMA produces a `PrecisionReal` number, with a precision (almost) equivalent to the precision of a floating point number.
> 
> So, what we are trying to understand to match what the likely WMA behavior is; when in WMA this conversion is not possible.
> 

OK, but to find a mpmath.fpn as the value of a `MachineReal` object would make rant yourself some months later, when you forget why you did that :)

> I understand in _theory_ that it can vary from machine to machine. However, we should shoot for most _likely_ values. These are the kinds of values you find in a laptop, in the cloud, on a workstation, etc.
> 
> I suspect that on most implementations to cutover point mathissa value will be larger than the IEEE double standard. I would not even be surprised if that were not also the case on a mobile phone implementation as well.
> 
> > The change of behavior (both in formatting as in the result of `Precision`) corresponds to this change in the argument, from a number that can be approximately represented by a floating point (float) number, and a number that must be treated as a (non-native) mpmath number. You can also see the difference if you print both numbers: in WMA
> > ```
> > In[2]:= {N[3^646], N[3^647]}                                                                                                                                                  
> > 
> >                    308                      308
> > Out[2]= {1.66085 10   , 4.982551584070027 10   }
> > ```
> 
> Yes, I noticed that. In the back of my mind, I was asking if this is a result of some WMA mantissa overflow, exactly what is that overflow value? If we know that, we can easily adjust the Mathics3 code to match that.
> 
> And in the end, we can document all of this behavior.

OK. At this point, I leave this to your expert criteria.

mmatera (approved):
LGTM

rocky:
> > Ok. But be aware that CPU "native" and "fast" are extremely subjective in the context of Python, and in the context of alternate implementations of Mathematica that attempt to conform to Mathematica.
> > A Python `int` is not the same thing as a native CPU register. The `int` and `float` "objects" have some sort of wrapping in , when you look at the lower-level implementation, much in the same way as our Integer wraps the _Python_ int and pre-this patch `MachineReal` wraps `float`. I would find it hard to believe that WMA `MachineReal` wraps a CPU float, because 1989, since `N[3^200]` has given results that, according to this explanation, do not fit in an IEEE double float .
> 
> 3^200 ~ 2.65614e95 and 95 fits in a 11-bits exponent, so its approximate value fits in a IEEE double float. 

If there were a WMA standard (there isn't) and we were only concerned with implementing that non-existent standard instead of trying to match WMA's current _behavior_, then, yes, you are correct: we can approximate it more poorly with an IEEE double float than WMA currently does.

> Remember that `N[x, MachinePrecision]` gives a float number that _approximates_ the value of `x`.

Yes, and if we can more closely approximate it the same way WMA approximates this, that's better. 
> 
> > Also, note that Python's underlying `int` representation automatically grows as the integer does.
> 
> OK, but there is no `MachineInteger` type or something like that in WMA.
> 
> > > When this conversion is not possible, WMA produces a `PrecisionReal` number, with a precision (almost) equivalent to the precision of a floating point number.
> > 
> > 
> > So, what we are trying to understand to match what the likely WMA behavior is; when in WMA this conversion is not possible.
> 
> OK, but to find a mpmath.fpn as the value of a `MachineReal` object would make rant yourself some months later, when you forget why you did that :)
> 

This part I regret!

If you can find what I wrote back then and it is of interest, I will try to argue with myself! (Or explain the shift in position.)

As the code has been evolving and as I have come to learn more about WMA and its implementations and _our_ implementation, my thoughts about things have been changing. For this, I apologize for my nasty prior behavior.

There are two events, I recall, and what I recall saying about these. First, there was the question of whether we could use a numpy value as the representation of some type. I think it was for an _array_ type. I believe I said no.
 
But I want to clarify that. A numpy array type is not remotely close to a Mathics3 list. It is open-ended how close it matches a NumericArray. 

But in this kind of use case, we _could_ use a numpy float, _if_ it matched what we need to represent better. (It doesn't here. decimal would be the next best, I think)

There was never a problem about storing a numpy representation as a `numpy` attribute instead of `value`.  And in that situation, I think having a `numpy` attribute was either sufficient or preferable. 

The kind of thing we don't want is a client of a numeric object having to be aware that it is implemented in SymPy, numpy, or mpmath.  If the type object wrapper can make that transparent, then that is good. I think to some extent, some of these packages do try to provide methods that match existing Python methods on the corresponding object

The other situation where this came up, and I have been wavering with how to represent Complex numbers.

Right now, we use a Python complex as the value field. To a large extent, that works. But something in the future, as equally subtle as we see here, might break, and we'll have to address it. And in that, we might switch to another representation from a library like SymPy, NumPy, mpmath, decimal, or who knows? 

As I am sure you know, WMA Complex numbers are more general than Python's complex value: the real and imaginary parts are arbitrary (non-complex) numbers. 

So here, as I think you indicate, the `value` field is only approximate.  And we have this weird method called `exact_value` which has the full story; the `Re` and `Imag`  builtins make use of the namedtuple that the `exact_value` property returns for Mathics3 Complex objects.  Again, what happens in the future here will depend on what we uncover and how we might be deficient. 

Thank you for being a good sport about this.

PR #1762, opened 2026-03-29
Fix Rayleigh expansion rules to only match half-integer orders

cxzhong:
The Rayleigh formulas for`BesselI`, `BesselK`, `BesselJ`, and `BesselY` are only valid for half-integer orders (nu = n + 1/2). The condition `IntegerQ[2*nu]` matches both half-integers and integers, causing wrong results for integer orders like `BesselK[2, I]`.

Test for genuine half-integer orders. Integer orders correctly fall through to the `mpmath` numerical evaluation.

### DIFF
--- a/mathics/builtin/specialfns/bessel.py
+++ b/mathics/builtin/specialfns/bessel.py
@@ -330,6 +330,10 @@ class BesselI(_Bessel):
         The special case of half-integer index is expanded using Rayleigh's formulas:
         >> BesselI[3/2, x]
          = Sqrt[2] Sqrt[x] (-Sinh[x] / x ^ 2 + Cosh[x] / x) / Sqrt[Pi]
+
+        Integer orders are not expanded and remain symbolic:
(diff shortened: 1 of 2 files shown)


cxzhong:
CC: @mkoeppe @mmatera @rocky

cxzhong:
I should add a test about it

rocky:
> I should add a test about it

Please do. Detailed bug-related test would go in `Mathics3/mathics-core/test/builtin/specialfns/test_bessel.py`.

However, you might also consider adding to the documentation, informing or educating users of BesslI, etc., of this property.  See for example: https://github.com/Mathics3/mathics-core/blob/master/mathics/builtin/specialfns/bessel.py#L330-L332

The screwy documentation markup is described at https://mathics-development-guide.readthedocs.io/en/latest/extending/developing-code/extending/documentation-markup.html#markup-in-documentation

cxzhong:
Thanks for reviewing

PR #1764, opened 2026-03-29
Fix handling non-ASCII characters in render mathml

mmatera:
In WMA, MathMLForm, disregarding the value of ``$CharacterEncoding", transforms non-ASCII characters from the internal encoding into escaped character codes in the form  `&#{code};`.  Tests are adjusted.

### DIFF
--- a/test/format/format_tests.yaml
+++ b/test/format/format_tests.yaml
@@ -71,7 +71,7 @@
     System`StandardForm: "\\text{$\\pi$ is a trascendental number}"
     System`TraditionalForm: "\\text{$\\pi$ is a trascendental number}"
   mathml:
-    System`InputForm: "<ms>&#960;&nbsp;is&nbsp;a&nbsp;trascendental&nbsp;number</ms>"
+    System`InputForm: '<ms>\[Pi]&nbsp;is&nbsp;a&nbsp;trascendental&nbsp;number</ms>'
     System`OutputForm: "<mtext>&#960;&nbsp;is&nbsp;a&nbsp;trascendental&nbsp;number</mtext>"
     System`StandardForm: "<mtext>&#960;&nbsp;is&nbsp;a&nbsp;trascendental&nbsp;number</mtext>"
     System`TraditionalForm: "<mtext>&#960;&nbsp;is&nbsp;a&nbsp;trascendental&nbsp;number</mtext>"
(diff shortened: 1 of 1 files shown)


mathics/format/render/mathml.py
```diff
@@ -69,10 +69,27 @@ def convert_inner_box(box, **options):
...
+    """
+    Convert a string into a MathML code.
+    * Escape special characters
+    * Replace non-ASCII characters by character codes.
+    """
     text = text.replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;")
     text = text.replace('"', "&quot;").replace(" ", "&nbsp;")
     text = text.replace("\n", '<mspace linebreak="newline" />')
-    return text
+    # Now, handle non-ansi characters using the
```
```suggestion
    # Now, handle non-ASCII characters using the
```

mathics/format/render/mathml.py
```diff
@@ -69,10 +69,27 @@ def convert_inner_box(box, **options):
...
+    # the form '&#{code};'
+    # Notice that this should happend after escaping special characters:
```
```suggestion
    # This must happen after escaping special characters:
```

rocky:
The description and code are good. Some small suggestions added. 

As I look at the bigger picture, I realize I don't understand the test framework. 

Consider this abbreviated example: 

```
Complex[1.09*^12, 3.]:
  msg: Complex number
  ...
  mathml:
    System`InputForm: <mn>1.09*^12&nbsp;+&nbsp;3.*I</mn>
    System`OutputForm: '<mtext>1.09&#215;10^12&nbsp;+&nbsp;3.&nbsp;I</mtext>'
    System`StandardForm: "<mrow>\n <mrow>\n  <mn>1.09</mn>\n  <mtext>*^</mtext>\n  <mn>12</mn>\n </mrow>\n <mo>+</mo>\n <mrow>\n  <mn>3.</mn>\n  <mo>&nbsp;</mo>\n  <mi>I</mi>\n </mrow>\n</mrow>"
    System`TraditionalForm: "<mrow>\n <mrow>\n  <mn>1.09</mn>\n  <mo>&#215;</mo>\n  <msup>\n   <mn>10</mn>\n   <mn>12</mn>\n  </msup>\n </mrow>\n <mo>+</mo>\n <mrow>\n  <mn>3.</mn>\n  <mo form=\"prefix\" lspace=\"0\" rspace=\"0.2em\">&#8290;</mo>\n  <mi>I</mi>\n </mrow>\n</mrow>"
  text:
    System`InputForm: 1.09*^12 + 3.*I
    System`OutputForm: "1.09\xD710^12 + 3. I"
    System`StandardForm: 1.09`*^12+3.` I
    System`TraditionalForm: "1.09`\xD710^12+3.`\u2062I"
```

What exactly do ``System`InputForm``,  ``System`OutputForm``, and ``System`StandardForm`` mean? No WMA output produces exactly that.:

```mathematica
In[1]:= Complex[1.09*^12, 3.] //MathMLForm                                

Out[1]//MathMLForm= <math>
                     <mrow>
                      <mrow>
                       <mn>1.09</mn>
                       <mo>&#215;</mo>
                       <msup>
                        <mn>10</mn>
                        <mn>12</mn>
                       </msup>
                      </mrow>
                      <mo>+</mo>
                      <mrow>
                       <mn>3.</mn>
                       <mtext> </mtext>
                       <mi>&#8520;</mi>
                      </mrow>
                     </mrow>
                    </math>

In[2]:=                                                                   
```
Notice there are surrounding `<math> ... </math>` tags. What this branch produces is: 

```mathematica
In[1]:= Complex[1.09*^12, 3.] //MathMLForm  
Out[1]//MathMLForm= <mrow>
                     <mrow>
                      <mn>1.09</mn>
                      <mo>&#215;</mo>
                      <msup>
                       <mn>10</mn>
                       <mn>12</mn>
                      </msup>
                     </mrow>
                     <mo>+</mo>
                     <mrow>
                      <mn>3.</mn>
                      <mo form="prefix" lspace="0" rspace="0.2em">&#8290;</mo>
                      <mi>I</mi>
                     </mrow>
                    </mrow>

In[2]:= 
```

without the surrounding `<math>` tags. 

And what is the input that is supposed to match the other forms listed?

rocky:
We should check to see if this changes building LaTeX or PDF docs.

mmatera:
> The description and code are good. Some small suggestions added.
Good
> 
> As I look at the bigger picture, I realize I don't understand the test framework.
> 
> Consider this abbreviated example:
> 
> What exactly do `` System`InputForm ``, `` System`OutputForm ``, and `` System`StandardForm `` mean? No WMA output produces exactly that.:
> 

In WMA, MathMLForm does not play very well with forms, probably because it does not use it for anything. On the other hand, we need this form works, because we use it in MathicsDjango. Something similar happen with TeXForm and the PDF documentation. These tests try to check that the output stays consistent.


> Notice there are surrounding `<math> ... </math>` tags. What this branch produces is:
> ...
> 
> without the surrounding `<math>` tags.
> 
> And what is the input that is supposed to match the other forms listed?

We are not adding these tags for a while. We should add them, but if we do that, we have to synchronize it with MathicsDjango.

Yeah, I agree.  BTW, right now in Mathics-Django, handling MLForm is broken when you look at the gallery example. I tried to fix this in https://github.com/Mathics3/Mathics3-django/pull/249 but other stuff broke.

> In any case, this is not related with this PR.

It is related, or I am concerned about it in this respect: I want to understand and limit the amount of _wrong_ test results we are checking against. This is one example of "not-quite-right" test results that are likely to change in the future. In this _particular_ situation, I think I agree with you that addressing this now is not a good idea, and this PR doesn't make things worse.

mmatera:
> We should check to see if this changes building LaTeX or PDF docs.

Done, and seems to be OK.

mathics/format/render/mathml.py
```diff
@@ -69,10 +69,27 @@ def convert_inner_box(box, **options):
...
     text = text.replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;")
     text = text.replace('"', "&quot;").replace(" ", "&nbsp;")
     text = text.replace("\n", '<mspace linebreak="newline" />')
-    return text
+    # Now, handle non-ASCII characters using the
+    # the form '&#{code};'
+    # This must happen after escaping special characters:
+    result = ""
+    for c in text:
+        code = ord(c)
```
rocky, line 85: Isn't this and the next line doing the same thing as the Python builtin method `isascii`?

rocky:
@mmatera, I see that I erroneously edited your remark rather than reply to it. (I think I did that once before, and I didn't mean to. My apologies for both times I did this.

Coming back to this: 
> In any case, this is not related with this PR.

Something slightly important is what exactly do ``System`InputForm``, ```System`OutputForm``, and ```System`StandardForm``` mean in the context of this testing framework.

Yes, aren't _changing_ that here, I should have asked about this when the thing goes in. But it would be good to understand before changes appear via a PR.

mmatera:
> @mmatera, I see that I erroneously edited your remark rather than reply to it. (I think I did that once before, and I didn't mean to. My apologies for both times I did this.
> 
> Coming back to this:
> 
> > In any case, this is not related with this PR.
> 
> Something slightly important is what exactly do `` System`InputForm ``, `` System`OutputForm``, and ``System`StandardForm``` mean in the context of this testing framework.
> 
> Yes, aren't _changing_ that here, I should have asked about this when the thing goes in. But it would be good to understand before changes appear via a PR.

Right now, changes in these tests requires to check that the MathicsDjango interface produces the expected output. I did this in this case.

mathics/format/render/mathml.py
mmatera: maybe, but in general we still want to know the value of  `code`.

rocky:
LGTM

PR #1769, opened 2026-03-31
Add Trig option to Numerator/Denominator

rocky:
Add `Trig` option to `Numerator` and `Denominator`.
Also move functions to `mathics.eval` with an eye towards replacing `_Expand()`with `_Algebriac()`.


mathics/builtin/numbers/algebra.py
```diff
@@ -561,24 +603,37 @@ class Denominator(Builtin):
...
+
     See also <url>
      :'Numerator':
      /doc/reference-of-built-in-symbols/integer-and-number-theoretical-functions/algebraic-transformations/numerator/</url>.
     """
 
     attributes = A_LISTABLE | A_PROTECTED
+    options = {
+        "Trig": "False",
+        "Modulus": "0",  # Modulus > 0 supported yet.
```
mmatera: "Mudulus >0 *not* supported yet", right?

mathics/eval/numbers/algebra/fraction.py
```diff
@@ -0,0 +1,82 @@
...
+        case func if func in (cosh, csch, coth):
+            return sinh
+        case func if func in (tanh,):
+            return cosh
+        case _:
+            return None
+
+
+def get_fractional_parts(expr, algebraic_options, want_numerator):
+    sympy_expr = expr.to_sympy()
```
mmatera, line 39: Also here expressions like `F[x]/G[x]` should be handled, by collecting factors with negative powers, isn't it?
For example  `Denominator[(1+F[x]) (1-G[x])^-1]` should return `1-G[x]`

mathics/builtin/numbers/algebra.py
rocky: Yes, you are correct. Thanks!

mathics/eval/numbers/algebra/fraction.py
rocky: For that particular example (non-trig cases), we do this. I am not confident, though, as to how to extend to the trig situations. 

What is a concrete example when the Trig option is True that demonstrates the situation?

mmatera: Let's say `Tan[x]^2` or `(1-Tan[x]) Sec[x]`

rocky: > Let's say `Tan[x]^2` or `(1-Tan[x]) Sec[x]`

For the first example, we do not give the same answer as WMA `Cos[x]^2]` .

However, for the second, we do `Cos[x]`, 

I am not confident right now, that I would know how to fix. So I'm happy to leave this hang out here then.

mmatera: I think we can handle it by adding two rules: one for `Times`, that apply the functions to each factor, and multiplied them back,  and the other for `Power`, that according to the sign of the exponent, retrieves the result of applying numerator or denominator. But it could come after this.

rocky: Ok. Then I leave it up to you whether to merge or hold back.

mmatera: Lgtm. Let's merge this and add it later

PR #1771, opened 2026-04-02
adjust mathics3 dependency

mmatera:


mmatera:
Detected when I was trying to figure out why Mathics3-scanner doctest workflow failed

.github/workflows/ubuntu.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.13', '3.12', '3.11', '3.10']
+        python-version: ['3.10', '3.13', '3.12', '3.11']
```
rocky: FYI - Previously, we couldn't go to 3.14 because there was a library used by Mathics3-Module-nltk that had not been updated to use 3.14. That I think has changed.

I will be getting the master branches updated with "critical" fixes so I can try a mathics-omnibus docker build using 3.14. If that passes, we can drop 3.10 (which might be causing problems in the stupid licensing incompatibility lingo in pyproject.toml that Python packaging has foisted upon us)

PR #1766, opened 2026-03-29
Handle encodings in docpipeline

mmatera:
This PR introduces the changes to make the docpipeline work in a transparent way with the system character encoding.  
Reference results in doctests are now in the internal encoding. Then, when docpipeline compares the reference with the result of a test, in the system encoding, the comparison is done against the result of applying the system encoding to the reference.  
Therefore, running
```
MATHICS_CHARACTER_ENCODING="ASCII" python mathics-docpipeline.py
```
is equivalent to running
```
python mathics-docpipeline.py
```

PR #1774, opened 2026-04-04
Handle encodings in docpipeline0

mmatera:
This is the part of #1766 that does not imply changes in docstrings. `docpipeline.py` still requires setting the `MATHICS_CHARACTER_ENCODING="ASCII" ` encoding in tests.
Also, this adds a missing image file in the documentation, and comment out a font choice that produced errors in the compilation of  latex documentation.


rocky:
LGTM

PR #1775, opened 2026-04-04
Ensure date formats have a year in them.

rocky:
In Python 3.14 we get warning when a date specification without a year is used. In 3.15 it will be mandatory.

Therefore, if a year is not provided via "%y" or "%Y", we will add year 2000 along with the specifier %Y.


mmatera:
LGTM

Issue #1767, opened 2026-03-30
Graph formatting in LaTeX mode is not occurring from doctest


rocky:
## Description

The graphs from the Graph module, such as for StarGraph no longer appear in the LaTeX PDF.

Tracking what is going wrong,  it appears that docpipeline in running:

```
$ MATHICS_CHARACTER_ENCODING="UTF-8" python mathics/docpipeline.py  --output --keep-going --load-module pymathics.graph
```

the `format_graph()` routine is not getting called for `form="latex". 

## How to Reproduce and Output Given

Build PDF, or just the LaTeX  `documentation.tex` look for a built-in that produces a tree graph, like `StarGraph`, and notice there is no test output.

## Priority

We can't release until this is fixed. 

## Additional context

A lot has changed in the way formatting is done. Also, I recall there was "duplicate code" removed, I think, in the area around how Graphs are handled. I can't find where that was, though.

rocky:
Was fixed just before release.

PR #1777, opened 2026-04-04
add missing file for documentation

mmatera:


rocky:
logo-heptatom-Mathics3.pdf is a derived file. We should correct https://github.com/Mathics3/mathics-core/blob/master/mathics/doc/latex/Makefile#L64-L66

rocky:
@mmatera Something broken in the documentation is #1767, but the source of the problem is the graph data is not getting inserted into the Pickle `.pcl` file.

mmatera:
> @mmatera Something broken in the documentation is #1767, but the source of the problem is the graph data is not getting inserted into the Pickle `.pcl` file.

I got the same error in master. For some reason, this file is not being created. Maybe is something local to my computer. I have to check.

rocky:
> I got the same error in master. For some reason, this file is not being created. Maybe is something local to my computer. I have to check.

No, sorry, this is my fault. 

Here is the longer story. Previously, we were not able to recreate the Mathics logo. In revising for Mathics3, I didn't want to have this kind of situation again, where we had a PDF that we couldn't edit the derived source.

In this process, I realized we should separate more _static_ PDFs, such as the logos, from the documentation-generated PDFs used in plots and images. So I created that `images` directory. (I realize now the same would be better called something like `static-images`). 

The basic idea is we should be able to remove dynamic images in the figures, but _not_ the static images like the logos. 

I messed up because I didn't adjust everything. And I didn't notice my mistake, because I had previously created these PDFs.

#1779 tries to address the mistakes I made.

I am sorry for the confusion it caused you.

mmatera:
Fixed in #1779

PR #1780, opened 2026-04-04
Limit-inspired changes

rocky:
Improve argument checking in builtins with options to emit "nonopt" tags when appropriate

Improve from_sympy so it understands SymPy reporting Indeterminate

Go over Limit docs, and autoload rules

Expand limit tests


mmatera:
LGTM

PR #1782, opened 2026-04-05
Add more argument checking.

rocky:
* In `Args`, remove the "method" option. The current thinking is that we should handle this transparently.
* Add expected args to functions in `mathics.builtin.arithmetic`.
* Fill in more `sympy` and `mpath` name correspondences.
* Rename `test.builtins.arithmetic` to `test.builtins.arthfns`, since that's where it is located.
* Test argument checking in `mathics.builtin.arithmetic`.


mmatera:
LGTM

PR #1783, opened 2026-04-05
Note Limit's limits and sympy AccumuateBounds problem

rocky:


mathics/builtin/numbers/calculus.py
```diff
@@ -1237,6 +1237,9 @@ def eval_D(self, func, domain, var, evaluation: Evaluation, options: dict):
         )
 
 
+# FIXME: Limit has changes a bit so that it now can return
```
rocky:
```suggestion
# FIXME: Limit has changed a bit, so that it now can return
```

mmatera (approved):
LGTM

PR #1784, opened 2026-04-05
Add `Path` and `$CharacterEncoding` options to `Get`

rocky:
* `Get` now properly includes directories in `$Path` in searching.
* Handle `nonopt`  tag message properly when exected_args is a `range` (instead of a `tuple`).
* Add `CharacterEncoding` option for `Get`
* Mathics->Mathics3

### DIFF
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -23,7 +23,7 @@ jobs:
         sudo apt update -qq && sudo apt install llvm-dev remake
         python -m pip install --upgrade pip setuptools wheel
 
-    - name: Run mypy
+    - name: Install mathics-core
       run: |
         pip install mypy==1.13 sympy==1.12
         # Adjust below for right branch
@@ -36,4 +36,6 @@ jobs:
         touch ./Mathics3-scanner/mathics_scanner/py.typed
         python -m pip install setuptools wheel
         make develop
+    - name: Run mypy on mathics-core
+      run: |
         mypy --install-types --ignore-missing-imports --non-interactive mathics


mathics/eval/files_io/files.py
```diff
@@ -122,6 +122,7 @@ def eval_Get(
     evaluation: Evaluation,
     trace_fn: Optional[Callable] = DEFAULT_TRACE_FN,
     path_directories: Optional[Sequence[str]] = None,
+    encoding="UTF-8",
```
mmatera: I wonder if this should be `SYSTEM_CHARACTER_ENCODING`

mathics/eval/strings.py
```diff
@@ -19,6 +20,54 @@
 from mathics.core.symbols import Symbol, SymbolTrue
 from mathics.format.box import format_element
 
+CHARACTER_ENCODING_MAP: Final[Dict[str, str]] = {
```
mmatera: My guess is that it should be in `mathics.eval.encoding`

mmatera:
Apart from the comments, LGTM

mathics/eval/strings.py
rocky: I think so too. It will be in the next commit.

mathics/eval/files_io/files.py
rocky: It probably should. Is in next commit.

rocky:
> Apart from the comments, LGTM

I need to find a way to test this.

rocky:
This is getting large. I will open another PR for just testing and possible corrections/additions.

Issue #1654, opened 2026-01-25
`TraditionalForm` is replacing square brackets with parenthesis when it should not


rocky:
## Description

For the parameters of some boxing functions like `SuperscriptBox` and `SubscriptBox`, TraditionalForm uses parentheses when it should use square brackets.

## How to Reproduce and Output

```
In[1]:= SuperscriptBox["x", "2"] //TraditionalForm
Out[1]//TraditionalForm= SuperscriptBox(x, 2)

In[2]:= SuberscriptBox["x", "2"] //TraditionalForm
Out[2]//TraditionalForm= SuberscriptBox(x, 2)
```

## Expected behavior

```
In[1]:= SuperscriptBox["x", "2"] //TraditionalForm
Out[1]//TraditionalForm= SuperscriptBox[x, 2]

In[2]:= SuberscriptBox["x", "2"] //TraditionalForm
Out[2]//TraditionalForm= SuberscriptBox[x, 2]
```

# Context and Workarounds

This is a recent behavior. The 9.0.0 does not have this behavior. Are there other forms that use parentheses where square brackets should be used?

rocky:
@mmatera: I looked at this, and the situation is much more involved.  So I removed the 10.0 milestone here.

 `TraditionalForm` is off by a wide margin, so matching it's behavior here is kind of silly.

rocky:
It appears that when TraditionalForm is applied to a box like SubscriptBox, its traditional-form-ness is lost ,and it may be that the end result is the same as StandardForm, which is why we get square brackets.

mmatera:
What about defining a MakeBoxes rule for this kind Box expressions? Something
`MakeBoxes[SuperscriptBox[a_,b_],form_]:=RowBox[{"SuperscriptBox", "[", MakeBoxes[a,form], "," MakeBoxes[b,form], "]"}]`?


Also, notice that in WMA,
```
In[9]:= Superscript["x",F[x]]//TraditionalForm                                  

Out[9]//TraditionalForm= 
 
>   FormBox[TemplateBox[{"x", F(x)}, Superscript, 
 
>     SyntaxForm -> SuperscriptBox], TraditionalForm]
```

So it seems that this kind of rules seems to be tricky also for them...

rocky:
This is a possibility. But let's revisit this after release. 

The kind of thing brought up in this issue is more of an edge case. And we are a good way away from handling `TraditionalForm` in any reasonable manner in the normal use cases.

PR #1791, opened 2026-04-07
Boxing experiment

rocky:
Items noticed in investigating #1654 

* Remove a hard-coded precedence value.
* Some Mathics -> Mathics3 renamings. 
* Add a `__str__` function to a Box class to as an experiment to see what this helps
* Use Symbols instead of "System`..." names
* Guard against a **kwargs call that does not include an "encoding" key


mmatera (commented):
Lgtm

PR #1818, opened 2026-04-26
Misc small bugs or improvements

rocky:
* Guard against uninitialized System`$CharacterEncoding
* An error message was erroneously tagged as "Put" when it should be "Get"
* Move SymbolString definition to systemsymbols


mmatera (approved):
LGTM

PR #1819, opened 2026-04-27
Try newer pyodide-build, newer Python and newer nodejs

rocky:
Bump Python to 3.14, Pyodide-build to ">=314.0.0", and Nodejs to 24. Also, remove unneeded JSON building. (It's now in the scanner).


rocky:
@davidar, if this is okay with you, we can try to update Mathics3-notebooks-frontend as well.

PR #1823, opened 2026-04-28
Cache sympy values

rocky:
Cache sympy values for Complex numbers, and Symbol.

PrecisionReal values are already a SymPy value, so allow `.sympy`  method access.

Some linting has been done.

### DIFF
--- a/mathics/core/convert/sympy.py
+++ b/mathics/core/convert/sympy.py
@@ -355,36 +355,38 @@ sympy_conversion_by_type = {
 #    return sympy_conversion_by_type.get(type(expr), old_from_sympy)(expr)
 
 
-def old_from_sympy(expr) -> BaseElement:
+def from_sympy(sympy_expr) -> BaseElement:
     """
     converts a SymPy object to a Mathics3 element.
     """
-    if isinstance(expr, (tuple, list)):
-        return to_mathics_list(*expr, elements_conversion_fn=from_sympy)
-    if isinstance(expr, int):
-        return Integer(expr)
-    if isinstance(expr, float):
-        return Real(expr)
-    if isinstance(expr, complex):
-        return Complex(Real(expr.real), Real(expr.imag))
-    if isinstance(expr, str):
-        return String(expr)
-    if expr is None:
+    if isinstance(sympy_expr, (tuple, list)):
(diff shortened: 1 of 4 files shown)


mathics/core/convert/sympy.py
```diff
@@ -355,36 +355,38 @@ def from_sympy_matrix(
 #    return sympy_conversion_by_type.get(type(expr), old_from_sympy)(expr)
 
 
-def old_from_sympy(expr) -> BaseElement:
+def from_sympy(sympy_expr) -> BaseElement:
```
mmatera, line 358: I remember at some point I had to write a version of this using a dispatch table instead of this very long chain of if's. Was it lost?

rocky: Git blame reports this is [commit 4d0bcb9a86 ](https://github.com/Mathics3/mathics-core/commit/4d0bcb9a86#diff-44d51dbd16f6d63c9c39a7c1e3f4c4589157eb4381c7ce26340e225dbae8ad15)

mathics/core/convert/sympy.py (Juan Mauricio Matera         2023-05-27 346)

You are in a better position here to explain why the code is the way it is. I am guessing you tried it and something did not work out. 

Or it wasn't completed?

mmatera: It seems it was just commented out, until we find the time to discuss it. It is just above this code. We can leave it as it is, but if you are looking at this part of the code, maybe it's worth having a look. The idea was to follow the same pattern we use in format.

rocky: You commented it out. 

A guess is that things could not be put in a strict table-lookup format. But now that we're on [3.10 and greater, probably we can use switch-case](https://docs.python.org/3/whatsnew/3.10.html) for things that aren't table lookup. 

This PR can hang out for a while.

rocky: Or isolate the changes to `mathics.core.convert.sympy` to a different PR.

mmatera: I am OK with the changes in this PR. I just mentioned the dispatch table implementation because it seems related.

mmatera (approved):
LGTM

PR #1824, opened 2026-04-29
Go over YAML files

rocky:
Make these yamllint clean.
pyodide.yml: disable use of cache. This has been failing recently

### DIFF
--- a/.github/workflows/consistency-checks.yml
+++ b/.github/workflows/consistency-checks.yml
@@ -1,10 +1,11 @@
+---
 name: Mathics3 (Consistency Checks)
 
 on:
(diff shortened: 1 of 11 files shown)


rocky:
CI tests for Pyodide have been failing for a couple of days. In this PR, we disable the cache. 

Because this blocks stuff, if this PR passes CI, it will be merged.

However, we can come back and adjust other stuff should there be concern over this particular PR.

In particular, adding a .precommit hook to run `check-jsonschema --check-metaschema .github/workflows/*.yml` would be a good idea. However,  I'll do that in a separate PR.

PR #1825, opened 2026-05-01
Revise `Import`, `Export`, `ImportString`, and `ExportString`

rocky:
Use the PyPI package `python_mime` in `FileFormat`. This does a better job than what we have now. 

Pull out internal Import/Export methods and evaluation code under `mathics.eval.files_io`.  DRY this.

Remove cut-and-paste copy (in `mathics.builtin.pymimesniffer`), an early Python Package to detect MIME types, which included its embedded unittest code inside `mathics.builtin. pymimesniffer`. 

Remove _some_ (but not all yet) MIME translation tables and instead rely on lists in `python_mime`. 

We now detect ZIP files stored with a "mx" extension. This is the kind of file that seems to be created when running  `<< https://paclets.github.io/PacletServer/Install.wl`


rocky:
Handling Import*, Export* needs a major revision. 

Watch this space...

rocky:
@mmatera, there are still some hard-coded tables that can be reduced. But this is already a heavy-enough lift. Future reduction will come after the PR clears.

mmatera (approved):

Issue #1032, opened 2024-05-10
Slowness of RSolve for third-order relations and above


pie3636:
## Description

Mathics hangs when using `RSolve` on a linear fourth-order recurrence relation with simple coefficients, which should be able to be handled in seconds. The CPU usage is high while the command is running, but no output is produced even after 15+ minutes on a decent CPU (Ryzen 5 5600H).

I also tried a third-order recurrence instead, and the correct output was produced after a total of 70 seconds, which still seems excessive. Dropping the `+1` to make the relation homogeneous doesn't cut down on the processing time either.

The commands used are as follows:
Fourth-order: `RSolve[{a[n] == a[n-1] + a[n-4] + 1, a[0] == 0, a[1] == 1, a[2] == 2, a[3] == 3}, a[n], n]`
Third-order: `RSolve[{a[n] == a[n-1] + a[n-3] + 1, a[0] == 0, a[1] == 1, a[2] == 2}, a[n], n]`

## How to Reproduce

$mathics -e 'RSolve[{a[n] == a[n-1] + a[n-4] + 1, a[0] == 0, a[1] == 1, a[2] == 2, a[3] == 3}, a[n], n]'
$mathics -e 'RSolve[{a[n] == a[n-1] + a[n-3] + 1, a[0] == 0, a[1] == 1, a[2] == 2}, a[n], n]'

## Output Given

None after 15 minutes for the fourth-order relation.

## Expected behavior

The closed form expression of the sequence in a reasonable time.

## Your Environment

Mathics 6.0.4
on CPython 3.11.9 (main, Apr 27 2024, 21:16:11) [GCC 13.2.0]
using SymPy 1.12, mpmath 1.3.0, numpy 1.24.4, cython Not installed

OS: Ubuntu 24.04 running on WSL2

## Priority

Very low

rocky:
I looked at this a little bit. The slowness seems completely inside [`sympy.rsolve`](https://docs.sympy.org/latest/modules/solvers/solvers.html#sympy.solvers.recurr.rsolve). 

The input to sympy is: 

``` 
f = _Mathics_User_Global`a(_Mathics_User_Global`n)
   - _Mathics_User_Global`a(_Mathics_User_Global`n - 3) 
   - _Mathics_User_Global`a(_Mathics_User_Global`n - 1) 
   - 1
y = _Mathics_User_Global`a(_Mathics_User_Global`n)
[...]

PR #1839, opened 2026-06-10
Trace sympy.rsolve

rocky:
Add tracing for Sympy.Rsolve.

Changes made/noticed in conjunction with trying to understand what's up with #1032 

Also: changes to get Windows CI working. Apparently, matplotlib font caching is messing things up on Windows.
And we can't run doctest assuming the bash way to set environment variables.


mmatera (approved):
LGTM

PR #1840, opened 2026-06-14
Improve PredefinedExpressions and add Python dict to Mathics3 Association conversion

rocky:
* Add a way that  `from_python()` can convert a dictionary to an `Association`.
* Remove `ExpressionInfinity`. This is a duplicate of `MATHICS3_INFINITY`.
* Fill in more properties for predefined constants;  set Python and SymPy corresponding values
* Define `ELEMENTS_FULLY_EVALUATED` for use on fully-evaluated constants.

### DIFF
--- a/mathics/builtin/list/eol.py
+++ b/mathics/builtin/list/eol.py
@@ -40,7 +40,7 @@ from mathics.core.exceptions import (
     PartError,
     PartRangeError,
 )
-from mathics.core.expression import Expression, ExpressionInfinity
+from mathics.core.expression import Expression
 from mathics.core.expression_predefined import MATHICS3_INFINITY
 from mathics.core.list import ListExpression
 from mathics.core.rules import Rule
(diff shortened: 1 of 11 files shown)


mmatera:
OK, this does not seem to have any to do with Association, but looks good.

mmatera (approved):
LGTM

rocky:
> OK, this does not seem to have any to do with Association, but looks good.

The association with Association: 
    

```python
def association_from_dict(arg: dict, options: ToPythonOptions) -> BaseElement:

    entries = [
        Expression(
            SymbolRule,
            from_python(key, options),
            from_python(value, options),
        )
        for key, value in arg.items()
    ]
    return Expression(SymbolAssociation, *entries)
```

This will be used by `Import` of a ZIP file in showing Summary items. For example: 

```mathematica
In[1]:= Import["t1.zip", "Elements"]                                                    
Out[1]= {FileNames, Summary}

In[2]:= Import["t1.zip", "Summary"]                                                     
Out[2]= Dataset[<|Format -> ZIP, FileName -> t1.zip, FileSize -> 5.675 kilobytes|>, 
 >    TypeSystem`Struct[{Format, FileName, FileSize}, 
 >     {TypeSystem`Atom[String], TypeSystem`Atom[String], TypeSystem`Atom[1 kilobyte]}]\
 >     , <||>]
```

PR #1841, opened 2026-06-15
Revise `Import`. Add `ZIP` import - Part 1.

rocky:
Start redoing `Import` 
* Add `eval_FindFile` and `eval_FileFormat`. Use these instead of `Expression(...).evaluate`.
* Start handling `ZIP` imports.
* Move `Formats` under `SystemFiles` which is where WMA put this. Start reducing differences in these files; but the files are renamed to `.wl` to avoid confusion with Objective C's `.m` files.


rocky:
@mmatera this is starting to get large. Should I finish this off and leave more stuff for another PR. Split this into smaller pieces? Or continue in this PR?

mathics/builtin/files_io/importexport.py
```diff
@@ -1208,43 +1212,102 @@ class Import(Builtin):
...
+    """
+
+    summary_text = "import ZIP file"
+
+    def eval(self, path: String, evaluation: Evaluation):
+        "%(name)s[path_String]"
+        return eval_ZIPImport(path.value)
+
+    def eval_elements(self, path: String, evaluation: Evaluation):
+        """%(name)s[path_String]["Elements"]"""
```
mmatera: I wonder if this get applied... My guess is that once the head of the expression get evaluated, it produce a different expression, which does not match with this rule.

mmatera (approved):

mathics/builtin/files_io/importexport.py
rocky: You are correct. It does nothing. It also appears elsewhere. Magically, though, we can extract elements.

I will investigate later and revise.

rocky: I see what was up that should have been 

```python
 """%(name)s[path_String, "Elements"]"""
```

I've tested that the _this_ pattern matches.  Good catch!

But more will come, and more tests for this later.

PR #1842, opened 2026-06-19
filesystem rename autoload -> Autoload, data->Data as it is in WMA

rocky:
I think it is helpful to more closely match the filesystem organization of the Wolfram distribution.


mmatera (approved):
LGTM

PR #1845, opened 2026-06-25
Add more import eval fns

rocky:
Add mathics.eval: eval_DeleteFile and eval_FileExtension
    
This simplifies deleting (temporary) files and picking out the file extension in implementing Import as well as in other builtins.


mmatera:
LGTM

PR #1847, opened 2026-06-28
Improve builtin Lookup

rocky:
`Lookup[]` using rules is failing on examples like: 

```mathematica
Lookup[<|a -> 1, b -> 2|>, a]
Lookup[<|a -> 1, b -> 2|>, c]
Lookup[<|a -> 1, b -> 2|>, c, 0]
```

This corrects that.  It is not perfect: threading over lists of associations and multiple query keys are not supported. But it is better.

The operator form of Lookup is needed to support Paclet install.


mathics/eval/list/associations.py
```diff
@@ -0,0 +1,49 @@
...
+def eval_Lookup(assoc, key, default, evaluation: Evaluation):
+    """Evaluation method for Lookup."""
+    if assoc.has_form("Association", None):
+        # Search through association elements (rules)
+        for element in assoc.elements:
+            if element.has_form(("Rule", "RuleDelayed"), 2):
+                if element.elements[0] == key:
+                    return element.elements[1]
+
+        # Key not found
```
mmatera, line 17: When a key is not in the association, WMA returns `Missing[KeyAbsent, <key>]`. For instance,

```WMA
In[1]:= a=Association[{"F":>1,"G":>2}] ;
In[2]:=Lookup[a,"H"]
Out[2]= Missing[KeyAbsent, H]
In[3]:=Lookup[a,{"P","G"}]
Out[3]= {Missing[KeyAbsent, P], 2}
```

rocky: Thanks for checking and catching. The problem was not in the code, but instead in setting Lookup's attributes.

PR #1852, opened 2026-07-06
Address a StandardOutput Format bug; 1849: 

mmatera:
The origin of the issue was that when `string_render_output_form` does not get a string, fall back to the default format function. In the example, `
```
 mathics -c 'xx="a string"; StringTake[xx[0]]'
```
produced an expression with a `String` as the head:
```
In[1]:= xx="a string"
Out[1]= "a string"

In[2]:= xx[0] //FullForm
Out[2]//FullForm= "a string"[0]
````
When `render_output_form` asks for the `lookup_name` of the callback function, it get `System`String`. But that function just processes `String` atoms, not expressions with a `String` as its head. This patch just raise the exception that produces to the parent function to use the default render function.

So, now we get

````
StringTake::argr: StringTake called with 1 argument; 2 arguments are expected.
StringTake[a string[0]]
```

Notice that if instead of a string, the head was an integer, the result was already right:
```
(Mathics3) mauricio@mauricio-T15thinkpad:~/Projects/mathics-core$ mathics -c 'xx=7; StringTake[xx[0]]'
StringTake::argr: StringTake called with 1 argument; 2 arguments are expected.
StringTake[7[0]]
```

### DIFF
--- a/mathics/format/form/outputform.py
+++ b/mathics/format/form/outputform.py
@@ -3,6 +3,15 @@ This module implements the "OutputForm" textual representation of expressions.
 
 OutputForm is two-dimensional keyboard-character-only output, suitable for CLI
 and text terminals.
+
+The entry point of this module is the function `render_output_form`. This function takes an expression and an evaluation object as parameters, and returns a string representing the expression in its OutputForm.
+
+To do this, `render_output_form` looks for the head of the expression, and uses its name to lookup one of the registered functions
+to process specific kind of expressions. Callback functions are registered using the decorator `@register_outputform([lookupname])`.
+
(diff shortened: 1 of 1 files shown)


mathics/format/form/outputform.py
```diff
@@ -844,6 +854,9 @@ def _slotsequence_outputform_text(expr: Expression, evaluation: Evaluation, **kw
 def string_render_output_form(expr: String, evaluation: Evaluation, **kwargs) -> str:
```
rocky: Note that the annotation `expr: String` in the `def` signature above is not correct.

rocky:
LGTM

PR #1853, opened 2026-07-07
MS Windows does not support GNU Readline

rocky:
Miminal change to get this working in Powershell on MS Windows.


rocky:
There will be more. But let's start here.

PR #1856, opened 2026-07-10
Handle Import[] for a WL file.

rocky:
ZIP import with a WL member is supported, so: 

```mathematica
<< https://paclets.github.io/PacletServer/Install.wl
```

now works properly.

(There may be other bugs to get paclet handling working, though.)


mmatera:
LGTM

rocky:
There is a lot more that could or should be done.  

Post-import filtering is not handled by Register{Import,Export}. 

There are more importers to add, and there is much more to do for  (metadata) "elements" to bring conformance more in line with WL. 

There is a lot of additional work that could be done here. It will probably be driven by need.

PR #1859, opened 2026-07-12
Implement ByteArrayQ and the usual associated cleanup

rocky:
Add `ByteArrayQ`, and argument checking in the Test class.

`ByteArrayQ` is used in the CodeParser package.


mmatera (approved):
LGTM

PR #1861, opened 2026-07-15
Preserve state of eval.was_boxing.

rocky:
We need to preserve `expression.is_boxing` for Mathics-Module-Trepan to work.


mmatera (approved):
LGTM

PR #1867, opened 2026-07-20
Refactor Rules part 2

rocky:
More Python linting around Rules.

Better document the weirdnesses around our Python Rules code.

### DIFF
--- a/mathics/builtin/list/associations.py
+++ b/mathics/builtin/list/associations.py
@@ -26,7 +26,7 @@ from mathics.eval.list.associations import (
 from mathics.eval.lists import list_boxes
 
 
-class Association(Builtin):
+class Association_(Builtin):
     """
     <url>
     :WMA link:
@@ -57,10 +57,11 @@ class Association(Builtin):
      = {1, 3}
     """
 
-    error_idx = 0
-
     attributes = A_HOLD_ALL_COMPLETE | A_PROTECTED
 
+    error_idx = 0
+
+    name = "Association"
     summary_text = "make an association between keys and values"
 
     def eval_makeboxes(self, rules, f, evaluation: Evaluation):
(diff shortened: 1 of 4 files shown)


mathics/builtin/patterns/rules.py
```diff
@@ -510,20 +512,39 @@ class Rule_(InfixOperator):
...
 
-    def eval_rule(self, elems, evaluation):
+    def eval(self, elems, evaluation: Evaluation):
         """Rule[elems___]"""
+
         num_parms = len(elems.get_sequence())
         if num_parms != 2:
             evaluation.message("Rule", "argrx", "Rule", Integer(num_parms), Integer2)
+
+        # "Rule" is a somewhat generic term in WMA and thus Mathics3.
```
mmatera, line 539: Maybe `Rule` and `RuleDelayed` should not be evaluated, but keep as expressions, until a consumer use it.

mathics/builtin/list/associations.py
```diff
@@ -26,7 +26,7 @@
 from mathics.eval.lists import list_boxes
 
 
-class Association(Builtin):
+class Association_(Builtin):
```
rocky, line 29: In the next PR there will be a `mathics.core.atoms.association.Association` class. I thought I had stashed away the changes, but this one bleeds through.

mathics/builtin/patterns/rules.py
rocky: Yes. This is apparently how things have been working. And we will continue to do this. I note this explicitly in the comments below, starting on line 540, added to this PR.

mmatera: Looking again, what we do here is just to check that Rule and RuleDelayed are invoked with two parameters.

rocky:
After this PR, I'll move on to Association. Then I'll come back and make another round in Rules, if for no other reason, to remove the evaluation parameter on BaseRule objects.

mmatera (approved):
LGTM

PR #1868, opened 2026-07-23
Add an Mathics3 Atom for an `Association` 

rocky:
First cut at converting an Association Expression into a more performant Association Atom.

The Association Atom is an object that wraps a Python dictionary of BaseElements and includes the additional Mathics3 Atom methods typically associated with Atoms.

Since it is based on Python dictionaries, it handles key lookup and update more efficiently.

For now, we still have several eval (complicated and inefficient) functions for an S-Expression-based representation. And that is because there still may be situations where we encounter S-Expression-based Associations. 

Over time, this code will probably be revised to avoid this, and in conjunction with that, the additional eval functions will be removed as they become unnecessary.

More `eval` code has been moved from `mathics.builtin` to `mathics.eval`.

Argument checking added to affected Modules.

Several bugs in Association handling were fixed. In particular, the Ordering properties of Associations with respect to other objects like Strings. Some output routines were not displaying as an Association properly.


mmatera:
@rocky thanks for facing this. LGTM

rocky:
> @rocky thanks for facing this. LGTM

Now we can move on to putting Information into an association, which is what motivated this.

mmatera:
I am curious: have you run some benchmarks to see it there was any performance improvement by handling associations with native dictionaries?

rocky:
> I am curious: have you run some benchmarks to see it there was any performance improvement by handling associations with native dictionaries?

Sadly, no, I haven't run any benchmarks.

rocky:
Also, there are several Association commands that can now be easily implemented, such as AssociationTo, KeyMemberQ, KeySelect, KeyDropFrom, etc. 

I should make an issue for all of that stuff.

PR #1872, opened 2026-07-24
Fix/distribute over list bug

### LINKED ISSUE #1871: Distribute Bug
## Description

Newly added  (by me) `Distribute` function is not quite correct.

# Current and Expected behavior:

We currnetly have: 

```mathematica
In[1]:= Distribute[Table[{1, 2}, {2}], List]
Out[1]= {{1, 2}, {1, 2}}
```

We should get:

```mathematica
Out[1]={{1, 1}, {1, 2}, {2, 1}, {2, 2}}
```

## Priority

low

## Additional context

This form of Distribute is used in CombinatoricaV201.

rocky:
The bug in the code shown by #1871 is that `distribute_across_multiple_position` is only called when the operator appears in multiple argument positions of an expression. But with List as the operator on a list `{1, 2}`, we have multiple elements that are themselves Lists, so we need cartesian product handling.

Fixes #1871

### DIFF
--- a/mathics/builtin/exp_structure/miscelleneous.py
+++ b/mathics/builtin/exp_structure/miscelleneous.py
@@ -42,6 +42,9 @@ class Distribute(Builtin):
     >> Distribute[{a(b + c), d(e + f)}]
      = {a b + a c, d e + d f}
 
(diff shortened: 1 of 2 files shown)


mmatera (approved):
LGTM

PR #1873, opened 2026-07-24
More `Association` refactoring

rocky:
Split out Association into its own module and sections to facilitate filling this out. (And it makes more sense too. Associations are not Lists.)

Add `KeyExistQ`.


mmatera (approved):
LGTM

PR #1874, opened 2026-07-25
Fix gmpy 2.2.2 incompatibility

mmatera:
This PR fixes one incompatibility I found running doctests on gmpy2 v. 2.2.2: in this version,the `is_integer()` seems to be removed from the API, raising an error. Instead of using it, this PR uses an `isinstance` check.


rocky:
LGTM

PR #1876, opened 2026-07-27
Improve encoding strings and get ride MATHICS_CHARACTER_ENCODING="ASCII" in pytests

mmatera:
This PR go over on how the encoding is used in ToString, and other parts of the code. Also, include a proposal on how to make tests more transparent on the system encoding.


mathics/builtin/files_io/files.py
```diff
@@ -1721,10 +1727,17 @@ def eval(self, channel, expr, evaluation: Evaluation):
             return None
 
         exprs = []
+        encoding = from_python_encoding(stream.encoding) or "UTF-8"
```
mmatera, line 1730: Ensure that the encoding exists in mathics.

mathics/format/form/inputform.py
```diff
@@ -176,6 +176,8 @@ def _infix_expression_to_inputform_text(
     num_ops = len(ops_lst)
     for index, operand in enumerate(operands[1:]):
         curr_op = ops_lst[index % num_ops]
+        # In InputForm, operators are encoded as ascii strings:
+        curr_op = UNICODE_CHARACTER_TO_ASCII.get(curr_op, curr_op)
```
mmatera, line 180: In InputForm, use the ASCII form of the operators, to improve the copy and paste consistency among systems.

test/helper.py
```diff
@@ -21,6 +22,21 @@
 data_dir = canonic_os_path(osp.normpath(osp.join(osp.dirname(__file__), "data")))
 
 
+def canonical_encoding(string, enc="ASCII") -> str:
```
mmatera, line 25: This function ensures that the string to be compared is always in the proper encoding.

test/builtin/import_export/test_importexport.py
```diff
@@ -47,6 +47,15 @@ def check_data(
     )
     if expected_data is None:
         expected_data = file_data
+
+    print(
```
mmatera: Here we have a problem: when the system encoding is set to "ASCII", `Export["file.txt", "\[AAcute]", CharacterEncoding->"UTF8"]` is storing the ASCII encoded version of the string.

test/builtin/test_comparison.py
```diff
@@ -13,14 +13,14 @@
             "Null",
             None,
         ),
-        ('"a"==A', "a == A", "comparable expressions"),
+        ('"a"==A', "a ⩵ A", "comparable expressions"),
```
mmatera, line 16: Here, and in many other places, the expect_string must be in the "UTF-8" enconding, which can be converted into ASCII or any other encoding. The opposite is not possible, because the conversion to ASCII is not invertible.

test/builtin/import_export/test_importexport.py
rocky: When I went over Import, I looked a little at Export, and although it didn't seem as big a mess as Import, but still a mess.

I didn't have the heart to even attempt to start to clean this up &mdash; we will have a bit more to do to fill out Import.

So as long as the problem is noted, we can deal with it when we get around to dealing with the larger Export mess.

<strike>In the meantime though, we should remove the "print()" statement.</strike>  Correction: I see this is just in a test.

rocky:
It is good to see progress here. Overall, this is a big step in the right direction.

test/builtin/test_comparison.py
rocky: I was wondering about that. Thanks for the explanation.

mmatera:
What is missing now is to do the same with docpipeline. This comes in a subsequent PR.

rocky:
LGTM

PR #1877, opened 2026-07-28
Add `Import` for Wolfram Language (`.wl`, `.m`) files.

rocky:
It works the same as `Get`, but has the `Import` interface.


rocky:
Apparently, this was code that probably should have been committed with Import was last revised. The code in `mathics.eval` for WL imports is already in there from before.

rocky:
<strike>Later today, I'll promote the test to a doctest and add some commentary around that.</strike> Done.

mmatera (approved):
LGTM

PR #1880, opened 2026-07-29
Cache Python-equivalent values in an Association.

rocky:
Cache Python-equivalent values in an Association. Extend unit test a little.


mmatera (approved):
LGTM

PR #1882, opened 2026-07-30
Load encoding files

mmatera:
This PR start allowing to load character encodings from the WL encoding files. Also, generate Python encodings that can be used to read/write file operations.


rocky:
LGTM. I'm kind of glad to see this happening. 

Algunos comentarios están en español, pero probablemente sea bueno para practicar.

mathics/eval/encoding.py
```diff
@@ -1,15 +1,65 @@
...
+  when the table is (almost) bijective, and we need to *delete* the
+  non-invertible entries from it.
+- Because `.to_python(string_quotes=False)` is called on Strings already
+  parsed by the WL tokenizer, `\\:UUUU` and `\\[Name]` are *already*
+  resolved to real Unicode characters by the time we see them here — no
+  separate named-character lookup table is needed in this context (unlike
+  a standalone script parsing the .wl file as raw text).
 """
 
+from __future__ import annotations
```
rocky: What does this do?

mathics/eval/encoding.py
```diff
@@ -52,112 +102,254 @@
...
-# Map WMA encoding names to Python encoding names
+# This character is used in encoding in WMA, and differs from what we
+# have in the mathics-scanner tables:
+UNICODE_CHARACTER_TO_ASCII.update({operator_to_unicode["Times"]: r" x "})
 
-CHARACTER_ENCODING_MAP = {
-    # see https://docs.python.org/2/library/codecs.html#standard-encodings
-    "ASCII": "ascii",
-    "CP949": "cp949",
-    "CP950": "cp950",
```
rocky, line 62: Home-grown table removed - Yay!.

mathics/builtin/atomic/strings.py
```diff
@@ -272,12 +277,23 @@ def eval_set(self, value, evaluation):
...
+            not isinstance(value, String)
+            or value.value not in CHARACTER_ENCODING_MAP.keys()
+        ):
+            evaluation.message("$CharacterEncoding", "charcode", value)
+            return
+        try:
+            load_encoding_table(value.value, evaluation)
+            evaluation.definitions.set_ownvalue("System`$CharacterEncoding", value)
+        except EncodingNameError:
+            pass
```
rocky: It would be nice to start moving code like this into `mathics.eval.atomic.strings`

(Whether the parameter checking gets moved there is kind of squishy; I don't really know. It would depend on how it might be used other places.)

mathics/builtin/files_io/files.py
```diff
@@ -465,6 +466,10 @@ def eval(self, path: String, evaluation: Evaluation, options: dict):
                 # "noopen" matches WMA. This is nonsensical.
                 evaluation.message("Get", "noopen", encoding)
                 py_encoding = py_current_encoding
+            try:
```
rocky, line 469: Notice that this code is pretty similar to the code mentioned above that was suggested to be moved into `mathics.eval`.  So maybe there should be a function for the `try` block. 

Not sure how to handle the small differences, though.

mathics/eval/strings.py
```diff
@@ -17,23 +17,23 @@
...
-        return String(boxes.to_text(evaluation=evaluation, encoding=encoding))
-    except EncodingNameError:
-        # Mimic the WMA behavior. In the future, we can implement the mechanism
-        # with encodings stored in .m files, and give a chance with it.
-        evaluation.message("Get", "noopen", String("encodings/" + encoding + "." + "m"))
-
-    return String(boxes.to_text(evaluation=evaluation, encoding="Unicode"))
+
+    if encoding not in ("utf8", "utf-8", "unicode"):
+        try:
```
rocky: Another `try` block wrapping `load_encoding_table`; this should definitely be isolated in a function in eval.

test/format/test_charmap_code.py
```diff
@@ -1,17 +1,14 @@
 """
-Tests para wl_charmap_codec.py como toolkit puro: NO lee archivos .wl/.m
-(esa responsabilidad es de encoding.load_encoding_table, via el
-interprete real de Mathics3 -- ver test_encoding.py para tests de
-integracion end-to-end contra los archivos reales). Aca las `Entry`
-se arman a mano, con los mismos valores que ya verificamos que produce
-el interprete real para Klingon.wl / ISO8859-1.m / ISO8859-8.m.
-
-Correr con: python3 test_wl_charmap_codec.py
+Tests for wl_charmap_codec.py as pure toolkit: does not read .wl files
```
rocky, line 2: I looked at _all_ of the translations in commit aaf000c0a9bf0e954be94214334cf8e2b685d0fa. They are pretty good!

mmatera:
> LGTM. I'm kind of glad to see this happening.
> 
> Algunos comentarios están en español, pero probablemente sea bueno para practicar.

Ja, sorry. I just forgot to translate them to English, or just remove them. I was doing experiments with Claude IA to build a part of this code, and as my interactions were in Spanish, it left comments in Spanish. Also it seems it left some "ideosincratic" lines, like add "__futures__" import :)

test/format/test_charmap_code.py
mmatera: Thanks :)

mathics/builtin/atomic/strings.py
mmatera: I guess it is fine to do parameter checks inside the `eval_` methods, when it helps to decide if the parameters are acceptable for some of the implemented `eval_` functions. An `eval_` function should always receive suitable parameters. For this particular case, I do not know either: This just sets the encoding,  so adding an extra call may be excessive.

mathics/builtin/files_io/files.py
mmatera: What I tried to do here was avoid the extra call for "native" encodings. But it probably should be written differently, using an exception.

mathics/builtin/atomic/strings.py
rocky: >  An `eval_` function should always receive suitable parameters. 

What I've come to see from use is that in some cases it can be rather involved to check validity. This parameter is only valid when the data is of a certain form, which is easier to check at the Python level, as opposed to at the S-Expression level.  

Checking for compatibility of matrices in matrix multiplication would be an example of this.

> For this particular case, I do not know either: This just sets the encoding, so adding an extra call may be excessive.

You have to consider also whether somewhere else in the code, it will be desirable to do the equivalent of the WL construct as part of its evaluation.

Here, that would be: 
```mathematica
$CharacterEncoding := ...
```

While this is probably unlikely, I can imagine it might happen that some other built-in, as a side effect sets the $CharacterEncoding.

But rather than spend mental effort on stuff like this, I think it is better just to follow this pattern and not have to worry about such stuff. It also makes it easier for someone else to follow the code and pattern and do things that way.

rocky:
> > LGTM. I'm kind of glad to see this happening.
> > Algunos comentarios están en español, pero probablemente sea bueno para practicar.
> 
> Ja, sorry. I just forgot to translate them to English, or just remove them.

It's better to translate the comments or leave them in than remove them. Especially if AI is assisting, because it can make mistakes and the comments help verify whether the intent was correct or incorrect.

mmatera:
> .

Sure! But some of the comments were related to the prompt I used to generate the code. In any case, I supervised most of the process: I just wanted to see how Claude produces code compatible with stdlib codec functions.

mathics/eval/strings.py
```diff
@@ -17,23 +17,27 @@
...
     try:
         return String(boxes.to_text(evaluation=evaluation, encoding=encoding))
     except EncodingNameError:
-        # Mimic the WMA behavior. In the future, we can implement the mechanism
-        # with encodings stored in .m files, and give a chance with it.
-        evaluation.message("Get", "noopen", String("encodings/" + encoding + "." + "m"))
+        pass
 
-    return String(boxes.to_text(evaluation=evaluation, encoding="Unicode"))
+    # If the encoding does not already exists, try to load it and do it again.
```
mmatera, line 34: I think this way makes more sense...

mathics/builtin/files_io/files.py
mmatera: I am trying to rewrite this code in order to use the stdlib custom encodings that now produce `wl_charmap_codec.py`. It is included in the following PR.

PR #1885, opened 2026-08-01
Remove `in_scope` parameter on `replace_vars()`

rocky:
Remove unused "in_scope" parameter on replace_vars.

### DIFF
--- a/mathics/builtin/atomic/strings.py
+++ b/mathics/builtin/atomic/strings.py
@@ -94,7 +94,7 @@ def _evaluate_match(s, m, evaluation):
     replace = dict(
         (_decode_pname(name), String(value)) for name, value in m.groupdict().items()
     )
-    return s.replace_vars(replace, in_scoping=False).evaluate(evaluation)
+    return s.replace_vars(replace).evaluate(evaluation)
 
 
 def _parallel_match(text, rules, flags, limit):
--- a/mathics/builtin/atomic/strings.py
+++ b/mathics/builtin/atomic/strings.py
@@ -35,7 +35,7 @@ from mathics.eval.encoding import (
 from mathics.eval.strings import eval_StringContainsQ, eval_ToString
 from mathics.settings import SYSTEM_CHARACTER_ENCODING
 
-# covers all of the variations. Here we just give some minimal basics
+# Covers all of the variations. Here we just give some minimal basics
 
 # Data taken from:
 #   https://unicode-org.github.io/cldr-staging/charts/37/summary/root.html
@@ -157,7 +157,7 @@ def mathics_split(patt, string, flags):
     Python >= 3.5, <= X and works as expected for Python >= X, where 'X' is
     some future version of Python (> 3.6).
 
-    For these reasons we implement our own split.
+    For these reasons, we implement our own split.
     """
     # (start, end) indices of splits
     indices = list((m.start(), m.end()) for m in re.finditer(patt, string, flags))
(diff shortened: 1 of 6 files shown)


rocky:
@mmatera Why was the "in_scoping" parameter needed?  Removing it does not affect any of the tests, and it seems that a weird thing to want to do: replace a variable's value independent of its scope.

mmatera:
> @mmatera Why was the "in_scoping" parameter needed? Removing it does not affect any of the tests, and it seems that a weird thing to want to do: replace a variable's value independent of its scope.

I guess the idea was to avoid (in some cases) using variables that are already defined in the scope as replacing variables. Maybe the best now would be to comment out the code instead of just removing it, and leave a comment to check if we can fully get ride of it.

mathics/builtin/scoping.py
```diff
@@ -517,7 +518,7 @@ def eval(self, vars, expr, evaluation: Evaluation):
             if new_def is not None:
                 evaluation.definitions.set_ownvalue(new_name, new_def.copy())
             replace[name] = Symbol(new_name)
-        new_expr = expr.replace_vars(replace, in_scoping=False)
```
mmatera, line 521: This is an explicit call with "in_scoping=False".

mathics/builtin/atomic/strings.py
```diff
@@ -94,7 +94,7 @@ def _evaluate_match(s, m, evaluation):
     replace = dict(
         (_decode_pname(name), String(value)) for name, value in m.groupdict().items()
     )
-    return s.replace_vars(replace, in_scoping=False).evaluate(evaluation)
+    return s.replace_vars(replace).evaluate(evaluation)
```
mmatera, line 97: And this is the other case. Let's make some tests to check against WMA and Mathics, to see the change in the behavior.

rocky: > And this is the other case. Let's make some tests to check against WMA and Mathics, to see the change in the behavior.

I wouldn't know how to.

mmatera: Later I can try to write some examples. Just let's keep comments to see if we need to reinstall this in another way.

rocky: For now, let's mark this as a draft. #1888 has what remains, and one more small omission I just noticed.

mmatera: OK. I check and for this case, it can safelty removed: the parameter is just used for Block, Module and Width, not on StringFind/Replace...

mathics/builtin/scoping.py
mmatera: Here, the parameter does nothing, because replace is empty...

mathics/core/expression.py
```diff
@@ -1672,39 +1672,18 @@ def descend(expr):
                 expr = Expression(head, *expr._elements)
             return expr, new_applied[0]
 
-    def replace_vars(
-        self, vars, options=None, in_scoping=True, in_function=True
-    ) -> "Expression":
+    def replace_vars(self, vars, options=None, in_function=True) -> "Expression":
```
mmatera, line 1674: @rocky, after looking in detail, it seems safe to remove the `in_scoping` parameter.

mmatera:
LGTM

rocky:
> LGTM

For isolation, I'll merge in #1888 and then rebase this off of master.  Then we can commit this smaller thing, and the history will have two separate things going on.

mathics/builtin/atomic/strings.py
rocky: Thanks for investigating, testing, and for the information.

rocky:
@mmatera another chance to reconsider this, or what has gone in for #1888 which was split off from this.

PR #1866, opened 2026-07-19
`Rule` refactor part1

rocky:
* Class `mathics.core.rules.Rule` is now `mathics.core.RewriteRule`.
* Add a `is_rule()` function to replace `expr.has_form(...)`
* Remove the `system` parameter on Rule creation.
* Go over docstring comments.

PR #1887, opened 2026-08-01
Improve Monomial Ordering

rocky:
Fixes #1866

### DIFF
--- a/mathics/core/keycomparable.py
+++ b/mathics/core/keycomparable.py
@@ -117,8 +117,23 @@ class Monomial:
                 other_exps[var] -= dec
                 if not other_exps[var]:
                     del other_exps[var]
-        self_exps = sorted((var, exp) for var, exp in self_exps.items())
-        other_exps = sorted((var, exp) for var, exp in other_exps.items())
+
+        # Sort variables so that symbols later in the alphabet (e.g. 'z')
+        # are considered higher-precedence (variables) than earlier ones
+        # (e.g. 'a'). Mathematica treats later symbols as variables and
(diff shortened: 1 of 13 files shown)


mmatera (approved):
LGTM

mathics/core/keycomparable.py
```diff
@@ -117,8 +117,23 @@ def __cmp(self, other) -> int:
...
+
+        # Sort variables so that symbols later in the alphabet (e.g. 'z')
+        # are considered higher-precedence (variables) than earlier ones
+        # (e.g. 'a'). Mathematica treats later symbols as variables and
+        # earlier symbols as coefficients when comparing monomials, so
+        # compare variable names in reverse alphabetical order here.
+        # If `var` is not a plain string (e.g. a Symbol object), fall back
+        # to its string representation for ordering.
+        def _var_key(pair):
+            var, _ = pair
```
mmatera, line 126: Maybe it worth leaving in the comment, or in some other place,  that the name of the variable has priority over the associated exponent. For example,
```
In[1]:= Sort[{a^3 ,z^2}]                                                        

          3   2
Out[1]= {a , z }


In[2]:= Sort[{a^2 ,z^3}]                                                        

          2   3
Out[2]= {a , z }

```

rocky: Yes, it is worth mentioning. All of this is rather weird and non-intuitive.  A challenge is where and how to describe this. 

The monomial takes precedence over exponents is not part of the monomial key comparable property since a monomial doesn't contain an exponent. It is part of the `element_order()` description. So I have updated that and referred back to this comment here for the monomial part.

PR #1889, opened 2026-08-02
`mathics.core.eval` routines should not have `Builtin` or `self` as a parameter.

rocky:
Make a pass over eval_assignment to remove Builtin parameter. Annotate more types, disambiguate names.

Add argument checking on Set-like builtin functions.


mmatera (approved):
LGTM

PR #1890, opened 2026-08-07
Parse URLs from Python Builtin docs.

rocky:
Also, handle Mathics3 Modules in definition lookup. 
And the usual cleanup:
  * Put in alphabetical order the methods in `BasePattern`
  * Strengthen/correct some annotation types
  
Handle `Information[xxx,  "ObjectType"]``


rocky:
This too is getting large and merging now. 

Part 3, to come, will build on this and promises to be equally large. Now that we have the components that go into the InformationData or InformationGridData wrapper around an Association, there is the final step to do this.

However, this will mean ripping out a lot of existing code and formatting routines.

PR #1892, opened 2026-08-15
makeboxes eval_infix should use always the internal utf encoding.

mmatera:
MakeBoxes should produce strings that do not depend on `$SystemCharacterEncoding`. The encoding must be taken into account just at the rendering step (when converting the Box expression into a string or LaTeX output).   This PR fixes the remaining place where this pattern was not followed and produced errors when call `pytest  /test/format/test_format.py` with "ASCII" as the system encoding. With this adjustment, tests pass regardless of the system character encoding.


mathics/builtin/box/layout.py
```diff
@@ -688,8 +688,8 @@ def elements(self):
         if self._elements is None:
             # self.box_options
             self._elements = elements_to_expressions(
+                self,
```
rocky, line 691: Pyrefly is showing a number of type errors around this section of code. But that can be addressed some other time.

mathics/builtin/box/layout.py
rocky, line 137: I think `self.boxes` should be `self.inner_box` now.

But again, this is not central to this PR; just something noticed.

rocky:
LGTM. I'm glad we are starting to address some of the many boxing issues we have.

PR #1893, opened 2026-08-16
Improve Character encoding for ToString and File Handlers

mmatera:
This PR improves the encoding in file operations, by
*  registering custom WMA encodings as Python character encodings. 
* organizing the mathics.eval.encoding and mathics.eval.wl_charmap_codec into a package.
* `FromCharacterCode` and `ToCharacterCode` loads the encoding if needed.
* The list of `$CharacterEncodings` now takes into account the files included in `SystemData/CharacterEncodings`
* Improve compatibility in the interpretation of  WL encoding definition files:
    *  (reversibility) is now ignored, just as in WMA.
    * `None`   in the second field is interpreted as a shorthand for associate the charcode with a displaced charcode.
* Tests were adjusted accordingly.

The current implementation only works with 7Bit and 8Bit  "Custom" encodings. For 16Bit (variable-length) encodings, we rely on Python's implementation.

### DIFF
--- a/mathics/eval/encoding/__init__.py
+++ b/mathics/eval/encoding/__init__.py
@@ -1,0 +1,45 @@
+"""
+Character Encodings
+
+This package contains the high-level and low-level functions used to work with different character encodings,
+and contains two sub-modules:
+
+* encoding: high-level interface to list, load and use Python-native and WL-custom character encodings.
+* wl_charmap_codec: low-level interface to build and register Python codecs from information gathered
+  from WL custom character encoding files, under SystemData/CharacterEncoding/*.wl
+
+"""
+
+from .encoding import (
+    CHARACTER_ENCODING_MAP,
+    ROOT_DIR,
+    WMA_DECODE_TABLES,
+    WMA_UNICODE_CHARACTER_MAPS,
+    EncodingNameError,
+    available_character_encodings,
+    decode_bytes_value,
+    encode_string_value,
+    from_python_encoding,
+    get_encoding_table,
+    load_encoding_table,
+    to_python_encoding,
+)
+from .wl_charmap_codec import _REGISTERED_TABLES, register_escape_error_handler
+
+__all__ = [
+    "CHARACTER_ENCODING_MAP",
+    "EncodingNameError",
+    "ROOT_DIR",
+    "WMA_DECODE_TABLES",
+    "WMA_UNICODE_CHARACTER_MAPS",
+    "_REGISTERED_TABLES",
+    "available_character_encodings",
+    "decode_bytes_value",
+    "encode_string_value",
+    "from_python_encoding",
+    "get_encoding_table",
+    "load_encoding_table",
+    "register_escape_error_handler",
+    "to_python_encoding",
+    "to_python_encoding",
+]
--- a/mathics/eval/encoding/__init__.py
+++ b/mathics/eval/encoding/__init__.py
@@ -41,5 +41,4 @@ __all__ = [
     "load_encoding_table",
     "register_escape_error_handler",
     "to_python_encoding",
-    "to_python_encoding",
 ]
--- a/mathics/eval/encoding/encoding.py
+++ b/mathics/eval/encoding/encoding.py
@@ -40,7 +40,8 @@ Design notes (see discussion that led to this rewrite):
 
 import codecs
 import os
-from typing import Final, List
+from collections.abc import Iterator
+from typing import Final
 
 from mathics_scanner.characters import UNICODE_CHARACTER_TO_ASCII
 
@@ -152,7 +153,7 @@ class EncodingNameError(Exception):
     pass
 
 
-def available_character_encodings() -> List[str]:
+def available_character_encodings() -> Iterator[str]:
     """
     List all the available character encodings, including the
     default Python encodings and the custom encodings defined in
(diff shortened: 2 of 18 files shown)


mathics/eval/encoding/encoding.py
```diff
@@ -148,6 +152,23 @@ class EncodingNameError(Exception):
     pass
 
 
+def available_character_encodings() -> List[str]:
```
rocky: `List` is not quite right here. Either cast `map()` to `list(map(...))` or I think better return an `Iterator` and change `from typing import List` to `from typing import Iterator`.

mathics/eval/encoding/encoding.py
```diff
@@ -316,6 +347,12 @@ def load_encoding_table(encoding, evaluation):
         None,
     )
     if etl is None or not etl.has_form("List", 2):
+        print(
```
rocky: Do we really want to use a print? This could instead be stored as a string in the raised message? 

Not sure. Your thoughts?

mathics/eval/encoding/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+    "WMA_UNICODE_CHARACTER_MAPS",
+    "_REGISTERED_TABLES",
+    "available_character_encodings",
+    "decode_bytes_value",
+    "encode_string_value",
+    "from_python_encoding",
+    "get_encoding_table",
+    "load_encoding_table",
+    "register_escape_error_handler",
+    "to_python_encoding",
```
rocky, line 43:
```suggestion
```
Remove duplicate entry.

mathics/eval/encoding/wl_charmap_codec.py
```diff
@@ -0,0 +1,460 @@
...
+    Check that all the values in a substitution table are pure ASCII.
+    """
+    offending = {
+        ch: repl for ch, repl in substitution.items() if any(ord(c) > 127 for c in repl)
+    }
+    if offending:
+        sample = dict(list(offending.items())[:5])
+        raise ValueError(
+            f"{table_name}: {len(offending)} valores de sustitucion no son "
+            f"ASCII puro (ejemplos: {sample!r}) -- el resultado ya no seria "
```
rocky: ¿Cómo?

mathics/eval/encoding/encoding.py
```diff
@@ -148,6 +152,23 @@ class EncodingNameError(Exception):
...
+    """
+    List all the available character encodings, including the
+    default Python encodings and the custom encodings defined in
+    SystemFiles/CharacterEncodings/*.wl
+    """
+    default_encodings = set(CHARACTER_ENCODING_MAP.keys())
+    encodings = default_encodings.union(
+        {
+            filename[:-3]
+            for filename in os.listdir(f"{ROOT_DIR}/SystemFiles/CharacterEncodings/")
```
rocky:
```suggestion
            for filename in os.listdir(os.path.join(ROOT_DIR), "SystemFiles/CharacterEncodings")
```

Is more portable than hardcoding "/"?

mathics/eval/encoding/encoding.py
```diff
@@ -344,6 +404,21 @@ def load_encoding_table(encoding, evaluation):
...
+    try:
+        py_name = register_codec_from_tables(encoding, decoding_table, encode_table)
+    except ValueError as e:
+        evaluation.message("$CharacterEncoding", "charfile", String(encoding))
+        raise EncodingNameError(str(e)) from e
+    CHARACTER_ENCODING_MAP[encoding] = py_name
+    REVERSE_CHARACTER_ENCODING_MAP[py_name] = encoding
+
 
 def to_python_encoding(encoding) -> str:
```
rocky:
```suggestion
def to_python_encoding(encoding) -> str | None:
```

mathics/eval/encoding/encoding.py
```diff
@@ -39,21 +39,25 @@
 """
 
 import codecs
-from typing import Final
+import os
+from typing import Final, List
```
rocky:
```suggestion
from typing import Final
```

rocky:
There is a lot of very good and needed improvements here. Some small stuff.

mathics/eval/encoding/encoding.py
mmatera: This was just for debug, because the messages are not very explicit about what is wrong with the file. Now I think it can be just removed.

mathics/eval/encoding/wl_charmap_codec.py
mmatera: @rocky, please check if the translation makes sense for you. The idea is just that in any 8-bit encoding, the characters with codes <128 are invariant, while the codes >=128 correspond to different Unicode characters.

mathics/eval/encoding/encoding.py
rocky: Something I now keep in mind. In some cases, it might be useful to keep the print statement commented out. This way, if there's a problem in the code in the future, we can recover useful debug information.

mathics/eval/encoding/wl_charmap_codec.py
rocky: This is great! (And as far as I can tell matches the other explanation.) Thanks!

rocky:
LGTM - and thanks!

mathics/eval/encoding/encoding.py
mmatera: Good. I restored it.

rocky:
With this PR, I realize that `/mathics-core/mathics/SystemFiles/CharacterEncodings/README.rst` should be changed. When it was written, the files in that directory weren't used.

Also, we can now add many more encodings.

I guess another PR for this.

PR #1894, opened 2026-08-17
In eval_Get, AbortInterrupt and MessageException must be handled on each query

### LINKED ISSUE #1884: Infinite recursion found in Rubi-based `SetDelayed` assignments for `EqQ` and `Star`
## Description

Rubi is not loading, failing with: 

```
$RecursionLimit::reclim: Recursion depth of xxx  exceeded.
``` 

It is failing in `IntegrationUtilityFunctions.m` which is read before any Integration rules are read in.

Run/Import/Get this file `star-bug.wl`: 

```mathematica 
(*Extracted from Rubi IntegrationUtilityFunctions.m Integration Utility Functions*)

(* If u-v equals 0, EqQ[u,v] returns True; else it returns False. *)

EqQ[args___] := Null /; EqQ[args]

(* Star[u,v] displays as u*v, and returns the product of u and v with u
   distributed over the terms of v." *)

Star[u_,v_] := (
  Message[Star::error];
  0 ) /;
EqQ[u,0]

Star[u_,Star[v_,w_]] :=
  Star[u*v,w]
Print["Succeeded"]
```

The following is output instrumented to show recursive calls to the internal function `inc_recursion_depth` of Evaluation
```mathematica
In[1]:= Get["star-bug.wl", Trace->True]
Reading file: ./star-bug.wl

    2: (* If u-v equals 0, EqQ[u,v] returns True; else it returns False. *)

    4: EqQ[args___] := Null /; EqQ[args]

    6: (* Star[u,v] displays as u*v, and returns the product of u and v with u
        distributed over the terms of v." *)
    9: Star[u_,v_] := (
  Message[Star::error];
  0 ) /;
EqQ[u,0]

   14: Star[u_,Star[v_,w_]] :=
     Star[u*v,w]

(u:_) Star(v:_) Star (w:_) := u * v Star w

(u:_) Star(v:_) Star (w:_) := u * v Star w

(u:_) Star(v:_) Star (w:_) := u * v Star w
```

## Expected behavior

Should print "Succeeded"

## Additional context

As stated above,  Rubi is broken as a result of this bug.  I've Mathics 10.0 and 9.0, and these both give a maximum recursion error exception as well.

mmatera:
This should fix #1884. In any case, it seems there is a problem in the example that is also shown in WMA

### DIFF
--- a/test/data/star-bug.m
+++ b/test/data/star-bug.m
@@ -12,6 +12,11 @@ Star[u_,v_] := (
   0 ) /;
 EqQ[u,0]
 
+(*
+This definition produces a recursion error:
+Since `Star` was already defined, the assignment
+evaluates the LHS recursively until the raise a RecursionLimit error...
+*)
 Star[u_,Star[v_,w_]] :=
   Star[u*v,w]
 "OK!"
--- a/test/data/star-bug.m
+++ b/test/data/star-bug.m
@@ -1,0 +1,17 @@
+(*Extracted from Rubi IntegrationUtilityFunctions.m Integration Utility Functions*)
+
+(* If u-v equals 0, EqQ[u,v] returns True; else it returns False. *)
+
+EqQ[args___] := Null /; EqQ[args]
+
+(* Star[u,v] displays as u*v, and returns the product of u and v with u
+   distributed over the terms of v." *)
+
+Star[u_,v_] := (
+  Message[Star::error];
+  0 ) /;
+EqQ[u,0]
+
+Star[u_,Star[v_,w_]] :=
+  Star[u*v,w]
+"OK!"
--- a/mathics/eval/files_io/files.py
+++ b/mathics/eval/files_io/files.py
@@ -19,11 +19,12 @@ import mathics
 import mathics.core.parser
 import mathics.core.streams as streams
 from mathics.core.atoms import Integer, String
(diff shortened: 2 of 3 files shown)


test/data/star-bug.m
```diff
@@ -0,0 +1,22 @@
...
+  Message[Star::error];
+  0 ) /;
+EqQ[u,0]
+
+(*
+This definition produces a recursion error:
+Since `Star` was already defined, the assignment
+evaluates the LHS recursively until the raise a RecursionLimit error...
+*)
+Star[u_,Star[v_,w_]] :=
```
mmatera, line 20: To work properly, the LHS should be wrapped in `HoldPattern`.

test/data/star-bug.m
```diff
@@ -0,0 +1,22 @@
+(*Extracted from Rubi IntegrationUtilityFunctions.m Integration Utility Functions*)
+
+(* If u-v equals 0, EqQ[u,v] returns True; else it returns False. *)
+
+EqQ[args___] := Null /; EqQ[args]
```
mmatera, line 5: This does not make sense: A good definition should be something like ```EqQ[args__]:=Null; EqQ[u_,v_]:=u==v```.

rocky:
LGTM

PR #1895, opened 2026-08-17
fix Information

mmatera:
This PR fixes an error that slipped into the Information refactor. Without this, `?? Plus` raises an error.


rocky:
Thanks!

PR #1897, opened 2026-08-18
More on makeboxes

mmatera:
Another round of test-driven small fixes.


mathics/builtin/box/layout.py
```diff
@@ -564,6 +564,7 @@ class StyleBox(BoxExpression):
     """
 
     options = {
+        "ImageSizeMultipliers": "Automatic",
```
mmatera, line 567: without a true "option", the `eval_` method that uses an option parameter fails...

mmatera:
With these changes, we pass from 46 to 28 xfailed tests in test_makeboxes, which is a kind of progress...

rocky:
LGTM

PR #1900, opened 2026-08-23
Add builtin `Refine`

rocky:
Start adding Built-in Function Refine.

To do this, we make use of SymPy's `sympy.assumptions.assume.AppliedPredict` and conversion from WMA to SymPy for this.
Currently, we handle relational expressions (==, >, etc.) and element domains (Reals, Integers, etc.).

There is much more to do, but this is a start.

We also start allowing "has_form" to accept a Symbol parameter _in addition_ to a string. Over time, I expect a PR to convert from `str`-argument `has-form` to Symbol-argument `has_form`.

`Assume` and `$Assumption` have been moved from the generic `mathics.builtin.arithmetic` to `mathics.builtin.assumptions` which is where the new `Refine` lives.

The motivation for this work stems from the fact that Rubi uses and calls `Refine`.

Some type checking was made more stringent.


mmatera:
Awesome! This has been waiting for a long time...

rocky:
For reference, there are bugs in SymPy that may be addressed and will help here. See https://github.com/sympy/sympy/issues/30324

What's cool about the PR to fix this, https://github.com/sympy/sympy/pull/30325, is that it is being done in SymPy by _improving its SAT solver!_  Thanks, @tkim601 !

One other thing of interest to me is that while I really like @ad-si [Woxi](https://github.com/ad-si/Woxi), as best as I can tell, right now it doesn't have access to such a beast yet. It has to do all of the algebraic manipulation inside the Woxi code. Do I have this correct, Adrian?  Possibly Woxi can find a similar SAT solver for such purposes? 

But it shows the differences in approaches right now. (Possibly over time the two projects will converge).

ad-si:
Yeah, no external CAS or SAT solver. It all has to be written in Rust in the repo, partly on purpose since the same binary runs on wasm32 for the playground and JupyterLite kernel.

Refine, Assuming and $Assumptions are already in, but it's just propagation of signs, bounds, domains and parity, not a decision procedure. Your example from the sympy issue works:

```sh
$ woxi eval 'Assuming[x >= 0 && y < 0, If[TrueQ[Refine[x - y > 0]], Refine[Sqrt[x^2 y^2]], 0]]'
-(x*y)
```

Same as wolframscript, but only because that's a case I handle. Messier combinations of hypotheses will fail where yours won't.

A SAT solver would help me most in `SatisfiableQ`, `TautologyQ`, and `BooleanMinimize`, which currently walk the whole truth table and give up past 24 variables. I should check if https://github.com/sarsko/CreuSAT or similar could be suitable.

On converging: Rubi 4.17.3.0 loads unmodified in Woxi and integrates, 20/30 on my benchmark against wolframscript, and none of the failures are `Refine` related.

rocky:
> A SAT solver would help me most in `SatisfiableQ`, `TautologyQ`, and `BooleanMinimize`, which currently walk the whole truth table and give up past 24 variables. I should check if https://github.com/sarsko/CreuSAT or similar could be suitable.
> 
:+1: 

> On converging: Rubi 4.17.3.0 loads unmodified in Woxi and integrates, 20/30 on my benchmark against wolframscript, and none of the failures are `Refine` related.

Woxi certainly is impressive!

PR #1904, opened 2026-08-24
Simplify/generalize Sympy expression to Sympy predicate

rocky:
Now handles Or, Not and swapped relational expressions

to_sympy() function added for Element.

All System Symbols are now marked Final.

A couple more tests added. (But many more are needed).


rocky:
(There will be a couple more iterations of this.)

mathics/core/convert/sympy.py
```diff
@@ -4,18 +4,46 @@
 Converts expressions from SymPy to Mathics3 expressions.
 Conversion to SymPy is handled directly in BaseElement descendants.
 """
-from typing import TYPE_CHECKING, Dict, List, Optional, Sequence, Tuple, Union, cast
+
+from typing import (
```
mmatera, line 8: This starts to be quite large. Maybe in another round, we should split this into submodules.

mmatera (approved):
LGTM

mathics/core/convert/sympy.py
rocky: Noted.

rocky:
In the next round, I will split out the SymPy predicate code. And probably more tests. 

In the round _after_ that, maybe we'll be able to get to Assume!

PR #1905, opened 2026-08-24
Handle patterns of the form H[a_Type,a_Type,...]

mmatera:
This PR handles another corner case of patterns with typed elements. Now, more tests pass.

### DIFF
--- a/test/timings/test_patterns.py
+++ b/test/timings/test_patterns.py
@@ -3,8 +3,24 @@ from test.helper import session
 
 import pytest
 
+from mathics.core import pattern as pattern_module
 from mathics.core.pattern import BasePattern
 from mathics.core.symbols import strip_context
+
+
+def _hold_pattern(pattern_text):
+    """Build a BasePattern from an unevaluated held expression."""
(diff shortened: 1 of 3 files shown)


test/timings/test_patterns.py
```diff
@@ -336,3 +335,392 @@ def impl():
             _ = precedence_pattern.pattern_precedence
 
     benchmark.pedantic(impl, rounds=10, iterations=2)
+
+
+# ---------------------------------------------------------------------------
+# 7. get_match_candidates() vs get_match_candidates_count()
```
rocky, line 341: 7. Smells like AI. <strike>Probably should remove the 7. Since I am not seeing 1. to 6.</strike>. I see them now.  Not complaining, but noting.

rocky:
LGTM

test/timings/test_patterns.py
mmatera: Indeed, I am trying to use IA (Claude Sonnet) to detect what is going wrong with this module. Most of the tests and benchmarks it suggested sound reasonable to me, so I included them.

PR #1906, opened 2026-08-24
Split out Mathics3 to SymPy predictate conversion

rocky:
Split out Mathics3 to SymPy predictate conversion

More tests.


rocky:
After this, I think we can finally move on to Assumptions.

mmatera (approved):
LGTM

rocky:
Ok - now to assuming (when I get a chance).

PR #1907, opened 2026-08-24
Improve expression cache documentation

mmatera:
* Improve the documentation of ``ExpressionCache`` and ``Structure`` classes. 
* Improve `ExpressionPointer.replace()` to notify parent expressions that the cache and ExpressionProperties got invalidated.

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -174,6 +174,72 @@ class BoxError(Exception):
 
 
 class ExpressionCache:
+    """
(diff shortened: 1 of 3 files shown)


rocky:
LGTM.

However, I would like to see this removed or severely reduced/isolated.

mmatera:
> LGTM.
> 
> However, I would like to see this removed or severely reduced/isolated.

Me too, but to do that, I first want to understand it better.

PR #1910, opened 2026-08-26
Split order and orderless patterns

mmatera:
This PR is about reorganizing the large module `mathics.core.pattern` into submodules. Also, to split the code in ExpressionPattern into specialized classes that handle the Ordered and Orderless cases.

Benchmarks comparisons in https://mathics3.github.io/mathics-core/cache/


mathics/core/pattern/common.py
```diff
@@ -0,0 +1,180 @@
+# cython: language_level=3
+# cython: profile=False
+# -*- coding: utf-8 -*-
+"""
+Common auxiliar functions for pattern matching.
```
rocky:
```suggestion
Common auxiliary functions for pattern matching.
```

mathics/core/pattern/common.py
```diff
@@ -0,0 +1,180 @@
...
+from mathics.core.evaluation import Evaluation
+from mathics.core.expression import Expression
+from mathics.core.systemsymbols import SymbolDefault, SymbolOptional
+
+from .base import AtomPattern, BasePattern, ExpressionPattern
+
+
+def _is_option_like(candidate: BaseElement) -> bool:
+    """Same shape check OptionsPattern.get_match_candidates() uses."""
+    return candidate.has_form(("Rule", "RuleDelayed"), 2) or candidate.has_form(
```
rocky: I'd appreciate it if we could start using the Symbols instead of strings in new `has_form()` uses.

And things like candidate_has_form("List", None) is really now `isinstance(candiate, ListExpression)`

mathics/core/pattern/orderless.py
```diff
@@ -0,0 +1,436 @@
+# cython: language_level=3
```
rocky, line 1: Experience has shown that using `# cython:`  hasn't speed things up in any significant way. In fact, I've found it a pain.

It is possible that _thoughtful_ use of `@cython.locals` and `@cython.cfunc`  (replacing the older `cdef` and `cpdef`) might help a lot.  But that's not this. 

In fact, if I read the benchmarks correctly, right now I am not seeing any _significant_ speedups from this code over what was there before.  And if there is a speedup, I am not seeing it attributed to cython.

So please let's not complicate things in advance of using Cython thoughtfully.

rocky:
This is a major rewrite (looks like with AI's help, which is probably better than the alternative). 

I suppose we should try it. 

But it tends to push us away from a algorithmy [Trie](https://en.wikipedia.org/wiki/Trie)-based approach, but that would also be tricky to get right.

mmatera:
Actually, I tried to use the IA to do this, but in the end I had to do it mostly by hand. Claude Sonnet 4 helped me to rewrite the docstrings, organize tests and detect some flaws in the previous code.

mmatera:
> But it tends to push us away from a algorithmy [Trie](https://en.wikipedia.org/wiki/Trie)-based approach, but that would also be tricky to get right.

Actually, most of the changes here are related to decoupling the existing handle-every-case-in-a-single-method classes into more specialized classes. The pattern match algorithm is almost the same as before, except that the Ordered/Orderless branches are now split into separate modules.

mathics/core/pattern/orderless.py
mmatera: This is just copy and paste. I plan to see if these modules would compile with Cython later, but it is not the goal here. BTW, I tried to test the code with USE_CYTHON, and at least in the current master, it doesn't work (it produced compilation failures).

mathics/core/pattern/common.py
mmatera: This form with two heads is not working right now with Symbols...

mmatera: Regarding `isinstance(candiate, ListExpression)`, there could be corner cases in which we make an expression of the form `List[...]` and that the expression is not evaluated to `ListExpression`.

rocky: > Regarding `isinstance(candiate, ListExpression)`, there could be corner cases in which we make an expression of the form `List[...]` and that the expression is not evaluated to `ListExpression`.

If or when that happens, then let us _fix_ it rather than keep this baggage.

rocky: > This form with two heads is not working right now with Symbols...

My bad. 

But couldn't we replace this with `is_rule()` from `mathics.core.rules`? 

(And fixing `has_form()` to handle a tuple or list of symbols would be nice too.)

rocky:
> > But it tends to push us away from a algorithmy [Trie](https://en.wikipedia.org/wiki/Trie)-based approach, but that would also be tricky to get right.
> 
> Actually, most of the changes here are related to decoupling the existing handle-every-case-in-a-single-method classes into more specialized classes. The pattern match algorithm is almost the same as before, except that the Ordered/Orderless branches are now split into separate modules.

Right. At this point, should we rewrite this or reconsider everything using ideas from that project that focused on this aspect?

mathics/core/pattern/orderless.py
rocky: Then let's remove Cython altogether, since this is a "rewrite".

The Cython regression test is slow. It is not clear that anyone is actually using it or that it is yielding much performance gain. It certainly can't be used from WebAssembly.

When we can simplify, let's simplify.

rocky: > Then let's remove Cython altogether, since this is a "rewrite".
> 

My take on the whole Cython mess. The Mathics code was extremely slow due to several _conceptual_ flaws and sloppy or misguided coding.  Rather than use better practices, the thought then was: hey, I know, let's get Cython to magically fix everything up!

I do not doubt that Cython, if used judiciously and properly, _could_ have a big impact.  But that would mean having a better understanding of Cython and specifically what's wrong in Mathics3 that Cython can reasonably address. 

And _that_ should come first.

mathics/core/pattern/common.py
mmatera: I am working on a fix for has_form too, but I am trying to do not mix too much stuff, in an already large refactor.

mathics/core/pattern/orderless.py
mmatera: Checking were Cython does not run or produces slower code allows me to find some of these inconsistencies. At the moment, I am using Cython as another debugging tool.

mmatera:
> Actually, most of the changes here are related to decoupling the existing handle-every-case-in-a-single-method classes into more specialized classes. The pattern match algorithm is almost the same as before, except that the Ordered/Orderless branches are now split into separate modules.

The first thing I am trying to check if this decoupling does not make performance worse. According to the tests, it doesn't and should help to analyze the code.

mathics/core/pattern/orderless.py
rocky: > Checking were Cython does not run or produces slower code allows me to find some of these inconsistencies. At the moment, I am using Cython as another debugging tool.

There are better tools that are more targeted for finding inconsistencies.

mathics/core/pattern/common.py
rocky: Can we slip the small stuff first?  It helps to work bottom-up rather than doing a major refactor before doing all the stuff that should have been addressed at the core.

mmatera: OK, #1923 introduces the required changes.

mathics/core/pattern/orderless.py
mmatera: Yes, but I see that to detect inconsistencies in type conversions, it works better than mypy...

rocky: Have you tried `pyrefly`?

mmatera: Nop. But it also works with implicit typing, right?

mmatera:
> This is a major rewrite (looks like with AI's help, which is probably better than the alternative).
> 
> I suppose we should try it.
> 
> But it tends to push us away from a algorithmy [Trie](https://en.wikipedia.org/wiki/Trie)-based approach, but that would also be tricky to get right.

Before changing the algorithm, I would like to make the current implementation as clear as possible.

mathics/core/pattern/orderless.py
rocky: > Nop. But it also works with implicit typing, right?

Yes. And it is _designed_ to do for this kind of thing explicitly. If there is a problem, it is a bug; report it. 

With Cython, if there is a bug or typing problem and it just happens not to be a problem with Cython, it is probably it is questionable if Cython would fix it since this kind of thing is not core to what Cython is about.

rocky:
> > This is a major rewrite (looks like with AI's help, which is probably better than the alternative).
> > I suppose we should try it.
> > But it tends to push us away from a algorithmy [Trie](https://en.wikipedia.org/wiki/Trie)-based approach, but that would also be tricky to get right.
> 
> Before changing the algorithm, I would like to make the current implementation as clear as possible.

Ok. But I was under the impression that Boxing and Form handling were next on the list of things.

mmatera:
> Trie

I was looking at the Wikipedia entry on Trie, and also consulted Claude IA about it: Trie seems be about to find a pattern that matches an expression among a bunch of possible candidate patterns. This is similar to what `rewrite_apply_eval_step` does, or what `Dispatch` should do. This PR is about how to represent and implement individual patterns. Looking at the code in master, it becomes clear that a single ExpressionPattern class handling all possible behaviors is slow and harder to analyze. 
So, I disagree with this push to use a Trie implementation: a clearer implementation of individual patterns would help to build the dispatch tables.

mmatera:
> > > This is a major rewrite (looks like with AI's help, which is probably better than the alternative).
> > > I suppose we should try it.
> > > But it tends to push us away from a algorithmy [Trie](https://en.wikipedia.org/wiki/Trie)-based approach, but that would also be tricky to get right.
> > 
> > 
> > Before changing the algorithm, I would like to make the current implementation as clear as possible.
> 
> Ok. But I was under the impression that Boxing and Form handling were next on the list of things.

Regarding this, I take this because, looking at the tests for handling MakeBoxes, I found a couple of them that fail at the level of the behavior of the pattern matching. This would also help to debug it.

rocky:
> > Trie
> 
> I was looking at the Wikipedia entry on Trie, and also consulted Claude IA about it: Trie seems be about to find a pattern that matches an expression among a bunch of possible candidate patterns. This is similar to what `rewrite_apply_eval_step` does, or what `Dispatch` should do. This PR is about how to represent and implement individual patterns. Looking at the code in master, it becomes clear that a single ExpressionPattern class handling all possible behaviors is slow and harder to analyze. So, I disagree with this push to use a Trie implementation: a clearer implementation of individual patterns would help to build the dispatch tables.

I haven't looked at this in detail, so I don't know. For dispatch to be fast, usually there is a particular data structure that facilitates this. 

An alternate approach for pattern matching and selection is to have a series of functions or methods, more or less like the way parsing works in Mathics3. (Note there is also a table-driven approach to parsing as well) .

As far as I can tell, the functions or methods approach is how Mathics has been pattern matching. It's not wrong, just different.  Depending on what needs to be pattern matched and how the functions have been written, there are definitely situations where using the data-structure and table approach has been a big win.

rocky:
By the way, one thing that looks wrong about our selection in rewrite/eval is that we are looking at Options-like rules for rewriting, and as far as I can tell, these are never applied. So Options-like rules should be segregated from rewrite rules. 

Here is an example: 

```mathematica
In[1]:= foo ->"bar"]
Out[1]= foo ⇾ bar
In[2]:= foo
Out[2]= foo
```

If there is never a possibility in evaluation that option rule foo->"bar" will be used in an expression evaluation, then it should not be considered or appear in the list of rules that are considered in application. And we are currently doing this.

rocky:
LGTM

mmatera:
> By the way, one thing that looks wrong about our selection in rewrite/eval is that we are looking at Options-like rules for rewriting, and as far as I can tell, these are never applied. So Options-like rules should be segregated from rewrite rules.
> 
> Here is an example:
> 
> ```mathematica
> In[1]:= foo ->"bar"]
> Out[1]= foo ⇾ bar
> In[2]:= foo
> Out[2]= foo
> ```
> 
> If there is never a possibility in evaluation that option rule foo->"bar" will be used in an expression evaluation, then it should not be considered or appear in the list of rules that are considered in application. And we are currently doing this.

Indeed, there is something that becomes apparent from this refactor round. More changes are coming, but it seems easy to implement on a less convoluted code.

rocky:
> > By the way, one thing that looks wrong about our selection in rewrite/eval is that we are looking at Options-like rules for rewriting, and as far as I can tell, these are never applied. So Options-like rules should be segregated from rewrite rules.
> > Here is an example:
> > ```mathematica
> > In[1]:= foo ->"bar"]
> > Out[1]= foo ⇾ bar
> > In[2]:= foo
> > Out[2]= foo
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If there is never a possibility in evaluation that option rule foo->"bar" will be used in an expression evaluation, then it should not be considered or appear in the list of rules that are considered in application. And we are currently doing this.
> 
> Indeed, there is something that becomes apparent from this refactor round. More changes are coming, but it seems easy to implement on a less convoluted code.

Well, I actually noticed this in the Rule refactor. I wasn't 100% sure, though, until I just tried it now.

mmatera:
> > > By the way, one thing that looks wrong about our selection in rewrite/eval is that we are looking at Options-like rules for rewriting, and as far as I can tell, these are never applied. So Options-like rules should be segregated from rewrite rules.
> > > Here is an example:
> > > ```mathematica
> > > In[1]:= foo ->"bar"]
> > > Out[1]= foo ⇾ bar
> > > In[2]:= foo
> > > Out[2]= foo
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > If there is never a possibility in evaluation that option rule foo->"bar" will be used in an expression evaluation, then it should not be considered or appear in the list of rules that are considered in application. And we are currently doing this.
> > 
> > 
> > Indeed, there is something that becomes apparent from this refactor round. More changes are coming, but it seems easy to implement on a less convoluted code.
> 
> Well, I actually noticed this in the Rule refactor. I wasn't 100% sure, though, until I just tried it now.

It means that it slipped into this branch....

PR #1914, opened 2026-08-27
rename *.m to *.wl. Add remaining CharacterEncoding files

rocky:
* Rename files with extension `.m` to `.wl`;  The `.m` is ambiguous. It could refer to Objective-C header files.


mmatera:
All these warnings should disappear now.

PR #1917, opened 2026-08-27
Add remaining CharacterEncoding files

rocky:
Add remaining CharacterEncoding files now that we have "7Bit", "8Bit", and "16Bit" tags implemented


mmatera:
One detail is that we are not currently support 16bit custom encodings, but OK, let's have them as a reference

PR #1918, opened 2026-08-27
replace string-arg-parameter has_form() with symbol. Part 1

rocky:
Start to replace the string first argument of `has_form` with a symbol first argument.


mmatera (approved):
LGTM

PR #1920, opened 2026-08-27
Move the `get_float_value()` method from the `BaseElement` class and move it to the `Number` and `Numeric` classes.

rocky:
* Remove get_float_value from the `BaseElement` class and add it to the `Rational` class.
* Go over the `Power` built-in e.g., add `eval_Power()`, and handle the case when the base is zero more correctly. 
Use SymPy's zero detection instead of Python's float zero test for better accuracy.
Go over some comments, and make the code more stringent for `mypy`

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -1516,7 +1516,9 @@ class Expression(BaseElement, NumericOperators, EvalMixin):
         elif isinstance(evaluation, sympy.core.numbers.NaN):
             return None
         else:
-            value = self.create_expression(SymbolN, self).evaluate(evaluation)
+            from mathics.eval.nevaluator import eval_N
+
+            value = eval_N(self, evaluation)
         if hasattr(value, "round") and hasattr(value, "get_float_value"):
             value = value.round()
             return value.get_float_value(permit_complex=permit_complex)
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -11,6 +11,7 @@ import sympy
 from mathics_scanner.location import SourceRange, SourceRange2
 
 from mathics.core.atoms import String
+from mathics.core.atoms.numerics import Number
 from mathics.core.attributes import (
     A_FLAT,
     A_HOLD_ALL,
@@ -1512,17 +1513,19 @@ class Expression(BaseElement, NumericOperators, EvalMixin):
         """
 
         if evaluation is None:
-            value = self
+            value: Optional[BaseElement] = self
         elif isinstance(evaluation, sympy.core.numbers.NaN):
             return None
         else:
             from mathics.eval.nevaluator import eval_N
 
             value = eval_N(self, evaluation)
-        if hasattr(value, "round") and hasattr(value, "get_float_value"):
-            value = value.round()
-            return value.get_float_value(permit_complex=permit_complex)
-        return None
+
+        if value is None or not isinstance(value, Number):
+            return None
+
+        rounded = value.round()
+        return rounded.get_float_value(permit_complex=permit_complex)
 
     def sameQ(self, other: BaseElement) -> bool:
         """Mathics3 SameQ"""
--- a/mathics/core/number.py
+++ b/mathics/core/number.py
@@ -15,6 +15,7 @@ from mathics.core.symbols import (
     SymbolMaxPrecision,
     SymbolMinPrecision,
 )
+from mathics.core.systemsymbols import SymbolDirectedInfinity
 
 LOG2_10 = mpmath.log(10.0, 2.0)  # ~ 3.3219280948873626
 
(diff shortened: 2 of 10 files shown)


rocky:
> # ⚠️ **Performance Alert** ⚠️
> Possible performance regression was detected for benchmark **'Mathics3 Core Benchmarks'**. Benchmark result of this commit is worse than the previous benchmark result exceeding threshold `2`.
> 
> Benchmark suite	Current: [22994bb](https://github.com/Mathics3/mathics-core/commit/22994bbe222f15e5e912a0044034ee782342181f)	Previous: [740def2](https://github.com/Mathics3/mathics-core/commit/740def2d634bd7a9530667e002364209a5478f9f)	Ratio
> `test/timings/test_regressions.py::test_regression_benchmark[arithmetic::1+2*3]`	`4.487958629365868` iter/sec (`stddev: 0.0008224954643661209`)	`9.477931981135386` iter/sec (`stddev: 0.00009665495529283073`)	`2.11`
> This comment was automatically generated by [workflow](https://github.com/Mathics3/mathics-core/actions?query=workflow%3ABenchmark) using [github-action-benchmark](https://github.com/marketplace/actions/continuous-benchmark).

@mmatera I don't believe this here. Can you translate what precisely it is saying has gotten worse and elucidate why? This PR involves exponentiation, mypy and comment tweaks.

mmatera:
> @mmatera I don't believe this here. Can you translate what precisely it is saying has gotten worse and elucidate why? This PR involves exponentiation, mypy and comment tweaks.

Probably is just a fluctuation in the server load during the benchmark. I was trying to reduce them to make this usable. Let's rerun the benchmark again and see if there are changes in this result.

rocky:
> > @mmatera I don't believe this here. Can you translate what precisely it is saying has gotten worse and elucidate why? This PR involves exponentiation, mypy and comment tweaks.
> 
> Probably is just a fluctuation in the server load during the benchmark. I was trying to reduce them to make this usable. Let's rerun the benchmark again and see if there are changes in this result.

How does one rerun the benchmark?

mmatera:
Just like when a github-action CI test fails, you can go to the Actions tab, choose the action, and ask to rerun it.
In any case, this warning was produced before I adjusted the benchmarks and the test to handle the variations and speed of the server load.

mmatera:
This is the log of the benchmark action:
```

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
============================= 73 passed in 59.22s ==============================
python3 admin-tools/normalize_benchmarks.py output.json
✅ Normalized benchmarks saved in output.json.
python3 admin-tools/check_benchmarks.py --threshold 200 output.json
📊 Comparing 73 benchmarks...
   Thresold: 200.0x (regression if ratio > 200.0)

✅ Significant regressions were not detected.
   Detected improvements: 1 benchmarks
  🚀 p=Plot[Sin[2 Pi x],{x,0,3}];]: 0.00 vs 0.00 (ratio 0.67)
```
So, no performance issues are reported now.

mmatera:
Thanks for your patience. LGTM

mathics/core/number.py
```diff
@@ -57,15 +58,15 @@ def __init__(self, name) -> None:
 
 
 def _get_float_inf(value, evaluation) -> Optional[float]:
-    value = value.evaluate(evaluation)
-    if value.has_form("DirectedInfinity", 1):
-        if value.elements[0].get_int_value() == 1:
+    evaled_value = value.evaluate(evaluation)
```
mmatera: evalued?

mathics/core/expression.py
```diff
@@ -1512,15 +1513,19 @@ def round_to_float(
...
+            value: Optional[BaseElement] = self
         elif isinstance(evaluation, sympy.core.numbers.NaN):
             return None
         else:
-            value = self.create_expression(SymbolN, self).evaluate(evaluation)
-        if hasattr(value, "round") and hasattr(value, "get_float_value"):
-            value = value.round()
-            return value.get_float_value(permit_complex=permit_complex)
-        return None
+            from mathics.eval.nevaluator import eval_N
```
mmatera, line 1520: Great! That has been hanging fruit for a long time

mathics/eval/arithfns/basic.py
```diff
@@ -107,6 +139,52 @@ def append_last():
...
+                    evaluation.message("Power", "indet", expr)
+                elif result is SymbolComplexInfinity:
+                    evaluation.message("Power", "infy", expr)
+                elif result is Integer0:
+                    pass
+                return result
+
+    if isinstance(base, Complex) and base.real.is_zero:
+        yhalf = Expression(SymbolTimes, exponent, RationalOneHalf)
+        factor = self.eval(Expression(SymbolSequence, base.imag, exponent), evaluation)
```
mmatera: This could be left as Expression(SymbolPower,base.image, exponent), or probably call this function again.

mathics/core/number.py
rocky: Would you prefer `evaluated_value`?  Or suggest something else.  I don't really care, as long as we don't overwrite "value".

mathics/eval/arithfns/basic.py
rocky: I tried without the wrapping `self.eval()`.  We fail ` test/builtin/arithfns/test_basic.py::test_power`. 

In particular, for `2 (-1) ^ (1 / 3) / 3` we get the same back, but  `(-1) ^ (1 / 3)` is expected.

I _do_ want to remove passing `self` though, so I'll look into other alternatives.

rocky:
> Just like when a github-action CI test fails, you can go to the Actions tab, choose the action, and ask to rerun it. In any case, this warning was produced before I adjusted the benchmarks and the test to handle the variations and speed of the server load.

Thanks. I did look at the actions tab initially; I missed it amid the large number of tests that we currently have.

rocky:
@mmatera: I  have tried to address your suggestions. If I am remiss, though, let me know, and we will resolve these in a future PR.

PR #1923, opened 2026-08-30
has_form now supports iterables with Symbols.

mmatera:
In this PR, the implementation of `has_form` was improved to work with symbols as heads.


rocky:
Thanks. When things stabilize, merge. 

The performance alert thing is getting to be annoying though.

mmatera:
> Thanks. When things stabilize, merge.
> 
OK

> The performance alert thing is getting to be annoying though.

I  just disabled it. The benchmarks are run, but the results only shows explicitly in the actions log.

mmatera:
Good. There are more places to change, but it is for another round.

PR #1925, opened 2026-08-31
Options refactor

rocky:
Reorganize the Options Management section and make it more akin to a Guide section.

As such, the location of a number of Mathics3 Builtin Functions and Mathics3 Options has been moved to more logical places that correspond better with the Wolfram docs.

Moved functions include `Automatic` and `Default` (into Options) and `FormatType` into drawing options.

* Add `is_option_rule()` option.
* Simplify a ListExpression check.

Move more eval code to `mathics.eval.options`
Options are `Predefined` and do not need to be `Builtins`.

All of this is in preparation for adding `CheckArguments`


mmatera (approved):
LGTM

PR #1926, opened 2026-08-31
Add mypy pre-commit hook and sync up benchmark.yml to v6

rocky:
* `pre-commit-config.yml`: run `mypy` over Python code in the pre-commit hook.
* `workflows.benchmark.yml` : use v6 in checkout and Python setup like the other yml files.


mmatera:
LGTM

PR #1928, opened 2026-09-01
Add check arguments

rocky:
Add `CheckArguments[]`.

CheckArguments is used in Rubi.

The association option "OptionsMode" and possibly the tag used in some failure messages (which change depending on whether an option was defined or how many arguments are given) may be different too. 

However, this is good enough for Rubi (the infinite recursion seen before is now gone).


mmatera (approved):
LGTM

PR #1929, opened 2026-09-02
Fix remaining makeboxes tests

mmatera:
This is another round on the remaining fixes for MakeBoxes. With this, the number of xfailing tests reduced to 9.


rocky:
LGTM and Thanks!

PR #1930, opened 2026-09-02
add TagUnset buitlin

mmatera:
Looking over the MakeBoxes stuff, I realize that this basic built-in symbol is still missing. TagUnset allows surgically removing rules from UpValues.


rocky:
LGTM - @mmatera, thanks for fixing the merge commit problem.

PR #1934, opened 2026-09-03
Change `get_int_value()` to a function in `mathics.core.atoms.numerics`. Remove it as an method of `Element`

rocky:
Change `get_int_value()` to a function in `mathics.core.atoms.numerics`. Remove it as an method of `Element`.

`get_int_value()` is a highly custom kind of function, not a fundamental property of an element. 

This is part of a long-desired reorganization of `mathics.core` to make it comprehensible and logical, and more in line with modern Python conventions.

`.int_value` is now a property on methods only where it makes sense.

### DIFF
--- a/mathics/builtin/atomic/numbers.py
+++ b/mathics/builtin/atomic/numbers.py
@@ -263,7 +263,7 @@ class IntegerLength(Builtin):
     def eval(self, n, b, evaluation):
         """IntegerLength[n_, b_]"""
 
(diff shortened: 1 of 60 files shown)


rocky:
@mmatera: Before we make what will look like massive changes to the code pattern and/or assignment code, I'd like this cleanup to be in place first.

mmatera:
LGTM. Thanks!

PR #1936, opened 2026-09-05
Simplify and Clarify. Remove duplicate assert fail messages

rocky:
Segregate long tests with mostly no expected messages from tests with no expected messages and the few that do. 

Disambiguate duplicate assert fail messages by describing the differences less vaguely

Go over the wording of  some assert messages.

Not done:  assert messages should be independent and not rely on understanding the history of tests that occurred before. This is a fine thing to do in a doctest for pedagogical reasons, but not in a pytest.


mmatera (approved):
LGTM

PR #1938, opened 2026-09-06
Fix `Verbatim` patterns

### LINKED ISSUE #1937: `Verbatim` doesn't allow setting DownValues
## Description

Builtin function `Verbatim` does not support setting DownValues

```mathematica
In[1]:= Verbatim[F][x_Integer]:=3; DownValues[F]
System`CompoundExpression[System`SetDelayed[System`Verbatim[Global`F][System`Pattern[Global`x, System`Blank[System`Integer]]], 3], System`DownValues[Global`F]]
SetDelayed::write: Tag Verbatim in Verbatim[F][x_Integer] is Protected.
Out[1]= {}
```

## Expected behavior

```mathematica
In[1]:= Verbatim[F][x_Integer]:=3; DownValues[F]                                              
Out[1]= {HoldPattern[Verbatim[F][x_Integer]] :> 3}
```

## Priority

Should be fixed before release.

## Additional context

This impacts correcting the implementation of `Condition` #1935  which impacts fixing #1932

mmatera:
This PR 
* Fixes the issue with storing rules that use `Verbatim` in patterns. 
* Reimplements the code that determines the position of a rule inside a definition.
* For patterns, it adds the `get_tag_position` method, to take advantage of the specific BasePattern subclasses structure.
* `Definition` now stores an attribute `symbol`, with the associated symbol.
* add tests for this

Fixes #1937


mathics/builtin/assignments/clear.py
```diff
@@ -268,7 +269,7 @@ def eval(self, expr, evaluation):
                 empty = []
             evaluation.definitions.set_values(symbol, expr.get_head_name(), empty)
             return SymbolNull
-        name = expr.get_lookup_name()
+        name = get_lookup_reference_name(expr)
```
mmatera, line 272: `get_lookup_reference_name` handles  `Condition` and `Verbatim` in a proper way, which `get_lookup_name` doesn't.

test/core/test_definitions.py
```diff
@@ -7,6 +7,40 @@
 
 from mathics.core.definitions import get_tag_position
 from mathics.core.parser import parse_builtin_rule
+from mathics.core.pattern import BasePattern
+
+
+def load_pattern_objects():
```
mmatera, line 13: To test PatternObjects, we need to be sure that we have already populated `pattern_objects`. To keep this test as a "lightweight" test, I avoided loading all the definitions here.

test/core/test_definitions.py
```diff
@@ -56,8 +93,34 @@
         ),
         # Just one argument, must be an upvalue
         ("N[A[s_]]", "A", "upvalues"),
+        # Verbatim
```
mmatera, line 96: I checked all these cases by hand against WMA, looking in which set of rules the assignment with `Set` and `TagSet` WMA stores the rule. For example, to check "Verbatim[F][Verbatim[G][x_]]",
```
In[1]:= F/:Verbatim[F][Verbatim[G][x_]]:=1                                      

In[2]:= DownValues[F]                                                           

Out[2]= {HoldPattern[Verbatim[F][Verbatim[G][x_]]] :> 1}

In[3]:= G/:Verbatim[F][Verbatim[G][x_]]:=1                                      

In[4]:= UpValues[G]                                                             

Out[4]= {HoldPattern[Verbatim[F][Verbatim[G][x_]]] :> 1}

In[5]:= DownValues[G]                                                           

Out[5]= {}

```

rocky:
This is a big improvement. There is much to like about this PR. 

Some suggestions to consider fall more under the category of style rather than substance.

First, `test_assignment` is already 700 lines long and something of a kitchen sink, while there are no explicit Verbatim tests in `tests/builtin/patterns/composite`. Let's move all the `Verbatim` tests to `test/builtin/patterns/composite` and the `HoldAll` tests to `test/builtin/test_attributes.py`. 

A more modern style of Python would have `get_tag_position` return an enumeration with `None` being one of the enumeration values.  This also allows for better and stricter typing.

mathics/builtin/assignments/clear.py
rocky: Does this suggest that, in the future,  we migrate other code to `get_lookup_reference_name`?

rocky:
Here, we introduce the function `get_tag_position()`.  As you know, I am fussy about names, and this is an important (and much-needed) function.  (I really love the fact that rewriting the condition is being removed.)

Here's my problem with the name. The collection of UpValue, DownValue, OwnValue, etc. is, in WMA, called a kind of Value. A tag is used in an error message. 

It seems, I think, that we are not _getting_ the position, but rather _using_ the position to determine the Value kind. 

So something along the lines of `get_value_role_from_position`.  Or better, I'd like to capitalize "value":  `get_Value_role_from_position` if linters will now tolerate that.

Here is a suggestion for the enumeration: 

```python
class SymbolValueRole(Enum):
    NONE = "None"
    OWN_VALUE = "OwnValues"
    DOWN_VALUE = "DownValues"
    UP_VALUE = "UpValues"
    SUB_VALUE = "SubValues"
    N_VALUE = "NValues"
    FORMAT_VALUE = "FormatValues"
```

Your thoughts?

rocky:
I don't like the name of the function `get_lookup_reference_name`.

"get" is a verb; "lookup" is a verb too. Pick one. "get" is generic, while "lookup" is often used to find an item in a dictionary or symbol table, which better matches here.

And it would be nice if this returned the Symbol, not a string name.

Begin rant:

There was a misunderstanding very early on in this project that thought that strings and symbols are the same thing. We also had confusion between "Elements" and "leaves", but we managed to get rid of that. We've never really removed this confusion between symbols and strings, though.

End rant.


I suppose we can address this in a separate PR. It is a pity it wasn't caught when `get_lookup_reference_name` was introduced.

mathics/builtin/assignments/clear.py
mmatera: This applies for the special case of assignment. In other contexts, `Vertabim[_]` should behave like any other expression.

mmatera:
Since this PR got large, I avoided changing names or moving code around. Let's do that in a subsequent round.

mmatera:
> I don't like the name of the function `get_lookup_reference_name`.
> 
> "get" is a verb; "lookup" is a verb too. Pick one. "get" is generic, while "lookup" is often used to find an item in a dictionary or symbol table, which better matches here.
> 
Regarding this: `get_lookup_reference_name` comes from the attribute in `BaseElemen` `get_lookup_name`. Here, `lookup_name` is used as a "noun" referring to the name of the symbol/definition where replacement rules should be asked. We could to a round of adjusting these names to more grammaticaly correct English nouns.

> And it would be nice if this returned the Symbol, not a string name.
> 
Same about this. Some years ago, I tried to rewrite the `Definitions` object to use Symbol objects instead of strings, but it required implementing hundreds of changes over all the base code. Since what I am doing here, and in the next PR is pretty large, I propose postponing this change for another round. Moreover, I would avoid changing this before the release, because it would imply an important change in the API.

> Begin rant:
> 
> There was a misunderstanding very early on in this project that thought that strings and symbols are the same thing. We also had confusion between "Elements" and "leaves", but we managed to get rid of that. We've never really removed this confusion between symbols and strings, though.
> 
> End rant.
> 
> I suppose we can address this in a separate PR. It is a pity it wasn't caught when `get_lookup_reference_name` was introduced.

mmatera:
> Here, we introduce the function `get_tag_position()`. As you know, I am fussy about names, and this is an important (and much-needed) function. (I really love the fact that rewriting the condition is being removed.)
> 
> Here's my problem with the name. The collection of UpValue, DownValue, OwnValue, etc. is, in WMA, called a kind of Value. A tag is used in an error message.
> 
> It seems, I think, that we are not _getting_ the position, but rather _using_ the position to determine the Value kind.
> 
> So something along the lines of `get_value_role_from_position`. Or better, I'd like to capitalize "value": `get_Value_role_from_position` if linters will now tolerate that.
> 
> Here is a suggestion for the enumeration:
> 
> ```python
> class SymbolValueRole(Enum):
>     NONE = "None"
>     OWN_VALUE = "OwnValues"
>     DOWN_VALUE = "DownValues"
>     UP_VALUE = "UpValues"
>     SUB_VALUE = "SubValues"
>     N_VALUE = "NValues"
>     FORMAT_VALUE = "FormatValues"
> ```
> 
> Your thoughts?

Again, regarding the return type, I try to avoid change the API. But yes, to use an Enum class would make a lot of sense. Regarding the name,  `Tag` is used in WMA as the explicit Symbol where a rule is attached (see `TagSet` for instance). `get_tag_position` refers to the position (the name of the rule list) inside the definition of the`tag` Symbol. 
An alternative name that fully describes what the function does would be `get_value_role_from_symbol` or
`get_value_role_from_tag`. On the other hand, maybe what is wrong is the choice of the second parameter: `target` should be `tag`. I can make that change.

mmatera:
> Some suggestions to consider fall more under the category of style rather than substance.
> 
> First, `test_assignment` is already 700 lines long and something of a kitchen sink, while there are no explicit Verbatim tests in `tests/builtin/patterns/composite`. Let's move all the `Verbatim` tests to `test/builtin/patterns/composite` and the `HoldAll` tests to `test/builtin/test_attributes.py`.
> 

Regading this, notice that the changes I did in `Verbatim` was mostly related on how `Verbatim` is handled in *Set*
expressions. This is not related directly with pattern matching (which is what we test on `test/builtin/patterns/composite`) but about how a rule is associated to a Definition object. In any case, we can split the `test/builtin/assignment/test_assignment.py` test module in smaller units.

> A more modern style of Python would have `get_tag_position` return an enumeration with `None` being one of the enumeration values. This also allows for better and stricter typing.

I tried to make these changes work in a way that keeps compatible with the 10.0.1 API. I was tempted to change the return type to an enumeration, but I refrained because that reason.

mmatera:
> First, `test_assignment` is already 700 lines long and something of a kitchen sink, while there are no explicit Verbatim tests in `tests/builtin/patterns/composite`. Let's move all the `Verbatim` tests to `test/builtin/patterns/composite` and the `HoldAll` tests to `test/builtin/test_attributes.py`.

This is not fully related with this PR, but now I split the assignment pytest module in smaller parts. However, notice that these tests are not related to the way in which `Verbatim[...]` behaves  in the standard evaluation sequence, or in pattern matching, but how it is handled as a part of the LHS in the evaluation of  a *Set* expression. This is why I think these tests belongs to the `test.builtin.assigment` unit test.

rocky:
> Since this PR got large, I avoided changing names or moving code around. Let's do that in a subsequent round.

Sure. Let's do this in a subsequent round.

rocky:
> Regarding this: `get_lookup_reference_name` comes from the attribute in `BaseElemen` `get_lookup_name`. Here, `lookup_name` is used as a "noun" referring to the name of the symbol/definition where replacement rules should be asked.

There is no "lookup name". There is a "name" or more precisely a "symbol name".

>  We could to a round of adjusting these names to more grammaticaly correct English nouns.

Yes, let's do this in a separate round.

> 
> > And it would be nice if this returned the Symbol, not a string name.
> 
> Same about this. Some years ago, I tried to rewrite the `Definitions` object to use Symbol objects instead of strings, but it required implementing hundreds of changes over all the base code.

I've been looking for incremental ways to do this. For example, `has_form` now allows symbols as well as strings. When we have removed the last use of a `has_form` with a string, we can then remove the code from `has_symbols`.

I guess what I am suggesting is keeping in mind not to increase but to decrease, the use of strings when we can.

 Since what I am doing here, and in the next PR is pretty large, I propose postponing this change for another round. Moreover, I would avoid changing this before the release, because it would imply an important change in the API.

We already have API breakage. The changes in MakeBoxes and Information do this, but I am sure there are other places.

rocky:
> 
> Again, regarding the return type, I try to avoid change the API

And again, the API has already changed. And those changes are more user impacting than get_tag_position

>  But yes, to use an Enum class would make a lot of sense. Regarding the name, `Tag` is used in WMA as the explicit Symbol where a rule is attached (see `TagSet` for instance). `get_tag_position` refers to the position (the name of the rule list) inside the definition of the`tag` Symbol. 

> An alternative name that fully describes what the function does would be `get_value_role_from_symbol` or `get_value_role_from_tag`. 

Right. The point is that "position" is an input parameter; "from" in the function name would help when there are _several_  `get_value_role...` functions. That's not the situation here. 

So how about <strike>`get_value_role`</strike> `deterimine_value_role`.

> On the other hand, maybe what is wrong is the choice of the second parameter: `target` should be `tag`. I can make that change.

<strike>Yes, that would be better. </strike>

rocky:
>  On the other hand, maybe what is wrong is the choice of the second parameter: `target` should be `tag`. I can make that change.

I've been looking at the documentation to understand what "Tag" refers to in "TagSet".  Here, I don't think it refers to a noun, as it does in a message _tag_.  Here, I think it is synonymous with "associate with". So "TagSet" is meant to convey "Set adding an association (or tagging the Value role) to Symbol ...". 

So changing `target` to `tag` is likely to lead to confusion. Instead, something like symbol_tagged or tagged_symbol or name_to_be_tagged. It is important to convey that we mean tag as an association or tagging of something _else_.

mmatera:
> > On the other hand, maybe what is wrong is the choice of the second parameter: `target` should be `tag`. I can make that change.
> 
> I've been looking at the documentation to understand what "Tag" refers to in "TagSet". Here, I don't think it refers to a noun, as it does in a message _tag_. Here, I think it is synonymous with "associate with". So "TagSet" is meant to convey "Set adding an association (or tagging the Value role) to Symbol ...".
> 
> So changing `target` to `tag` is likely to lead to confusion. Instead, something like symbol_tagged or tagged_symbol or name_to_be_tagged. It is important to convey that we mean tag as an association or tagging of something _else_.

@rocky, thanks for the research. So, after this, what would be a reasonable name for these functions? Maybe something like `get_value_kind_relative_to_a_symbol`? `get_value_role_in_definition`?

rocky:
> > > On the other hand, maybe what is wrong is the choice of the second parameter: `target` should be `tag`. I can make that change.
> > 
> > 
> > I've been looking at the documentation to understand what "Tag" refers to in "TagSet". Here, I don't think it refers to a noun, as it does in a message _tag_. Here, I think it is synonymous with "associate with". So "TagSet" is meant to convey "Set adding an association (or tagging the Value role) to Symbol ...".
> > So changing `target` to `tag` is likely to lead to confusion. Instead, something like symbol_tagged or tagged_symbol or name_to_be_tagged. It is important to convey that we mean tag as an association or tagging of something _else_.
> 
> @rocky, thanks for the research. So, after this, what would be a reasonable name for these functions? Maybe something like `get_value_kind_relative_to_a_symbol`? `get_value_role_in_definition`?

```python
 determine_value_role(self, tagged_symbol: Symbol)
```

(Changed "get" to "determine" because, at least in some cases, we are not getting a precomputed Value role is or a previously set Value role, but instead _computing_ or _determining_ what the Value role is.

And with this change, we can merge this in.

It's not my intention to hold this PR back. So I will be adding a bunch of follow-up chores to be done that this either uncovers (or uncovers at least for me).  Some of it is legacy stuff; some of it was introduced in prior massive PRs. 

I spend a lot of time trying to clean up code. It looks to me like this trend is going to continue for a while. Sigh.

test/builtin/assignments/test_assign_set_and_clear.py
```diff
@@ -0,0 +1,338 @@
...
+            "This clears A and B, but not $Context",
+            ("Special symbol $Context cannot be cleared.",),
+        ),
+        (
+            "A=1; B=2; ClearAll[A, $Context, B];{A,$Context,B}",
+            "{A, Global`, B}",
+            "This clears A and B, but not $Context",
+            ("Special symbol $Context cannot be cleared.",),
+        ),
+        (
```
rocky, line 309: Please see chore https://github.com/Mathics3/mathics-core/issues/1941

mmatera:
> > > > On the other hand, maybe what is wrong is the choice of the second parameter: `target` should be `tag`. I can make that change.
> > > 
> > > 
> > > I've been looking at the documentation to understand what "Tag" refers to in "TagSet". Here, I don't think it refers to a noun, as it does in a message _tag_. Here, I think it is synonymous with "associate with". So "TagSet" is meant to convey "Set adding an association (or tagging the Value role) to Symbol ...".
> > > So changing `target` to `tag` is likely to lead to confusion. Instead, something like symbol_tagged or tagged_symbol or name_to_be_tagged. It is important to convey that we mean tag as an association or tagging of something _else_.
> > 
> > 
> > @rocky, thanks for the research. So, after this, what would be a reasonable name for these functions? Maybe something like `get_value_kind_relative_to_a_symbol`? `get_value_role_in_definition`?
> 
> ```python
>  determine_value_role(self, tagged_symbol: Symbol)
> ```
> 
> (Changed "get" to "determine" because, at least in some cases, we are not getting a precomputed Value role is or a previously set Value role, but instead _computing_ or _determining_ what the Value role is.
> 
> And with this change, we can merge this in.
> 
> It's not my intention to hold this PR back. So I will be adding a bunch of follow-up chores to be done that this either uncovers (or uncovers at least for me). Some of it is legacy stuff; some of it was introduced in prior massive PRs.
> 
> I spend a lot of time trying to clean up code. It looks to me like this trend is going to continue for a while. Sigh.

It is perfectly fine. Just that if we are going to change names to make them clearer, I prefer not to do it alone.

test/builtin/assignments/test_assign_set_and_clear.py
mmatera: OK, in this round, I just split the tests into modules to make the job of reviewing and refactoring them easier. The relation with this PR is just that I have added more tests to this module. But I guess that changing the tests of unrelated modules would make it harder to follow the changes. Let's do this in the next round.

rocky:
> > > > > On the other hand, maybe what is wrong is the choice of the second parameter: `target` should be `tag`. I can make that change.
> > > > 
> > > > 
> > > > I've been looking at the documentation to understand what "Tag" refers to in "TagSet". Here, I don't think it refers to a noun, as it does in a message _tag_. Here, I think it is synonymous with "associate with". So "TagSet" is meant to convey "Set adding an association (or tagging the Value role) to Symbol ...".
> > > > So changing `target` to `tag` is likely to lead to confusion. Instead, something like symbol_tagged or tagged_symbol or name_to_be_tagged. It is important to convey that we mean tag as an association or tagging of something _else_.
> > > 
> > > 
> > > @rocky, thanks for the research. So, after this, what would be a reasonable name for these functions? Maybe something like `get_value_kind_relative_to_a_symbol`? `get_value_role_in_definition`?
> > 
> > 
> > ```python
> >  determine_value_role(self, tagged_symbol: Symbol)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (Changed "get" to "determine" because, at least in some cases, we are not getting a precomputed Value role is or a previously set Value role, but instead _computing_ or _determining_ what the Value role is.
> > And with this change, we can merge this in.
> > It's not my intention to hold this PR back. So I will be adding a bunch of follow-up chores to be done that this either uncovers (or uncovers at least for me). Some of it is legacy stuff; some of it was introduced in prior massive PRs.
> > I spend a lot of time trying to clean up code. It looks to me like this trend is going to continue for a while. Sigh.
> 
> It is perfectly fine. Just that if we are going to change names to make them clearer, I prefer not to do it alone.

Um, this misses the point that coming up with clear names is not something you _can_ do alone. It's probably related to the ability to come up with simple, focused examples.

PR #1939, opened 2026-09-06
Fix `Condition` appearing in a `Set` function

mmatera:
This PR follows #1938, and fixes the issue with `Condition[...]` as part of the LHS of *Set* expressions.

### DIFF
--- a/mathics/eval/assignments/assignment.py
+++ b/mathics/eval/assignments/assignment.py
@@ -7,7 +7,7 @@ found in module mathics.builtin.assignments.assignment
 """
 
 from functools import reduce
-from typing import Optional, Tuple
+from typing import Optional
 
 from mathics.core.assignment import (
     get_symbol_list,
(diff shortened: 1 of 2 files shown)


mathics/eval/assignments/assignment.py
```diff
@@ -1400,6 +1404,90 @@ def get_lookup_reference_name(expr: BaseElement) -> str:
...
+    Parameters
+    ----------
+    lhs : Expression
+        The left-hand side expression to be processed.
+    evaluation : Evaluation
+        The evaluation context used when evaluating the elements of the expression.
+
+    Returns
+    -------
+    Tuple[BaseElement, BaseElement]
```
rocky: In Python 3.9, in PEP 585,  the form `tuple[BaseElement, BaseElement]` was introduced as an alternative to `Tuple[BaseElement, BaseElement]` and this no longer requires `from typing import Tuple`. 

I've been slowly migrating code to use this form, since we no longer support Python 3.8.

rocky:
LGTM. Merge in after 1938 is merged.

rocky:
LGTM - thanks!

Issue #1932, opened 2026-09-03
Infinite Recursion in running `??Int` after Rubi is loaded.


rocky:
## Description

After `Rubi` is loaded, `??Int` (or `Information["Int", LongForm -> True]` goes into an infinite recursion

In `MatchQ`:

```mathematica
     file '/tmp/Mathics3/mathics-core/mathics/core/expression.py' at line 636, column 9
E:1 (4) Expression.rewrite_apply_eval_step.<locals>.eval_elements <class 'mathics.core.expression.Expression'>
         eval_elements() file '/tmp/Mathics3/mathics-core/mathics/core/expression.py' at line 1270, column 17
E:2 (5) Expression.rewrite_apply_eval_step <class 'mathics.core.expression.Expression'>
         rewrite_apply_eval_step(MatchQ[_, Alternatives[_ Blank, _ BlankSequence, _ BlankNullSequence]],
	)
     file '/tmp/Mathics3/mathics-core/mathics/core/expression.py' at line 1284, column 13
E:3 (7) Expression.evaluate <class 'mathics.core.expression.Expression'>
         evaluate(MatchQ[_, Alternatives[_ Blank, _ BlankSequence, _ BlankNullSequence]],
	)
     file '/tmp/Mathics3/mathics-core/mathics/core/expression.py' at line 660, column 36
E:4 (10) Expression.rewrite_apply_eval_step.<locals>.eval_elements <class 'mathics.core.expression.Expression'>
          eval_elements() file '/tmp/Mathics3/mathics-core/mathics/core/expression.py' at line 1270, column 17
E:5 (11) Expression.rewrite_apply_eval_step <class 'mathics.core.expression.Expression'>
          rewrite_apply_eval_step(Not[MatchQ[_, Alternatives[_ Blank, _ BlankSequence, _ BlankNullSequence]]],
	)
     file '/tmp/Mathics3/mathics-core/mathics/core/expression.py' at line 1284, column 13
E:6 (13) Expression.evaluate <class 'mathics.core.expression.Expression'>
          evaluate(Function[Not[MatchQ[#1, Alternatives[_ Blank, _ BlankSequence, _ BlankNullSequence]]]][_],
	)
     file '/tmp/Mathics3/mathics-core/mathics/core/expression.py' at line 660, column 36
```
## How to Reproduce and Output Given

Git clone Rubi from GitHub and cd into the project root. Then.

```mathematica
$ mathics3 --post-mortem --full-form

[...]

PR #1940, opened 2026-09-07
Now `Repeated` and `RepeatedNull` can use regular format rules

mmatera:
This PR provides a first payoff of the previous fixes #1938 and #1939: we can move the special cases of Repeated and RepeatedNull format rules that had to be handled in a hard-coded way to the regular format rules mechanism.
The issue with the infinite recursion in Condition seems to be related to the infinite recursion these patterns produced.

### DIFF
--- a/mathics/builtin/patterns/composite.py
+++ b/mathics/builtin/patterns/composite.py
@@ -577,7 +577,7 @@ class Repeated(PostfixOperator, PatternObject):
             "expected at position `1` in `2`."
         )
     }
-
+    needs_verbatim = True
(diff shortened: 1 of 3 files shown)


rocky:
LGTM. Merge after #1938 and #1939

rocky:
@mmatera Note that after all of this, I do not think #1932 is fully addressed. I'll check again, but I am still getting an infinite recursion error. 

Also, something as simple as `Int[x, x]` is not working in Rubi, whereas it did in 7.0. I will log specific issues for this after investigation. 

(That said, I think that after the dust settles a bit, all of this will be great progress. Thanks!)

PR #1950, opened 2026-09-08
Use evaluation for creating rules

mmatera:
This PR starts to use the evaluation object to instantiate patterns in rules defined during a session. This avoid to use `DeferredExpressionPattern` when attributes for patterns are already available.

### DIFF
--- a/mathics/core/rules.py
+++ b/mathics/core/rules.py
@@ -138,11 +138,11 @@ class BaseRule(KeyComparable, ABC):
         pattern: BaseElement,
         system: bool = False,
         attributes: Optional[int] = None,
+        evaluation: Optional[Evaluation] = None,
     ) -> None:
         self.location: Optional[Callable] = None
         self.pattern = BasePattern.create(
-            pattern,
-            attributes=attributes,
+            pattern, attributes=attributes, evaluation=evaluation
         )
 
     def apply(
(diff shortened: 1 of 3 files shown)


rocky:
LGTM, but I would like to understand what's up. 

From the changes, all we are doing is adding an evaluation parameter, presumably for something down the call chain to use. What is that something called that is going to use this? 

Presumably was motivated by for performance, right? If so, do those benchmark statistics show anything?

mmatera:
> LGTM, but I would like to understand what's up.
> 
> From the changes, all we are doing is adding an evaluation parameter, presumably for something down the call chain to use. What is that something called that is going to use this?
> 
This evaluation parameter allows to  `Rule` to create the right kind of Pattern.

> Presumably was motivated by for performance, right? If so, do those benchmark statistics show anything?

By now, having rules that use the right kind of ExpressionPattern allows for better logic in pattern matching. 
For example, a lot of logic in the general ExpressionPattern was related to handling special attributes like Orderless, or  OneIdentity, but most of the rules do not have patterns with these properties. In performance, we now win a little bit, but we could win more when we start implementing dispatch tables, as WMA does, to find the proper rules at evaluation time.

rocky:
> > LGTM, but I would like to understand what's up.
> > From the changes, all we are doing is adding an evaluation parameter, presumably for something down the call chain to use. What is that something called that is going to use this?
> 
> This evaluation parameter allows to `Rule` to create the right kind of Pattern.
> 
> > Presumably was motivated by for performance, right? If so, do those benchmark statistics show anything?
> 
> By now, having rules that use the right kind of ExpressionPattern allows for better logic in pattern matching. For example, a lot of logic in the general ExpressionPattern was related to handling special attributes like Orderless, or OneIdentity, but most of the rules do not have patterns with these properties. In performance, we now win a little bit, but we could win more when we start implementing dispatch tables, as WMA does, to find the proper rules at evaluation time.

Thanks for the explanation.

Issue #1943, opened 2026-09-07
Rename functions `get_lookup_reference_name` and `get_lookup_name`


rocky:
`get_lookup_name` should probably be `lookup_name`. 

I am not sure what `get_lookup_reference_name`  should be. In #1942 its docstring will probably be revised. The name will probably fall out of that revision.

mmatera:
In Spanish, the translation of `get lookup reference name` would be (according to google translator) "obtener nombre de referencia de búsqueda". `get_reference_lookup_name` would be closer to what I would say in Spanish ("obtener el nombre de búsqueda de la referencia")  If I understand it right, in English, `Lookup` is a verb, but can work as a modifier of a noun,  so "lookup name" would be translated to Spanish as "nombre de búsqueda" (to lookup a definition in the Definitions object). And what this function returns is the lookup name corresponding to the reference expression. `Expression` objects have a `get_lookup_name` that returns the "lookup name" of the Expression, i.e., the lookup name of the definition where rules that could apply to the expression must be looked up.
For LHS, the place where the rules must be stored comes from the lookup name of a certain subexpression, the reference expression.

So, I propose to rename this function as `get_reference_lookup_name`.  @rocky, would that work for you?

rocky:
> In Spanish, the translation of `get lookup reference name` would be (according to google translator) "obtener nombre de referencia de búsqueda". `get_reference_lookup_name` would be closer to what I would say in Spanish ("obtener el nombre de búsqueda de la referencia") If I understand it right, in English, `Lookup` is a verb, but can work as a modifier of a noun, so "lookup name" would be translated to Spanish as "nombre de búsqueda" (to lookup a definition in the Definitions object). And what this function returns is the lookup name corresponding to the reference expression. `Expression` objects have a `get_lookup_name` that returns the "lookup name" of the Expression, i.e., the lookup name of the definition where rules that could apply to the expression must be looked up. For LHS, the place where the rules must be stored comes from the lookup name of a certain subexpression, the reference expression.
> 
> So, I propose to rename this function as `get_reference_lookup_name`. [@rocky](https://github.com/rocky), would that work for you?

Let me think about this. 

Somewhere in WMA docs, there has to be a description of this process of doing this kind of thing: how the left-hand side of some kind of assignment is treated. We should try to match that terminology.

There is no concept of a "reference expression" in WMA. If there is some close concept, let's use that.  If there is not a close concept, but things are framed in different concepts, then we are probably doing something wrong.

[...]

PR #1954, opened 2026-09-12
Revise variables and functions, part 3

rocky:
*  `tags` -> `target_symbol_names`
* `tag_name` -> `target_symbol_name`
* `lhs_reference` -> `lhs_target`
* `tag_symbol` -> `target_symbol`
* `process_tags_and...` -> `collect_lhs_names_and..`
* revise `eval_assign_recursion_limit()`

See also #1943. 

In WMA, the use of " tag " in `TagSet` is intended as an adjective for Symbol, synonymous with "targeted" (as in targeted Symbol) in an assignment function. In an error message, the parameter with the tag "tagfn" is a Symbol, though. 

In our code, "tag" is used inconsistently to refer to this tagged or targeted symbol. In every other place in the code, "tag" refers to a label, like the tag on an error message (but not the error message symbol being referred to), or the tag of a Box (not the boxing symbol). 

Avoid this confusing use.

### DIFF
--- a/mathics/builtin/assignments/assignment.py
+++ b/mathics/builtin/assignments/assignment.py
@@ -282,13 +282,19 @@ class TagSet(Builtin):
     ) -> Optional[BaseElement]:
         "f_ /: lhs_ = rhs_"
 
-        tag_name = f.get_name()
-        if not tag_name:
+        target_symbol_name = f.get_name()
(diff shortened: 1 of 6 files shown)


mmatera (approved):
LGTM

PR #1957, opened 2026-09-14
`has_form()` with `str` arg removal - part1

rocky:
`has_form()` with `str` arg removal &mdash; part 1
Also, `xxx.has_form("List", None)`  -> `isinstance(xxx, ListExpression)`


mathics/builtin/assumptions/assumptions.py
```diff
@@ -42,7 +42,9 @@ def eval_assuming(self, assumptions, expr, evaluation: Evaluation):
         assumptions = assumptions.evaluate(evaluation)
         if assumptions is SymbolTrue:
             cond = []
-        elif isinstance(assumptions, Symbol) or not assumptions.has_form("List", None):
+        elif isinstance(assumptions, Symbol) or not isinstance(
```
mmatera, line 45: Regarding `assumptions.has_form("List", None)-> isinstance(assumptions,ListExpression)`,
my only concern is about something like   ```Integrate[Q[x], {x,0,2}, Assumptions-> List@@F[x]]``` produces an intermediate expression of the form ``List[___]``` that is not a ListExpression.

mmatera (approved):
Besides the comment, LGTM

mathics/builtin/assumptions/assumptions.py
rocky: We should fix that then. 

I will look into this.

mmatera: I mean, I do not know if this is important in this particular case. But what is true is in some intermediate evaluations, we can have an `Expression` with `List` as its head.

rocky: Those should be found and fixed.

rocky: @mmatera I did a little investigation on this. One place that this can occur is in `Map..[f, ...]` functions where F is `List`.

Aside from the obvious approach of making the Map evaluation functions sensitive to this change, another approach to addressing this is to catch it inside the `Expression()` object creator like this: 

```python
    def __new__(
        cls,
        *args, **kwargs
    ):
        if len(args) > 0 and args[0] is SymbolList and cls is not ListExpression:
            return ListExpression(args[0], *args[1:], **kwargs)
        return super().__new__(cls)
```

I experimented with this a little bit, but the above might not be quite right. 

(Also, I did find what looks like a little bug in `Map`, that I'll report and put in a small PR for later.)

But given this, right now, my feeling is to declare yet another temporary defeat, add it as a chore and revert any `isinstance(xxx, ListExpression)` additions. 

That way, we'll make a little bit of progress on this front, even if there are many other fronts that remain needy.

Your thoughts?

PR #1963, opened 2026-09-17
Go over `mathics.core.atoms.numerics`

rocky:
Go over `mathics.core.atoms.numerics`

* Go over the use of `@cache` and `@cache_property` (not used yet)
*  Start introducing Python slots to save memory and speed up access.
* Calling SymPy is slow, so prefer to initialize it lazily
* Complex `real` and `imag` attributes are now properties, so they can't be changed.
* Go over types and adjust comments
* `expr.nan` is now a property


rocky:
This is in preparation for removing the `is_inexact`  method from BaseElement. There will be an `in_exact()` function that accepts any object as a parameter.

mmatera (approved):
LGTM

PR #1964, opened 2026-09-18
move Element.is_inexact method to mathics.core.atoms.numerics.is_inexact

rocky:
Move `Element.is_inexact()` method to the function `mathics.core.atoms.numerics.is_inexact`.

 `is_inexact` is now a property on Number (not Element). 
Use the function `is_inexact(expr)` for arbitrary elements. 

Also, convert more`has_form` `str` parameters to `Symbol`.
Spelling and grammar were gone over in some of the changed files.


mmatera (approved):
LGTM

rocky:
After this we are in a position to (finally) move method `Element.get_precision` to property method `.precision` and `mathics.core.numerics.precision()`.  And the same with `.is_zero`.

I tried these earlier and ran into these the is_inexact situation.

PR #1966, opened 2026-09-19
Remove LaTeX PDF generation

rocky:
This is now located in GitHub repository Mathics3/Mathics-Documentation-LaTeX


mathics/doc/documentation/1-Manual.mdoc
```diff
@@ -1,1427 +0,0 @@
-
```
mmatera, line 1: This file is used in `/test/doc/test_latex.py`.  Maybe we can put a part of this inside the test folder, or we can move these tests to the new MathicsDocumentation repo. In that case,, I would also move `mathics.doc.latex_doc` there.

rocky: Locally `test/doc/test_latex.py` has been moved to the Mathics3-Documentation-LaTeX repo. I just haven't pushed this branch yet.  I will push that branch soon though. 

It will be a bit messy and will still be draft-like. 

I think in the middle I want to finish converting `has_form` and moving numeric functions out of BaseElement.

